PR #3, opened 2021-11-18
Add squashfs support in archive handlers.

QKaiser:
Working draft for squashfs chunk identification, carving, and extraction:
- supports squashfs version 3 to 4 included.
- supports little endian and big endian encodings.


QKaiser:
Created integration tests for each version and endianness (note that v4 does not support big endian). If all tests pass I think we can merge after a quick review.

PR #17, opened 2021-11-22
Add pre-commit.ci instead of action/pre-commit

kukovecz:
The pre-commit action is in mainteance mode
(https://github.com/pre-commit/action), they are suggesting to use
pre-commit.ci


kukovecz:
Deleted the branch and closed the pull request, because I decided we are not going into that direction (yet).

PR #25, opened 2021-11-23
Limited file reader

kissgyorgy:
A subclass of `io.BufferedIOBase` which ensures we don't seek earlier than the start offset.

### DIFF
--- a/tests/test_file_utils.py
+++ b/tests/test_file_utils.py
@@ -1,4 +1,5 @@
 import io
+from unittest.mock import MagicMock
 
 import pytest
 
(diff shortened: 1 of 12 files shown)


tests/test_file_utils.py
```diff
@@ -15,3 +17,37 @@
...
+        assert reader.tell() == 5
+
+    def test_seek_to_end_of_file(self, fake_file):
+        reader = LimitedStartReader(fake_file, 5)
+        reader.seek(-1, io.SEEK_END)
+        assert reader.tell() == len(fake_file.getvalue()) - 1
+
+    def test_read_only(self, fake_file):
+        with pytest.raises(TypeError):
+            reader = LimitedStartReader(fake_file, 5)
```
kukovecz, line 52: This line can be out of the `pytest.raises` context so we would be sure that the `write` function raises the error.

unblob/file_utils.py
```diff
@@ -9,3 +10,38 @@ def snull(content: bytes):
...
+    def seek(self, offset: int, whence=io.SEEK_SET):
+        new_pos = self._file.seek(offset, whence)
+        if new_pos < self._start:
+            new_pos = self._file.seek(self._start, io.SEEK_SET)
+        return new_pos
+
+    def write(self, *args, **kwargs):
+        raise TypeError("MUST NOT call write method")
+
+    def detach(self, *args, **kwargs):
```
kukovecz, line 34: Don't know if it is a good / effective idea, but I would write simple test cases for each of these functions, at least check them that they return with the result of calling the same function (with the same parameters) on the `self._file`

unblob/models.py
```diff
@@ -86,10 +86,13 @@ class Handler(Protocol):
 
     NAME: str
     YARA_RULE: str
+    # We need this, because not every match reflects the actual start
+    # (e.g. tar magic is in the middle of the header)
+    YARA_MATCH_OFFSET: int
```
kukovecz, line 91: This `YARA_MATCH_OFFSET` is not added everywhere, please add those!
(while you were working on this branch, new `filesystem` handlers were introduced and they would fail because of this )

tests/test_file_utils.py
kissgyorgy: Good catch!

PR #33, opened 2021-11-24
Add RAR handler

tests00:
Add handler and tests for RAR. 

The handler will find two distinct RAR magics: from v1.5 to v4 (called $magic_v4), and v5 (called $magic_v5). 

There's only a test for RAR v5 at the moment.


unblob/handlers/archive/rar.py
```diff
@@ -1,175 +1,46 @@
...
-            header.flags
-            & [x[0] for x in FILE_HEADER_FLAGS if x[1] == "LHD_LONG_BLOCK"][0]
-            or header.type == 0x74
-        ):
-            header = cparser.base_vol_header_data(file)
-            file.seek(offset)
-            extra_data = header.datasize
+def calculate_rar_size(file: io.BufferedIOBase, start_offset: int):
+    rar_limited = LimitedStartReader(file, start_offset)
+    rar_file = rarfile.RarFile(rar_limited)
```
QKaiser, line 27: I'd recommend catching the rarfile exceptions here (see https://rarfile.readthedocs.io/api.html?#exceptions). 

- rarfile.BadRarFile: header is f'ed
- rarfile.NotRarFile: not a rar file, unlikely to happen but still
- rarfile.NeedFirstVolume - yara match in between ?

I'd also take password protected RAR files into acocunt. They can be detected using `rar_file.needs_password()`.

unblob/handlers/archive/rar.py
```diff
@@ -1,175 +1,46 @@
...
-    return Chunk(start_offset, end_offset)
+    return ValidChunk(
+        start_offset=start_offset,
+        end_offset=rar_end_offset,
+    )
 
 
 def make_extract_command(inpath: str, outdir: str) -> List[str]:
-    return ["unrar", "e", inpath, outdir]
+    return ["unar", inpath, "-o", outdir]
```
QKaiser, line 46: This will hang waiting for stdin if it's a password protected rar file. Similar to ZIP handling, I'd call it this way:

```
unar -p '' -o outdir inpath
pwprotect.rar: RAR 5
Archive parsing failed! (Missing or wrong password.)
No files extracted.
```

unblob/handlers/archive/rar.py
kissgyorgy: I checked for all possible exceptions and checking for password protected files and issue a warning in that case. It works for both encrypted and password protected files too.

vlaci (approved):

unblob/handlers/archive/rar.py
```diff
@@ -0,0 +1,51 @@
...
+class RarHandler(Handler):
+    NAME = "rar"
+
+    YARA_RULE = r"""
+        strings:
+            $rar_magic_v4 = { 52 61 72 21 1A 07 00 }
+            $rar_magic_v5 = { 52 61 72 21 1A 07 01 00 }
+
+        condition:
+            $rar_magic_v4 or $rar_magic_v5
```
QKaiser, line 29: We could simplify this YARA rule this way:

```suggestion
        strings:
            // RAR v4.x ends with 00, RAR v5.x ends with 01 00
            $magic_rar = { 52 61 72 21 1A 07 ( 00 | 01 00 ) }
        condition:
            $magic_rar
```

QKaiser (commented):
Have a look at the YARA rule simplification and rebase on main :)

unblob/handlers/archive/rar.py
kissgyorgy: Done in a separate PR: https://github.com/IoT-Inspector/unblob/pull/60

PR #34, opened 2021-11-24
Fail on missing integration tests

### LINKED ISSUE #28: Fail on missing integration tests
When we have a handler which don't have integration tests in the folder, the test suite should fail.
Currently we are only collecting existing tests.

kukovecz:
Makes the test suite fail if there is a `handler` without any integration tests.

Closes #28


tests/test_handlers.py
```diff
@@ -53,3 +56,34 @@ def test_all_handlers(input_dir: Path, output_dir: Path, tmp_path: Path):
...
+@pytest.mark.parametrize(
+    "handler",
+    (
+        pytest.param(handler, id=handler.NAME)
+        for handler_map in _ALL_MODULES_BY_PRIORITY
+        for handler in handler_map.values()
+    ),
+)
+def test_missing_handlers_integrations_tests(handler):
+    handler_name = handler.NAME
```
kissgyorgy, line 72: You can use `handler.NAME` instead of renaming the variable. It's less indirect and has the same length :)

tests/test_handlers.py
```diff
@@ -53,3 +56,34 @@ def test_all_handlers(input_dir: Path, output_dir: Path, tmp_path: Path):
...
+    handler_path = (
+        Path(inspect.getfile(handler.__class__))
```
kissgyorgy, line 74: I would do it multiple steps, because a little bit hard to glance.
```suggestion
        handler_module_path = Path(inspect.getfile(handler.__class__))
        handler_test_path = handler_module_path.relative_to(HANDLERS_PACKAGE_PATH ).with_suffix("")
```

tests/test_handlers.py
```diff
@@ -7,18 +7,21 @@
...
 
+from unblob.handlers import _ALL_MODULES_BY_PRIORITY
 from unblob.processing import DEFAULT_DEPTH, process_file
 
 TEST_DATA_PATH = Path(__file__).parent / "integration"
 TEST_INPUT_DIRS = list(TEST_DATA_PATH.glob("**/__input__"))
 TEST_CASE_DIRS = [p.parent for p in TEST_INPUT_DIRS]
 TEST_OUTPUT_DIRS = [p / "__output__" for p in TEST_CASE_DIRS]
 TEST_IDS = [p.name for p in TEST_CASE_DIRS]
+HANDLER_MODULE_PATH = Path(__file__).parent.parent / "unblob" / "handlers"
```
kissgyorgy, line 26: A more future-proof and less-verbose way would be:
```suggestion
from unblob import handlers
HANDLERS_PACKAGE_PATH = Path(handlers.__file__).parent
```
Also it's not a module, it's a package.

tests/test_handlers.py
```diff
@@ -53,3 +58,34 @@ def test_all_handlers(input_dir: Path, output_dir: Path, tmp_path: Path):
...
+)
+def test_missing_handlers_integrations_tests(handler):
+    handler_name = handler.NAME
+    handler_path = (
+        Path(inspect.getfile(handler.__class__))
+        .relative_to(HANDLER_MODULE_PATH)
+        .with_suffix("")
+    )
+
+    if handler_name == handler_path.name:
```
kissgyorgy, line 79: I know I was the one pushing for this 😄, but maybe I would not do this, because if we add new tests next to the existing, the files have to move. Which would cause unnecessary diffs in git commits. With Git LFS, we should embrace producing more test files with edge cases. Test directory structure would also be more consistent with no exception.

tests/test_handlers.py
```diff
@@ -53,3 +58,34 @@ def test_all_handlers(input_dir: Path, output_dir: Path, tmp_path: Path):
...
+        # when there is 1 handler class in the handler module, with the same NAME as the module
+        expected_test_path = handler_path
+    else:
+        expected_test_path = handler_path.joinpath(handler_name)
+
+    test_path = TEST_DATA_PATH.joinpath(expected_test_path)
+    found_tests = list(test_path.glob("**/__input__"))
```
kissgyorgy, line 86: Maybe you can just simply call `test_path.exists()` instead.
If there is a path there, it probably has tests.

kissgyorgy (changes requested):
I would just squash all the "move test files" in one commit.  
I would name squashfs_v3 tests `little_endian` and `big_endian` instead, simply `le` and `be` might be a little bit ambiguous.

kissgyorgy (approved):

PR #35, opened 2021-11-24
Add ZIP

tests00:
Add ZIP handler.

This PR also adds a new helper function called `find_first()`, which can be used to find the first instance of a certain byte pattern in a stream. It reads in chunks, and returns -1 if the pattern isn't found.


unblob/file_utils.py
```diff
@@ -12,6 +13,24 @@ def round_up(size: int, alignment: int):
     return alignment * math.ceil(size / alignment)
 
 
+def find_first(file: io.BufferedIOBase, pattern: bytes) -> int:
```
QKaiser, line 16: I'd recommend making the chunk_size an optional parameter with default value set 0x1000, making the function more generic as we might want to re-use it in the future for other formats but need to read by smaller or larger increment.

unblob/handlers/archive/zip.py
```diff
@@ -98,78 +98,76 @@
...
-    content = io.BytesIO(file.read(zip_end - start_offset))
 
-    with ZipFile(content) as z:
-        logger.info("Found ZIP filenames", filenames=[x.filename for x in z.infolist()])
-        for g in z.infolist():
-            if g.flag_bits & 0b0001:
-                encrypted_files.add(g.filename)
-            file_names.add(g.filename)
+    # In our case, we want to find the first instance of the EOCD record header, not the last!
+    zip_end = find_first(file, b"\x50\x4b\x05\x06") + start_offset
```
QKaiser, line 115: Please make things explicit by using a variable name explaining what we are actually looking for.

```suggestion
    eocd_marker = b"\x50\x4b\x05\x06"
    zip_end = find_first(file, eocd_marker) + start_offset
```

unblob/handlers/archive/zip.py
```diff
@@ -98,78 +98,76 @@
...
+    with zipfile.ZipFile(this_zip_chunk) as zip:
+        for zipinfo in zip.infolist():
+            if zipinfo.flag_bits & 0b0001:
+                encrypted_files.add(zipinfo.filename)
+            all_files.add(zipinfo.filename)
+
+    if len(encrypted_files) > 0:
+        # TODO: We can't handle encrypted ZIP files yet, so we fall back to the UnknownChunk in the
+        # cases where there are encrypted files in the ZIP.
+        return UnknownChunk(
```
QKaiser, line 159: Let's talk about encrypted files within ZIP archives here. With the current logic, an archive with multiple plaintext files and one encrypted file will not be processed by unblob. This is problematic.

If we provide the `-P` argument with an empty value, unzip will simply skip encrypted files with wrong password without triggering exceptions:

```
unzip -d lol -P '' fruits.zip
Archive:  fruits.zip
   skipping: fruits/banana.txt       incorrect password
```

I'd be in favor of using this in order to emit a valid chunk even if an encrypted file within an archive is found. This would make unblob carve out the whole zip so that an analyst can take a deeper look at it.

Given the extraction process won't be lossless in these cases (encrypted files won't be extracted), I'd be in favor of logging a warning message saying "X files from zip chunk Y are encrypted but we don't have a password, better have a look !" or something.

unblob/handlers/archive/zip.py
```diff
@@ -98,78 +98,76 @@
...
     return ValidChunk(
         start_offset=start_offset,
-        end_offset=start_offset + size,
+        end_offset=end_of_zip,
     )
 
 
 def make_extract_command(inpath: str, outdir: str) -> List[str]:
+    # TODO: This will just hang waiting for user input if any the ZIP is encrypted.
     return ["unzip", inpath, "-d", outdir]
```
QKaiser, line 173: See the previous comment on `-P` operator to get rid of hangs when processing archives with encrypted content.

kissgyorgy: We are using 7z with the same `-p ""`  argument.

kukovecz (approved):

PR #37, opened 2021-11-24
Logging improvements

kissgyorgy:


unblob/logging.py
```diff
@@ -2,22 +2,45 @@
...
+            elif isinstance(value, cstruct):
+                event_dict[key] = dumpstruct(value)
+
         return event_dict
 
-    return add_path
+    return convert_type
+
+
+class LazyField:
```
kukovecz, line 28: I believe this (commit: 97f5700fea3f4e4805f0c9f9e77e2ac2e745f54a) was left here by mistake?

unblob/logging.py
```diff
@@ -2,22 +2,45 @@
...
-    def add_path(logger, method_name: str, event_dict: structlog.types.EventDict):
+def pretty_print_types(extract_root: Path):
+    def convert_type(logger, method_name: str, event_dict: structlog.types.EventDict):
         use_absolute_path = event_dict.pop("_absolute_path", False)
         for key, value in event_dict.items():
             if isinstance(value, Path):
                 path = value if use_absolute_path else value.relative_to(extract_root)
                 event_dict[key] = str(path)
+
+            elif isinstance(value, cstruct):
```
kukovecz, line 20: I really, really love this idea!

However, I wanted to see for myself, and it was not working. I check the `zip` handler's debug logs, they were in a "normal" format, so the `dumpstruct` was not running for them.

With this modification I could make it work:
```suggestion
            elif isinstance(value, cstruct) or isinstance(value, Instance):
```

kukovecz, line 20: I really, really love this idea!

However, I wanted to see for myself, and it was not working. I checked the `zip` handler's debug logs, they were in a "normal" format, so the `dumpstruct` was not running for them.

With this modification I could make it work:
```python
from dissect.cstruct.types.instance import Instance
...

elif isinstance(value, cstruct) or isinstance(value, Instance):
```

kukovecz (changes requested):
Here are my suggestions and I just clicked "Request changes" to see what will happen with the pull request this way :sweat_smile:

unblob/logging.py
kissgyorgy: I wanted to keep this in the history and reverted this commit.

kukovecz: But you reverted the reverted commit, and it is in there, right now, and it should not be.

unblob/logging.py
kissgyorgy: Thanks, I changed it to check only for `Instance`, because it seems the parser methods will emit the instance of that.

PR #39, opened 2021-11-24
Add 7z handler.

tests00:
This has also caused me to add an extra property to the LimitedStartReader, so that py7zr can access the name property of the file (otherwise it will throw an assertation error).


kissgyorgy:
We decided not to use an external library for this, as the 7zip format is simple.

kukovecz (approved):

vlaci (approved):

PR #40, opened 2021-11-24
Handler base class

kissgyorgy:
Big refactoring of the Handler implementation


unblob/models.py
```diff
@@ -81,21 +85,40 @@ def __repr__(self):
...
+
+    def __init__(self):
+        if self.C_STRUCTURES:
+            self.cparser = cstruct()
+            self.cparser.load(self.C_STRUCTURES)
+
+            if self.HEADER_STRUCT:
+                self._header_parser = getattr(self.cparser, self.HEADER_STRUCT)
+
+    def parse_header(self, file: io.BufferedIOBase):
```
QKaiser, line 110: If we add an optional endianness argument that defaults to little endian, we could rely on parse_header in CPIO v3 handlers, and be ready for future use cases where we need to handle big endian transparently without rewriting parse_header in the subclass.

```suggestion
    def parse_header(self, file: io.BufferedIOBase, endian= "<"):
      self.cparser.endian = endian
```

unblob/handlers/filesystem/squashfs.py
```diff
@@ -0,0 +1,149 @@
...
+        self.cparser_be.load(self.C_STRUCTURES)
+
+    def calculate_chunk(
+        self, file: io.BufferedIOBase, start_offset: int
+    ) -> Union[ValidChunk, UnknownChunk]:
+
+        # read the magic and derive endianness from it
+        magic_bytes = file.read(4)
+        magic = struct.unpack(">I", magic_bytes)[0]
+        cparser = self.cparser_be if magic == BIG_ENDIAN_MAGIC else self.cparser_le
```
QKaiser, line 90: If we use the optional endian parameter in parse_header in the Handler base class, we can do:

```suggestion
        endian = ">" if magic == BIG_ENDIAN_MAGIC else "<"
        header = self.parse_header(file, endian)
```

Of course we need to define `HEADER_STRUCT` for both CPIO v3.

QKaiser (changes requested):
Just two comments related to endianness support in parse header, otherwise everything looks great !

unblob/models.py
kissgyorgy: There will be only one `Handler` class instance during the whole run, so we still must not change any state of the instance. The __init__ is there, so the subclasses can customize some common behavior or initialize extra attributes needed.

vlaci: What is the overhead of `cparser` initialization? Is it worth storing it in the instance?

kissgyorgy:
```
>>> In [10]: %timeit parser=cstruct(); parser.load(C_STRUCTURES)
2.17 ms ± 63.1 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

kukovecz (approved):
The refactoring looks good to me. From my PoV, this can be merged (when concerns raised by others are attended)

unblob/models.py
QKaiser: Ok that makes sense. Let's leave it this way then.

QKaiser (approved):

PR #42, opened 2021-11-25
New StructHandler base class

kissgyorgy:
We need to be able to parse structs with different endianness dynamically.
The StructParser class creates the cparser for the given endian lazily.

Handlers which want to use this parser must subclass StructHandler.


QKaiser (approved):
Looks good to me. Learned about underscores in integer values in Python today :)

kukovecz (approved):

PR #45, opened 2021-11-25
No unknownchunk

### LINKED ISSUE #21: We should stop using UnkownChunk for matched structures with invalid headers
Right now we return an UnknownChunk for any YARA match where its handler failed to compute a ValidChunk. This can be due to YARA matching on something that is not an archive format, the archive header being corrupted, etc.

To me, an unknown chunk should represent chunks in-between valid chunks for which we did not match. We should not emit structures for something that our handlers failed to process.

This can be problematic when running unblob on a 2.5Gb ISO image of Ubuntu. We will have a valid ISO9660 chunk, followed by hundreds of UnknownChunk being printed out on the console. Mostly "fake" ZIP files in my case.

My solution would be to simply return `None` when a format handler cannot properly identify a chunk.

kissgyorgy:
Fixes #21

### DIFF
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -18,9 +18,5 @@ class TestChunk:
         assert chunk1.contains(chunk2) is result
 
     def test_range_hex(self):
-        chunk = UnknownChunk(start_offset=3, end_offset=10, reason="Doesnt matter")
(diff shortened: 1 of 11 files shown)


QKaiser (approved):
Developed in pair session so I agree with this !

kukovecz (approved):

PR #46, opened 2021-11-26
Add ARJ

tests00:
This PR adds the ARJ format. There's no Python library we can use here, so I'm parsing the file in a very simple way myself to walk file nodes to the end.

This also introduces the `read_until_null()` helper function, which will read a stream until the first null byte. Due to the way that the ARJ headers work, it's not possible for us to know the length of the filename and comment strings in the header, but these strings are null-terminated. So we use the header `first_hdr_size` value to read past the header to where these strings start.


.github/workflows/tests.yml
```diff
@@ -19,6 +19,10 @@ jobs:
             pip install poetry
             poetry install
 
+      - name: Install 3rd party from apt
+        run: |
+            sudo apt install unar
+
```
QKaiser, line 25: Rebase onto main, this dependency is now part of the project and can be removed from this commit.

unblob/handlers/archive/arj.py
```diff
@@ -0,0 +1,184 @@
...
+
+class ARJExtendedHeaderException(Exception):
+    pass
+
+
+class ARJFileLengthException(Exception):
+    pass
+
+
+class ARJHandler(Handler):
```
QKaiser, line 25: Rebase onto main and move to a StructHandler to have access to functions like `parse_header`.

unblob/handlers/archive/arj.py
```diff
@@ -0,0 +1,184 @@
...
+            extended_header = cparser.extended_header(file)
+            if extended_header.size != 0:
+                raise ARJExtendedHeaderException
+
+            # Read past the file contents
+            _ = file.read(file_header.compressed_size)
+
+    def calculate_chunk(
+        self, file: io.BufferedIOBase, start_offset: int
+    ) -> Union[ValidChunk, UnknownChunk]:
```
QKaiser, line 151: Rebase onto main, we moved away from UnknownChunk. Signature is now:

````python
-> Optional[ValidChunk]
````

File: unblob/handlers/archive/arj.py [hunk 4]
```diff
@@ -0,0 +1,184 @@
...
+        except ARJExtendedHeaderException:
+            return UnknownChunk(
+                start_offset=start_offset,
+                reason="Main ARJ header contains extended_header, which we don't currently handle.",
+            )
+        except EOFError:
+            return UnknownChunk(
+                start_offset=start_offset,
+                reason="File ends before ARJ file resolves.",
+            )
```
QKaiser, line 175: All of this can be replaced by warning messages now that we are no longer emitting UnknownChunk structures.

unblob/handlers/archive/arj.py
```diff
@@ -0,0 +1,184 @@
...
+            _ = file.read(file_header.compressed_size)
+
+    def calculate_chunk(
+        self, file: io.BufferedIOBase, start_offset: int
+    ) -> Union[ValidChunk, UnknownChunk]:
+
+        # Not using the inbuild self.parse_header() here, because we need access to a few different
+        # structures at various points.
+        cparser = cstruct()
+        cparser.load(self.C_STRUCTURES)
```
QKaiser, line 156: We should discuss this with @kissgyorgy and see if we can make some generalization.

kissgyorgy (approved):

QKaiser (approved):

PR #47, opened 2021-11-26
Error log about obviously invalid Chunks

kissgyorgy:
When a Chunk overflows the end of the file, it's obviously invalid,
it's either a bug in our code or a corrupt file/header or something.

We can continue parsing, because it might be a bug only in one of the parsers.


unblob/strategies.py
```diff
@@ -38,9 +40,15 @@ def search_chunks_by_priority(path: Path, file: io.BufferedReader) -> List[Chunk
                 real_offset = offset + handler.YARA_MATCH_OFFSET
                 limited_reader = LimitedStartReader(file, real_offset)
                 chunk = handler.calculate_chunk(limited_reader, real_offset)
+
                 # We found some random bytes this handler couldn't parse
                 if chunk is None:
                     continue
+
+                if chunk.end_offset >= file_size:
```
kukovecz, line 48: Would something like this make sense?

```suggestion
                if chunk.end_offset >= file_size or chunk_size.start_offset < 0:
```

unblob/strategies.py
```diff
@@ -38,9 +40,15 @@ def search_chunks_by_priority(path: Path, file: io.BufferedReader) -> List[Chunk
...
+                    logger.error("Chunk overflows file", chunk=chunk)
```
kukovecz, line 49: It would make sense to put the file size to the log context as well.

```suggestion
                    logger.error("Chunk overflows file", chunk=chunk, file_size=format_hex(file_size))
```

`format_hex` since the `Chunk` has its `__repr__` in hex)

unblob/strategies.py
kukovecz: We fixed this by using the 2nd diff

unblob/strategies.py
kukovecz: The `file_size` is already logged when processing the file.

kukovecz (approved):

PR #51, opened 2021-11-26
Add support for ARC archive format.

QKaiser:
We use a strong matching Yara rule to get rid of potential false positive due the lack of "magic" signature field in the ARC format definition.
We rely on unar to extract carved files given that the 'arc' Unix utility does not support explicit output directories.

If you want to create test ARC archive files, you can use the 'arc' unix utility:

```
# create archive
arc a archive.arc directory/*
# list archive content
arc l archive.arc directory/*
# extract archive content
arc x archive.arc
```


.github/workflows/tests.yml
```diff
@@ -19,6 +19,10 @@ jobs:
             pip install poetry
             poetry install
 
+      - name: Install 3rd party from apt
+        run: |
+            sudo apt install unar
```
kukovecz, line 24: This is already added in the `main` branch, so this modification will not be needed once this branch is rebased on top of the current `main`,

unblob/handlers/archive/arc.py
```diff
@@ -0,0 +1,70 @@
...
+        short crc;	/* cyclic redundancy check */
+        long length;	/* true file length */
+    };
+    """
+
+    HEADER_STRUCT = "heads"
+
+    def calculate_chunk(
+        self, file: io.BufferedIOBase, start_offset: int
+    ) -> Union[ValidChunk, UnknownChunk]:
```
kukovecz, line 50: This type annotation is changed in the `main`, so this will needs to be updated once this this is rebased.
```suggestion
    ) -> Optional[ValidChunk]:
```

unblob/handlers/archive/arc.py
```diff
@@ -0,0 +1,70 @@
...
+
+        # we loop from header to header until we reach the end header
+        offset = start_offset
+        while True:
+            file.seek(offset)
+            if file.read(2) == END_HEADER:
```
kukovecz, line 56: We can (easily?) stuck in an infinite loop, if/when this condition is never `True`.  Can we do something to avoid this?

.github/workflows/tests.yml
QKaiser: Rebased on top of the current `main` so it's no longer there.

unblob/handlers/archive/arc.py
QKaiser: Fixed !

kissgyorgy (approved):

PR #55, opened 2021-11-26
Add support for Apple DMG archive format.

QKaiser:
Detect DMG archives based on trailer signature, the carved chunk is extracted with 7z.

Instructions on how to create DMG files:

```
# install dependencies
sudo apt install hfsprogs
# create random file
dd if=/dev/zero of=/tmp/fruits.hfs bs=1M count=16 status=progress
# format random file as HFS+
mkfs.hfsplus -v fruits /tmp/fruits.hfs
# mount HFS+
udisksctl loop-setup -f fruits.hfs
# write files to the mounted HFS+ filesystem
echo apple > /media/$USER/fruits/apple.txt
echo cherry > /media/$USER/fruits/cherry.txt
udisksctl unmount -b /dev/loopXX
# convert to DMG using libdmg-hfsplus
/home/$USER/libdmg-hfsplus/dmg/dmg dmg fruits.hfs fruits.dmg
```

You'll need to build [libdmg-hfsplus](https://github.com/planetbeing/libdmg-hfsplus).


QKaiser:
@kissgyorgy need advice on why the test is failing during the workflow checks even though pytest succeeds on my machine.

PR #57, opened 2021-11-26
Add support for Apple DMG archive format.

QKaiser:
Detect DMG archives based on trailer signature, the carved chunk is extracted with 7z.

Instructions on how to create DMG files:

```
# install dependencies
sudo apt install hfsprogs
# create random file
dd if=/dev/zero of=/tmp/fruits.hfs bs=1M count=16 status=progress
# format random file as HFS+
mkfs.hfsplus -v fruits /tmp/fruits.hfs
# mount HFS+
udisksctl loop-setup -f fruits.hfs
# write files to the mounted HFS+ filesystem
echo apple > /media/$USER/fruits/apple.txt
echo cherry > /media/$USER/fruits/cherry.txt
udisksctl unmount -b /dev/loopXX
# convert to DMG using libdmg-hfsplus
/home/$USER/libdmg-hfsplus/dmg/dmg dmg fruits.hfs fruits.dmg
```

You'll need to build [libdmg-hfsplus](https://github.com/planetbeing/libdmg-hfsplus).


QKaiser:
Tests are failing due to the presence of `[HFS+ Private Data]` directory in extracted content, which cannot be tracked by git. Need to find a way around this.

QKaiser:
The test sample here is an HFS+ filesystem packed in DMG archive, which is used a lot in the wild. I fixed the 7zip extraction command to not extract that specific directory (`[HFS+ Private Data]`). I don't think we will ever need information from it. The reasons behind this directory's existence are explained here https://newbedev.com/hfs-private-directory-data

Mentioning it here so we don't forget about it when implementing HFS/HFS+ handlers in the future.

PR #62, opened 2021-11-29
Optimize search_chunks_by_priority: Skip calculate_chunk if the offset is already used

### LINKED ISSUE #23: Stop running handlers when a chunk matches the whole file.
Whenever a chunk matching the whole file is found, we should just return it and stop processing others.

kukovecz:
Fixes #23

### DIFF
--- a/unblob/strategies.py
+++ b/unblob/strategies.py
@@ -33,12 +33,13 @@ def search_chunks_by_priority(  # noqa: C901
             match = result.match
             sorted_matches = sorted(match.strings, key=itemgetter(0))
             for offset, identifier, string_data in sorted_matches:
+                real_offset = offset + handler.YARA_MATCH_OFFSET
                 logger.info(
                     "Calculating chunk for YARA match",
                     start_offset=format_hex(offset),
+                    real_offset=format_hex(real_offset),
                     identifier=identifier,
                 )
-                real_offset = offset + handler.YARA_MATCH_OFFSET
                 limited_reader = LimitedStartReader(file, real_offset)
                 chunk = handler.calculate_chunk(limited_reader, real_offset)
 
--- a/unblob/strategies.py
+++ b/unblob/strategies.py
@@ -31,7 +31,7 @@ def search_chunks_by_priority(  # noqa: C901
         for result in yara_results:
             handler = result.handler
             match = result.match
-            sorted_matches = sorted(match.strings, key=itemgetter(0), reverse=True)
+            sorted_matches = sorted(match.strings, key=itemgetter(0))
             for offset, identifier, string_data in sorted_matches:
                 logger.info(
                     "Calculating chunk for YARA match",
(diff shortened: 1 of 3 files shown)


unblob/strategies.py
```diff
@@ -55,9 +55,9 @@ def search_chunks_by_priority(  # noqa: C901
                 log.info("Found valid chunk")
                 all_chunks.append(chunk)
 
-                if chunk.size == file_size:
-                    log.info("This Chunk represents the whole file")
-                    return [chunk]
+                if sum([chunk.size for chunk in all_chunks]) == file_size:
```
kissgyorgy, line 58: This is not enough, the chunks has to be aligned.
Write a new function so you can unit test this.

kissgyorgy (changes requested):

unblob/strategies.py
kukovecz: Yes, thank you.

tests/test_strategies.py
```diff
@@ -85,3 +89,29 @@ def test_remove_inner_chunks(chunks, expected, explanation):
 )
 def test_calculate_unknown_chunks(chunks, file_size, expected):
     assert expected == calculate_unknown_chunks(chunks, file_size)
+
+
+@pytest.mark.parametrize(
+    "all_chunks, offset, expected, explanation",
+    [
+        ([], 0x0, False, "Empty all_chunks, offset 0x0"),
```
kissgyorgy, line 31: `pytest.param(id=)`

unblob/strategies.py
```diff
@@ -16,6 +16,14 @@
 logger = get_logger()
 
 
+def chunk_already_calculated(all_chunks: List[Chunk], offset: int) -> bool:
```
kissgyorgy, line 19: I would put this method on Chunk

unblob/strategies.py
```diff
@@ -34,6 +42,10 @@ def search_chunks_by_priority(  # noqa: C901
             sorted_matches = sorted(match.strings, key=itemgetter(0))
             for offset, identifier, string_data in sorted_matches:
                 real_offset = offset + handler.YARA_MATCH_OFFSET
+
+                if chunk_already_calculated(all_chunks, real_offset):
```
kissgyorgy, line 46: so you could use that Chunk method like this:
```
for chunk in all_chunks:
    if chunk.contains_offset(offset):
        continue
```

unblob/strategies.py
```diff
@@ -71,27 +71,6 @@ def search_chunks_by_priority(  # noqa: C901
     return all_chunks
 
 
-def remove_inner_chunks(chunks: List[Chunk]):
```
kissgyorgy, line 65: We still need this function because when formats are not inside the other we predicted with the priority, it can happen that we still found chunks inside another.

kissgyorgy (changes requested):

kissgyorgy (approved):

PR #67, opened 2021-11-29
Use separate file for handlers

### LINKED ISSUE #43: Multiple handler should be able to use the same file at the same time
We must ensure they can seek and read from the same file independently, so they can be run in parallel.

kukovecz:
Fixes #43


kukovecz:
We closed this in favor of #71 and #72 - we are going with the parallelization in a different direction.

PR #79, opened 2021-12-02
Don't read the whole file into the memory when carving out chunks

### LINKED ISSUE #74: Don't read the whole file into the memory when carving out chunks
We are currently reading the whole file into memory when carving out chunks, which is obviously terrible. There is a function for that in standard library: `shutil.copyfileobj`, we should use that instead.

kukovecz:
Fixes #74

### DIFF
--- a/tests/test_file_utils.py
+++ b/tests/test_file_utils.py
@@ -1,4 +1,5 @@
 import io
+from typing import List
 from unittest.mock import MagicMock
 
 import pytest
@@ -9,6 +10,7 @@ from unblob.file_utils import (
     StructParser,
     convert_int32,
     find_first,
+    iterate_file,
     round_up,
 )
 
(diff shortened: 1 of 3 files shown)


tests/test_file_utils.py
```diff
@@ -156,3 +157,36 @@ def test_find_first_smaller_chunksize_than_pattern_doesnt_hang():
...
+        pytest.param(b"0123456789", 0, 5, 3, [b"012", b"34"], id="size_buffersize"),
+        pytest.param(b"0123456789", 0, 5, 100, [b"01234"], id="big_buffersize"),
+        pytest.param(b"0123456789", 0, 10, 10, [b"0123456789"], id="real_all"),
+        pytest.param(b"0123456789", 0, 10, 100, [b"0123456789"], id="big_buffersize"),
+        pytest.param(b"0123456789", 0, 100, 10, [b"0123456789"], id="big_size"),
+        pytest.param(b"0123456789", 5, 3, 2, [b"56", b"7"], id="middle_offset"),
+        pytest.param(b"0123456789", 10, 3, 2, [], id="offset_bigger_than_file"),
+    ),
+)
+def test_iterate_file(content, start_offset, size, buffer_size, expected):
```
kissgyorgy, line 177: put type hints to the test signatures so pyright will be able to catch errors even in tests.

kissgyorgy (changes requested):

tests/test_file_utils.py
```diff
@@ -156,3 +157,36 @@ def test_find_first_smaller_chunksize_than_pattern_doesnt_hang():
...
+        pytest.param(b"", 0, 1, 10, [], id="empty_file"),
+        pytest.param(b"0123456789", 0, 6, 3, [b"012", b"345"], id="from_start"),
+        pytest.param(b"0123456789", 0, 0, 3, [], id="zero_size"),
+        pytest.param(b"0123456789", 0, 5, 3, [b"012", b"34"], id="size_buffersize"),
+        pytest.param(b"0123456789", 0, 5, 100, [b"01234"], id="big_buffersize"),
+        pytest.param(b"0123456789", 0, 10, 10, [b"0123456789"], id="real_all"),
+        pytest.param(b"0123456789", 0, 10, 100, [b"0123456789"], id="big_buffersize"),
+        pytest.param(b"0123456789", 0, 100, 10, [b"0123456789"], id="big_size"),
+        pytest.param(b"0123456789", 5, 3, 2, [b"56", b"7"], id="middle_offset"),
+        pytest.param(b"0123456789", 10, 3, 2, [], id="offset_bigger_than_file"),
```
kissgyorgy, line 174: Shouldn't this be a ValueError too?

unblob/file_utils.py
```diff
@@ -65,6 +66,35 @@ def find_first(
         bytes_searched += chunk_size - compensation
 
 
+def iterate_file(
+    file: io.BufferedIOBase,
+    start_offset: int,
+    size: int,
+    # default buffer size in shutil for unix based systems
+    buffer_size: int = 64 * 1024,
```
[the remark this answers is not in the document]
kukovecz: Yeah, it would be a cool idea.

It has a  value we would use:
```python
COPY_BUFSIZE = 1024 * 1024 if _WINDOWS else 64 * 1024
```

Sadly, I just realized `shutil` does not offers it, just use it internally. (so I can not use `shutil.COPY_BUFSIZE`).

Should it remain as it is? Should I add this line and write a comment that it is from `shutil`?....

tests/test_file_utils.py
kukovecz: Technically, it is possible to `seek` a bigger offset and do a `read` - it would just return empty bytes, which would stop the file iteration. So I thought it would be reasonable to keep this ("default") behavior.

kissgyorgy: Wow:
```python
>>> from pathlib import Path
>>> path.stat().st_size
744
>>> fp = path.open("rb")
>>> fp.seek(1000)
1000
>>> fp.tell()
1000
>>> fp.read()
b''
```
Ok.

unblob/file_utils.py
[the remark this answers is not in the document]
kukovecz: We discussed it offline and I created a separate commit for it - it has some `pyright` problems, but we can use that import.

kissgyorgy (approved):

PR #80, opened 2021-12-02
Fix missing FAT32 and DMG integration tests

QKaiser:
This pull request adds integration test files for the following formats:
- FAT 32
- DMG
- DMG (compressed)

DMG files were generated on OSX using these commands:

```
hdiutil create /tmp/fruits.dmg -ov -volname "fruits" -fs HFS+ -srcfolder "/tmp/fruits/"
hdiutil convert /tmp/fruits.dmg -format UDZO -o fruits.compressed.dmg
```

FAT 32 file was generated on Linux using these commands:

```
mkfs.vfat -F32 -n FRUITS fat32.img
```


QKaiser:
@kissgyorgy tests are failing because empty directories were not pushed (`.Trashes`, `.HFS+ Private Directory Data`), should I put a `.gitkeep` file in these directories ? I think they may be caught by `diff -r`.

kissgyorgy (approved):

PR #84, opened 2021-12-02
Add support for bzip2 compression format.

QKaiser:
We took inspiration from bzip2recover implementation to write the function in charge of finding the chunk end offset.

Need to verify if and how we should mention the original libbzip2 license (simple copyright, no custom licensing).

bzip2 can only act on files so in order to generate content, I simply used this Lorem Ipsum online service:

```
curl http://metaphorpsum.com/sentences/100 > lorem.txt
```

### DIFF
--- a/unblob/file_utils.py
+++ b/unblob/file_utils.py
@@ -16,6 +16,19 @@ class Endian(enum.Enum):
     BIG = ">"
 
 
+def bits(file: io.BufferedIOBase) -> Iterator[int]:
+    """bit-wise reading of file in little-endian mode"""
+    try:
+        cur_bytes = file.read(8)
+        while cur_bytes:
+            for b in cur_bytes:
+                for i in range(7, -1, -1):
+                    yield (b >> i) & 1
+            cur_bytes = file.read(8)
+    except EOFError:
+        yield 2
+
+
 def snull(content: bytes):
     """Strip null bytes from the end of the string."""
     return content.rstrip(b"\x00")
--- a/unblob/file_utils.py
+++ b/unblob/file_utils.py
@@ -10,6 +10,8 @@ from dissect.cstruct import cstruct
 
 from .logging import format_hex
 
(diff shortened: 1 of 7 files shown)


unblob/file_utils.py
```diff
@@ -16,6 +16,19 @@ class Endian(enum.Enum):
...
+def bits(file: io.BufferedIOBase) -> Iterator[int]:
+    """bit-wise reading of file in little-endian mode"""
+    try:
+        cur_bytes = file.read(8)
+        while cur_bytes:
+            for b in cur_bytes:
+                for i in range(7, -1, -1):
+                    yield (b >> i) & 1
+            cur_bytes = file.read(8)
+    except EOFError:
```
vlaci, line 28: File read won't raise EOFError ever, just returns an empty string if reached the end of file

vlaci (commented):
I have to better understand this `bzip2_recover` thingy so I am not finished as of now :)

unblob/handlers/compression/bzip2.py
```diff
@@ -0,0 +1,128 @@
...
+                            block_end=block_end[curr_block],
+                        )
+                else:
+                    curr_block -= 1
+                break
+
+            buff_hi = (buff_hi << 1) | (buff_lo >> 31)
+            buff_hi = buff_hi & 0xFFFFFFFF
+            buff_lo = (buff_lo << 1) | (b & 1)
+            buff_lo = buff_lo & 0xFFFFFFFF
```
vlaci, line 92: This whole lo-hi stuff can be significantly improved as Python's int is arbitrary precision so holding 6 bytes there shouldn't be an issue.
```suggestion
            buff = (buff << 1 | b) & 0xFFFF_FFFF_FFFF 
```
Having a single buffer would reduce clutter here significantly. On the other hand I'd like to look into how this code performs to see if it is even feasible to implement it in Python

vlaci (approved):

PR #87, opened 2021-12-02
Fail unblob with non-0 exit status when there were errors

### LINKED ISSUE #48: Fail unblob with non-0 exit status when there were errors
We are skipping some errors happened during the run, but we should keep it and exit at the end with non-0.
For example when we couldn't extract an archive, `unblob` process itself should exit with an error code.

kukovecz:
Fixes #48

### DIFF
--- a/unblob/cli.py
+++ b/unblob/cli.py
@@ -43,15 +43,20 @@ def cli(files: Tuple[Path], extract_root: Path, depth: int, verbose: bool):
 
 def main():
     try:
-        ctx = cli.make_context("unblob", sys.argv[1:])
-    except click.ClickException as e:
-        e.show()
-        sys.exit(e.exit_code)
-    except click.exceptions.Exit as e:
-        sys.exit(e.exit_code)
+        try:
+            ctx = cli.make_context("unblob", sys.argv[1:])
+        except click.ClickException as e:
+            e.show()
+            sys.exit(e.exit_code)
+        except click.exceptions.Exit as e:
+            sys.exit(e.exit_code)
 
-    with ctx:
-        cli.invoke(ctx)
+        with ctx:
+            cli.invoke(ctx)
+    except SystemExit as e:
+        sys.exit(e.code)
+    except Exception:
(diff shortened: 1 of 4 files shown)


unblob/cli.py
```diff
@@ -54,5 +55,21 @@ def main():
...
+    except SystemExit as e:
+        EXIT_CODE.set(e.code)
+    except Exception:
+        EXIT_CODE.set(1)
+        logger.exception("Unhandled exception during unblob")
+    finally:
+        default_exit_code = 0
+        exit_code = EXIT_CODE.get(default_exit_code)
+        logger_func = logger.info if exit_code == default_exit_code else logger.error
+        logger_func(f"Exiting with code: {exit_code}")
```
kissgyorgy, line 70: We don't need to log this at all. CI will fail, Docker container will show the exit status, shell sets the `$?` variable, so it's pretty much redundant. The code in this except block would be very simple if you don't do this.

unblob/state.py
```diff
@@ -0,0 +1,3 @@
+from contextvars import ContextVar
+
+EXIT_CODE: ContextVar[int] = ContextVar("EXIT_CODE")
```
kissgyorgy, line 3: This is not a constant value at all, so should be `exit_code_var`

unblob/cli.py
```diff
@@ -54,5 +55,21 @@ def main():
         cli.invoke(ctx)
 
 
+def main():
```
kissgyorgy, line 58: I don't see the point of having two "main" functions. Just wrap the `cli.invoke` with try-except above. It would be much more linear.

kissgyorgy (changes requested):

unblob/cli.py
kukovecz: I made the changes, but I wrapped everything, not just the `cli.invoke`. The whole point is to be able to handle any possible exception raised - even in `cli.make_context`, which is currently only handling specific exceptions.

kissgyorgy (approved):

PR #89, opened 2021-12-03
Add support for LZO compression format.

QKaiser:
The LZO file format holds a compressed size value in its header that we can use to calculate the right end offset.

lzop utilities are a requirement to decompress carved out chunks, we added it to Github Action tests.

To compress a file with lzop with highest level of compression (9):

```
lzop -9 -olorem.lzo lorem.txt
```

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -15,7 +15,7 @@ jobs:
 
       - name: Install 3rd party from apt
         run: |
-            sudo apt install unar zlib1g-dev liblzo2-dev
+            sudo apt install unar zlib1g-dev liblzo2-dev lzop
 
       - name: Install dependencies
         run: |
--- a/unblob/handlers/compression/lzo.py
+++ b/unblob/handlers/compression/lzo.py
@@ -1,0 +1,104 @@
+import io
+from typing import List, Optional
+
+from structlog import get_logger
+
+from ...models import StructHandler, ValidChunk
+
+logger = get_logger()
+
+
+F_H_FILTER = 0x00000800
+
+
+class LZOHandler(StructHandler):
+    NAME = "lzo"
+
+    YARA_RULE = r"""
+        strings:
+            $lzo_magic = { 89 4C 5A 4F 00 0D 0A 1A 0A }
+        condition:
+            $lzo_magic
+    """
+
+    C_DEFINITIONS = r"""
+        struct lzo_header_no_filter
+        {
+            char magic[9];
+            uint16 version;
+            uint16 libversion;
+            uint16 reqversion;
+            uint8 method;
+            uint8 level;
+            uint32 flags;
+            //uint32 filter;              // only if flags & F_H_FILTER
+            uint32 mode;
+            uint32 mtime;
+            uint32 gmtdiff;
+            uint8 filename_len;
+            /** The filename length is variable and set by filename_len,
+            so we don't know what's the exact filename char array length
+            at parsing time. Filename parsing is handled in calculate_chunk */
+            //char filename[];
+        }
+
+        struct lzo_header_filter
+        {
+            char magic[9];
+            uint16 version;
+            uint16 libversion;
+            uint16 reqversion;
+            uint8 method;
+            uint8 level;
+            uint32 flags;
+            uint32 filter;              // only if flags & F_H_FILTER
+            uint32 mode;
+            uint32 mtime;
+            uint32 gmtdiff;
+            uint8 filename_len;
+            /** The filename length is variable and set by filename_len,
+            so we don't know what's the exact filename char array length
+            at parsing time. Filename parsing is handled in calculate_chunk */
+            //char filename[];
+        }
+
+        struct lzo_size_crc {
+            uint32 original_crc;        // (CRC32 if flags & F_H_CRC32 else Adler32)
+            uint32 uncompressed_size;
+            uint32 compressed_size;
+            uint32 uncompressed_crc;
+            uint32 compressed_crc;      // (only if flags & F_ADLER32_C or flags & F_CRC32_C)
+        }
+    """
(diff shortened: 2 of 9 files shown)


unblob/handlers/compression/lzo.py
```diff
@@ -0,0 +1,98 @@
...
+    HEADER_STRUCT = "lzo_header"
+
+    def calculate_chunk(
+        self, file: io.BufferedIOBase, start_offset: int
+    ) -> Optional[ValidChunk]:
+
+        header = self.cparser_be.lzo_header_no_filter(file)
+
+        if header.flags & F_H_FILTER:
+            file.seek(start_offset)
```
kukovecz, line 76: Why do we need this `seek` here? Does the above header calculation mess with the file pointer?

unblob/handlers/compression/lzo.py
```diff
@@ -0,0 +1,98 @@
...
+            header = self.cparser_be.lzo_header_filter(file)
+
+        filename = file.read(header.filename_len)
+        logger.debug("LZO header parsed", header=header, filename=filename)
+
+        size_crc_header = self.cparser_be.lzo_size_crc(file)
+        logger.debug("CRC header parsed", header=size_crc_header)
```
kukovecz, line 83: Until now, we did not call the `self.cparser_be` directly, as far as I can tell that's what `self.parse_header` is for. I can see why can't that be used in these cases for LZO... but what do you think is there a room to refactor here?

Would not be better to call it like this?
```python
size_crc_header = self.parse_header(file=file, endian=Endian.BIG, struct_name="lzo_size_crc")
```

kukovecz (commented):
I added a few comments here, usually just to satisfy my curiosity - I am just trying to understand the `cstruct` parsing logic :sweat_smile:

unblob/handlers/compression/lzo.py
kissgyorgy: No, the current implementation is better. We are using a reference to an object method. What you are proposing is called "string typing", when you are calling methods by strings, which weakens the strong type requirements of the codebase.

unblob/handlers/compression/lzo.py
kissgyorgy: The cparser moves the file pointer when using any of the methods on it.

.github/workflows/tests.yml
```diff
@@ -15,7 +15,7 @@ jobs:
 
       - name: Install 3rd party from apt
         run: |
-            sudo apt install unar zlib1g-dev liblzo2-dev
+            sudo apt install unar zlib1g-dev liblzo2-dev lzop
```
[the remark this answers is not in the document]
kissgyorgy: It doesn't matter, does it? I would not add any of the non-required dependencies to checks, it just takes longer to set them up. `zlib1g-dev` `liblzo2-dev` were needed only because pyright couldn't find dependencies.
Anyway, we should solve this in a different issue/PR., not here.

unblob/handlers/compression/lzo.py
vlaci: Is there any IDE help for these dynamic fields like  `lzo_size_crc` BTW? It seems like `__getattribute__` magic to me which are strings too just more hidden.

unblob/handlers/compression/lzo.py
```diff
@@ -0,0 +1,98 @@
...
+            uint16 reqversion;
+            uint8 method;
+            uint8 level;
+            uint32 flags;
+            //uint32 filter;              // only if flags & F_H_FILTER
+            uint32 mode;
+            uint32 mtime;
+            uint32 gmtdiff;
+            uint8 filename_len;
+            //char filename[];
```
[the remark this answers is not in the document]
QKaiser: I kept it here for documentation but it would be better to have a single line comment explaining why. The filename length is variable and set by `filename_len`, so we don't know what's the exact filename char array length at parsing time. That's why it is handled dynamically at https://github.com/IoT-Inspector/unblob/pull/89/files#diff-9c572eb32097246fa6c46bd103c604618f9c899900cc72215f40d1e486511aeeR79

unblob/handlers/compression/lzo.py
```diff
@@ -0,0 +1,98 @@
...
+    """
+    HEADER_STRUCT = "lzo_header"
+
+    def calculate_chunk(
+        self, file: io.BufferedIOBase, start_offset: int
+    ) -> Optional[ValidChunk]:
+
+        header = self.cparser_be.lzo_header_no_filter(file)
+
+        if header.flags & F_H_FILTER:
```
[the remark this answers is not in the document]
QKaiser: Good catch, will check how to create such a file with lzop :)

unblob/handlers/compression/lzo.py
QKaiser: Exactly.

unblob/handlers/compression/lzo.py
QKaiser: I'm not sure about this. I thought about recommending something similar when we merged the ARJ handler (see https://github.com/IoT-Inspector/unblob/pull/46/files#diff-45cd7bc944a29cf37415ad82360ef8280dd3f11d4ca2218014281fa8e5310952R100) but in the end I don't really mind us using the `cparser_be` and `cparse_le` objects when it's called for.

I'll leave the debate up to you guys, let me know if we need to change this.

unblob/handlers/compression/lzo.py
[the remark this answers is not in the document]
QKaiser: Switched to clear inline comment.

unblob/handlers/compression/lzo.py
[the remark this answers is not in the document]
QKaiser: Done !

vlaci (approved):

PR #91, opened 2021-12-03
Improve ARC handler

### LINKED ISSUE #85: Improve ARC handling
The ARC handler fails when run against randomly generated content:

```
2021-12-02 15:25.12 [info     ] Calculating chunk for YARA match identifier=$arc_magic real_offset=0x554816 start_offset=0x554816
2021-12-02 15:25.12 [debug    ] Header parsed                  header=
00000000  1a 03 1b da e2 92 33 be  29 e5 72 b0 7b 71 fe c8   ......3.).r.{q..
00000010  80 11 57 34 68 96 3e 58  2a 9a 0a cc 26            ..W4h.>X*...&

struct heads:
- archive_marker: 0x1a
- header_type: 0x3
- name: b'\x1b\xda\xe2\x923\xbe)\xe5r\xb0{q\xfe'
- size: 0x571180c8
- date: 0x6834
- time: 0x3e96
- crc: 0x2a58
- length: 0x26cc0a9a
2021-12-02 15:25.12 [error    ] Unhandled Exception during chunk calculation 
  File "/home/quentin/unblob/unblob/strategies.py", line 50, in search_chunks_by_priority
  File "/home/quentin/unblob/unblob/handlers/archive/arc.py", line 67, in calculate_chunk
    header = self.parse_header(file)
  File "/home/quentin/unblob/unblob/models.py", line 119, in parse_header
    header = self._struct_parser.parse(self.HEADER_STRUCT, file, endian)
  File "/home/quentin/unblob/unblob/file_utils.py", line 143, in parse
    return struct_parser(file)
  File "/home/quentin/.cache/pypoetry/virtualenvs/unblob-U71Ryf-L-py3.8/lib/python3.8/site-packages/dissect/cstruct/types/base.py", line 16, in __call__
    return self.read(*args, **kwargs)
  File "/home/quentin/.cache/pypoetry/virtualenvs/unblob-U71Ryf-L-py3.8/lib/python3.8/site-packages/dissect/cstruct/types/base.py", line 62, in read
    return self._read(obj)
  File "<compiled heads>", line 14, in _read
EOFError
```

We should capture the EOFError early in the handler and simply not return a chunk.

QKaiser:
Fix #85 by catching errors triggered when we seek past the file.

### DIFF
--- a/unblob/handlers/archive/arc.py
+++ b/unblob/handlers/archive/arc.py
@@ -36,7 +36,7 @@ class ARCHandler(StructHandler):
         int8 archive_marker;
         int8 header_type;
         char    name[13];		/* file name */
-        long size;		/* size of file, in bytes */
+        ulong size;		/* size of file, in bytes */
         ushort date;	/* creation date */
         ushort time;	/* creation time */
         short crc;	/* cyclic redundancy check */
@@ -54,11 +54,7 @@ class ARCHandler(StructHandler):
         offset = start_offset
         while True:
             file.seek(offset)
-            try:
-                read_bytes = file.read(2)
-            except EOFError:
-                logger.warning("Potential ARC header is missing")
-                return
+            read_bytes = file.read(2)
 
             if read_bytes == END_HEADER:
                 offset += 2
--- a/unblob/finder.py
+++ b/unblob/finder.py
@@ -67,6 +67,13 @@ def search_chunks_by_priority(  # noqa: C901
                 limited_reader = LimitedStartReader(file, real_offset)
                 try:
                     chunk = handler.calculate_chunk(limited_reader, real_offset)
+                except EOFError as exc:
+                    logger.debug(
+                        "File ends before header could be read",
+                        exc_info=exc,
+                        handler=handler.NAME,
+                    )
+                    continue
                 except Exception as exc:
                     exit_code_var.set(1)
                     logger.error(
--- a/unblob/handlers/archive/arj.py
+++ b/unblob/handlers/archive/arj.py
@@ -148,11 +148,6 @@ class ARJHandler(StructHandler):
         except ARJError as exc:
             logger.warning("Invalid ARJ file", reason=exc.__doc__)
             return
-        except EOFError:
-            logger.warning(
-                "Invalid ARJ file", reason="File ends before ARJ file resolves."
-            )
-            return
 
         return ValidChunk(
             start_offset=start_offset,


unblob/handlers/archive/arc.py
```diff
@@ -50,22 +50,26 @@ def calculate_chunk(
...
+
+                if read_bytes == END_HEADER:
+                    offset += 2
+                    break
+                file.seek(offset)
+                header = self.parse_header(file)
+                offset += len(header) + header.size
+
+        except (EOFError, OSError):
+            # OSError triggered when seeking past the EOF
```
kukovecz, line 68: Can you show an example? I believe this is not true. We just talked about this with @kissgyorgy, that `seek` does not trigger any error in this case.

His example to prove:
```python
>>> from pathlib import Path
>>> path = Path("pyproject.toml")
>>> path.stat().st_size
744
>>> fp = path.open("rb")
>>> fp.seek(1000)
1000
>>> fp.tell()
1000
>>> fp.read()
b''
```

vlaci: >       lseek() allows the file offset to be set beyond the end of the
>       file (but this does not change the size of the file).

QKaiser:
Fixed and rebased !

unblob/handlers/archive/arc.py
```diff
@@ -50,22 +50,26 @@ def calculate_chunk(
...
+
+                if read_bytes == END_HEADER:
+                    offset += 2
+                    break
+                file.seek(offset)
+                header = self.parse_header(file)
+                offset += len(header) + header.size
+
+        except (EOFError):
+            # OSError triggered when seeking past the EOF
```
vlaci, line 68: On a related note: what throws the `EOFError`? You should always be able to read from an open file. After some digging I found that cstruct can raise it. I'd adjust this comment accordingly
```suggestion
        except EOFError:
            # EOFError triggered when file ends during header parsing
```

vlaci (changes requested):

unblob/handlers/archive/arc.py
kissgyorgy: If you put the try: except: block around the self.parse_header call, you don't need the comment. EOFError can't be raised from other parts of this code.

unblob/handlers/archive/arc.py
QKaiser: Finally found what triggered the OSError: negative seeking !

```
>>> with open("/tmp/garbage.bin", "rb") as f:
...     f.seek(-1)
... 
  File "<stdin>", line 2, in <module>
```

QKaiser: Fixed this by switching the ARC header size from `long `to `ulong`. Having a negative size is not right.

unblob/handlers/archive/arc.py
QKaiser: Got rid of the comment and placed the try/except block around `parse_header` :)

unblob/handlers/archive/arc.py
```diff
@@ -54,17 +54,19 @@ def calculate_chunk(
...
+            read_bytes = file.read(2)
 
             if read_bytes == END_HEADER:
                 offset += 2
                 break
             file.seek(offset)
-            header = self.parse_header(file)
+            try:
+                header = self.parse_header(file)
+            except (EOFError):
```
vlaci, line 65:
```suggestion
            except EOFError:
```

QKaiser: Done.

kissgyorgy (approved):
We fixed the comments in pair programming, the whole team was there.

PR #98, opened 2021-12-07
Refactor processing, get rid of strategies.py

kissgyorgy:
We decided that we don't want to implement the strategy abstraction right now.
Moved search chunk related functionality into finder.py and deleted strategies.py and more small refacts.

Added unit tests and bumped test coverage requirement.


unblob/finder.py
```diff
@@ -19,8 +28,68 @@
...
+        yara_results = search_yara_patterns(yara_rules, handler_map, path)
+
+        for result in yara_results:
+            handler, match = result.handler, result.match
+
+            sorted_matches = sorted(match.strings, key=itemgetter(0))
+            for offset, identifier, string_data in sorted_matches:
+                real_offset = offset + handler.YARA_MATCH_OFFSET
+
+                if any(chunk.contains_offset(real_offset) for chunk in all_chunks):
```
vlaci, line 54: What does this mean? What is the decision we are making here. In other words: what does the loop's body do below this line?

unblob/finder.py
```diff
@@ -19,8 +28,68 @@
...
+                    start_offset=offset,
+                    real_offset=real_offset,
+                    identifier=identifier,
+                )
+
+                limited_reader = LimitedStartReader(file, real_offset)
+                try:
+                    chunk = handler.calculate_chunk(limited_reader, real_offset)
+                except Exception as exc:
+                    exit_code_var.set(1)
```
vlaci, line 68: So this function should only be called from a CLI? It is very hard to use it from a library by masking errors. A possible way forward is to encapsulate errors into the chunks list (with a no-op or exception raising handler perhaps?)

vlaci: This can be considered out of scope for now oc

unblob/finder.py
```diff
@@ -19,8 +28,67 @@
...
+    Collect all the registered handlers by priority, search for YARA patterns and run
+    Handler.calculate_chunk() on the found matches.
+    We don't deal with offset within already found ValidChunks and invalid chunks are thrown away.
+    """
+    all_chunks = []
+
+    for priority_level, handlers in enumerate(_ALL_MODULES_BY_PRIORITY, start=1):
+        logger.info("Starting priority level", priority_level=noformat(priority_level))
+        yara_rules = make_yara_rules(tuple(handlers.values()))
+        yara_results = search_yara_patterns(yara_rules, handlers, path)
```
vlaci: Just so I won't forget: We make multiple passes over an input file in the hope that the first priority level would match most of the time. This assumption only works in the first few layers, as we are recursing into the contents eventually the opposite will happen. It would make more sense to me to aggregate all rules, do one paths and process the results according to priority.

vlaci (commented):
I have left a few general comments on the code to bring up points where potential further refactoring is due

unblob/finder.py
kissgyorgy: Added a comment about this.

unblob/finder.py
kissgyorgy: I made a separate issue for this: https://github.com/IoT-Inspector/unblob/issues/101

unblob/finder.py
kissgyorgy: I added this case for the performance measurements:
https://github.com/IoT-Inspector/unblob/issues/100

vlaci (approved):

PR #99, opened 2021-12-07
Add GZIP handler

tests00:
This commit adds the GZIP handler.

It also adds the `find_first_backwards()` helper function, which is a lot like `find_first()`, except backwards.

There are a lot of comments in the handler to forewarn people from the future who might suspect that they can do something more clever or efficient.

### DIFF
--- a/unblob/handlers/compression/gzip.py
+++ b/unblob/handlers/compression/gzip.py
@@ -60,7 +60,10 @@ class GZIPHandler(StructHandler):
 
             with gzip.open(file) as g:
                 g.seek(0, os.SEEK_END)
+            
             end_offset = file.tell()
+
+            gzip_size = end_offset - start_offset
 
         except gzip.BadGzipFile as gze:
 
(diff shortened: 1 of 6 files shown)


unblob/handlers/compression/gzip.py
```diff
@@ -0,0 +1,235 @@
...
+    def calculate_chunk(
+        self, file: io.BufferedIOBase, start_offset: int
+    ) -> Optional[ValidChunk]:
+
+        try:
+            # If this is just a valid gzip file, we should be able to just read it into a
+            # gzip.py GzipFile, and just get the end offset like that. Easy!
+
+            with gzip.open(file) as g:
+                g.seek(0, os.SEEK_END)
```
kissgyorgy, line 62: seek is using `self.read()` until the file is exhausted. If we would use `_GzipReader.read()` directly, we can stop at the proper offset, because `_read_gzip_header` will raise the `gzip.BadGzipFile` and we can stop there.

Calculating the correct offset can be copied from `_compression.DecompressReader.seek()` method, it's doing some adjustments and returning the correct offset.

PR #105, opened 2021-12-08
Add support for LZMA compression format.

QKaiser:
We take advantage of Python 3's lzma built-in library (available since version 3.3) to identify the end offset of a given chunk.

LZMA file format standard: https://svn.python.org/projects/external/xz-5.0.3/doc/lzma-file-format.txt

Two test files are provided to action every branch in the `calculate_chunk `function:
- one with an LZMA stream only
- one with an LZMA stream and garbage at the end

These files were built this way:

```
curl http://metaphorpsum.com/sentences/100 > lorem.txt
lzma -k -z lorem.txt
echo 'AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAa' > a.bin
cat lorem.txt.lzma a.bin > lorem.txt.unused.lzma
```


kissgyorgy (approved):
Pair programmed.

PR #109, opened 2021-12-09
Add support for LZIP compression format.

QKaiser:
LZIP handler properly detects the LZIP magic and apply a naive approach derived from the LZIP standard  (https://datatracker.ietf.org/doc/html/draft-diaz-lzip) to identify the end offset of identified chunk.

The idea is to read an 8 bytes uint64 every 2 bytes alignment (this is the alignment of LZMA stream which is the central part of any LZIP compressed file) until we read a value that corresponds to 'Member size'.

Member size is the 'total size of the member, including header and trailer.', so when we read this value it _must_ correspond to our position in the file, minus the start offset.

We use `lziprecover` as extraction binary because it's the only lzip extractor that supports and explicit user controlled output file.

Test files can be created with lzip:

```
lzip yourfile.txt
file yourfile.txt.lz
yourfile.txt.lz: lzip compressed data, version: 1
```

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -15,7 +15,7 @@ jobs:
 
       - name: Install 3rd party from apt
         run: |
-            sudo apt install unar zlib1g-dev liblzo2-dev lzop
+            sudo apt install unar zlib1g-dev liblzo2-dev lzop lziprecover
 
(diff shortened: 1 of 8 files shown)


kissgyorgy (approved):
We did it in pair programming.

PR #119, opened 2021-12-15
Add --show-external-dependencies option (and remove poetry from Docker container)

### LINKED ISSUE #113: Show external dependencies
We need a function which can list all the required third party dependencies which needed to extract all file types we support.

- [x] `--show-external-dependencies` list all required scripts with new lines (maybe suggest Ubuntu package names to install?)
  Put a checkmark or a cross next to every dependency, exit with exit code 1 if something is missing, 0 if we have every dependency.
  (eager option with a separate function in Click)
- [x] Put a comma separated list to the end of the help "You also need these commands to be able to extract the supported file types"
- [x] Run this command in the Docker build GitHub action
- [x] Put the command name in the error message when the command not found 😄

kukovecz:
Fixes #113

### DIFF
--- a/.github/workflows/publish-image.yml
+++ b/.github/workflows/publish-image.yml
@@ -44,6 +44,9 @@ jobs:
       - name: Check unblob - help
         run: docker run --rm ${{ env.DOCKER_IMAGE }} --help
 
+      - name: Check unblob - show-external-dependencies
+        run: docker run --rm ${{ env.DOCKER_IMAGE }} --show-external-dependencies
+
       - name: Check unblob - run for a file with --verbose
         run: docker run --rm -v "$(pwd)"/tests/integration/archive/zip/regular:/test ${{ env.DOCKER_IMAGE }} -v -e /tmp /test
 
--- a/.github/workflows/publish-image.yml
+++ b/.github/workflows/publish-image.yml
@@ -17,6 +17,12 @@ jobs:
       - name: Setup git lfs
         uses: ./.github/actions/setup-git-lfs
 
+      - name: Setup 3rd party dependencies
+        uses: ./.github/actions/setup-dependencies
+
+      - name: Poetry build
+        run: UNBLOB_BUILD_RUST_EXTENSION=1 poetry build --format wheel
+
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v1
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,9 @@
 FROM python:3.8-slim
 
-WORKDIR /app
 RUN mkdir -p /data/input /data/output
-RUN useradd --home-dir /app unblob
+RUN useradd unblob
 
-ENTRYPOINT ["/app/.venv/bin/unblob"]
+ENTRYPOINT ["unblob"]
 
 RUN apt-get update && apt-get install --no-install-recommends -y \
     unar \
@@ -17,16 +16,11 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
     squashfs-tools \
     p7zip-full
 
-COPY pyproject.toml /app
-COPY poetry.lock /app
-COPY unblob/ /app/unblob/
 
-# This will make the Project virtualenv in /app/.venv
-# See: https://python-poetry.org/docs/configuration/#virtualenvsin-project-boolean
-ENV POETRY_VIRTUALENVS_IN_PROJECT=true
-RUN pip install poetry
-RUN poetry install --no-dev
+# You MUST do a poetry build before to have the wheel to copy & install here (CI action will do this when building)
+COPY dist/*.whl /tmp/
+RUN pip install /tmp/unblob*.whl
 
 WORKDIR /data/output
-RUN chown -R unblob /app /data
+RUN chown -R unblob /data
 USER unblob
--- a/unblob/cli.py
+++ b/unblob/cli.py
@@ -6,6 +6,8 @@ from typing import Tuple
 import click
 from structlog import get_logger
 
+from .dependencies import get_dependencies, pretty_format_dependencies
+from .handlers import ALL_HANDLERS
 from .logging import configure_logger, noformat
 from .processing import DEFAULT_DEPTH, process_file
 from .state import exit_code_var
@@ -13,7 +15,34 @@ from .state import exit_code_var
 logger = get_logger()
 
 
-@click.command()
+def show_external_dependencies(
+    ctx: click.Context, param: click.Option, value: bool
+) -> None:
+    if not value or ctx.resilient_parsing:
+        return
+
+    dependencies = get_dependencies(ALL_HANDLERS)
+    text = pretty_format_dependencies(dependencies)
+    exit_code = 0 if all(dep.is_installed for dep in dependencies) else 1
+
+    click.echo(text)
+    ctx.exit(code=exit_code)
+
+
+def get_help_text():
+    dependencies = get_dependencies(ALL_HANDLERS)
+    lines = [
+        "A tool for getting information out of any kind of binary blob.",
+        "",
+        "You also need these extractor commands to be able to extract the supported file types:",
+        ", ".join(dep.command for dep in dependencies),
+        "",
+        "NOTE: Some older extractors might not be compatible.",
+    ]
+    return "\n".join(lines)
+
+
+@click.command(help=get_help_text())
 @click.argument(
     "files",
     nargs=-1,
@@ -35,6 +64,14 @@ logger = get_logger()
     help="Recursion depth. How deep should we extract containers.",
 )
 @click.option("-v", "--verbose", is_flag=True, help="Verbose mode, enable debug logs.")
+@click.option(
+    "--show-external-dependencies",
+    help="Shows commands needs to be available for unblob to work properly",
+    is_flag=True,
+    callback=show_external_dependencies,
+    expose_value=False,
+    is_eager=True,
+)
 def cli(files: Tuple[Path], extract_root: Path, depth: int, verbose: bool):
     configure_logger(verbose, extract_root)
     logger.info("Start processing files", count=noformat(len(files)))
--- a/unblob/dependencies.py
+++ b/unblob/dependencies.py
@@ -1,0 +1,34 @@
+import shutil
+from typing import List, Type
+
+import attr
+
+from .models import Handler
+
+
+@attr.define
+class Dependency:
+    command: str
+    is_installed: bool
+
+
+INSTALLED = "✓"
+NOT_INSTALLED = "✗"
+
+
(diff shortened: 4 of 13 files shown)


.github/workflows/publish-image.yml
```diff
@@ -17,6 +17,12 @@ jobs:
       - name: Setup git lfs
         uses: ./.github/actions/setup-git-lfs
 
+      - name: Setup 3rd party dependencies
+        uses: ./.github/actions/setup-dependencies
+
+      - name: Poetry build
+        run: poetry build --format wheel
```
vlaci, line 24:
```suggestion
        run: UNBLOB_BUILD_RUST_EXTENSION=1 poetry
```

kukovecz: Ok, I am just checking this new `UNBLOB_BUILD_RUST_EXTENSION`. 

I just rebased it to main, I am adding this & doing some test run with the container publish.

kukovecz: Cool, it is okay now, I felt Like I had to try it:
https://github.com/IoT-Inspector/unblob/runs/4533288685

I had to increase the minor version in the base image of the Dockerfile to work:
https://github.com/IoT-Inspector/unblob/pull/119/commits/67161481730b25adbdf449a8e25fc2868c1ed97a

unblob/dependencies.py
```diff
@@ -0,0 +1,57 @@
...
+
+def _get_dependency_map() -> Dict[str, bool]:
+    return {cmd: _is_installed(cmd) for cmd in get_dependencies()}
+
+
+def _pretty_format_dependencies() -> Tuple[int, str]:
+    return_code = 0
+    dependency_map = _get_dependency_map()
+    longest_legth = len(max(dependency_map.keys(), key=len))
+    text = "The following executables found installed, which are needed by unblob:"
```
kissgyorgy, line 36: Python text concatenation 1x1: put the lines in a list and join them with new line. It's a more idiomatic way. I think I learned it from here: https://waymoot.org/home/python_string/ but there are many newer articles why it's better:
https://dock2learn.com/tech/how-to-efficiently-concatenate-strings-in-python/
https://blog.ganssle.io/articles/2019/11/string-concat.html
https://betterprogramming.pub/python-and-string-concatenation-31772a10fed
https://grski.pl/fstrings-performance.html

In this specific case, the speed wouldn't matter at all, but I find still easier to read `"\n".join(lines)` than figuring out in my head where and when to put new line characters.

unblob/cli.py
```diff
@@ -6,13 +6,25 @@
...
 from .logging import configure_logger, noformat
 from .processing import DEFAULT_DEPTH, process_file
 from .state import exit_code_var
 
 logger = get_logger()
 
 
+def override_cli_help_text(f):
+    commands = get_dependencies()
+    f.__doc__ = """A tool for getting information out of any kind of binary blob.
```
kissgyorgy, line 19: `click.command()` has a help kwarg which you can use instead of this `__doc__` mambo-jambo.

tests/test_dependencies.py
```diff
@@ -0,0 +1,57 @@
...
+            ("some_random_command_surely_not_found", False),
+            ("which", True),
+        ),
+    )
+    def test_is_installed(self, command, expected):
+        assert expected == dependencies._is_installed(command)
+
+    @pytest.mark.parametrize("found", (True, False))
+    def test_get_dependency_map(self, monkeypatch, found):
+        monkeypatch.setattr(dependencies, "_is_installed", lambda command: found)
```
kissgyorgy, line 30: We should talk about testing 1x1. Tests should be written for public API only. That makes refactorings very easy.
Sometimes we can break that if we have a very complex function, but this is not the case here.

tests/test_dependencies.py
```diff
@@ -0,0 +1,57 @@
...
+
+    @pytest.mark.parametrize(
+        "installed, exit_code, mark",
+        (
+            (True, 0, "✓"),
+            (False, 1, "✗"),
+        ),
+    )
+    def test_pretty_format_dependencies(self, monkeypatch, installed, exit_code, mark):
+        monkeypatch.setattr(dependencies, "_is_installed", lambda command: installed)
```
kissgyorgy, line 46: same, don't test internal API. Test functions *throughout* public APIs.

tests/test_dependencies.py
```diff
@@ -0,0 +1,57 @@
...
+        assert expected == dependencies._is_installed(command)
+
+    @pytest.mark.parametrize("found", (True, False))
+    def test_get_dependency_map(self, monkeypatch, found):
+        monkeypatch.setattr(dependencies, "_is_installed", lambda command: found)
+        result = dependencies._get_dependency_map()
+        assert result == {name: found for name in self.DEPENDENCIES}
+
+    def test_pretty_print_marks(self):
+        assert dependencies.INSTALLED == "✓"
```
kissgyorgy, line 35: Why? Just put that unicode character into the variable.

kissgyorgy (changes requested):

kissgyorgy:
There are a lot of comments here, but my biggest problem is that you wrote silly tests, but did not test important functions and suppressed coverage for those. It should be the other way around!

Dockerfile
```diff
@@ -17,15 +17,11 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
...
-COPY pyproject.toml /app
-COPY poetry.lock /app
-COPY unblob/ /app/unblob/
 
-# This will make the Project virtualenv in /app/.venv
-# See: https://python-poetry.org/docs/configuration/#virtualenvsin-project-boolean
-ENV POETRY_VIRTUALENVS_IN_PROJECT=true
-RUN pip install poetry
-RUN poetry install --no-dev
+COPY unblob/ /app/unblob/
```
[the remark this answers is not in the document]
kukovecz: YUP, this somehow remained here, thanks!

unblob/dependencies.py
```diff
@@ -0,0 +1,57 @@
...
+        else:
+            mark = NOT_INSTALLED
+            return_code = 1
+
+        text += f"\n    {command:<{longest_legth}}    {mark}"
+
+    return return_code, text
+
+
+def show_external_dependencies(
```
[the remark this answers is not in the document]
kukovecz: Totally right. I am doing a refactor on this dependencies stuff all-together .... I believe I will keep the `dependencies˙ module, but this function will be in the `cli.py`. Really, `click` should not be imported from anywhere else. But in this manner, it is also true that the `cli.py` should not have any side-calculation, as the cli-related code should be separated.

unblob/dependencies.py
```diff
@@ -0,0 +1,57 @@
...
+def _is_installed(command: str) -> bool:
+    return bool(which(command))
+
+
+def _get_dependency_map() -> Dict[str, bool]:
+    return {cmd: _is_installed(cmd) for cmd in get_dependencies()}
+
+
+def _pretty_format_dependencies() -> Tuple[int, str]:
+    return_code = 0
```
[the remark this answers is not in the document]
kukovecz: I would do both of these steps here, because here we decide which `mark` should we print, based on if the dependency installed or not. Since we have this knowledge here, we can return with it and not make the caller call an other function to decide on the exit code the very same way this function does.

Maybe the naming is not right, I will try to rewrite this a little bit, but I can't see a good logic to slip this up. After I push a new version, we check that if it satisfy your requirements or not.

Dockerfile
```diff
@@ -17,15 +17,11 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
...
+# You MUST do a poetry build before to have the wheel to copy & install here (CI action will do this when building)
```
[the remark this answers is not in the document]
kukovecz: Right now, we do the pip/poetry install (using cache) in the github action, then can easily execute a `poetry build`.

I would not do multi staged Dockerfile, installing poetry just for the sake of doing a `poetry build` in the docker container, and copy the files from there.

unblob/dependencies.py
```diff
@@ -0,0 +1,57 @@
...
+            return_code = 1
+
+        text += f"\n    {command:<{longest_legth}}    {mark}"
+
+    return return_code, text
+
+
+def show_external_dependencies(
+    ctx: Context, param: Option, value: bool
+) -> None:  # pragma: no cover
```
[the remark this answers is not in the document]
kukovecz: I created a separate Issue about this, to test the CLI interface: #135

unblob/cli.py
```diff
@@ -6,14 +6,48 @@
...
+
+def get_help_text():
+    dependencies = get_dependencies(ALL_HANDLER_CLASS)
+    lines = [
+        "A tool for getting information out of any kind of binary blob.",
+        "",
+        "You also need these extractor commands to be able to extract the supported file types:",
+        ", ".join(dependencies),
+        "",
+        "NOTE: The extract extractor commands can be seen in the handlers, for version compatibility.",
```
kukovecz, line 45: I added this note as we talked about this, but I created a separate issue to solve the versions: #134

unblob/handlers/__init__.py
```diff
@@ -44,3 +44,7 @@
         xz.XZHandler,
     ),
 ]
+
+ALL_HANDLER_CLASSES = [
```
kissgyorgy, line 48: The name of this variable is inconsistent with the variable it is using. You can simply name it `ALL_HANDLERS`

unblob/dependencies.py
```diff
@@ -0,0 +1,24 @@
...
+from typing import List, Type
+
+from .models import Handler
+
+INSTALLED = "✓"
+NOT_INSTALLED = "✗"
+
+
+def is_installed(command: str) -> bool:
+    return bool(shutil.which(command))
```
kissgyorgy, line 11: From the documentation: 
> Return the path to an executable which would be run if the given cmd was called. If no cmd would be called, return None.

It would be more precise to use the `None` return value telling the absence, but also, this function is so simple, you don't need a function for it. I would just inline this as a variable. It is only used once.

tests/test_dependencies.py
```diff
@@ -0,0 +1,87 @@
...
+class TestDependencies:
+    @pytest.mark.parametrize(
+        "command, expected",
+        (
+            ("some_random_command_surely_not_found", False),
+            ("sh", True),
+        ),
+    )
+    def test_is_installed(self, command, expected):
+        assert expected == dependencies.is_installed(command)
```
kissgyorgy, line 39: You are essentially testing the standard library.

kissgyorgy (changes requested):
Added some more comments after the refact.

tests/test_dependencies.py
```diff
@@ -0,0 +1,87 @@
+import pytest
```
kissgyorgy, line 1: You wrote a lot of tests, but still not testing the ONLY function which should be tested: show_external_dependencies

kissgyorgy (approved):

PR #120, opened 2021-12-15
Add support for LZH/LHA format.

QKaiser:
Handler for LZH/LHA format based on standard documented at https://web.archive.org/web/20021005080911/http://www.osirusoft.com/joejared/lzhformat.html.

The handler will create valid chunks for each LZH stream instead of concatenating sequential streams into an overall ValidChunk.

Test files were generated using jlha, which can be installed with `apt install jlha-tools` on Ubuntu.

```
jlha uf0 apple.lvl0.lzh *.txt
jlha uf0 apple.lvl1.lzh *.txt
jlha uf0 apple.lvl2.lzh *.txt
```

There is a single test file for each LZH 'level identifier' (0, 1, 2).


unblob/handlers/compression/lzh.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+class LZHHandler(StructHandler):
+
+    NAME = "lzh"
+
+    YARA_RULE = r"""
+        strings:
+            // 2 header bytes followed by method ID
+            $magic = /[\x00-\xff][\x00-\xff](-lh0-|-lzs-|-lz4-|-lh1-|-lh2-|-lh3-|-lh4-|-lh5-|-lh6-|-lh7-|-lh8-|-lhd-)/
```
kukovecz, line 36: 2 comments here for this line.

1.  rename the `$magic` (here and below) to `$lhz_magic` or something. 
This is just to honor [this pull request](https://github.com/IoT-Inspector/unblob/pull/117), I think this make sense to have them in separate names. 

2. I found it a little bit awkward to have the same magic ids defined here as we have stored in the `METHOD_IDS` variable. We can make it work with this change for the YARA rule, so the IDs are filled dynamically:

```python
    YARA_RULE = r"""
        strings:
            // 2 header bytes followed by method ID
            $lhz_magic = /[\x00-\xff][\x00-\xff]({magic_ids})/
        condition:
            $lht_magic
    """.format(magic_ids="|".join([char.decode() for char in METHOD_IDS]))
```
But **I do not want to force this solution**. If you think the duplication is fine for readability (my version is not too reader-friendly), then just leave it!

unblob/handlers/compression/lzh.py
```diff
@@ -0,0 +1,102 @@
...
+        };
+    """
+    HEADER_STRUCT = "lzh_default_header"
+
+    def calculate_chunk(
+        self, file: io.BufferedIOBase, start_offset: int
+    ) -> Optional[ValidChunk]:
+
+        header = self.parse_header(file, Endian.LITTLE)
+        file.seek(-len(header), io.SEEK_CUR)
```
kukovecz, line 70: This `seek` is needed only for the `read` like 10-ish lines below. Please put this line right before the `read` - in that way it will be easier to understand what is happening.

unblob/handlers/compression/lzh.py
```diff
@@ -0,0 +1,102 @@
...
+
+        if (
+            header.method_id not in METHOD_IDS
+            or header.level_identifier not in LEVEL_IDENTIFIERS
+        ):
+            return
+
+        if header.level_identifier == 0x2:
+            # magic trick, just lookup the header definition of level 2
+            header_size = header.header_size + header.header_checksum << 8
```
kukovecz, line 80: I am sure this is only issue for me, because I am soooo not used to do bitwise operations .... but I wanted to make this whole `if` block simpler, and could not ....  (kudos for you for creating cool integration tests!!!)

.... I just found out that the `+` operator has higher dependency than the `<<` :sweat_smile: 

So just for the sake of dumb people like me, could you use parenthesis like this?
```suggestion
            header_size = (header.header_size + header.header_checksum) << 8

```

unblob/handlers/compression/lzh.py
```diff
@@ -0,0 +1,102 @@
...
+        file.seek(-len(header), io.SEEK_CUR)
+
+        if (
+            header.method_id not in METHOD_IDS
+            or header.level_identifier not in LEVEL_IDENTIFIERS
+        ):
+            return
+
+        if header.level_identifier == 0x2:
+            # magic trick, just lookup the header definition of level 2
```
kukovecz, line 79: Is there any official header definition? Maybe the one you [linked in the description of the PR](https://web.archive.org/web/20021005080911/http://www.osirusoft.com/joejared/lzhformat.html)?
If so, you might want to add here in the comment as well since you are referring it.

unblob/handlers/compression/lzh.py
QKaiser: Ahah, nice catch ! It worked because the size was lower than `0xff` for all my test files, what I really want to do is this:

```
header_size = header.header_size + ( header.header_checksum) << 8)
```

The idea is that for level 0 and level 1, the header size is stored as a `uint8 `and the header_checksum as `uint8`. However, for level2 the header size is stored as a `uint16`, and crc is absent. Instead of parsing the header again using the level 2 definition, we can use this trick. Let's say you have a level 2 header with a header size of `0xffff`, we will have a default lzh header with size `0xff` and checksum `0xff`, so we do `size = 0xff + (0xff << 8) = 0xff`

QKaiser: Fixed it and explained the magic trick with an inline comment.

unblob/handlers/compression/lzh.py
QKaiser: Explained the magic trick with an inline comment and referenced the level 2 definition I kept in `C_DEFINITIONS` for documentation purposes.

unblob/handlers/compression/lzh.py
QKaiser: Makes perfect sense. Done.

unblob/handlers/compression/lzh.py
QKaiser: I like how you think. I used your proposition with renamed yara variable 😺

unblob/handlers/compression/lzh.py
kukovecz: Oh, ok! I was just confused about the operator precedence, because of the test files. But I see the testfiles are also changed :smile:

unblob/handlers/compression/lzh.py
```diff
@@ -0,0 +1,106 @@
...
+from typing import List, Optional
+
+from structlog import get_logger
+
+from ...file_utils import Endian
+from ...models import StructHandler, ValidChunk
+
+logger = get_logger()
+
+PADDING_LEN = 2
```
kukovecz, line 11: This is not used (sorry I did not catch earlier)

unblob/handlers/compression/lzh.py
```diff
@@ -0,0 +1,106 @@
...
+    b"-lh4-",
+    b"-lh5-",
+    b"-lh6-",
+    b"-lh7-",
+    b"-lh8-",
+    b"-lhd-",
+]
+
+
+class LZHHandler(StructHandler):
```
kukovecz, line 29: Now it looks cool! I just have 1 missing piece I try to understand:

We have nice test files for all levels (0-1-2). 
I think they look completely the same (each level is one `.lzh` file containing three text files). HOWEVER, the output is different. 
* In level 0-1 case, the `LZHHandler` is running 3 different times, extracting the each text file separately 
* In level 2, it runs only once, extracting the 3 files in 1 step. This is sooo much cleaner.

Why? :sweat_smile:

unblob/handlers/compression/lzh.py
```diff
@@ -0,0 +1,106 @@
+import io
+from typing import List, Optional
+
+from structlog import get_logger
+
+from ...file_utils import Endian
+from ...models import StructHandler, ValidChunk
+
+logger = get_logger()
```
kukovecz, line 9: This is also not used. If we don't log anything, don't deal with the logging at all.

QKaiser: Removed.

unblob/handlers/compression/lzh.py
QKaiser: Now it's used.

unblob/handlers/compression/lzh.py
QKaiser: Hands down best reviewer here. You are totally right and this was due to a wrong calculation of header size when in level 2 (I was converting as big endian instead of little endian). It matched the chunk size out of pure luck. It's fixed now and we have a separate chunk for each LZH stream.

kukovecz (approved):

unblob/handlers/compression/lzh.py
```diff
@@ -0,0 +1,102 @@
...
+        if header.level_identifier == 0x2:
+            # with level 2, the header size is a uint16 rather than uint8 and there
+            # is no checksum. We use this magic trick so we don't parse the header
+            # again. See the level_2_header definition in C_DEFINITIONS
+            header_size = header.header_size + (header.header_checksum << 8)
+        else:
+            header_size = header.header_size + PADDING_LEN
+
+        file.seek(-len(header), io.SEEK_CUR)
+        file.read(header_size + header.compressed_size)
```
kissgyorgy, line 84: I'm not sure what's going on here. You don't even need to seek to calculate `end_offset` if I understand correctly. Also reading a possibly huge file into memory.

PR #123, opened 2021-12-15
Add support for Unix compress (LZW) compression format.

QKaiser:
This module implements Unix compress'ed chunk identification.

We identify the end offset of any identified unix compress'ed chunk by performing Lempel-Ziv-Welch decompression on a chunk starting from the identified start offset, and ending at the end of the whole file being analyzed.

If we reach an invalid code or the stream ends in the middle of a code, we recursively call the decompression, this time with a chunk ending at the failing position - 1.

Once the decompression procedure works without errors, that means we have a valid chunk and can return its current end offset.

We use a small heuristic to return the right end offset. This heuristic tends to work well when arbitrary data appended at the end of the stream is made of random bytes (no repeating letters, no large set of ASCII letters).

It obviously can be wrong from time to time, leading to a compress'ed chunk that we can decompress (obviously), but uncompressed data will contain garbage bytes at the end.

Sadly, there is no way we can identify with 100% probability the end offset of a compress'ed stream with byte precision if it is followed by other content.

The good news is that because of this behavior, it's highly unlikely we will observe compress'ed chunks followed by other chunks in the wild. The only ones I observed were followed by null bytes sentinels, which helps identifying the exact end offset.

🔴 we need to set up a pair programming session to work on the LZW decompressor. Current cyclomatic complexity is through the roof (18!), so we clearly need to rewrite it. On top of cyclomatic complexity, the decompressor keep tracks of decompressed bytes while we only need to know the end offset of that stream. Just pushing the branch here so it's out there and can work on something else :)

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -19,3 +19,19 @@ AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
 LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.
+
+--
+
+Logic for Unix Compress (unblob/handlers/compression/compress.py) is
+ported from the unlzw project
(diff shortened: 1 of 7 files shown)


vlaci (approved):

vlaci:
:fireworks:

Issue #124, opened 2021-12-15
Inexact seek back on error when parsing malformed AR archives


QKaiser:
So I'm starting to fuzz each formats at the moment.

A really dumb technique to fuzz the AR format:

```
#!/bin/bash 
while true; do
    printf '\x21\x3C\x61\x72\x63\x68\x3E\x0A' > /tmp/file.bin;
    dd if=/dev/random count=58 bs=1 >> /tmp/file.bin;
    printf '\x60\x0A' >> /tmp/file.bin
    dd if=/dev/random count=1 bs=1M >> /tmp/file.bin;
    poetry run unblob /tmp/file.bin;
    rm -rf file.bin*
done
```

The idea is to craft a file that will trigger a Yara match, with random bytes wherever we can put them.

The errors are well handled and do not lead to crashes, however we noticed an error in the exception handling. On error, the handler seek back of `HEADER_LENGTH` which is set to 60, but the exact size of header is actually 68.

```python
except arpy.ArchiveFormatError as exc:
    logger.debug(
        "Hit an ArchiveFormatError, we've probably hit some other kind of data",
        exc_info=exc,
    )
    # Since arpy has tried to read another file header, we need to wind the cursor back the
    # length of the header, so it points to the end of the AR chunk.
    ar.file.seek(-HEADER_LENGTH, os.SEEK_CUR)
```

This is obvious when unknown chunk is logged:

```
2021-12-15 15:57.37 [warning  ] Found unknown Chunks           chunks=[0x8-0x100044]
```

We should fix this by changing the HEADER_LENGTH value.

Issue #126, opened 2021-12-15
AR archive handler end_offset computation is wrong.


QKaiser:
The AR archive handler calculates the end offset as if the AR archive always starts at the beginning of the file being analyzed. This is wrong.

Issue #125, opened 2021-12-15
AR archive handler emits empty valid chunks on malformed AR archives


QKaiser:
The AR archive handler emits empty valid chunks with start offset 0 and end offset 0 when an exception is triggered by malformed archive.

This is due to the current implementation that emits valid chunks regardless of the exception state:

```
try:
    ar.read_all_headers()
except arpy.ArchiveFormatError as exc:
    # debug log
    ...
offset = ar.file.tell()
return ValidChunk(
    start_offset=start_offset,
    end_offset=start_offset + offset,
)
```

PR #127, opened 2021-12-15
Fix issues #124, #125, and #126 by rewriting the AR archive handler.

QKaiser:
This also adds AR archive test files created for fuzzing purposes:

- a file containing a malformed AR archive followed by a valid AR archive
- a file containing a valid AR archive followed by a malformed AR archive
- a file with two sequential AR archives
- a file with a valid AR archive starting at an offset > 0
- a file with a malformed AR archive starting at an offset > 0


QKaiser:
Fix issues #124, #125, and #126

vlaci (approved):

Issue #129, opened 2021-12-15
ARC handler emits lots of false positives due to bad handler/yara rule combination.


QKaiser:
The ARC handler matches on the following YARA rule:

```
$arc_magic = /\x1A[\x00-\x07][\S]{12}[\x00|\xf0-\xff][\x00-\xff]{4}[\x00-\x8d][\x00-\x8f][\x00-\xc7][\x00-\x9f]/
```

And then does this:

```python
offset = start_offset
while True:
    file.seek(offset)
    read_bytes = file.read(2)
    
    if read_bytes == "\x1A\x00":
        offset += 2
        break
    file.seek(offset)
    header = self.parse_header(file)
    offset += len(header) + header.size
```

I identified the issue by generating random ARC files. If the header starts with `\x1A\x00`, it will fail early on the first loop and the header object won't be set. Which ends up with us generating lots of false positive ARC chunks of length 2.

We can fix it by moving the Yara rule to this one:

```
$arc_magic = /\x1A[\x01-\x07][\S]{12}[\x00|\xf0-\xff][\x00-\xff]{4}[\x00-\x8d][\x00-\x8f][\x00-\xc7][\x00-\x9f]/
```

We won't loose anything, a second byte set to "\x00" in the ARC header means "no compression" and we should only use this "\x1A\x00" to match the archive's end. Interestingly, we stop having false positives when doing so. Magic 🌟

PR #130, opened 2021-12-15
Fix issue #129 by rewriting the ARC handler yara rule.

QKaiser:
See #129

### DIFF
--- a/unblob/handlers/archive/arc.py
+++ b/unblob/handlers/archive/arc.py
@@ -3,6 +3,8 @@ from typing import List, Optional
 
 from structlog import get_logger
 
+from unblob.file_utils import snull
+
 from ...models import StructHandler, ValidChunk
 
 logger = get_logger()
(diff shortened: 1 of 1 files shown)


QKaiser:
I added two enhancements to ARC handler that will help with #129. Moved the size to an unsigned integer (should always be positive), and added a name validation function that verify if the file name is unicode encoded.

unblob/handlers/archive/arc.py
```diff
@@ -40,12 +42,19 @@ class ARCHandler(StructHandler):
...
-        long length;	/* true file length */
+        ulong length;	/* true file length */
     };
     """
 
     HEADER_STRUCT = "heads"
 
+    def valid_name(self, name: bytes) -> bool:
+        try:
+            snull(name.decode("utf-8"))
```
vlaci, line 53: We first want to strip nulls, _then_ convert to string:
```suggestion
            snull(name).decode("utf-8")
```
This'll fix pyright and the test failure too

vlaci (changes requested):

vlaci:
Hm, a file doesn't get extracted now :(

vlaci:
From the docstring of the yara rule:

>             Then a null-byte or unitialized-byte terminated filename string of 13 bytes, the
>             uninitialized byte is always set between 0xf0 and 0xff.

This matches with the output of the test:

> struct heads:
> - archive_marker: 0x1a
> - header_type: 0x2
> - name: b'apple.txt\x00\x00\x00\xfd'
> - size: 0x6
> - date: 0x5379
> - time: 0x6cef
> - crc: 0x3038
> - length: 0x6

This will trip on the `valid_name` check.

QKaiser:
Fixed `valid_name` by moving to this:

```python
# we don't care about the terminating byte (null or unitialized)
snull(name[:-1]).decode("utf-8")
```

vlaci:
Yaay, null bytes are valid unicode code points :D

vlaci (approved):

PR #133, opened 2021-12-16
Fix #132 by making sure the computed real_offset is not negative.

### LINKED ISSUE #132: Unblob chunk finder can trigger OSError on negative seek
This is a case that came up when writing the ExtFS handler. The handler expects the Yara magic to be found at offset 0x438, so the actual start offset is the match offset - 0x438. This isn't unique to ExtFS and was implemented in unblob using the `YARA_MATCH_OFFSET` variable.

If we have a false positive match at an offset lower than the defined YARA_MATCH_OFFSET for a given handler, we end up with a negative real_offset value, which triggers an OSError when seeking negatively.

```
2021-12-16 11:04.40 [error    ] Unhandled exception during unblob 
  File "/home/quentin/unblob/unblob/cli.py", line 61, in main
    cli.invoke(ctx)
  File "/home/quentin/unblob/unblob/cli.py", line 43, in cli
  File "/home/quentin/unblob/unblob/processing.py", line 53, in process_file
    all_chunks = search_chunks_by_priority(path, file, size)
  File "/home/quentin/unblob/unblob/finder.py", line 67, in search_chunks_by_priority
    limited_reader = LimitedStartReader(file, real_offset)
  File "/home/quentin/unblob/unblob/file_utils.py", line 178, in __init__
    self._file.seek(start)
```

We should add some defense against it at https://github.com/IoT-Inspector/unblob/blob/main/unblob/finder.py#L53, something like:

```diff
diff --git a/unblob/finder.py b/unblob/finder.py
index c0749f3..9244b84 100644
--- a/unblob/finder.py
+++ b/unblob/finder.py
@@ -52,6 +52,8 @@ def search_chunks_by_priority(  # noqa: C901
             for offset, identifier, string_data in sorted_match_strings:
                 real_offset = offset + handler.YARA_MATCH_OFFSET
 
+                if real_offset < 0:
+                    continue
                 # Skip chunk calculation if this would start inside another one,
                 # similar to remove_inner_chunks, but before we even begin calculating.
                 if any(chunk.contains_offset(real_offset) for chunk in all_chunks):
```

QKaiser:
Fix #132

### DIFF
--- a/unblob/finder.py
+++ b/unblob/finder.py
@@ -52,6 +52,11 @@ def search_chunks_by_priority(  # noqa: C901
             for offset, identifier, string_data in sorted_match_strings:
                 real_offset = offset + handler.YARA_MATCH_OFFSET
 
+                # Skip chunk calculation if the match is found too early in the file,
+                # leading to a negative real offset once YARA_MATCH_OFFSET is applied.
+                if real_offset < 0:
+                    continue
+
                 # Skip chunk calculation if this would start inside another one,
                 # similar to remove_inner_chunks, but before we even begin calculating.
                 if any(chunk.contains_offset(real_offset) for chunk in all_chunks):


unblob/finder.py
```diff
@@ -52,6 +52,8 @@ def search_chunks_by_priority(  # noqa: C901
             for offset, identifier, string_data in sorted_match_strings:
                 real_offset = offset + handler.YARA_MATCH_OFFSET
 
+                if real_offset < 0:
+                    continue
```
kukovecz, line 56: Just a quick question:

The empty `continue` is enough here?
I mean, this is not a "real error" which should be logged and change the exit code of the whole `unblob` from something other than `0`, but just a "precheck" that the given `handler` should not deal with this `offset`, right?

QKaiser: Yes this is a pre-check in order to discard it if it's not valid. I put an inline comment on top of it to make things clear.

kukovecz: Thank you, just wanted to make sure.

kukovecz (approved):

PR #140, opened 2021-12-17
Remove global state handling by introducing `ErrorReport`

### LINKED ISSUE #101: Make unblob usable as a library
https://github.com/IoT-Inspector/unblob/pull/98#discussion_r763856196

- Make something with the `exit_code_var` so global state will not be stored
- Check the logging (`noformat`, ...)

kukovecz:
Fixes #101

### DIFF
--- a/tests/test_handlers.py
+++ b/tests/test_handlers.py
@@ -39,7 +39,7 @@ def test_all_handlers(input_dir: Path, output_dir: Path, tmp_path: Path):
         list(input_dir.iterdir()) != []
     ), f"Integration test input dir should contain at least 1 file: {input_dir}"
 
-    process_file(
+    all_reports = process_file(
         path=input_dir,
         extract_root=tmp_path,
         entropy_depth=0,
@@ -65,6 +65,8 @@ def test_all_handlers(input_dir: Path, output_dir: Path, tmp_path: Path):
     except subprocess.CalledProcessError as exc:
         runnable_diff_command = shlex.join(diff_command)
         pytest.fail(f"\nDiff command: {runnable_diff_command}\n{exc.stdout}\n")
+
+    assert all_reports == [], f"Unexpected error reports: {all_reports}"
 
 
 @pytest.mark.parametrize(
--- a/tests/test_pool.py
+++ b/tests/test_pool.py
@@ -1,0 +1,59 @@
+import multiprocessing
+
+import pytest
+
+from unblob.pool import MultiPool, SinglePool
+
+
+def test_singlepool():
+    results = []
+    handled = []
+
+    def _handler(i):
+        handled.append(i)
+        return i
+
+    def _callback(pool, result):
+        results.append(result)
+        if result > 0:
+            pool.submit(result - 1)
+
+    with SinglePool(handler=_handler, result_callback=_callback) as pool:
+        pool.submit(5)
+        pool.process_until_done()
+
+    assert handled == [5, 4, 3, 2, 1, 0]
+    assert results == [5, 4, 3, 2, 1, 0]
+
+
+@pytest.mark.parametrize("process_num", [-1, 0])
+def test_multipool_dummy_process_num(process_num: int):
+    def _dummy(*args):
+        pass
+
+    with pytest.raises(ValueError):
+        MultiPool(process_num=process_num, handler=_dummy, result_callback=_dummy)
+
+
+@pytest.mark.parametrize("process_num", range(1, 5))
+def test_multipool(process_num: int):
+    handled = multiprocessing.Manager().list()
+    results = []
+
+    def _handler(i):
+        handled.append(i)
+        return i
+
+    def _callback(pool, result):
+        results.append(result)
+        if result > 0:
+            pool.submit(result - 1)
+
+    with MultiPool(
+        process_num=process_num, handler=_handler, result_callback=_callback
+    ) as pool:
+        pool.submit(5)
+        pool.process_until_done()
+
+    assert list(handled) == [5, 4, 3, 2, 1, 0]
+    assert list(results) == [5, 4, 3, 2, 1, 0]
--- a/unblob/extractor.py
+++ b/unblob/extractor.py
@@ -112,17 +112,11 @@ def carve_unknown_chunks(
     return carved_paths
 
 
-def extract_valid_chunk(
-    extract_dir: Path,
-    file: io.BufferedIOBase,
-    chunk: ValidChunk,
-    task_result: TaskResult,
+def carve_valid_chunk(
+    extract_dir: Path, file: io.BufferedIOBase, chunk: ValidChunk
 ) -> Path:
     filename = f"{chunk.start_offset}-{chunk.end_offset}.{chunk.handler.NAME}"
     carve_path = extract_dir / filename
     logger.info("Extracting valid chunk", path=carve_path, chunk=chunk)
     carve_chunk_to_file(carve_path, file, chunk)
-    extracted = extract_with_command(
-        extract_dir, carve_path, chunk.handler, task_result
-    )
-    return extracted
+    return carve_path
--- a/unblob/extractor.py
+++ b/unblob/extractor.py
@@ -10,8 +10,8 @@ from typing import List
 from structlog import get_logger
 
 from .file_utils import iterate_file
-from .models import Chunk, Handler, UnknownChunk, ValidChunk
-from .state import exit_code_var
+from .models import Chunk, Handler, TaskResult, UnknownChunk, ValidChunk
+from .report import ExtractCommandFailedReport
 
 logger = get_logger()
 
@@ -47,7 +47,7 @@ def fix_permissions(outdir: Path):
 
 
 def extract_with_command(
-    extract_dir: Path, carved_path: Path, handler: Handler
+    extract_dir: Path, carved_path: Path, handler: Handler, task_result: TaskResult
 ) -> Path:
     content_dir = extract_dir / (carved_path.name + APPEND_NAME)
     # We only extract every blob once, it's a mistake to extract the same blog again
@@ -57,7 +57,8 @@ def extract_with_command(
     outdir = content_dir.expanduser().resolve()
     cmd = handler.make_extract_command(str(inpath), str(outdir))
 
-    logger.info("Running extract command", command=shlex.join(cmd))
+    command = shlex.join(cmd)
+    logger.info("Running extract command", command=command)
     try:
         res = subprocess.run(
             cmd,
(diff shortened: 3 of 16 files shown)


unblob/handlers/archive/zip.py
```diff
@@ -80,18 +80,16 @@ def _calculate_zipfile_end(self, file: io.BufferedIOBase, start_offset: int) ->
         self.cparser_le.end_of_central_directory_t(file)
         return file.tell()
 
-    def is_valid(self, file: io.BufferedIOBase) -> bool:
+    def is_valid(self, file: io.BufferedIOBase) -> Tuple[bool, bool]:
```
kissgyorgy, line 83: One function should do one thing at a time. This should be split into two: `is_valid` and `has_encrypted_files`

unblob/finder.py
```diff
@@ -98,13 +110,29 @@ def search_chunks_by_priority(  # noqa: C901
                     continue
 
                 if chunk.start_offset < 0 or chunk.end_offset < 0:
-                    exit_code_var.set(1)
-                    logger.error("Chunk has negative offset", chunk=chunk)
+                    error_text = "Chunk has negative offset"
+                    error_queue.put(
```
kissgyorgy, line 114: Naming of the same thing should be consistent throughout the codebase. `report_queue` would be the most generic IMO.

unblob/extractor.py
```diff
@@ -95,11 +110,20 @@ def carve_unknown_chunks(
 
 
 def extract_valid_chunk(
-    extract_dir: Path, file: io.BufferedIOBase, chunk: ValidChunk
-) -> Path:
+    extract_dir: Path, file: io.BufferedIOBase, chunk: ValidChunk, error_queue: Queue
```
kissgyorgy, line 113: The function name is lying. It doesn't always extract a ValidChunk, which is a code smell. Either the function is doing too much, or the name should be changed. Maybe this should be split up to 3 functions or called differently?

unblob/processing.py
```diff
@@ -149,13 +176,15 @@ def _process_task(  # noqa: C901
...
-            new_path = extract_valid_chunk(extract_dir, file, chunk)
-            task_queue.put(
-                Task(
-                    root=config.extract_root,
-                    path=new_path,
-                    depth=task.depth + 1,
-                )
+            new_path = extract_valid_chunk(extract_dir, file, chunk, result_queue)
+
+            if new_path is None:
```
kissgyorgy, line 181: This makes no sense to me. extract_valid_chunk should always extract a ValidChunk and the code should be at a different place deciding if it's indeed valid or not.

kissgyorgy (changes requested):
We were thinking about how `multiprocessing.pool.Pool` would made everything simpler by not requiring any queue to be passed around. We should finish this together with @vlaci.

unblob/handlers/archive/zip.py
kukovecz: I would not split this function. This is already going through the whole file to check if it is valid and it is encrypted, it just did not return with the encryption info.

Instead, I renamed it. Now calling it looks clear enough:


```python
    is_valid, has_encrypted_files = self.check_file(file)
```

tests/test_handlers.py
```diff
@@ -64,6 +64,19 @@ def test_all_handlers(input_dir: Path, output_dir: Path, tmp_path: Path):
         runnable_diff_command = shlex.join(diff_command)
         pytest.fail(f"\nDiff command: {runnable_diff_command}\n{exc.stdout}\n")
 
+    expected_errors = []
+
+    if str(input_dir).endswith("/filesystem/ntfs/__input__"):
```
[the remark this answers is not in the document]
kukovecz: This highlighted exception ("known error") case is removed, as we discussed this type of invalid processing format does not  mean processing error right now.

However for the error report checking, I would very much leave this here. As we run our integration tests, we want to check that no error is found during the test execution.

kukovecz:
We sat down together with @kissgyorgy  and @vlaci  to refactor the multiprocess handling in unblob and talked through / created together the reporting feature.

Here I prepared this branch for review, containing all things we done together and extending it with the original reflected changes (like avoiding encrypted chunk extraction, etc.)

Please look at the "fresh" pull request, commit by commit , should contain everything we need from this right now.

kissgyorgy (changes requested):

unblob/pool.py
```diff
@@ -0,0 +1,104 @@
...
+            self._result_callback(self, result)
+            self._input.task_done()
+
+
+class SinglePool(PoolBase):
+    def __init__(self, handler, *, result_callback):
+        self._handler = handler
+        self._result_callback = result_callback
+
+    def submit(self, args):
```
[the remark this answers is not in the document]
kukovecz: In your code, the handler and the callback would only run once, not for every submit.

But I get your point, I can do something like:
```python
    def submit(self, args):
        self._args.append(args)

    def process_until_done(self):
        for args in self._args:
            result = self._handler(args)
            self._result_callback(self, result)
```

unblob/models.py
```diff
@@ -155,3 +148,58 @@ def parse_header(self, file: io.BufferedIOBase, endian=Endian.LITTLE):
...
+
+@attr.define(kw_only=True)
+class Report:
+    """A common base class for different reports"""
+
+    # Stored in `str` rather than `Handler`, because the pickle picks ups structs from `C_DEFINITIONS`
+    handler: Optional[str] = None
+
+    def items(self) -> dict:
+        return attr.asdict(self)
```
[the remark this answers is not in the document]
kukovecz: No, I don't think so.

We want to log the Report's content in the caller side, so I did this:
```python
logger.error("Extract command failed", **error_report.items())
```
The caller does not need to know that the `Report` is an `attrs` class and it should not include `attrs` just for the sake of call `attrs.asdict`. I just called this `items` because of a `dict`-like representation looks easy enough to understand from the caller side.

unblob/processing.py
```diff
@@ -30,137 +35,146 @@ def process_file(
...
-                    root=task.root,
-                    path=path,
-                    depth=task.depth + 1,
-                )
+        log.info("Calculated file size", size=size)
+
+        self._process_regular_file(task, size, result)
+        return result
+
+    def _process_regular_file(self, task: Task, size: int, result: TaskResult) -> None:
```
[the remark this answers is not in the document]
kukovecz: As far as I know pyright can handle it. But I remove it, just for the sake of consistency.

unblob/processing.py
```diff
@@ -30,137 +35,146 @@ def process_file(
...
+        self._entropy_depth = entropy_depth
+        self._verbose = verbose
 
-    for p in worker_processes:
-        p.terminate()
-        p.join()
+    def process_task(self, task: Task) -> TaskResult:
+        try:
+            return self._process_task(task)
+        except Exception as exc:
```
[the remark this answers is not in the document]
kukovecz: BAH! I love this idea (and with the proposed `None` return values below, so we will always work with 1 `TaskResult` object for the whole `Task` to process.

unblob/processing.py
```diff
@@ -30,137 +35,146 @@ def process_file(
...
 
     root = path if path.is_dir() else path.parent
-    task_queue = multiprocessing.JoinableQueue()
-    task_queue.put(
-        Task(
-            root=root,
-            path=path,
-            depth=0,
-        )
+    root_task = Task(
```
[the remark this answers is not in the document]
kukovecz: I leave this up for debate :sweat_smile: 

We have 
* `Task` objects (usually the variable called `task` when we pass it around) 
* `TaskResult` objects, (usually the variable called `result` when we pass it around)

Modifying the `Task` class name wound also require modifications in the `TaskResult` class name and in the variable names ...

Honestly, I would leave it like this, but I am open to suggestions.

tests/test_pool.py
```diff
@@ -0,0 +1,58 @@
...
+def test_multipool(process_num: int):
+    handled = multiprocessing.Manager().list()
+    results = multiprocessing.Manager().list()
+
+    def _handler(i):
+        handled.append(i)
+        return i
+
+    def _callback(pool, result):
+        results.append(result)
```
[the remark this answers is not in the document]
kukovecz: Yes! Fixed.

unblob/pool.py
```diff
@@ -0,0 +1,117 @@
...
+    def __init__(self, handler, *, result_callback):
+        self._handler = handler
+        self._result_callback = result_callback
+        self._args = []
+
+    def submit(self, args):
+        self._args.append(args)
+
+    def process_until_done(self):
+        for args in self._args:
```
vlaci, line 104: This is incorrect, as the callback may call submit, changing the _args list while iterating over it via the for loop. Can be changed to a while loop and using pop though to retrieve the next element

unblob/processing.py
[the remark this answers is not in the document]
vlaci: or ProcessInfo <-> ProcessResult but I am also okay resolving this in a followup PR

unblob/models.py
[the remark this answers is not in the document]
vlaci: I also like the idea of not leaking the implementation detail that this class is made using attrs.

unblob/pool.py
kukovecz: Really? 
* The `handler` foes not have the `pool` itself, only the `result_callback`.
* In the tests, the `result_callback` is already submitting to the pool in the test, and it is passed.

What can't I see here?

vlaci: the callback is called from line 106 inside the for loop which iterates over the same array. I wouldn't count on this working every time on each supported python version and arbitrary number of appends.

unblob/pool.py
```diff
@@ -0,0 +1,118 @@
...
+class MultiPool(PoolBase):
+    def __init__(
+        self,
+        process_num: int,
+        handler: Callable[[Any], Any],
+        *,
+        result_callback: Callable[["MultiPool", Any], Any],
+    ):
+        if process_num <= 0:
+            raise ValueError("At process_num must be greater than 0!")
```
vlaci, line 58: I don't think that it worth to use this class with zero processes :)

unblob/pool.py
```diff
@@ -0,0 +1,115 @@
...
+
+        if process_num == 1:
```
vlaci, line 60: Using with 1 process is not strictly useful, I wouldn't disallow it as it works as intended and can be useful for e.g. debugging.

unblob/pool.py
kukovecz: Yes... hence the `ValueError` :sweat_smile: 

Should I those conditions and let the "user" (us....basically....) hand off these restrictions:

unblob/models.py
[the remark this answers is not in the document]
kissgyorgy: Ok, bought it, but the method name should be `asdict` then, because if you want to call items on the return value, it would look like this: `report.items().items()` which would be very confusing.

vlaci (approved):

Issue #118, opened 2021-12-14
Inexact padding calculation on specific TAR files


QKaiser:
The offset computation we came up with in the TarHandler at `_get_tar_end_offset` is inexact for specific files.

I surmise it has to do with block padding and block alignment. Without a fix, unblob finds the end offset past the file and discard the chunk.

I can provide a sample internally if someone wants to work on this:

```
sha1sum /tmp/firmware.img 
fc910cc004254fda5eef27c88f943296d3df967e  /tmp/firmware.img
```

vlaci:
I can look into it tomorrow

vlaci:
I've experimented a bit, and dove into GNU tar's source and found out that `tarfile.RECORDSIZE` is just `tarfile.BLOCKSIZE` times what GNU tar calls as `--blocking-factor`. It's default value is `20` matching the Python implementation, but it is not something we should expect and certainly not set in stone. It can be smaller and even greater.

It is not the full story as the example file here is properly aligned to consist of 2 tar blocks, however it ends with only 384 padding zeroes.

Regarding the final zeroes `info tar` says:

> At the end of the archive file there are two 512-byte blocks
> filled with binary zeros as an end-of-file marker.  A reasonable system
> should write such end-of-file marker at the end of an archive, but must
> not assume that such a block exists when reading an archive.  In
> particular GNU 'tar' always issues a warning if it does not encounter
> it.

Interestingly this file does not issue a warning when extracted by GNU tar, so this end block thingy can be considered optional?
On the other hand we shouldn't fail on tar files missing this feature as the documentation says `reasonable systems shouldn't expect its existence`.

So my proposal now is to either:
* guess the last block's end, then count down any as many 512 byte zero blocks as there are. 
* extract files without final zero blocks

PR #142, opened 2021-12-20
tar: support smaller than RECORDSIZE archives without 2x zero blocks

vlaci:
There are tar files (such as what found in #118) which are not padded
to `20*BLOCKSIZE` bytes and doesn't have `2*BLOCKSIZE` ending
zeros.

The above cases are independent:

1. `tarfile.RECORDSIZE` is just tarfile.BLOCKSIZE times what GNU tar
calls as `--blocking-factor`. It's default value is 20 matching the
Python implementation, but it is not something we should expect to be
used and certainly. It can be smaller and even greater albeit it is
discouraged to increase it from the default.

2. According to `info tar` extractors shouldn't rely on the
existence of ending zero blocks:

> At the end of the archive file there are two 512-byte blocks filled
> with binary zeros as an end-of-file marker. A reasonable system
> should write such end-of-file marker at the end of an archive, but
> must not assume that such a block exists when reading an archive. In
> particular GNU 'tar' always issues a warning if it does not
> encounter it.

In reality GNU tar doesn't warn when a file without 2 zero blocks is
extracted.

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -1,0 +1,111 @@
+import binascii
+import io
+
+from attr import dataclass
+from lark import Lark, Transformer
+from lark.visitors import Discard
+
+
+def unhex(hexdump: str) -> bytes:
+    """Unparses hexdump back to binary representation.
+
+    In addition to basic parsing the following extra features are supported:
+
+    * line comments starting with ``#``
+    * squeezing (repetition of previous line) indicated via ``*`` (see: man 1 hexdump)
+
+    Relative position of data is kept in the result object. This means
+    that the offset indicator at the beginning of each line is
+    significant, each line will be stored at the position indicated
+    relative to the start position.
+
+    The printable ASCII column is discarded during parsing.
+    """
+    parsed = _hexdump_parser.parse(hexdump)
+    return _HexDumpToBin().transform(parsed)
+
+
+_hexdump_parser = Lark(
+    """
+    COMMENT:    _SPACE* "#" /[^\n]/x* NEWLINE
+    %ignore COMMENT
+
+    %import common.NEWLINE
+    %import common.HEXDIGIT
+
+    start:   line (_NEWLINE line)* _NEWLINE?
+    line:    address [_SEPARATOR hex _SEPARATOR "|"? ascii "|"?]  -> canonical
+             | SQUEEZE                                            -> squeezed
+    address: HEXDIGIT+                                            -> join
+    hex:     HEXDIGIT+ (_SPACE* HEXDIGIT)+                        -> join
+    ascii:   CHAR+                                                -> join
+    CHAR:    /./
+    SQUEEZE: "*"
+
+    _SEPARATOR: ": " | "  "
+    _SPACE:     " "
+    _NEWLINE:   NEWLINE
+"""
+)
+
+
+@dataclass
+class _HexdumpLine:
+    offset: int
+    data: bytes
+
+    @classmethod
+    def from_bytes(cls, offset, data):
+        offset = int.from_bytes(binascii.unhexlify(offset), byteorder="big")
+        data = binascii.unhexlify(data) if data else b""
+        return cls(offset, data)
+
+    def __len__(self):
+        return len(self.data)
+
+
+class _HexDumpToBin(Transformer):
+    def __init__(self):
+        super().__init__(visit_tokens=False)
+        self._last_line = None
+        self._squeezing = False
+
+    def join(self, s):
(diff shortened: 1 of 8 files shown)


tests/helpers.py
```diff
@@ -0,0 +1,111 @@
...
+from lark.visitors import Discard
+
+
+def unhex(hexdump: str) -> bytes:
+    """Unparses hexdump back to binary representation.
+
+    In addition to basic parsing the following extra features are supported:
+
+    * line comments starting with ``#``
+    * squezing (repetion of previous lire) indicated via ``*`` (see: man 1 hexdump)
```
QKaiser, line 15:
```suggestion
    * squeezing (repetition of previous line) indicated via ``*`` (see: man 1 hexdump)
```

tests/helpers.py
```diff
@@ -0,0 +1,111 @@
...
+
+    Relative position of data is kept in the result object. This means
+    that the offset indicator at the begining of each line is
```
QKaiser, line 18:
```suggestion
    that the offset indicator at the beginning of each line is
```

tests/helpers.py
```diff
@@ -0,0 +1,111 @@
...
+    """
+    COMMENT:    _SPACE* "#" /[^\n]/x* NEWLINE
+    %ignore COMMENT
+
+    %import common.NEWLINE
+    %import common.HEXDIGIT
+
+    start:      line (_NEWLINE line)* _NEWLINE?
+    line:       address [_SEPARATOR hex _SEPARATOR "|"? ascii "|"?]  -> canonical
+                | SQUIZE                                             -> squized
```
QKaiser, line 38:
```suggestion
                | SQUEEZE                                             -> squeezed
```

tests/helpers.py
```diff
@@ -0,0 +1,111 @@
...
+    address:    HEXDIGIT+                                            -> join
+    hex:        HEXDIGIT+ (_SPACE* HEXDIGIT)+                        -> join
+    ascii:      CHAR+                                                -> join
+    CHAR:       /./
+    SQUIZE:     "*"
```
QKaiser, line 43:
```suggestion
    SQUEEZE:     "*"
```

tests/test_helpers.py
```diff
@@ -0,0 +1,63 @@
...
+        pytest.param(HEXDUMP_C, id="hexdump-C"),
+        pytest.param(WITH_COMMENTS, id="with-comments"),
+    ),
+)
+def test_hexdump(hexdump):
+    binary = unhex(hexdump)
+    assert binary == EXPECTED
+
+
+WITH_SQUIZED_DATA = """\
```
QKaiser, line 38:
```suggestion
WITH_SQUEEZED_DATA = """\
```

tests/test_helpers.py
```diff
@@ -0,0 +1,63 @@
...
+00: 0102 0304 0506 0708 090a 0b0c 0d0e 0f10  ................
+10: FF00 FF00 FF00 FF00 FF00 FF00 FF00 FF00  ................
+*
+30: 4142 4344 4546 4748 494a 4b4c 4d4e 4f44  ABCDEFGHIJKLMNOD
+"""
+
+
+def test_with_squized_data():
+    binary = unhex(WITH_SQUIZED_DATA)
```
QKaiser, line 47:
```suggestion
    binary = unhex(WITH_SQUEEZED_DATA)
```

tests/test_helpers.py
```diff
@@ -0,0 +1,63 @@
...
+    assert binary[:0x10] == EXPECTED[:0x10]
+    assert binary[0x10:0x30] == b"\xFF\x00" * 0x10
+    assert binary[0x30:] == EXPECTED[0x10:]
+
+
+WITH_SQUIZED_END = """\
```
QKaiser, line 53:
```suggestion
WITH_SQUEEZED_END = """\
```

tests/test_helpers.py
```diff
@@ -0,0 +1,63 @@
...
+00: FFFF FFFF FFFF FFFF FFFF FFFF FFFF FFFF  ................
+*
+40
+"""
+
+
+def test_with_squized_end():
+    binary = unhex(WITH_SQUIZED_END)
```
QKaiser, line 61:
```suggestion
    binary = unhex(WITH_SQUEEZED_END)
```

QKaiser (changes requested):
Excellent work as always, sorry for being ultra pedantic with spelling :)

vlaci:
Should be configuring a spell checker in my editor :)

PR #143, opened 2021-12-20
CLI Testing

### LINKED ISSUE #135: Test the CLI
We should test the unblob CLI.

We could do by simply executing the `unblob` command ... 
.... but we could also check this out, since we are using `click` for everything CLI-related: https://click.palletsprojects.com/en/8.0.x/testing/

kukovecz:
Fixes #135

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -1,7 +1,13 @@
+import unittest.mock as mock
+from pathlib import Path
+from typing import List
+
+import pytest
 from click.testing import CliRunner
 from conftest import TestHandler
 
 import unblob.cli
+from unblob.processing import DEFAULT_DEPTH
 
 
 class ExistingCommandHandler(TestHandler):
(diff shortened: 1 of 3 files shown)


tests/test_cli.py
```diff
@@ -39,3 +43,110 @@ def test_show_external_dependencies_not_exists(monkeypatch):
...
+            ["--verbose", "--extract-dir", "unblob", "--depth", "2", "tests", "--help"],
+            id="eager_2",
+        ),
+    ),
+)
+def test_help(params):
+    runner = CliRunner()
+    result = runner.invoke(unblob.cli.cli, params)
+    assert result.exit_code == 0
+    # NOTE: In practive, it writes "Usage: unblob ...", this is done in the `cli.main` with `click.make_context`
```
QKaiser, line 66:
```suggestion
    # NOTE: In practice, it writes "Usage: unblob ...", this is done in the `cli.main` with `click.make_context`
```

QKaiser (approved):
Looks good to me.

tests/test_cli.py
```diff
@@ -39,3 +43,110 @@ def test_show_external_dependencies_not_exists(monkeypatch):
...
+        pytest.param(
+            ["--verbose", "--extract-dir", "unblob", "--depth", "2"],
+            id="verbose+extract-dir+depth",
+        ),
+    ),
+)
+def test_without_file(params: List[str]):
+    runner = CliRunner()
+    result = runner.invoke(unblob.cli.cli, params)
+    assert result.exit_code == 0
```
vlaci, line 88: I think we'd want a non-zero exit code in that case

unblob/cli.py
```diff
@@ -74,7 +74,11 @@ def get_help_text():
 )
 def cli(files: Tuple[Path], extract_root: Path, depth: int, verbose: bool):
     configure_logger(verbose, extract_root)
-    logger.info("Start processing files", count=noformat(len(files)))
+    file_count = len(files)
+    if file_count <= 0:
```
vlaci, line 78: I'd just set `required=True` for the files argument

tests/test_cli.py
```diff
@@ -39,3 +43,110 @@ def test_show_external_dependencies_not_exists(monkeypatch):
...
+    in_path = (
+        Path(__file__).parent
+        / "integration"
+        / "archive"
+        / "zip"
+        / "regular"
+        / "__input__/"
+    )
+    new_params = params + ["--extract-dir", str(tmp_path), str(in_path)]
+    result = runner.invoke(unblob.cli.cli, new_params)
```
vlaci, line 124: I'd mock the `process_file` call and assert on its arguments. Wouldn't run this with any of the real integration test assets.

vlaci (changes requested):

unblob/cli.py
kukovecz: Yeah, good idea!

vlaci (approved):

PR #144, opened 2021-12-21
Fixing infinite loops in XZ handler

vlaci:
It is still a draft as I want to refactor `iterate_patterns` first

### DIFF
--- a/unblob/file_utils.py
+++ b/unblob/file_utils.py
@@ -100,6 +100,20 @@ def find_first(
     Seek the file pointer to the next byte of where we found the pattern or
     seek back to the initial position when we did not find it.
     """
+    try:
+        return next(iterate_patterns(file, pattern, chunk_size))
+    except StopIteration:
+        return -1
+
+
+def iterate_patterns(
+    file: io.BufferedIOBase, pattern: bytes, chunk_size: int = 0x1000
+) -> Iterator[int]:
+    """Iterate on the file searching for pattern until all occurences has been found.
+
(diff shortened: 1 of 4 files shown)


unblob/file_utils.py
```diff
@@ -105,44 +119,39 @@ def find_first(
...
+                file.seek(found_pos + len(pattern))
+                yield found_pos
+                # re-set position in-case it gets changed during iteration
+                file.seek(found_pos + len(pattern))
+                marker = data.find(pattern, marker + len(pattern))
+
+            file.seek(-compensation, os.SEEK_CUR)
+            bytes_searched += chunk_size - compensation
+    finally:
+        file.seek(initial_position)
```
vlaci, line 154: I like the idea to always reset the file pointer. A bunch of tests fail now because of this. What do you think, should I adjust the tests or restore the old behavior?

QKaiser: I like that idea too as it will make the behavior generic between the different cases (no pattern found, one pattern found, multiple pattern found). So I'd go for adjusting tests.

If I understand correctly, the iterator consumer must read all values from it for the pointer to be set back, right ?

vlaci: Or discard the iterator, like in the case of `find_first` Another possible way to go is to reset the pointer to the beginning at each yield, so it is the consumer's responsibility to seek.

vlaci: I have gone forward with the latter approach, only UBI depended on the seeking side-effect anyway.

QKaiser (approved):

PR #148, opened 2022-01-04
Add support for JFFS2 filesystem format.

QKaiser:
We add support for both JFFS2 new and JFFS2 old formats using a simple cparser, no external library required.

The extraction is performed by Jefferson (see https://github.com/sviehb/jefferson/). Given that it is MIT-licensed
we put it into unblob/extractors and applied some modifications to support JFFS2 old format too (only the magic
differs, the format stays the same).

Jefferson requires cstruct version 1.8 and fails with recent version such as 2.1 due to a breaking change in the API.
cstruct was therefore added to the poetry dependencies, enforcing version 1.8.

We also added the extractors directory to the flake8 exclusion list given that it would require a full rewrite of
any external tool we want to use (i.e. to severely reduce their cyclomatic complexity) in order to pass those checks.

Test files have been generated with mkfs.jffs2, with one version recompiled to use the old magic signature (0x1984).
Test files cover every possible test case in terms of signature, padding, endianness, and compression.

The following script was used to generate test files with two custom versions of mkfs.jffs2:

```bash
#!/bin/sh

KIND=(old new)
EXTFS=(jffs2)
ENDIAN=(l b)
COMPRESSION=(lzo zlib rtime)
LABEL="fruits"

rm -f /tmp/fruits*jffs2

mkdir -p "/tmp/fruits"
echo "AAAAAAAAAAAAAAAAAAAAAAAAA" > "/tmp/fruits/apple.txt"
echo "cherry" > "/tmp/fruits/cherry.txt"
echo "banana" > "/tmp/fruits/banana.txt"


get_endianness () {
    if [ "$1" == "b" ]; then
        echo "be";
    else
        echo "le";
    fi
}
    
# unpadded no compression

for ext in ${EXTFS[@]}; do
    for k in ${KIND[@]}; do
        for endian in ${ENDIAN[@]}; do
            e="$(get_endianness $endian)"
            echo "[+] crafting ${k} ${ext} fs with ${e} endian (not padded)"
            FILE="/tmp/fruits.${k}.${e}.nocomp.${ext}"
            ~/mtd-utils/mkfs.${ext}.${k} -v -x rtime -x lzo -x zlib -${endian} -d /tmp/fruits -o ${FILE}
            sleep 1
        done
    done
done

# unpadded with compression
for ext in ${EXTFS[@]}; do
    for k in ${KIND[@]}; do
        for endian in ${ENDIAN[@]}; do
            for comp in ${COMPRESSION[@]}; do
                e="$(get_endianness $endian)" 
                echo "[+] crafting ${k} ${ext} fs with ${e} endian and ${comp} compression (not padded)"
                FILE="/tmp/fruits.${k}.${e}.${comp}.${ext}"
                ~/mtd-utils/mkfs.${ext}.${k} -v -X $comp -y 100:${comp} -${endian} -d /tmp/fruits -o ${FILE}
                sleep 1
            done
        done
    done
done

# padded no compression

for ext in ${EXTFS[@]}; do
    for k in ${KIND[@]}; do
        for endian in ${ENDIAN[@]}; do
            e="$(get_endianness $endian)" 
            echo "[+] crafting ${k} ${ext} fs with ${e} endian (padded)"
            FILE="/tmp/fruits.${k}.${e}.nocomp.padded.${ext}"
            ~/mtd-utils/mkfs.${ext}.${k} -v -x rtime -x lzo -x zlib -p -${endian} -d /tmp/fruits -o ${FILE}
            sleep 1
        done
    done
done

# padded with compression
for ext in ${EXTFS[@]}; do
    for k in ${KIND[@]}; do
        for endian in ${ENDIAN[@]}; do
            for comp in ${COMPRESSION[@]}; do
                e="$(get_endianness $endian)" 
                echo "[+] crafting ${k} ${ext} fs with ${e} endian and ${comp} compression (padded)"
                FILE="/tmp/fruits.${k}.${e}.${comp}.padded.${ext}"
                ~/mtd-utils/mkfs.${ext}.${k} -v -p -X $comp -y 100:${comp} -${endian} -d /tmp/fruits -o ${FILE}
                sleep 1
            done
        done
    done
done
```

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -82,6 +82,14 @@ tomli = {version = "*", optional = true, markers = "extra == \"toml\""}
 toml = ["tomli"]
 
 [[package]]
+name = "cstruct"
+version = "2.1"
+description = "C-style structs for Python"
+category = "main"
+optional = false
+python-versions = "*"
+
+[[package]]
 name = "dissect.cstruct"
 version = "2.0"
 description = "Structure parsing in Python made easy."
@@ -127,6 +135,21 @@ description = "iniconfig: brain-dead simple config-ini parsing"
 category = "dev"
 optional = false
 python-versions = "*"
+
+[[package]]
+name = "jefferson"
+version = "0.3"
+description = ""
+category = "main"
+optional = false
+python-versions = "*"
+develop = false
+
+[package.source]
+type = "git"
+url = "https://github.com/IoT-Inspector/jefferson.git"
+reference = "216eee6c56d338e5a14a0af5d07f1117cff92b3b"
+resolved_reference = "216eee6c56d338e5a14a0af5d07f1117cff92b3b"
 
 [[package]]
 name = "lark"
@@ -394,7 +417,7 @@ python-versions = "*"
 [metadata]
 lock-version = "1.1"
 python-versions = "^3.8"
-content-hash = "b3663d45d1926a5a2239856d84088fd0ef130d410da2d43ce01afbf0576fec83"
+content-hash = "4ce7fb75dd8eacb9ff7d9fa4855310fe538fdb4973f994cfa639660986c4f2bd"
 
 [metadata.files]
 arpy = [
@@ -473,6 +496,10 @@ coverage = [
     {file = "coverage-6.2-pp36.pp37.pp38-none-any.whl", hash = "sha256:5829192582c0ec8ca4a2532407bc14c2f338d9878a10442f5d03804a95fac9de"},
     {file = "coverage-6.2.tar.gz", hash = "sha256:e2cad8093172b7d1595b4ad66f24270808658e11acf43a8f95b41276162eb5b8"},
 ]
+cstruct = [
+    {file = "cstruct-2.1-py2.py3-none-any.whl", hash = "sha256:ed4449bc672a37abbc0ecfbdaf4c26aa6ab1bafe7dfdb83eab2953cdbb153cf4"},
+    {file = "cstruct-2.1.tar.gz", hash = "sha256:9c2741f46fb12613a4f746369ec0a7b3b1f25712178c567437902d446f7c3648"},
+]
 "dissect.cstruct" = [
     {file = "dissect.cstruct-2.0-py3-none-any.whl", hash = "sha256:e150abcc7500a279126ebf1a0f87c34d857e126ac3d65a980844bfb3296a5492"},
     {file = "dissect.cstruct-2.0.tar.gz", hash = "sha256:4f1d3911da99270213feed0a0b22e70670b04fe53b16d95d2e04bafcd338f547"},
@@ -493,6 +520,7 @@ iniconfig = [
     {file = "iniconfig-1.1.1-py2.py3-none-any.whl", hash = "sha256:011e24c64b7f47f6ebd835bb12a743f2fbe9a26d4cecaa7f53bc4f35ee9da8b3"},
     {file = "iniconfig-1.1.1.tar.gz", hash = "sha256:bc3af051d7d14b2ee5ef9969666def0cd1a000e121eaea580d4a313df4b37f32"},
 ]
+jefferson = []
 lark = [
     {file = "lark-1.0.0-py2.py3-none-any.whl", hash = "sha256:10818563b44e18ea264b5013a80fe04406999ad51bacc74214f69f13d95af6e1"},
     {file = "lark-1.0.0.tar.gz", hash = "sha256:2269dee215e6c689d5ce9d34fdc6e749d0c1c763add3fc7935938ebd7da159cb"},
--- a/unblob/handlers/filesystem/jffs2.py
+++ b/unblob/handlers/filesystem/jffs2.py
@@ -1,0 +1,150 @@
+import io
+from typing import List, Optional
+
+from structlog import get_logger
+
+from unblob.file_utils import Endian, convert_int16, read_until_past, round_up
+
+from ...models import StructHandler, ValidChunk
+
+logger = get_logger()
+
+
+BLOCK_ALIGNMENT = 4
+JFFS2_MAGICS = [0x1985, 0x8519, 0x1984, 0x8419]
+
+# Compatibility flags.
+JFFS2_COMPAT_MASK = 0xC000
+JFFS2_NODE_ACCURATE = 0x2000
+JFFS2_FEATURE_INCOMPAT = 0xC000
+JFFS2_FEATURE_ROCOMPAT = 0x8000
+JFFS2_FEATURE_RWCOMPAT_COPY = 0x4000
+JFFS2_FEATURE_RWCOMPAT_DELETE = 0x0000
+
+DIRENT = JFFS2_FEATURE_INCOMPAT | JFFS2_NODE_ACCURATE | 1
+INODE = JFFS2_FEATURE_INCOMPAT | JFFS2_NODE_ACCURATE | 2
+CLEANMARKER = JFFS2_FEATURE_RWCOMPAT_DELETE | JFFS2_NODE_ACCURATE | 3
+PADDING = JFFS2_FEATURE_RWCOMPAT_DELETE | JFFS2_NODE_ACCURATE | 4
+SUMMARY = JFFS2_FEATURE_RWCOMPAT_DELETE | JFFS2_NODE_ACCURATE | 6
+XATTR = JFFS2_FEATURE_INCOMPAT | JFFS2_NODE_ACCURATE | 8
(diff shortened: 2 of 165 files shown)


QKaiser:
We need to have a discussion on external extractors. Specifically:
- should we run flake8 on them, and adapt them so they pass it ?
- should we include them into test coverage calculation, and rewrite them so they're tested ?

Another thing I'd like the framework to provide is some kind of global variable pointing to the unblob root so I can construct the path to external extractors properly.

vlaci:
I am not really comfortable with vendoring 3rd parties as is. Even if it is compatible licensing wise, it will quickly become a mess when we want to incorporate upstream changes or upstream our modifications. It just makes sense to keep these things separate. I'd keep the extractors directory for stuff, we write and maintain exclusively.

A way forward could be:

1. First add the original repository as a [git dependency ](https://python-poetry.org/docs/dependency-specification/#git-dependencies) (pinned to a well-known working revision, can be our fork to have the place for our changes) to pyproject.toml so we are not blocked on the below steps
2. Ask Stefan if he'd like to publish the original tool on pypi
    a. If he is willing, then we are all set
    b. If he doesn't want to bother with it, we should fork it permanently and optionally publish ourselves.
3. Then we can depend on it the same way as we depend on the ubireader stuff

QKaiser:
I forked it under the IoT-Inspector organization on Github (https://github.com/IoT-Inspector/jefferson) so I can work on some improvements. I contacted Stefan but it seems it's not super actively maintained so we better use our own fork for the time being.

Plan so far with jefferson fork:

- [ ] fix it to support cstruct 2.1
- [ ] add support for legacy signature 
- [ ] add LZO compression support
- [ ] improve speed (there is an interesting open PR using mmap)

QKaiser:
@vlaci I generated proper test files. I had to craft text files worth the compression otherwise mkfs.jffs2 falls back to no compression, mkfs.jffs2 is too smart.  I also forced no compression by disabling all compression formats for the 'nocompress' test cases.

Fun fact about compression: a text file containing "AAAAAAAAAAAAAAAAAAAAAAAAA" will trigger lzo, zlib, or rtime compression when enabled. Ratios and stuff :)

I removed jefferson from unblob/extractors so we don't vendor 3rd party scripts. Jefferson is installed by poetry using the current latest revision of our fork at https://github.com/IoT-Inspector/jefferson

unblob/handlers/filesystem/jffs2.py
```diff
@@ -0,0 +1,150 @@
...
+                    current_offset = read_until_past(file, b"\x00\xFF")
+                else:
+                    logger.debug("unexpected header magic", header_magic=header.magic)
+                break
+
+            if header.nodetype not in JFFS2_NODETYPE._value2member_map_:
+                logger.debug("Invalid JFFS2 node type", node_type=header.nodetype)
+                return
+
+            if header.totlen > eof:
```
vlaci, line 91: Should more generally the condition be:

```python
node_start_offset = current_offset  # at the begining of the loop
# ...
if node_start_offset + header.totlen > eof:
    # ...
```
My above suggestion about changing the while loop condition covers this issue more generally.

unblob/handlers/filesystem/jffs2.py
```diff
@@ -0,0 +1,150 @@
...
+    ) -> Optional[ValidChunk]:
+
+        file.seek(0, io.SEEK_END)
+        eof = file.tell()
+        file.seek(start_offset)
+
+        endian = self.guess_endian(file)
+        current_offset = start_offset
+
+        while True:
```
vlaci, line 73: The exit condition is always based on `current_offset`'s relation to `eof`. It would be easier to reason about this code ever returning if we'd use the exit condition as the loop condition: 

```python
while current_offset < eof:
    #  ...

if current_offset > eof:
    # corrupt file or last chunk isn't really jffs2
```

 Would it make sense, to return the length of the previous offset if we are past eof at the end of the loop?

unblob/handlers/filesystem/jffs2.py
```diff
@@ -0,0 +1,150 @@
...
+            if header.totlen < len(header):
+                logger.debug(
+                    "node length greater than header size", node_len=header.totlen
+                )
+                return
+
+            node_len = round_up(header.totlen, BLOCK_ALIGNMENT)
+            file.seek(node_len, io.SEEK_CUR)
+
+            current_offset += round_up(header.totlen, BLOCK_ALIGNMENT)
```
vlaci, line 108: I think `current_offset += node_len` or just `current_offset = file.tell()` would be easier to read.

unblob/handlers/filesystem/jffs2.py
```diff
@@ -0,0 +1,150 @@
...
+            uint32 hdr_crc;
+        } jffs2_unknown_node_t;
+    """
+
+    HEADER_STRUCT = "jffs2_unknown_node_t"
+
+    BIG_ENDIAN_MAGIC = 0x19_85
+
+    def guess_endian(self, file: io.BufferedIOBase) -> Endian:
+        magic = convert_int16(file.read(2), Endian.BIG)
```
vlaci, line 54: what happens if we are at the end of the file? I guess we'd raise from `convert_int16` and throw away this chunk's processing. This shouldn't be an issue, just wanted to think through it :)

Interestingly `file_utils.get_endian` has different behavior, as it would seek back more when it is near the end: https://github.com/IoT-Inspector/unblob/pull/157

vlaci (changes requested):

unblob/handlers/filesystem/jffs2.py
QKaiser: We call `guess_endian` only once on the beginning of our match. Given that our yara rule match on exactly 4 bytes, `convert_int16` will never throw an `EOFError`.

unblob/handlers/filesystem/jffs2.py
QKaiser: Fixed.

unblob/handlers/filesystem/jffs2.py
QKaiser: Fixed.

unblob/handlers/filesystem/jffs2.py
QKaiser: Good catch. Fixed.

unblob/handlers/filesystem/jffs2.py
```diff
@@ -0,0 +1,150 @@
...
+                # JFFS2 allows padding at the end with 0xFF or 0x00, usually
+                # to the size of an erase block.
+                if header.magic in [0x0000, 0xFFFF]:
+                    file.seek(-len(header), io.SEEK_CUR)
+                    current_offset = read_until_past(file, b"\x00\xFF")
+                else:
+                    logger.debug("unexpected header magic", header_magic=header.magic)
+                break
+
+            if header.nodetype not in JFFS2_NODETYPE._value2member_map_:
```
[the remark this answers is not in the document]
QKaiser: Moved to a set rather than an enum.

poetry.lock
```diff
@@ -16,29 +16,17 @@ python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*"
 
 [[package]]
 name = "attrs"
-version = "21.2.0"
+version = "21.4.0"
```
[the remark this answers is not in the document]
QKaiser: I applied your patch to poetry.lock, thx.

unblob/handlers/filesystem/jffs2.py
```diff
@@ -0,0 +1,150 @@
...
+            if header.magic not in JFFS2_MAGICS:
+                # JFFS2 allows padding at the end with 0xFF or 0x00, usually
+                # to the size of an erase block.
+                if header.magic in [0x0000, 0xFFFF]:
+                    file.seek(-len(header), io.SEEK_CUR)
+                    current_offset = read_until_past(file, b"\x00\xFF")
+                    continue
+                else:
+                    logger.debug("unexpected header magic", header_magic=header.magic)
+                    break
```
QKaiser, line 85: @vlaci I just fixed this so that we don't end up with hundreds of valid but separated JFFS2 chunks that actually belong to a larger chunk. Tested with a firmware holding two JFFS2 filesystems next to each other, with multiple sections of each filesystem being 0xFF padded.

vlaci (approved):

vlaci:
On a related note, we may still improve how jefferson and ubireader is handled, because right now tests and unblob itself depends on that the virtualenv's bin directory is added to the PATH, which is there if using `poetry run` but isn't if the environment is activated by hand.

Issue #134, opened 2021-12-16
Extractor version handling - documentation


kukovecz:
#113  Introduces the ability to check if the current environment has the needed extractors installed for each handler.

However, the CLI of these extractors can be varying based on the version - some subcommands / options may only be available from/to a specific version. Right now, we do not check if the right version of the extractor is installed, only that the extractor itself is available.

So, check the installed extractor version and match it with the handler.

kukovecz:
#131 also mentions this problem in great detail specific to `7z`.

kissgyorgy:
We agreed to only document this for v1.0.

vlaci:
Fixed in #251

Issue #131, opened 2021-12-16
Default version of p7zip-full is outdated


QKaiser:
At the moment we rely on the default version of 7zip that is available as `p7zip-full` in our Linux distributions.

These versions are quite outdated, case in point with latest 7z version on Ubuntu 20.04 and our Docker container (python3.8-slim):

```
7-Zip [64] 16.02 : Copyright (c) 1999-2016 Igor Pavlov : 2016-05-21
```

While the latest version is:

```
7-Zip (z) 21.06 (x64) : Copyright (c) 1999-2021 Igor Pavlov : 2021-11-24
```

This is problematic for a few reasons:
- we lack potential enhancements from 7z's team on decompression/extraction speed
- we might miss security patches from 7z latest, although they may get backported by distribution maintainers (but we should not rely on that)
- we lack support for specific file formats (**I can't send a pull request for ext[2-4] because my version of 7z does not support ext4 extraction**)

### Solution: download and install 7z manually

7zip only provides a download page and there is no link to get the latest version automatically so we would need to keep track of it or parse the history file.

To get the (current) latest version of 7zip as a stripped binary:

```
curl -s https://www.7-zip.org/a/7z2106-linux-x64.tar.xz --output - | tar Jxvf - 7zzs
```

Of course we can use dumb tricks like this:

```
LATEST_VERSION=$(curl -s https://www.7-zip.org/history.txt | head -n 4 | tail -n 1 | awk '{ print $1 }' | sed 's/\.//')
curl -s "https://www.7-zip.org/a/7z${LATEST_VERSION}-linux-x64.tar.xz --output - | tar Jxvf - 7zzs
```

But I'm 100% sure it will break at some point if we do this.

QKaiser:
To be discussed but I'd be in favor of either distributing 7zip in [unblob/extractors](https://github.com/IoT-Inspector/unblob/tree/main/unblob/extractors), or to have an install script somewhere in the repository that must be run prior to use unblob (could it be done with `poetry install` ?)

I think we're good on the licensing side (see https://www.7-zip.org/license.txt).

For our docker container it's just a line to add to the Dockerfile to fetch and install the latest version.

QKaiser:
Can be closed as soon as we land #158

PR #149, opened 2022-01-05
Add support for EXT filesystems.

QKaiser:
EXT filesystem handlers support ext2, ext3, and ext4 with any (valid) block size. Test files have been generated to cover ext2, ext3, and ext4 with 1024, 2048, and 4096 block sizes.

In order to support ext4 we need a recent version of 7zip.

We therefore install the latest version of 7zip in Github actions. We also had to fix calls to 7z on password protected archives since it changed in recent versions.

### DIFF
--- a/.github/actions/setup-dependencies/action.yml
+++ b/.github/actions/setup-dependencies/action.yml
@@ -5,6 +5,14 @@ runs:
       run: sudo apt install unar zlib1g-dev liblzo2-dev lzop lziprecover
       shell: bash
 
+    - name: Install 7zip
+      run: |
+        sudo apt remove -y p7zip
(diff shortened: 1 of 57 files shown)


QKaiser:
This branch solves issue #131 by installing the latest version of 7zip in Github actions.

We should take care of #134 before releasing v1.0 to make sure that end users have a 7z version equal or higher than our current requirement.

vlaci:
Shouldn't we update it in the Dockerfile to have a consistent version?

vlaci (approved):

QKaiser:
> Shouldn't we update it in the Dockerfile to have a consistent version?

Yeah that makes sense.

QKaiser:
@vlaci I updated the Dockerfile and made some improvements in #158.

Issue #141, opened 2021-12-20
Don't read a whole file into memory


kissgyorgy:
https://github.com/IoT-Inspector/unblob/pull/120

QKaiser:
I went through all of our existing handlers and fixed it whenever we used this bad design (read rather than seek).

PR #154, opened 2022-01-05
Seek rather than read in LZO handler

QKaiser:
See #141.


vlaci (approved):
Hm, general thought here: Is it possible to craft a malicious firmware image, which alters the data logged to the console through ansi escape codes? :D

QKaiser:
> Hm, general thought here: Is it possible to craft a malicious firmware image, which alters the data logged to the console through ansi escape codes? :D

Possibly.

PR #157, opened 2022-01-05
get_endian: reset file pointer even at the end of file

vlaci:


QKaiser (approved):
Nice trick, I like this :)

PR #158, opened 2022-01-07
Multiple Dockerfile improvements.

QKaiser:
- install 7zip version 21.07 manually
- upgrade pip to latest version to get rid of warning messages
- perform unblob installation as unblob user rather than system-wide install as root

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,10 @@
 FROM python:3.8-slim
 
 RUN mkdir -p /data/input /data/output
-RUN useradd unblob
+RUN useradd -m unblob
+RUN chown -R unblob /data
 
-ENTRYPOINT ["unblob"]
+WORKDIR /data/output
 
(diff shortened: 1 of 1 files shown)


vlaci:
What is the added benefit of unblob being _installed_ as a regular user? I can see an advantage in _running_ it as a restricted user but it just isn't clear why we install it as such.

QKaiser:
> What is the added benefit of unblob being _installed_ as a regular user? I can see an advantage in _running_ it as a restricted user but it just isn't clear why we install it as such.

It's a matter of defense in depth. If one of the pip packages that we rely upon gets compromised, that means we run their `setup.py` as root on install, then we distribute a compromised image to our users. Obviously this can still happen when installing with limited privileges of the `unblob` user, but impact will be limited.

vlaci (approved):

PR #160, opened 2022-01-10
bzip2: fixing off-by one error in end offset calculation

vlaci:
The end offset should point to the first byte which is _not_ part of the chunk, not to the last, that is.

Also I couldn't find it documented what is the meaning of the offsets in a chunk, so I tried to clarify them as well.


vlaci:
My confidence level is quite small around this bitwise machinery so it is possible that I just misunderstood what the code does. However the new tests fail with the original code.

unblob/handlers/compression/bzip2.py
```diff
@@ -81,7 +81,7 @@ def bzip2_recover(self, file: io.BufferedIOBase, start_offset: int) -> int:
                 # This can be negative, but we don't care, because we also count found blocks
```
QKaiser, line 81: This makes perfect sense to me, I don't know why we added 1 in the first place. Another optimization to get rid of the "this can be negative" message would be:

```python3
if buff == BLOCK_HEADER or buff == BLOCK_ENDMARK:
    blocks_found += 1

if buff == BLOCK_ENDMARK:
    current_block_end = bits_read - COMPRESSED_MAGIC_LENGTH
```

unblob/handlers/compression/bzip2.py
```diff
@@ -81,7 +81,7 @@ def bzip2_recover(self, file: io.BufferedIOBase, start_offset: int) -> int:
...
                 # and in the case we didn't found both of them, we don't use this value
                 # When there are two blocks found, this will reflect the correct end
-                current_block_end = bits_read - (COMPRESSED_MAGIC_LENGTH + 1)
+                current_block_end = bits_read - COMPRESSED_MAGIC_LENGTH
 
                 curr_block += 1
```
QKaiser, line 86: This `curr_block` value is never used.

QKaiser (changes requested):

unblob/handlers/compression/bzip2.py
vlaci: Just saw that myself while I am rewriting the code further :)

unblob/handlers/compression/bzip2.py
vlaci: In reality it no longer can be negative, because we have just found a complete block.

unblob/models.py
```diff
@@ -92,7 +102,11 @@ class Handler(abc.ABC):
     def calculate_chunk(
         self, file: io.BufferedIOBase, start_offset: int
     ) -> Optional[ValidChunk]:
-        """Calculate the Chunk offsets from the Blob and the file type headers."""
+        """
+        Calculate the Chunk offsets from the Blob and the file type headers.
+
+        Returns the index of the first byte which is not part of the chunk.
```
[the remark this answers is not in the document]
vlaci: you are right, I am intended to write this doc for `bzip2_recover`

QKaiser (approved):

PR #163, opened 2022-01-12
Add support for RomFS filesystems

QKaiser:
Based on kernel documentation available at https://www.kernel.org/doc/Documentation/filesystems/romfs.txt.

This commit adds a RomFS handler to unblob and a dedicated RomFS extractor.
The extractor supports any file type, but will only create block and character devices if the user has root privileges.
The extractor enforce checksum validation at both romfs header and inode headers to make sure the file was not corrupted.

A dedicated test file has been added under tests/extractors/test_unromfs to validate utility functions of our extractor.


QKaiser:
The test filesystem was created with this bash script:

```bash
#!/bin/bash

ROOT_DIR="/tmp/fruits"
sudo rm -rf "${ROOT_DIR}"

mkdir -p ${ROOT_DIR}

# create dirs and files
for i in {1..5}; do
    mkdir -p "${ROOT_DIR}/dir_${i}";
    for j in {1..5}; do
        echo "file${i}${j}" > "${ROOT_DIR}/dir_${i}/file_${j}.txt";
    done
done

# create hard links
for i in {1..5}; do
    mkdir -p "${ROOT_DIR}/dir_${i}";
    for j in {1..5}; do
        ln "${ROOT_DIR}/dir_${i}/file_${j}.txt" "${ROOT_DIR}/hlink_${i}_${j}"
    done
done

# create symlinks links
#for i in {1..5}; do
#    mkdir -p "${ROOT_DIR}/dir_${i}";
#    for j in {1..5}; do
#        ln -s "${ROOT_DIR}/dir_${i}/file_${j}.txt" "${ROOT_DIR}/slink_${i}_${j}"
#    done
#done

# create block dev
for i in {1..5}; do
    sudo mknod "${ROOT_DIR}/blkdev${i}" b 31 ${i};
done

# create char dev
for i in {1..5}; do
    sudo mknod "${ROOT_DIR}/chrdev${i}" b 31 ${i};
done

# create fifo
mkfifo /tmp/fruits/fifo0

sudo genromfs -d ${ROOT_DIR} -f /tmp/fruits.romfs
```

symlinks were commented out due to issues with diff in integration test results, but this should not pose any issues.

vlaci:
Hm, what was the issue exactly with `diff`? Would it help if we added `--no-dereference` argument to it?

QKaiser:
> Hm, what was the issue exactly with `diff`? Would it help if we added `--no-dereference` argument to it?

Turns out I can keep symlinks, the actual error was caused by FIFOs (see https://lists.debian.org/debian-devel/1996/05/msg00169.html). I removed it from our romfs sample.

QKaiser:
Turns out I had to use `--no-dereference` explicitly for tests to pass in Github runners. Seemed to be implicit behavior of diff on my machine somehow.

QKaiser:
I need to fix unromfs.py in order to protect against malicious symlinks and path traversal (related to https://github.com/IoT-Inspector/unblob/issues/179) before this can get merged.

QKaiser:
Need to be looked at during a pair programming session. I need guidance on how to write tests for the standalone RomFS extractor (unromfs.py) so that our coverage is kept above 90%.

unblob/extractors/unromfs.py
```diff
@@ -0,0 +1,398 @@
...
+        fd_inodes = {
+            k: v
+            for k, v in self.inodes.items()
+            if v.type in [FS_TYPE.FILE, FS_TYPE.DIRECTORY, FS_TYPE.FIFO, FS_TYPE.SOCKET]
+        }
+        for inode in sorted(fd_inodes.values(), key=lambda inode: inode.path):
+            self.create_inode(self.extract_root, inode)
+
+        # then we create links
+        links_inodes = {
```
vlaci, line 311: Cannot a harlink target be a character device? If so, what happens if we are running as a restricted user?

unblob/extractors/unromfs.py
```diff
@@ -0,0 +1,398 @@
...
+
+    with romfs_file.open("rb") as f:
+        header = RomFSHeader(f, extract_root, force)
+        header.validate()
+        header.recursive_walk(header.header_end_offset, None)
+        header.dump_fs()
+
+
+def main():
+    try:
```
vlaci, line 377: These shenanigans are not needed here. The only reason we have this in unblob is click's special handling of EOFError https://click.palletsprojects.com/en/8.0.x/exceptions/

pyproject.toml
```diff
@@ -39,6 +39,7 @@ profile = "black"
 
 [tool.poetry.scripts]
 unblob = "unblob.cli:main"
+unromfs = "unblob.extractors.unromfs:main"
```
vlaci, line 42: Should we add the extractors directory to the coverage measurement 3 lines below? Hopefully subprocess coverage is tracked well since the parallelization.

vlaci (changes requested):
I like this code very much :) I hope

tests/extractors/test_unromfs.py
```diff
@@ -0,0 +1,67 @@
...
+    (
+        (b"\x00\x00\x00\x00", b""),
+        (b"AAAAAAAAAAAAAAAA\x00", b"AAAAAAAAAAAAAAAA"),
+        (b"AAAAAAAAAAAAAAA\x00", b"AAAAAAAAAAAAAAA"),
+        (b"AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA\x00", b"AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA"),
+        (b"AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA\x00", b"AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA"),
+    ),
+)
+def test_get_string(content, expected):
+    with patch("builtins.open", mock_open(read_data=content)) as mock_file:
```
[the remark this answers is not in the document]
QKaiser: Pyright didn't like the idea, saying that `BytesIO` is not compatible with `io.BufferedReader`. I'm using `io.BufferedReader` in unromfs given that I'm simply opening the file with built-in `open()` and not relying on LimitedReader like unblob does.

If there's an easy way that doesn't trigger pyright let me know.

pyproject.toml
QKaiser: It's already covered by `--cov=unblob` since unromfs is in `unblob/extractors`.

Pytest coverage output:

```
---------- coverage: platform linux, python 3.8.10-final-0 -----------
Name                                             Stmts   Miss Branch BrPart  Cover
----------------------------------------------------------------------------------
unblob/extractors/__init__.py                        0      0      0      0   100%
unblob/extractors/unromfs.py                       269     39     91     14    81%
```

unblob/extractors/unromfs.py
QKaiser: Good point. A hard link can point to a character or block device, but the process needs root privileges to create the link.

We should go through char/block devices creation before the links creation. Then, in `create_hardlink`, we should check whether or not the target requires privileges to be linked to. A warning should be emitted if current privileges are not sufficient instead of raising an exception during `os.link()`

unblob/extractors/unromfs.py
QKaiser: Great ! I'll get rid of it.

tests/extractors/test_unromfs.py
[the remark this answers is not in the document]
vlaci: I guess the BufferedIOBase is the right interface that makes pyright happy (we use that extensively in unblob)

unblob/extractors/unromfs.py
QKaiser: Switched the order between link and device creation. I moved to imply capturing the PermissionError when unromfs attempts to create a hardlink pointing to a block/char device instead of adding pointless `if` conditions before.

tests/extractors/test_unromfs.py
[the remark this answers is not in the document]
QKaiser: It worked :)

vlaci (approved):

PR #169, opened 2022-01-18
Fix bzip2 handler to stop early

### LINKED ISSUE #164: bzip2 handler can produce invalid results
Currently `bzip2recover` scans the file until the last `BLOCK_HEADER` or `BLOCK_ENDMARK` found. This can produce invalid results if the file being processed contains multiple bzip2 files embedded with potentially arbitrary data between them.

QKaiser:
Fix #164 by making sure we return as soon as we find the end mark instead of reading until the end of the file. This was causing problems when a file is composed of multiple bzip2 streams with random data in between.

A test file representing this exact test case has been added to bzip2 integration tests, along with a pytest case in test_bzip2.py


vlaci:
We have spoken IRL about having to fine-tune this approach: we should process until we find an endmark without a block start immediately following it.

QKaiser:
@vlaci please approve so we can merge

vlaci (approved):

PR #171, opened 2022-01-18
Normalize C struct conventions in C_DEFINITIONS

### LINKED ISSUE #128: Normalize C struct convention.
We should normalize the C definitions in terms of types and structures. Ideally I'd like to:

- move from `struct name {};` to `typedef struct name_t {} name_t;`
- remove mentions of `short`, `long`, `ulong`, etc in favor of `(u)int8/16/32/64`

It's obviously me being pedantic but it'll make things cleaner, similar to what @kissgyorgy brought by renaming the Yara `$magic` vars that were everywhere 🤓

QKaiser:
Fix #128 by normalizing C struct convention of every C_DEFINITIONS entry (no uppercase type names, standard C11 data types and types definition with typedef struct).

### DIFF
--- a/unblob/handlers/archive/arc.py
+++ b/unblob/handlers/archive/arc.py
@@ -34,7 +34,7 @@ class ARCHandler(StructHandler):
     """
 
     C_DEFINITIONS = r"""
-    struct heads {			/* archive entry header format */
+    typedef struct arc_head {			/* archive entry header format */
         int8 archive_marker;
         int8 header_type;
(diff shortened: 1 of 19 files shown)


kissgyorgy (approved):
I like it, but have no idea why typedef is better in our case.

QKaiser:
> I like it, but have no idea why typedef is better in our case.

It's just a matter of convention, no speed or security gains from it. Just readability :) It will be part of dev guidelines in the wiki.

PR #176, opened 2022-01-19
Fix SquashFS handler negative seeking

### LINKED ISSUE #173: SquashFS handler negative seek triggers OSError
This was identified by structured fuzzing (valid signatures prefixed and suffixed by randomly generated content).

An `OSError` is triggered by squashfs handler, similar to this:

```
2022-01-19 12:40.20 [error    ] Unhandled exception during unblob 
  File "/home/quentin/unblob/unblob/cli.py", line 99, in main
    cli.invoke(ctx)
  File "/home/quentin/unblob/unblob/cli.py", line 81, in cli
  File "/home/quentin/unblob/unblob/processing.py", line 60, in process_file
    carve_unknown_chunks(extract_dir, file, unknown_chunks)
  File "/home/quentin/unblob/unblob/extractor.py", line 84, in carve_unknown_chunks
    carve_chunk_to_file(carve_path, file, chunk)
  File "/home/quentin/unblob/unblob/extractor.py", line 39, in carve_chunk_to_file
    for data in iterate_file(file, chunk.start_offset, chunk.size):
  File "/home/quentin/unblob/unblob/file_utils.py", line 178, in iterate_file
    file.seek(start_offset)
```

This is due to the fact that `chunk.start_offset` is negative. This `start_offset` is negative because the chunk is an `UnknownChunk` and these chunks are created in `calculate_unknown_chunks` by assigning the `start_offset` to the previous chunk's `end_offset`.

At the source of all this, we have a valid squashfs chunk (false positive) with a negative end_offset because the end_offset is calculated by squashfs handlers like this:

```python
size = round_up(header.bytes_used, PAD_SIZE)
end_offset = start_offset + size
```

`bytes_used` being an int64 (signed), it can be negative if the file is malformed.

We should:
- re-type `bytes_used` as a uint64
- implement defenses in `search_chunks_by_priority` against chunks with negative start_offset or end_offset

QKaiser:
Fix #173 by re-typing bytes_used as uint64 so that we no longer have negative offsets in squashfs chunks. Added a negative offsets check in search_chunks_by_priority so we can catch similar issues in the future.

### DIFF
--- a/unblob/handlers/filesystem/squashfs.py
+++ b/unblob/handlers/filesystem/squashfs.py
@@ -61,13 +61,13 @@ class SquashFSv3Handler(_SquashFSBase):
             uint32 block_size;
             uint32 fragments;
             uint32 fragment_table_start_2;
-            int64  bytes_used;
-            int64  uid_start;
-            int64  guid_start;
-            int64  inode_table_start;
-            int64  directory_table_start;
-            int64  fragment_table_start;
-            int64  lookup_table_start;
+            uint64  bytes_used;
+            uint64  uid_start;
+            uint64  guid_start;
+            uint64  inode_table_start;
+            uint64  directory_table_start;
+            uint64  fragment_table_start;
+            uint64  lookup_table_start;
         } squashfs3_super_block_t;
     """
     HEADER_STRUCT = "squashfs3_super_block_t"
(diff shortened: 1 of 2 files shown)


unblob/handlers/filesystem/squashfs.py
```diff
@@ -61,7 +61,7 @@ class SquashFSv3Handler(_SquashFSBase):
             uint32 block_size;
             uint32 fragments;
             uint32 fragment_table_start_2;
-            int64  bytes_used;
+            uint64  bytes_used;
```
vlaci, line 64: We may not use the other fields but aren't they supposed to be the same types?

QKaiser: They are supposed to be the same type. Fixed.

vlaci (approved):

PR #178, opened 2022-01-19
Fix #177

### LINKED ISSUE #177: LZ4 triggers ValueError on malformed files
This was identified by structured fuzzing (valid signatures prefixed and suffixed by randomly generated content).

The LZ4 handler performs bytes to integer conversion using `convert_int8` and `convert_int32`. On malformed files, the buffer provided to these functions can be empty or smaller than the required value, leading to a ValueError being raised.

We should capture these exceptions and return immediately from the LZ4 handler.

QKaiser:
Fix #177 by catching ValueError raised by convert_int functions.

### DIFF
--- a/unblob/handlers/compression/lz4.py
+++ b/unblob/handlers/compression/lz4.py
@@ -92,7 +92,11 @@ class LegacyFrameHandler(_LZ4HandlerBase):
             if raw_bsize == b"":  # EOF
                 break
 
-            block_compressed_size = convert_int32(raw_bsize, Endian.LITTLE)
+            try:
+                block_compressed_size = convert_int32(raw_bsize, Endian.LITTLE)
+            except ValueError:
+                return
+
             if block_compressed_size in FRAME_MAGICS:
                 # next magic, read too far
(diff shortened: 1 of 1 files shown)


vlaci:
I see a pattern here and in the other PRs. My idea that these utility functions should all raise a common exception, like `InvalidInputFormat` or something, which we could catch globally and not log the backtrace.

QKaiser:
> I see a pattern here and in the other PRs. My idea that these utility functions should all raise a common exception, like `InvalidInputFormat` or something, which we could catch globally and not log the backtrace.
> 
> I don't really like the pattern of spreading this `try..except..return` pattern in the code

This applies to #176 , #178 , and #175 and I do agree that a global catch of `InvalidInputFormat` is a good solution. I do think however that we should keep catching errors emitted by built-in libraries in handlers (specifically #192 and #193).

QKaiser:
I still have the fuzzing test cases so we could pair program this global error handling together, check it against those cases and close #176, #178, #175.

vlaci:
Agreed

PR #191, opened 2022-01-21
Rework recursive process_file core calls 

### LINKED ISSUE #181: Rework recursive `process_file` core calls
Currently we have the unblob running around recursively calling the `process_file` function. 
* This makes the investigations and extending the core code harder. 
* In this way the parameters of the `process_file` are mixed: it contains variables, configuration globals, local states, etc....

We had an idea that instead of recursively calling it over and over again, we could have a simple `Queue`, where we could put `Tasks` which needs to be processed. This would potentially improve the code quality and add the ability to easily extend with multiprocessing.

Somehow related to:
* #71 
* #100
* #101

kukovecz:
Fixes #181

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,6 +46,7 @@ addopts = "--cov=unblob --cov=tests --cov-branch --cov-fail-under=90"
 [tool.vulture]
 paths = ["unblob/"]
 exclude = ["unblob/_py/"]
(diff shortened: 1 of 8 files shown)


kissgyorgy (approved):

QKaiser:
🚀

Issue #185, opened 2022-01-21
Negative seek in ZIP handler on malformed file


QKaiser:
This issue was identified through mutation fuzzing (valid integration test files run through radamsa).

Corrupted ZIP files can lead to this exception that is not caught in the ZIP handler:

```
Traceback (most recent call last):
  File "/home/quentin/unblob/unblob/finder.py", line 74, in search_chunks_by_priority
    chunk = handler.calculate_chunk(limited_reader, real_offset)
  File "/home/quentin/unblob/unblob/handlers/archive/zip.py", line 144, in calculate_chunk
    with zipfile.ZipFile(this_zip_chunk) as zip:
  File "/usr/lib/python3.8/zipfile.py", line 1269, in __init__
    self._RealGetContents()
  File "/usr/lib/python3.8/zipfile.py", line 1354, in _RealGetContents
    fp.seek(self.start_dir, 0)
ValueError: negative seek value -5
```

Issue #183, opened 2022-01-20
Unhandled BadZipFile exception in ZIP handler when parsing corrupted ZIP files


QKaiser:
This issue was identified through structured fuzzing (valid integration test files + corruption in the middle).

Corrupted ZIP files can lead to this exception that is not caught in the ZIP handler:

```
Traceback (most recent call last):
  File "/home/quentin/unblob/unblob/finder.py", line 74, in search_chunks_by_priority
    chunk = handler.calculate_chunk(limited_reader, real_offset)
  File "/home/quentin/unblob/unblob/handlers/archive/zip.py", line 144, in calculate_chunk
    with zipfile.ZipFile(this_zip_chunk) as zip:
  File "/usr/lib/python3.8/zipfile.py", line 1269, in __init__
    self._RealGetContents()
  File "/usr/lib/python3.8/zipfile.py", line 1392, in _RealGetContents
    x._decodeExtra()
  File "/usr/lib/python3.8/zipfile.py", line 468, in _decodeExtra
    raise BadZipFile("Corrupt extra field %04x (size=%d)" % (tp, ln))
zipfile.BadZipFile: Corrupt extra field 8ed8 (size=35731)
```

PR #192, opened 2022-01-21
Fix uncaught exceptions in zip handler

### LINKED ISSUE #182: Unhandled UnicodeDecodeError in ZIP handler when parsing corrupted ZIP files
This issue was identified through structured fuzzing (valid integration test files + corruption in the middle).

Corrupted ZIP files can lead to this exception that is not caught in the ZIP handler:

```
  File "/home/quentin/unblob/unblob/finder.py", line 74, in search_chunks_by_priority
  File "/home/quentin/unblob/unblob/handlers/archive/zip.py", line 144, in calculate_chunk
    with zipfile.ZipFile(this_zip_chunk) as zip:
  File "/usr/lib/python3.8/zipfile.py", line 1269, in __init__
    self._RealGetContents()
  File "/usr/lib/python3.8/zipfile.py", line 1371, in _RealGetContents
    filename = filename.decode('utf-8')
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x93 in position 3: invalid start byte
```

QKaiser:
Fix #182, #183, #185 by adding a validation function catching exceptions raised by zipfile builtin library. If an exception is caught, that means the ZIP file is corrupted and we should not emit a valid chunk.

### DIFF
--- a/unblob/handlers/archive/zip.py
+++ b/unblob/handlers/archive/zip.py
@@ -84,6 +84,19 @@ class ZIPHandler(StructHandler):
         self.cparser_le.end_of_central_directory_t(file)
         return file.tell()
 
+    def is_valid(self, zip_chunk: io.BytesIO) -> bool:
+        has_encrypted_files = False
+        try:
+            with zipfile.ZipFile(zip_chunk) as zip:
+                for zipinfo in zip.infolist():
+                    if zipinfo.flag_bits & ENCRYPTED_FLAG:
+                        has_encrypted_files = True
+            if has_encrypted_files:
+                logger.warning("There are encrypted files in the ZIP")
+            return True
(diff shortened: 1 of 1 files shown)


unblob/handlers/archive/zip.py
```diff
@@ -134,20 +147,12 @@ def calculate_chunk(
             end_of_zip = self._calculate_zipfile_end(file, start_offset)
         except MissingEOCDHeader:
             return
-
         file.seek(start_offset)
 
-        has_encrypted_files = False
-
         zip_content = file.read(end_of_zip - start_offset)
```
vlaci, line 152: Hmm, we shouldn't read the whole data into memory... I've opened #202 to not forget this issue

vlaci (approved):

Issue #184, opened 2022-01-20
Unhandled UnicodeEncodeError in unblob through corrupted CAB files


QKaiser:
It seems that corrupted CAB files can lead to weird filenames on disk when extracted by 7zip. When unblob walk through the extraction dir and tries to process that file, a `UnicodeEncodeError` is raised:

```
2022-01-20 18:30.41 [error    ] Unhandled exception during unblob 
Traceback (most recent call last):
  File "/home/quentin/unblob/unblob/cli.py", line 99, in main
    cli.invoke(ctx)
  File "/home/quentin/.cache/pypoetry/virtualenvs/unblob-U71Ryf-L-py3.8/lib/python3.8/site-packages/click/core.py", line 1395, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/quentin/.cache/pypoetry/virtualenvs/unblob-U71Ryf-L-py3.8/lib/python3.8/site-packages/click/core.py", line 754, in invoke
    return __callback(*args, **kwargs)
  File "/home/quentin/unblob/unblob/cli.py", line 81, in cli
    process_file(root, path, extract_root, max_depth=depth)
  File "/home/quentin/unblob/unblob/processing.py", line 39, in process_file
    process_file(root, path, extract_root, max_depth, current_depth + 1)
  File "/home/quentin/unblob/unblob/processing.py", line 62, in process_file
    process_file(
  File "/home/quentin/unblob/unblob/processing.py", line 39, in process_file
    process_file(root, path, extract_root, max_depth, current_depth + 1)
  File "/home/quentin/unblob/unblob/processing.py", line 31, in process_file
    log.info("Start processing file")
  File "/home/quentin/.cache/pypoetry/virtualenvs/unblob-U71Ryf-L-py3.8/lib/python3.8/site-packages/structlog/_log_levels.py", line 124, in meth
    return self._proxy_to_logger(name, event, **kw)
  File "/home/quentin/.cache/pypoetry/virtualenvs/unblob-U71Ryf-L-py3.8/lib/python3.8/site-packages/structlog/_base.py", line 204, in _proxy_to_logger
    return getattr(self._logger, method_name)(*args, **kw)
  File "/home/quentin/.cache/pypoetry/virtualenvs/unblob-U71Ryf-L-py3.8/lib/python3.8/site-packages/structlog/_loggers.py", line 111, in msg
    until_not_interrupted(self._write, message + "\n")
[...]

PR #194, opened 2022-01-21
Fix uncaught encoding errors.

### LINKED ISSUE #186: Uncaught UnicodeDecodeError in unblob extraction on malformed CAB files
This issue was identified through mutation fuzzing (valid integration test files run through radamsa).

It seems that corrupted CAB files can lead to weird filenames on disk when extracted by 7zip. When unblob tries to decode the stderr output, a UnicodeEncodeError is raised:

```
  File "/home/quentin/unblob/unblob/cli.py", line 99, in main
    cli.invoke(ctx)
  File "/home/quentin/unblob/unblob/cli.py", line 81, in cli
  File "/home/quentin/unblob/unblob/processing.py", line 61, in process_file
    for new_path in extract_valid_chunks(extract_dir, file, outer_chunks):
  File "/home/quentin/unblob/unblob/extractor.py", line 95, in extract_valid_chunks
    extracted = extract_with_command(extract_dir, carve_path, chunk.handler)
  File "/home/quentin/unblob/unblob/extractor.py", line 56, in extract_with_command
    res = subprocess.run(
  File "/usr/lib/python3.8/subprocess.py", line 495, in run
    stdout, stderr = process.communicate(input, timeout=timeout)
  File "/usr/lib/python3.8/subprocess.py", line 1028, in communicate
    stdout, stderr = self._communicate(input, endtime, timeout)
  File "/usr/lib/python3.8/subprocess.py", line 1910, in _communicate
    stderr = self._translate_newlines(stderr,
  File "/usr/lib/python3.8/subprocess.py", line 905, in _translate_newlines
    data = data.decode(encoding, errors)
UnicodeDecodeError: 'utf-8' codec can't decode bytes in position 146-147: invalid continuation byte
```

This is different than https://github.com/IoT-Inspector/unblob/issues/184

QKaiser:
A few encoding errors were identified during the initial fuzzing campaigns. This commit fix these errors at multiple locations.
    
#### Logging
- pathlib.Path are represented as POSIX path encoded as utf-8 using surrogateescape in case of encoding error.
- str values are encoded as utf-8 using surrogateescape in case of encoding error.

Test cases were modified accordingly.
    
#### Extraction
subprocess calls output (stderr, stdout) are now encoded as utf-8 using surrogateescape in case of encoding error.
    
#### Processing

path provided to process_file are checked for potential encoding error. If an error is triggered, the file won't be processed.
    
Path triggering encoding errors are not processed because, even if they can be printed out in logs, they are not properly handled by yara-python. We took this decision because encoding errors like these only arise on badly corrupted files, or specifically crafted malicious files.
    
These fixes relates to issues #186 and #184. This commit fixes both as they both relates to encoding errors triggered by corrupted files.

### DIFF
--- a/unblob/file_utils.py
+++ b/unblob/file_utils.py
@@ -4,6 +4,7 @@ import math
 import os
 import shutil
 import struct
+from pathlib import Path
 from typing import Iterator, Tuple
 
 from dissect.cstruct import cstruct
@@ -287,3 +288,11 @@ def read_until_past(file: io.BufferedIOBase, pattern: bytes):
             return file.tell()
         if next_byte not in pattern:
             return file.tell() - 1
+
+
+def valid_path(path: Path) -> bool:
+    try:
+        path.as_posix().encode("utf-8")
+    except UnicodeEncodeError:
+        return False
+    return True
--- a/unblob/logging.py
+++ b/unblob/logging.py
@@ -31,10 +31,11 @@ def _format_message(value: Any, extract_root: Path) -> Any:
 
     elif isinstance(value, Path):
         try:
-            return str(value.relative_to(extract_root))
+            new_value = value.relative_to(extract_root)
         except ValueError:
             # original files given to unblob may not be relative to extract_root
-            return str(value)
+            new_value = value
+        return new_value.as_posix().encode("utf-8", errors="surrogateescape")
 
     elif isinstance(value, Instance):
         return dumpstruct(value, output="string")
--- a/unblob/processing.py
+++ b/unblob/processing.py
@@ -10,7 +10,7 @@ import plotext as plt
 from structlog import get_logger
 
 from .extractor import carve_unknown_chunks, extract_valid_chunk, make_extract_dir
-from .file_utils import iterate_file
+from .file_utils import iterate_file, valid_path
 from .finder import search_chunks_by_priority
 from .iter_utils import pairwise
 from .logging import multiprocessing_breakpoint, noformat
@@ -101,6 +101,10 @@ def _process_task(  # noqa: C901
     log = logger.bind(path=task.path)
     if task.depth >= config.max_depth:
         log.info("Reached maximum depth, stop further processing")
+        return
+
+    if not valid_path(task.path):
+        log.warn("Path contains invalid characters, it won't be processed")
         return
 
     log.info("Start processing file")
--- a/tests/test_file_utils.py
+++ b/tests/test_file_utils.py
@@ -1,4 +1,5 @@
 import io
+from pathlib import Path
 from typing import List
 from unittest.mock import MagicMock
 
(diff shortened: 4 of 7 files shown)


unblob/file_utils.py
```diff
@@ -287,3 +288,11 @@ def read_until_past(file: io.BufferedIOBase, pattern: bytes):
             return file.tell()
         if next_byte not in pattern:
             return file.tell() - 1
+
+
+def valid_path(path: Path) -> bool:
```
kukovecz, line 293: I would add some simple tests for this function.

unblob/processing.py
```diff
@@ -98,6 +98,10 @@ def _process_task(  # noqa: C901
         log.info("Reached maximum depth, stop further processing")
         return
 
+    if not valid_path(task.path):
+        log.warn("Path contains invalid characters, it won't be processed")
+        return
+
```
kukovecz, line 104: Important decisions / features like this, should be in a separate commit message which only contains this few lines (and of course the `valid_path` function in `file_utils` and its tests).

But I am okay leaving this with the other changes right now, as the commit message explicitly writes about this decision and its reason.

kukovecz (commented):
Apart from the small things / questions I mentioned, it looks good!

unblob/logging.py
```diff
@@ -31,16 +31,22 @@ def _format_message(value: Any, extract_root: Path) -> Any:
...
-            return str(value)
+            return value.as_posix().encode("utf-8", errors="surrogateescape")
 
     elif isinstance(value, Instance):
         return dumpstruct(value, output="string")
 
     elif isinstance(value, int):
         return format_hex(value)
+    elif isinstance(value, str):
+        return value.encode("utf-8", errors="surrogateescape")
```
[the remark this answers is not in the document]
QKaiser: There was a smarter way of doing this. I need it to properly encode the stdout/stderr in case of errors, so I moved the encoding to the logger call in extractor (see https://github.com/IoT-Inspector/unblob/pull/194/files#diff-9f4b3ab3c14d83d69adeba706796c94983b997912f147d92738229b69fc11a16R69) and got rid of it in logging.py

unblob/file_utils.py
QKaiser: Added a test to test_file_utils with two unit tests (valid and invalid paths).

unblob/processing.py
QKaiser: Understood, will keep this in mind !

unblob/logging.py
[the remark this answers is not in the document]
kukovecz: Ok, makes sense! Just fix the logging tests by removing your `str`->`byte` conversion on the non-path expected values, and we are good to go :smile:

kukovecz (approved):

unblob/logging.py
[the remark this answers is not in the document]
QKaiser: Ahah, done. I blame it on writing code before the week-end.

PR #196, opened 2022-01-24
Fix path traversal

QKaiser:
Attempt to fix all potential issues due to path traversal identified in https://github.com/IoT-Inspector/unblob/issues/179

Will be kept as draft until we manage to fix all these:

- [x] replace tar extractor by 7z extractor in tar handler
- [x] fix unromfs extractor path traversal (see https://github.com/IoT-Inspector/unblob/pull/163)
- [x] fix unromfs extractor link path traversal (see https://github.com/IoT-Inspector/unblob/pull/163)
- [x] fix ubireader_extract_files path traversal (see https://github.com/IoT-Inspector/ubi_reader/commit/3b30234ffe117eb3d4f2a7a4e455c3fb4bca3668)
- [x] fix ubireader_extract_files link path traversal (fixed within unblob)
- [x] fix jefferson path traversal (see https://github.com/IoT-Inspector/jefferson/commit/41f654eac078cfec240428be457c054613a80851)
- [x] fix jefferson link path traversal (fixed within unblob)
- [x] fix unar link traversal (fixed within unblob)
- [x] fix unsquashfs link traversal (fixed within unblob)
- [x] write an integration test suite for symlink fixup function

While path traversal is a real issue that should be fixed in our 3rd party extractors, link based traversal is only problematic in platforms like ours. Instead of fixing link based traversals in 3rd party extractors, we should rather do it in unblob once files have been extracted.

### DIFF
--- a/tests/test_extractor.py
+++ b/tests/test_extractor.py
@@ -1,7 +1,14 @@
 import io
-from pathlib import Path
+from pathlib import Path, PosixPath
 
-from unblob.extractor import carve_unknown_chunks
+import pytest
+
+from unblob.extractor import (
+    carve_unknown_chunks,
+    fix_extracted_directory,
+    fix_permission,
+    fix_symlink,
+)
 from unblob.models import UnknownChunk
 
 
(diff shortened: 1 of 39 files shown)


QKaiser:
Made some progress, need to write proper unit testing for this.

tests/test_extractor.py
```diff
@@ -30,3 +33,125 @@ def test_multiple_chunks(self, tmp_path: Path):
         assert sorted(tmp_path.iterdir()) == [written_path1, written_path2]
         assert written_path1.read_bytes() == content[:4]
         assert written_path2.read_bytes() == content[4:]
+
+
+def test_fix_permissions():
+
+    with tempfile.TemporaryDirectory() as tempdir:
+        tmp_dir = Path(tempdir)
```
kukovecz, line 41: There is a pytest fixture, which can help you achieve this without using the `tempfile` lib at all: https://docs.pytest.org/en/6.2.x/tmpdir.html

e.g. These two are the same:

```python
def test_fix_permissions():
    with tempfile.TemporaryDirectory() as tempdir:
        tmp_dir = Path(tempdir)
        test_dir = tmp_dir / "dir_a"
```

```python
def test_fix_permissions(tmp_path: Path):
    test_dir = tmp_path / "dir_a"
```

unblob/extractor.py
```diff
@@ -46,6 +47,45 @@ def fix_permissions(outdir: Path):
             path.chmod(0o664)
 
 
+def is_safe_path(basedir: Path, path: Path) -> bool:
```
kukovecz, line 50: Just a fun thing to keep in mind (this has already came up in `logger.py`):

In python 3.9, there is an `is_relative_to` function in `pathlib`:
https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.is_relative_to

unblob/extractor.py
```diff
@@ -46,6 +47,45 @@ def fix_permissions(outdir: Path):
...
+def fix_symlink(path: Path, outdir: Path) -> Path:
+    """Fix symlinks by rewriting absolute symlinks to make them point within
+    the extraction directory (outdir), if it's not a relative symlink it is
+    either removed it it attempts to traverse outside of the extraction directory
+    or rewritten to be fully portable (no mention of the extraction directory
+    in the link value)."""
+    target = Path(os.readlink(path))
+
+    if target.is_absolute():
+        target = Path(target.as_posix().lstrip("/"))
```
kukovecz, line 67: I honestly don't understand this. Why can't we use `path.resolve()` whether it is absolute or not since later we check that if it is relative to the outdir anyways?

QKaiser: So the idea is that calling `resolve()` will resolve the full path, which will always be absolute due to pathlib inner working.

Quick illustration with two symlinks extracted in `outdir = '/tmp/out'`

1. link points to `/etc/passwd`, resolving the path will give `/etc/passwd` (absolute)
2. link points to `dir_a/file_1`, resolving the path will give `/tmp/out/dir_a/file_1` (absolute)

That's why I call `os.readlink()` rather than `path.resolve()`. It's the only decent way I found to fix symlinks by rewriting **actual** absolute symlinks to make them point within the extraction directory.

Makes sense ? I may be missing an obvious way to do that in a cleaner way :)

unblob/extractor.py
```diff
@@ -46,6 +47,45 @@ def fix_permissions(outdir: Path):
...
+    else:
+        target = path.resolve()
+
+    safe = is_safe_path(outdir, target)
+
+    if not safe:
+        print(f"Path traversal attempt through symlink, removing {target}.")
```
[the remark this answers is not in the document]
QKaiser: I agree with you that we should create a Report about that issue (e.g. `MaliciousSymlinkRemoved`) and log it, but I think we should still get rid of it. It's a really weird edge case where someone crafted a file with links like '../../../../test/file` given that all the other links were normalized (even `/etc/passwd`), so we must delete that file.

However, this brings up another subject we need to explore in more depth with the team: 7z handling of 'malicious' symlinks.

7z delete any symlink that points outside of the extraction directory and we don't have control over that behavior (no command line flag to force it to create those links). It does that for any format it can extract. It's great and we follow its behavior by deleting such symlinks that may have been kept by other extractors.

However, 7z validation somehow acts differently.

It will delete links pointing to files that do not exists on disk:

```
ERROR: Dangerous link path was ignored : etc/resolv.conf : /mnt/tmp/resolv.conf
ERROR: Dangerous link path was ignored : lib/modules/4.9.44/build : /home/mdq/new_svn/sys/nt98312/NA51039_BSP_V1.01/BSP/linux-kernel
```

It will delete links having a '../' traversal in them, even if it's entirely valid. To me the symlink below should have been created:

```
ERROR: Dangerous symbolic link path was ignored : usr/bin/awk : ../../bin/busybox
```

I'll investigate how 7z acts in more details and document it. We should then take a decision: either we keep 7z as is and accept how it works, or fork and remove the check (I don't like that idea given the burden of keeping the repo up to date and potentially performing some backporting).

Fun times ahead !

unblob/extractor.py
kukovecz: Yes, makes sense! I can't think of a cleaner way. I was thinking to play around with the `path.relative_to` function, but I think it would not help with symlinks pointing to non-existing targets. So this is fine by me.

tests/test_extractor.py
```diff
@@ -30,3 +32,110 @@ def test_multiple_chunks(self, tmp_path: Path):
         assert sorted(tmp_path.iterdir()) == [written_path1, written_path2]
         assert written_path1.read_bytes() == content[:4]
         assert written_path2.read_bytes() == content[4:]
+
+
+def test_fix_permission(tmpdir: Path):
```
martonilles, line 37: I would add a testcase to see if `fix_extracted_directory` is fixing permissions recursively

an interesting edge case:

testdir/ 0o200
testdir/testdir2/ 0x000
testdir/testdir2/testfile 0o200

unblob/extractor.py
```diff
@@ -36,14 +37,52 @@ def carve_chunk_to_file(carve_path: Path, file: io.BufferedIOBase, chunk: Chunk)
...
+
+def is_safe_path(basedir: Path, path: Path) -> bool:
+    try:
+        basedir.joinpath(path).resolve().relative_to(basedir.resolve())
+    except ValueError:
+        return False
+    return True
+
+
+def fix_symlink(path: Path, outdir: Path) -> Path:
```
martonilles, line 55: do we want to fix infinite symlinks as well, eg:
```
marci@indigo:/tmp/sym$ ls -lR 
.:
total 4
drwxrwxr-x 2 marci marci 4096 febr  15 12:34 a

./a:
total 0
lrwxrwxrwx 1 marci marci 1 febr  15 12:34 a -> .
```

unblob/extractor.py
```diff
@@ -36,14 +37,52 @@ def carve_chunk_to_file(carve_path: Path, file: io.BufferedIOBase, chunk: Chunk)
...
+
+    if target.is_absolute():
+        target = Path(target.as_posix().lstrip("/"))
+    else:
+        target = path.resolve()
+
+    safe = is_safe_path(outdir, target)
+
+    if not safe:
+        logger.warning(f"Path traversal attempt through symlink, removing {target}.")
```
martonilles, line 71: I would at least give an error here, nobody cares about warnings usually

unblob/extractor.py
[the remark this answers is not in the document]
QKaiser: I'll go through all my logs next week to identify which formats push 7z to simply unlink "dangerous" link paths. I also want to check for each format whether we should move to something else like patool (https://github.com/wummel/patool), which seems to not really care about those.

Will keep you posted !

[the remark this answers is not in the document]
vlaci: I just wanted to chime saying that it is extremely easy using Nix to apply out-of-tree patches to any dependency without much hassle.

unblob/extractor.py
QKaiser: Added a protection against recursive links (`is_recursive_link`) and wrote multiple integration tests to check both recursive links and link chains (`test_fix_symlink_chain`, `test_fix_symlink_chain_traversal`, `test_fix_recursive_symlink`).

tests/test_extractor.py
QKaiser: The `test_fix_extracted_directory` test takes care of this. That's a nice edge case, had to fix the function :)

unblob/extractor.py
QKaiser: Makes sense. Moved to error.

vlaci:
I had one comment on `jefferson` https://github.com/IoT-Inspector/jefferson/pull/4/files#r810998315

vlaci (approved):

PR #198, opened 2022-01-25
tar: adding tests for truncated content handling

vlaci:


QKaiser:
200 bytes truncation was not enough, I dug into it and moved it to 0x180 bytes.

Interestingly, if we successfully parse a tar header with tarfile.TarFile, it's virtually impossible to have an empty members array. 

The tarfile builtin library will either fail on parsing (in `tarfile.TarFile` constructor), or create a `tarfile.TarFile` object with at least one member in `members`. It's just how the format works.

QKaiser (approved):

PR #199, opened 2022-01-25
Calculate entropy

### LINKED ISSUE #70: Calculate entropy for UnknownChunks
- [x] calculate shannon_entropy
- [x] log record
- [x] Pretty graph with entropy calculated for chunks
- [x] calculate entropy for input files if we couldn't find anything in them

kissgyorgy:
Calculate entropy for all `UnknownChunk`s and draw an ASCII text plot on verbose mode.

Fixes #70

### DIFF
--- a/tests/test_handlers.py
+++ b/tests/test_handlers.py
@@ -44,7 +44,7 @@ def test_all_handlers(input_dir: Path, output_dir: Path, tmp_path: Path):
         path=input_dir,
         extract_root=tmp_path,
         max_depth=DEFAULT_DEPTH,
-        entropy_depth=1,
+        entropy_depth=0,
     )
 
     diff_command = [
--- a/tests/test_handlers.py
+++ b/tests/test_handlers.py
@@ -44,6 +44,7 @@ def test_all_handlers(input_dir: Path, output_dir: Path, tmp_path: Path):
         path=input_dir,
         extract_root=tmp_path,
         max_depth=DEFAULT_DEPTH,
+        entropy_depth=1,
     )
 
     diff_command = [
--- a/unblob/cli.py
+++ b/unblob/cli.py
@@ -65,6 +65,17 @@ def get_help_text():
     show_default=True,
     help="Recursion depth. How deep should we extract containers.",
 )
+@click.option(
+    "-n",
+    "--entropy-depth",
+    type=click.IntRange(0),
+    default=1,
+    show_default=True,
+    help=(
+        "Entropy calculation depth. How deep should we calculate entropy for unknown files? "
+        "1 means input files only, 0 turns it off."
+    ),
+)
 @click.option("-v", "--verbose", is_flag=True, help="Verbose mode, enable debug logs.")
 @click.option(
     "--show-external-dependencies",
@@ -74,12 +85,20 @@ def get_help_text():
     expose_value=False,
     is_eager=True,
 )
-def cli(files: Tuple[Path], extract_root: Path, depth: int, verbose: bool):
+def cli(
+    files: Tuple[Path],
+    extract_root: Path,
+    depth: int,
+    entropy_depth: int,
+    verbose: bool,
+):
     configure_logger(verbose, extract_root)
     logger.info("Start processing files", count=noformat(len(files)))
     for path in files:
         root = path if path.is_dir() else path.parent
-        process_file(root, path, extract_root, max_depth=depth)
+        process_file(
+            root, path, extract_root, max_depth=depth, entropy_depth=entropy_depth
+        )
 
 
 def main():
--- a/unblob/cli.py
+++ b/unblob/cli.py
@@ -60,8 +60,9 @@ def get_help_text():
 @click.option(
     "-d",
     "--depth",
-    type=int,
     default=DEFAULT_DEPTH,
+    type=click.IntRange(1),
+    show_default=True,
     help="Recursion depth. How deep should we extract containers.",
 )
 @click.option("-v", "--verbose", is_flag=True, help="Verbose mode, enable debug logs.")
--- a/unblob/cli.py
+++ b/unblob/cli.py
@@ -97,7 +97,12 @@ def cli(
     for path in files:
         root = path if path.is_dir() else path.parent
         process_file(
-            root, path, extract_root, max_depth=depth, entropy_depth=entropy_depth
+            root,
+            path,
+            extract_root,
+            max_depth=depth,
+            entropy_depth=entropy_depth,
+            verbose=verbose,
         )
 
 
--- a/unblob/processing.py
+++ b/unblob/processing.py
@@ -7,7 +7,7 @@ from typing import List
 import plotext as plt
 from structlog import get_logger
 
-from .extractor import carve_unknown_chunks, extract_valid_chunks, make_extract_dir
+from .extractor import carve_unknown_chunks, extract_valid_chunk, make_extract_dir
 from .file_utils import iterate_file
 from .finder import search_chunks_by_priority
 from .iter_utils import pairwise
@@ -83,7 +83,8 @@ def process_file(  # noqa: C901
             for carved_path in carved_paths:
                 calculate_entropy(carved_path, draw_plot=verbose)
 
-        for new_path in extract_valid_chunks(extract_dir, file, outer_chunks):
+        for chunk in outer_chunks:
+            new_path = extract_valid_chunk(extract_dir, file, chunk)
             process_file(
                 extract_root,
                 new_path,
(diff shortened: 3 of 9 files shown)


QKaiser:
> and draw an ASCII text plot on verbose mode

nerd alert ! 🤓

unblob/cli.py
```diff
@@ -60,10 +60,22 @@ def get_help_text():
...
     help="Recursion depth. How deep should we extract containers.",
 )
+@click.option(
+    "-n",
+    "--entropy-depth",
+    type=click.IntRange(0),
+    default=1,
+    show_default=True,
+    help=(
+        "Entropy calculation depth. How deep we should calculate entropy for unknown files? "
```
QKaiser, line 75:
```suggestion
        "Entropy calculation depth. How deep should we calculate entropy for unknown files? "
```

unblob/cli.py
```diff
@@ -60,10 +60,22 @@ def get_help_text():
 @click.option(
     "-d",
     "--depth",
-    type=int,
-    default=DEFAULT_DEPTH,
```
QKaiser, line 64: I don't understand why we replaced the explicit constant definition by implicit `10` values being scattered around. Is there a reason behind it ?

unblob/processing.py
```diff
@@ -119,3 +149,49 @@ def calculate_unknown_chunks(
...
+    plt.title("Entropy distribution")
+    plt.xlabel("mB")
+    plt.ylabel("entropy %")
+
+    plt.scatter(percentages, marker="dot")
+    # 16 height leaves no gapes between the lines
+    plt.plot_size(80, 16)
+    plt.ylim(0, 100)
+    # Always show 0% and 100%
+    plt.yticks(range(0, 101, 10))
```
QKaiser, line 195: Adding a limit to the X axis would be a good thing too. Given that we compute entropy on 1Mb chunks, we can draw a tick every 1Mb on the x axis.

```suggestion
    plt.yticks(range(0, 101, 10))
    plt.xticks(range(0, len(percentages)+1))
```

unblob/processing.py
```diff
@@ -119,3 +149,49 @@ def calculate_unknown_chunks(
...
+def calculate_entropy(path: Path, *, draw_plot: bool):
+    """Calculate and log shannon entropy divided by 8 for the file in 1mB chunks."""
+    logger.info("Calculating entropy for file", path=path)
+    percentages = []
+
+    # We could use the chunk size instead of another syscall,
+    # but we rely on the actual file size written to the disk
+    file_size = path.stat().st_size
+
+    # Smaller chuk size would be very slow to calculate. This takes ~ 3sec for a 4,5 GB file.
```
QKaiser, line 163: What if we cut the file into a fixed amount of chunks, with lower and upper limit to the chunk size (say the upper limit at 1MB and the lower limit at 1KB) ? This would provide more granularity to users looking into small files (e.g. encrypted config file) but keep a "constant" time regardless of the file being analyzed.

QKaiser (changes requested):
I'll wait for @kukovecz or @vlaci to review this branch prior to approving it. I tend to miss things otherwise :)

QKaiser:
I've been playing with plotext on this branch and I honestly love this ❤️

unblob/processing.py
```diff
@@ -1,26 +1,32 @@
...
+from .math import shannon_entropy
 from .models import UnknownChunk, ValidChunk
 
 logger = get_logger()
 
-DEFAULT_DEPTH = 10
 
-
-def process_file(
+# TODO: this function became too complex when adding entropy calculation, but
```
vlaci, line 21: We should plan a bit about how to do this refactor considering #191 as well. It may worth to create a config dataclass in this branch for the parameters having static values (extract_root, max_depth, entropy_depth, verbose)

unblob/processing.py
vlaci: Seems like a very good idea

unblob/processing.py
kukovecz: Yes, #191 got this refactored a bit. There, the `process_file` function is not called recursively, its purpose is only to be called:
1. From the `cli.py` - when using unblob as CLI tool
2. Directly - when using unblob as a 3rd party library

So keeping mind the 2nd use case, I think it is not a problem to have more (default) parameters, like `max_depth` or the `entropy_depth` here. But on that PR, I already got rid off the unnecessary stuff like `root` and `current_depth`.

If this gets accepted sooner, I don't mind rebasing and fixing #191 on top of this.

unblob/processing.py
```diff
@@ -119,3 +149,49 @@ def calculate_unknown_chunks(
         unknown_chunks.append(unknown_chunk)
 
     return unknown_chunks
+
+
+def calculate_entropy(path: Path, *, draw_plot: bool):
+    """Calculate and log shannon entropy divided by 8 for the file in 1mB chunks."""
```
kukovecz, line 155: Why `divided by 8`?  For bit -> byte conversion?
Since we have this docstring here, could you elaborate it more?

tests/test_handlers.py
```diff
@@ -43,7 +43,8 @@ def test_all_handlers(input_dir: Path, output_dir: Path, tmp_path: Path):
         root=input_dir,
         path=input_dir,
         extract_root=tmp_path,
-        max_depth=DEFAULT_DEPTH,
+        max_depth=10,
+        entropy_depth=1,
```
kukovecz, line 47: Since we are just giving the `__input__` dir to this test, I don't think the code will do anything with `entropy_depth=1`.
I hope it won't make things (so much) slower, but I would pass the same number to `max_depth` and `entropy_depth` for the tests.

unblob/processing.py
```diff
@@ -119,3 +149,49 @@ def calculate_unknown_chunks(
...
+    plt.ylabel("entropy %")
+
+    plt.scatter(percentages, marker="dot")
+    # 16 height leaves no gapes between the lines
+    plt.plot_size(80, 16)
+    plt.ylim(0, 100)
+    # Always show 0% and 100%
+    plt.yticks(range(0, 101, 10))
+
+    logger.debug("Entropy chart", chart=plt.build())
```
kukovecz, line 197: The log looks a little bit awkward:
```
2022-01-26 11:36.32 [debug    ] Entropy chart                  chart=                              Entropy distribution  
<plot>
```

I think if you use `chart=plt.build().lstrip())` instead, it will remove the spaces before the "Entropy distribution" text to look better.

kukovecz (commented):
I really love the modifications here :+1:

unblob/processing.py
QKaiser: Either that or insert a newline at the beginning. The "Entropy distribution" string is the title that goes on top of the chart, right in the middle (hence the whitespaces) :)

unblob/processing.py
vlaci: Some context which could be used to rephrase the doc-string:

> Shannon entropy returns the amount of information (in bits) of some numeric sequence. We calculate the average entropy of byte chunks, which in theory can contain 0-8bits of entropy. We normalize it for visualization to a 0-100% scale.

unblob/cli.py
kissgyorgy: I removed it because the module did not use it where it was defined anymore, but you are right, I should have thought about this more :)

unblob/processing.py
kissgyorgy: I already have code for that, which was on this branch, but #191 is the exact reason I removed it from this branch so we can discuss it further.

unblob/processing.py
kissgyorgy: Yes, shannon entropy outputs the possibility of a value based on the input. So if there are 8 possible values, it will return a number between 1 and 8. When we divide this by 8, we get a "percentage" value, which is easier to interpret IMO. I put something more elaborate in the docstring.

tests/test_handlers.py
kissgyorgy: Hmm, actually thinking about this a bit more, I would just turn off entropy calculation for the integration tests, as we don't need it at all.

unblob/processing.py
kissgyorgy: I did this:

```python
def calculate_buffer_size(
    file_size, *, chunk_count: int, min_limit: int, max_limit: int
) -> int:
    """Split the file into even sized chunks, limited by lower and upper values."""
    # We don't care about floating point precision here
    buffer_size = file_size // chunk_count
    buffer_size = max(min_limit, buffer_size)
    buffer_size = min(buffer_size, max_limit)
    return buffer_size
```

kukovecz (approved):

QKaiser (approved):

PR #214, opened 2022-01-28
Improve LimitedFileReader to protect against offset size overflow.

QKaiser:
###  Improve LimitedFileReader to protect against offset size overflow.
    
File seeking APIs on Unix systems expects an offset of type `off_t`, which is a signed 64 bits integer on modern systems. If we provide an offset higher or lower than the upper and lower limits of that integer type, Python will trigger a `ValueError: "cannot fit 'int' into an offset-sized integer"`.
    
As a defense against malformed chunks and mistakes in handler that may have flown under the radar, we inserted an offset size check in LimitedFileReader on calls to seek() and raised an InvalidInputFormat on such cases.

### YAFFS handler improvement.
    
Improve the YAFFS handler by raising the newly introduced InvalidInputFormat exception when no files were identified in the (supposed to be) YAFFS chunk.
    
The loop condition has been adapted no to get offsets past the EOF.
    
Page sizes array has been sorted in increasing order for readability.

Kept to two separate commits so that YAFFS handler and framework modifications are kept separate.


unblob/file_utils.py
```diff
@@ -11,6 +11,7 @@
 from .logging import format_hex
 
 DEFAULT_BUFSIZE = shutil.COPY_BUFSIZE  # type: ignore
+MAX_INT64 = 2 ** 63
```
vlaci, line 14: There is `sys.maxsize()` that holds a value for C longs for a given platform

unblob/file_utils.py
```diff
@@ -211,6 +212,8 @@ def __init__(self, file: io.BufferedIOBase, start: int):
         self._file.seek(start)
 
     def seek(self, offset: int, whence=io.SEEK_SET):
+        if (MAX_INT64 <= offset) or (offset <= -(MAX_INT64)):
```
vlaci, line 215: As `0` is encoded as "positive" number, the minimum supported value is `-2^63` but the maximum is `2^63-1`

vlaci (changes requested):

vlaci (approved):

PR #216, opened 2022-01-28
UnixCompress: Speed-up Yara rule

vlaci:
Regular expressions are much slower than literal patterns. This
transformation causes a 6x performance increase of the given stage.

### DIFF
--- a/unblob/handlers/compression/compress.py
+++ b/unblob/handlers/compression/compress.py
@@ -47,7 +47,7 @@ class UnixCompressHandler(StructHandler):
     YARA_RULE = r"""
         strings:
             // magic
-            $compress_magic = /[\x1f][\x9d][\x00-\xff]/
+            $compress_magic = { 1f 9d }
         condition:
             $compress_magic
     """


vlaci:
Total performance increase for a 1GiB high-entropy sample is also 3 fold :)

vlaci:
It may worth redoing the other rules as well. For the same random sample, `/[\x00][\x00][\x00][\x00][\x00][\x00][\x00][\x00]/` is 40 times slower than ` /\x00\x00\x00\x00\x00\x00\x00\x00/` or `{ 00 00 00 00 00 00 00 00 }`

QKaiser:
Handlers using regex instead of literal patterns; compress, ntfs, bzip2, gzip, lzh.

The LZH handler yara rule is super dumb btw, can't believe I wrote this.

vlaci:
Interestingly not every regex is the same. E.g. arc seems complicated but in reality it is super fast to execute

QKaiser:
Interesting read on the subject: https://github.com/Neo23x0/YARA-Performance-Guidelines

kissgyorgy (approved):
I think this can be merged and other simplifications can be done on separate branches.

PR #219, opened 2022-01-31
Improve Chunk error handling 

kukovecz:


QKaiser (approved):

unblob/extractor.py
```diff
@@ -66,15 +66,15 @@ def extract_with_command(
             stdout=subprocess.PIPE,
             stderr=subprocess.PIPE,
         )
-        if res.returncode != 0:
+        if res.returncode == 0:
```
vlaci, line 69: Non-0 return code doesn't mean that no output is generated.

unblob/processing.py
```diff
@@ -233,6 +233,10 @@ def calculate_entropy(path: Path, *, draw_plot: bool):
     # We could use the chunk size instead of another syscall,
     # but we rely on the actual file size written to the disk
     file_size = path.stat().st_size
+
+    if file_size == 0:
```
kissgyorgy, line 237: I would rather blow up, because this should never happen.

kukovecz: Yeah, we should reveal the underlying issue earlier instead. In this case, I found it that we mix the `start_offset` and the `end_offset` of an `UnknownChunk`, so we create and empty file.

I put this PR to draft and do the error handling on the `Chunk` level and fix this issue, then reopen and ping you when it is ready.

unblob/extractor.py
kukovecz: Right, good catch! The, I will modify the `fix_permissions` to handle when the `path` does not exist.

kukovecz: I honestly don't know what mysterious thing triggered the error I created this patch for, but I can not find or reproduce it. Theoretically it can not happen, because:
1. The directory in the `outdir` object is always created here, so at least an empty dir should be there
2. Even if the `outdir` `Path` is invalid, the `Path("<invalid>").rglob("*")` should return with an empty list in the `fix_permissions`, so it would not cause any problems...

So I will simply just drop this commit ... :smiling_face_with_tear:

QKaiser (approved):
I really like the move from error to simple debug message on overflows.

PR #227, opened 2022-02-03
Add SIGINT (ctrl+c in terminal) handling to Pool

kukovecz:


unblob/pool.py
```diff
@@ -67,13 +89,16 @@ def __init__(
...
         ]
+        self._should_stop_processing = False
+        _register_signal_handler()
 
     def start(self):
         for p in self._procs:
             p.start()
 
     def close(self):
-        self._input.join()
```
vlaci, line 76: After thinking about it a while, I think this join here is unsound: We either called `process_until_done` and let it do its job, then the `is_empty` call returning False is equivalent to `join` returning without blocking, or we didn't call `process_until_done` or it failed with an exception, in which case the queue never gets emptied and this line will block forever.

TL; DR we should never call `join`. Also, by removing that line, everything will terminate properly and we just need to silence the traceback printed by handling KeyboardInterrupt in the children.

unblob/pool.py
```diff
@@ -82,20 +107,33 @@ def submit(self, args):
...
+        except ShutDownRequired:
+            logger.debug("Shutdown MultiPool.process_until_done")
+            self._should_stop_processing = True
 
 
 class SinglePool(PoolBase):
     def __init__(self, handler, *, result_callback):
         self._handler = handler
         self._result_callback = result_callback
+        self._should_stop_processing = False
```
vlaci, line 124: I guess, there is no issue in single processing mode and we don't need to extend this class at all?

unblob/pool.py
```diff
@@ -67,13 +89,16 @@ def __init__(
             )
             for _ in range(process_num)
         ]
+        self._should_stop_processing = False
+        _register_signal_handler()
```
vlaci, line 93: I am really not fond of registering signal handlers from other places than the vicinity of the application's entrypoint. Also, libraries should never mess with the process's signal handlers from which they are called from.

unblob/pool.py
```diff
@@ -1,11 +1,29 @@
...
 
+logger = get_logger()
+
+
+class ShutDownRequired(Exception):
+    pass
+
+
+def _handle_signal(signum: int, _frame):
+    logger.debug(f"Signal ({signal.Signals(signum).name}) received")
```
vlaci, line 20: By default, Python will raise `KeyboardInterrupt` on CTRL-C. What is the added value here raising our own exception?

vlaci (changes requested):

vlaci:
Also I want to emphasis that signals to the children are only propagated by the terminal to the processes part of the foreground prcess group so it wont work for `kill -SIGINT <PID>`. Even for SIGTERM, the child processes could possibly be leaked.

unblob/pool.py
kukovecz: Just to hide the traces, I guess. But we can leave it out, making the SinglePool cleaner.

unblob/pool.py
kukovecz: Totally agree!

In offline, we talked about how to handle this. Our (newest) idea is to save the original signal handlers for the signals we want to create our own, then when we handled the signals and exited revert back the signal handlers to the original.
This way we can (hopefully) be sure that does not mess with the caller's signal setup.

unblob/pool.py
kukovecz: Right, at this point, nothing, really. 
However we decided that we should handle SIGTERM as well so at this point, we will need some custom exception. And since we have that, we should used the same for SIGTERM and SIGINT.

unblob/pool.py
```diff
@@ -1,11 +1,30 @@
...
 from .logging import multiprocessing_breakpoint
 
+logger = get_logger()
+
+
+class ShutDownRequired(Exception):
+    pass
+
+
+def _handle_signal(signum: int, _frame):
```
vlaci, line 20: we should also call the original signal handlers from ours

unblob/pool.py
```diff
@@ -72,20 +92,41 @@ def start(self):
...
+            try:
+                result = self._output.get()
+                if result is None:
+                    logger.debug("Shutdown MultiPool.process_until_done: done")
+                    break
+
+                self._result_callback(self, result)
+                self._input.task_done()
+            except ShutDownRequired:
+                logger.debug("Shutdown MultiPool.process_until_done: initiated")
```
vlaci, line 128:
```suggestion
                logger.debug("Shutdown requested, remaining tasks will be cancelled")
```

unblob/pool.py
```diff
@@ -72,20 +92,41 @@ def start(self):
...
 
     def process_until_done(self):
         while not self._input.is_empty():
-            result = self._output.get()
-            self._result_callback(self, result)
-            self._input.task_done()
+            try:
+                result = self._output.get()
+                if result is None:
+                    logger.debug("Shutdown MultiPool.process_until_done: done")
```
vlaci, line 122:
```suggestion
                    logger.debug("Shutting down pool, remaining tasks cancelled")
```

unblob/processing.py
```diff
@@ -78,6 +78,8 @@ def process_task(self, task: Task) -> TaskResult:
         result = TaskResult()
         try:
             self._process_task(result, task)
+        except ShutDownRequired:
```
vlaci, line 81: This will be called in the child process, I think we shouldn't catch this exception here.

vlaci (changes requested):

unblob/processing.py
kukovecz: Yes! I will remove it from here (and from the finder.py)

unblob/pool.py
kukovecz: I just re-set the signal handlers at the end of the `pool.close` to point back to the original handlers.

But you are right, we need to re-raise (re-fire?) those signals as well to let the original handlers handle it as they please. I made a separate commit about this to show its importance.

unblob/pool.py
```diff
@@ -72,11 +83,30 @@ def __init__(
             for _ in range(process_num)
         ]
         self._tid = threading.get_native_id()
+        self._signals_fired = []
+
+    def _handle_signal(self, signum: int, frame):
+        logger.debug(f"Signal ({signal.Signals(signum).name}) received")
```
vlaci, line 89: Based on our investigation of the possible deadlock introduced by the process pool approach I am a bit hesitant about this approach: If there is a mutex involved for printing log messages, this log line here can introduce a deadlock. It is generally not safe to cause any side-effect from a signal handler because of these issues creeping in.

P.s.
This also involves somewhat true for raising an exception as well, because we don't know if we are in a `try: ... except: pass` block for example when the signal arrives. (I guess this is true for KeyboardInterrupt as well...).

unblob/pool.py
```diff
@@ -85,18 +115,31 @@ def close(self):
     def _clear_input_queue(self):
         try:
             while True:
+                logger.warning("getnowait")
```
vlaci, line 118: Is this really a warning case? It would be printed once even if the queue was empty and we are in a normal shutdown, wouldn't it?

unblob/pool.py
```diff
@@ -85,18 +115,31 @@ def close(self):
...
                 self._input.get_nowait()
         except queue.Empty:
             pass
 
     def _request_workers_to_quit(self):
         for _ in self._procs:
+            logger.warning("sentinel")
```
vlaci, line 125: Is this really a warning case? It wouid be printed `-p` times even for normal shutdown

unblob/pool.py
```diff
@@ -85,18 +115,31 @@ def close(self):
...
             self._input.put(_SENTINEL)
 
     def _wait_for_workers_to_quit(self):
         for p in self._procs:
+            logger.warning("join")
```
vlaci, line 130: Is this really a warning case? It wouid be printed `-p` times even for normal shutdown

unblob/pool.py
```diff
@@ -85,18 +115,31 @@ def close(self):
...
             p.join()
 
+        # Set back the original signal handlers
+        for sig, handler in self._original_signal_handlers.items():
+            signal.signal(sig, handler)
+
+        # Call the original signal handler with the fired and catched signal(s)
```
vlaci, line 137: Wait, we would call the original handler even for normal shutdown doesn't involving a signal as well...

unblob/pool.py
```diff
@@ -1,13 +1,23 @@
...
 from typing import Any, Callable, Union
 
+from structlog import get_logger
+
 from .logging import multiprocessing_breakpoint
 
+logger = get_logger()
+
+
+class ShutDownRequired(Exception):
```
vlaci, line 18: I'd inherit from `BaseException` so that it escapes `try ... except Exception` blocks like `KeyboardInterrupt` does.

unblob/pool.py
```diff
@@ -107,9 +150,17 @@ def submit(self, args):
...
+                result = self._output.get()
+                if result is _SENTINEL:
+                    logger.debug("Shutting down pool, remaining tasks cancelled")
+                    break
+
+                self._result_callback(self, result)
+                self._input.task_done()
+            except ShutDownRequired:
+                logger.debug("Shutdown requested, remaining tasks will be cancelled")
+                self._output.put(_SENTINEL)
```
vlaci, line 163: I know that I came up with this design in the first place but it doesn't make much sense: We can just move the try block outside of the loop so that we can exit the loop without touching the output queue at all.

unblob/pool.py
```diff
@@ -72,11 +83,30 @@ def __init__(
...
+        raise ShutDownRequired
+
+    def _register_signal_handler(self, sig: signal.Signals):
+        return signal.signal(sig, self._handle_signal)
 
     def start(self):
         for p in self._procs:
             p.start()
 
+        # Save the original signal handlers while setting our own
```
vlaci, line 100: What if we add this signal handler setup-teardown machinery to its own context manager and just use it in the `process_until_done` call? That way we could separate its complexity from the pool's own complexity

unblob/pool.py
kukovecz: Wow, sorry, these new warnings were just were some debug messages for me during development, which somehow slip through and remained here. Sorry, I will remove them.

unblob/pool.py
kukovecz: No, I don't think it would.

This iterates over the `self._signals_fired` list of the `MultiPool`, which is empty by default and elements are only added when a signal is caught. On a normal shutdown, this list is empty and nothing should happen.

unblob/pool.py
```diff
@@ -44,6 +54,7 @@ def is_empty(self) -> bool:
 
 
 def _worker_process(handler, input, output):
+    os.setpgrp()
```
kissgyorgy, line 57: It's important enough for a comment here.

unblob/pool.py
vlaci: Ugh, totally missed that list :O

unblob/pool.py
```diff
@@ -72,11 +83,30 @@ def __init__(
...
         ]
         self._tid = threading.get_native_id()
+        self._signals_fired = []
+
+    def _handle_signal(self, signum: int, frame):
+        logger.debug(f"Signal ({signal.Signals(signum).name}) received")
+        self._signals_fired.append((signum, frame))
+        raise ShutDownRequired
+
+    def _register_signal_handler(self, sig: signal.Signals):
```
kissgyorgy, line 93: Again a pointless abstraction. Just call `signal.signal()` directly where you would call this method.

unblob/pool.py
kukovecz: Separating this logic from the Pool's logic is a great idea, but as the part of "signal handler setup-teardown machinery", we are:
* registering our new signal handlers: this needs to be done as early as possible, but I can live with it if we want to do it before the `process_until_done`.
* calling the original signal handlers: sadly, we can not do to that right after the `process_until_done`, but after the last part of the `close` of the Pool, right now. 

This makes the whole context_manager over process_until_done concept not ideal, I think.

unblob/pool.py
kukovecz: Maybe we can talk about this more, but at his point, I don't understand how do we want to do it. Of course we don't have to log if that makes things harder, but we also can't raise the exception here? The whole concept if based on raising the custom exception in the signal handler. If that introduces this level of uncertainty, we may to rethink the idea?

vlaci: The exception raising can remain, it is just a remark about its limitation (which applies to the out of the box ctrl c handling as well)

unblob/pool.py
vlaci: Why need it to be in the close?

kukovecz: Because if we (re-)raise the signal to the original handler, it will be handling it during the close, without allowing the Pool to gracefully exit.

vlaci (approved):

Issue #223, opened 2022-02-01
TAR handler yara rule can be more efficient


QKaiser:
The current yara rule for tar files only checks for the presence of the "ustar" magic, which can match on lots of other content like random text files.

```
strings:
    $tar_magic = { 75 73 74 61 72 }

condition:
    $tar_magic
```

The actual magic is:

- `ustar\x0000`  (ustar, null byte, then "00" for version) - POSIX version
- `ustar\x20\x20\x00`- old GNU tar format

We should edit the yara rule to match on POSIX version including the null byte and "00" version.

QKaiser:
Padding seems to be different between posix and gnu format, yay.

QKaiser:
You can test the different formats by creating test files this way:

```
tar --format=posix -cvf archive.posix.tar dir/*
tar --format=gnu -cvf archive.gnu.tar dir/*
```

With my test archive, the posix format is padded to `0x000a000` bytes, while the gnu format pads to `0x0007800`. The difference is exactly tar.RECORDSIZE (512 * 20), so it seems that posix use a blocking factor of 40 (?) rather than 20 when padding.

QKaiser:
Turns out the padding is the same in posix and gnu: `tar.RECORDSIZE (512 * 20)`. The difference is that posix pads the end of the file **twice**, while gnu pads the end of the file **once**. We just have to adjust the end offset calculation function for this.

QKaiser:
Just for fun, an example of why the previous rule was bad:

```
struct posix_header:
- name: b'er\x00MAC-Adresse\x00L2TP-Subnetzmaske\x00M\xc3\xb6chten Sie "D-Link Router Web Management" als Lesezeichen setzen?'
- mode: b'\x00erweite'
- uid: b'rt\x00Benut'
- gid: b'zername '
- size: b'oder Kennwor'
- mtime: b't ist nicht '
- chksum: b'korrekt.'
- typeflag: b'\x00'
- linkname: b'(Zwischen 8 und 63 Zeichen)\x00Sitzungsdaten konnten nicht generiert werden. F\xc3\xbchren Sie bitte einen Ne'
- magic: b'ustart'
- version: b' d'
--snip--
```

It didn't lead to wrong chunks, just useless computation cycles :)

PR #229, opened 2022-02-03
Add support for posix/pax tar format (#223)

martonilles:
The end offset calculation was broken for posix format, instead of using our
own offset calculation we should just use python's tarfile implementation.


QKaiser (changes requested):
Can you modify the integration tests files for tar too ? Ideally there should be one file per format (posix, gnu). They're located under `tests/integration/archive/tar`.

They can be generated this way:

```
tar --format=posix -cvf cherry.posix.tar *
tar --format=gnu -cvf cherry.gnu.tar *
```

You'll find out that tar pads the end of the file twice in posix mode. `_find_end_of_padding` should be adjusted accordingly.

QKaiser (approved):
Looks good to me.

tests/handlers/archive/test_tar.py
```diff
@@ -86,24 +184,43 @@ def test_offset(contents: bytes, expected_length: int, message: str):
...
-            len(TAR_CONTENTS),
+            True,
             "File is truncated but valid parts should be recovered",
             id="truncated",
         ),
     ),
 )
-def test_truncated_files(contents: bytes, expected_length: int, message: str):
-    f = io.BytesIO(contents)
+def test_truncated_files(contents: bytes, start_complete: bool, message: str):
```
kissgyorgy, line 214: These are 2 completely different test cases. It would easier to follow and debug in case of a failure and the implementation would be 4x simpler if you would just implement them in 2 separate functions without any parametrization.

Before the change, only the `contents` was different, which is good for a parametrized test, but after the change literally everything changed, so it doesn't fit into one test case anymore.

kissgyorgy (commented):
The test can be simplified.

PR #230, opened 2022-02-04
Add different verbose levels

### LINKED ISSUE #228: Add different verbose levels
Currently we have a `verbose` option  (`-v` / `--verbose` in the CLI) to have `debug` logs enabled.

However, we have tremendous amount of logs (which are really-really useful at different points) which just floods the CLI. We should have multiple levels of verbosity to help this.

kukovecz:
Fixes #228

### DIFF
--- a/unblob/finder.py
+++ b/unblob/finder.py
@@ -39,7 +39,7 @@ def search_chunks_by_priority(  # noqa: C901
     all_chunks = []
 
     for priority_level, handler_classes in enumerate(ALL_HANDLERS_BY_PRIORITY, start=1):
-        logger.info("Starting priority level", priority_level=noformat(priority_level))
+        logger.debug("Starting priority level", priority_level=noformat(priority_level))
         yara_rules = make_yara_rules(handler_classes)
         handler_map = make_handler_map(handler_classes)
         yara_results = search_yara_patterns(yara_rules, handler_map, path)
@@ -62,11 +62,12 @@ def search_chunks_by_priority(  # noqa: C901
                 if any(chunk.contains_offset(real_offset) for chunk in all_chunks):
                     continue
 
-                logger.info(
+                logger.debug(
                     "Calculating chunk for YARA match",
                     start_offset=offset,
                     real_offset=real_offset,
                     identifier=identifier,
+                    _verbosity=2,
                 )
 
                 limited_reader = LimitedStartReader(file, real_offset)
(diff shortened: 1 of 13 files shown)


QKaiser (approved):
Looks great. I like the fact that we have a shared function to use in our internal extractors such as unromfs. Levels are well defined, we just have to keep this in mind when writing new handlers :)

unblob/finder.py
```diff
@@ -77,13 +78,15 @@ def search_chunks_by_priority(  # noqa: C901
                         "File format is invalid",
                         exc_info=exc,
                         handler=handler.NAME,
+                        verbosity=2,
```
vlaci, line 81: isn't it `_verbosity`?

unblob/finder.py
```diff
@@ -77,13 +78,15 @@ def search_chunks_by_priority(  # noqa: C901
...
                     )
                     continue
                 except EOFError as exc:
                     logger.debug(
                         "File ends before header could be read",
                         exc_info=exc,
                         handler=handler.NAME,
+                        verbosity=2,
```
vlaci, line 89: isn't it `_verbosity`?

vlaci:
I am thinking about that can't we add our own override methods to the logger via the `wrapper_class` machinery?
Level 1 can remain debug, level 3 could be trace and we'd need to come up with a name for level 2. This way it would be less prone to typeos. Another thing we could do is to override the `debug` method to make the verbosity the first mandatory argument, like `debug(2, "whatever")`.

QKaiser:
> I am thinking about that can't we add our own override methods to the logger via the `wrapper_class` machinery? Level 1 can remain debug, level 3 could be trace and we'd need to come up with a name for level 2. This way it would be less prone to typeos. Another thing we could do is to override the `debug` method to make the verbosity the first mandatory argument, like `debug(2, "whatever")`.

In terms of levels, Apache uses these:
- debug
- trace1
- trace2
- traceX

We would only need to implement trace1 and trace2 for levels 2 and 3 if we go this way.

vlaci:
We were looking it into more deeply with @kukovecz and there is no easy composable way of implementing this on top of structlog. There is one global configuration for logger wrapping, so we cannot adjust the interface just for ourselves.

unblob/finder.py
kukovecz: Yeah, typo :/

unblob/finder.py
kukovecz: Yeah, typo :/

vlaci (approved):

PR #234, opened 2022-02-04
Fix overlapping valid chunks.

### LINKED ISSUE #232: valid chunk can overlap the next valid chunk
Here's the backtrace:

```
  File "/home/quentin/unblob/unblob/processing.py", line 80, in process_task
  File "/home/quentin/unblob/unblob/processing.py", line 127, in _process_task
    self._process_regular_file(task, size, result)
  File "/home/quentin/unblob/unblob/processing.py", line 133, in _process_regular_file
    unknown_chunks = calculate_unknown_chunks(outer_chunks, size)
  File "/home/quentin/unblob/unblob/processing.py", line 213, in calculate_unknown_chunks
    breakpoint()
  File "<attrs generated init unblob.models.UnknownChunk>", line 4, in __init__
    self.__attrs_post_init__()
  File "/home/quentin/unblob/unblob/models.py", line 61, in __attrs_post_init__
    raise InvalidInputFormat(
unblob.file_utils.InvalidInputFormat: Chunk has higher start_offset than end_offset: 0x2dd7f2-0x2cd81c
```

Looking at it with pdb, we see that a valid chunk emitted by the LZMA handler overlaps with the next valid chunk emitted by the squashfs handler:

```
(Pdb) p chunk.handler
<unblob.handlers.compression.lzma.LZMAHandler object at 0x7fe888980d90>
(Pdb) p next_chunk.handler
<unblob.handlers.filesystem.squashfs.SquashFSv4Handler object at 0x7fe88a0d9a90>
(Pdb) p chunk 
0x17f5d4-0x2dd7f2
(Pdb) p next_chunk
0x2cd81c-0x62a81c
```

Overlapping chunks should not happen. We take care of removing valid chunks within larger valid chunks in `remove_inner_chunks` but we do not take into account potential overlaps between valid chunks.

In this case, the error obviously comes from a miscalculation in the LZMA handler. I see two potential ways of dealing with this, to be performed after the call to remove_inner_chunks:

Getting rid of chunks that overlaps onto the next one:

```python
sorted_by_offset = sorted(chunks, key=attrgetter("start_offset"))
for chunk, next_chunk in pairwise(sorted_by_offset):
    if chunk.end_offset > next_chunk.start_offset:
        # delete chunk
```

Resizing chunks that overlaps onto the next one by setting their end_offset just before the start_offset of the next chunk:

```python
sorted_by_offset = sorted(chunks, key=attrgetter("start_offset"))
for chunk, next_chunk in pairwise(sorted_by_offset):
    if chunk.end_offset > next_chunk.start_offset:
        chunk.end_offset = next_chunk.start_offset - 1
```
      
This is to be considered as a defense in depth, and handlers should be written not to end up in this situation in the first place. 

Please note that this will continue to happen, especially in situations where the current chunk is a valid compressed chunk and that the next_chunk starts with data that is considered to be valid compressed data (which will end up decompressing to garbage but whatever). Knowing that, I think option two is the best.

QKaiser:
Due to the very nature of compression streams, we can't identify the exact end offset of a stream. In some situations this can lead to a valid compressed chunk overlapping the next chunk because the next chunk starts with data that is valid to the decompressor (even though it will be meaningless garbage when decompressed).

To overcome this situation, we introduce the `fix_chunk_overlaps` function that will browse through the list of valid chunks sorted by increasing start_offset and adjust any overlapping chunk by setting its end_offset to the next chunk start_offset value.

Fix #232

### DIFF
--- a/unblob/processing.py
+++ b/unblob/processing.py
@@ -130,6 +130,7 @@ class Processor:
         with task.path.open("rb") as file:
             all_chunks = search_chunks_by_priority(task.path, file, size, result)
             outer_chunks = remove_inner_chunks(all_chunks)
+            outer_chunks = fix_chunk_overlaps(outer_chunks)
             unknown_chunks = calculate_unknown_chunks(outer_chunks, size)
             if not outer_chunks and not unknown_chunks:
                 # we don't consider whole files as unknown chunks, but we still want to
@@ -189,6 +190,22 @@ def remove_inner_chunks(chunks: List[ValidChunk]) -> List[ValidChunk]:
         removed_inner_chunk_count=noformat(removed_count),
     )
     return outer_chunks
+
+
+def fix_chunk_overlaps(chunks: List[ValidChunk]) -> List[ValidChunk]:
+    """Fix potential overlaps between valid chunks. This can happen with
+    compression streams followed by a chunk starting with data that is
+    considered valid compressed data by the compression algorithm."""
+
+    if not chunks:
+        return []
+
+    sorted_by_offset = sorted(chunks, key=attrgetter("start_offset"))
+    for chunk, next_chunk in pairwise(sorted_by_offset):
+        if chunk.end_offset > next_chunk.start_offset:
+            chunk.end_offset = next_chunk.start_offset - 1
+
+    return chunks
 
 
 def calculate_unknown_chunks(


unblob/processing.py
```diff
@@ -191,6 +192,22 @@ def remove_inner_chunks(chunks: List[ValidChunk]) -> List[ValidChunk]:
...
+    compression streams followed by a chunk starting with data that is
+    considered valid compressed data by the compression algorithm."""
+
+    if not chunks:
+        return []
+
+    sorted_by_offset = sorted(chunks, key=attrgetter("start_offset"))
+    for chunk, next_chunk in pairwise(sorted_by_offset):
+        if chunk.end_offset > next_chunk.start_offset:
+            chunk.end_offset = next_chunk.start_offset - 1
```
kukovecz, line 206: I will be honest, I do not like this idea.

Once we have calculated and returned with a `ValidChunk`, with valid start and end offset, I think we should **never** modify those values, otherwise the whole operation goes to some weird, unexpected roads.

In my option:  if there are overlapping chunks, it always means some problem.  In those cases we should emit some error messages to the user and possibly remove the erroneous chunk from the valid chunk list (....remove both? ....) and later check the handlers and fix them. 

Of course, you know better the nature of these things than me, but I really don't want to mix things and do these modifications later - this is what the search_chunks / calculate_chunk is for. 

I personally would be really happy if we could solve this in some other way - but of course, no idea how :sweat_smile:

QKaiser: I can only agree with you that this is not the best of solutions. This situation can only happen with compression streams so I looked at each compression handler we currently have. The only ones that may lead to such a situation are LZW (compress), and LZMA. With others we can derive the actual offset from header fields, match on a unique end marker (e.g. bzip2), or the stream won't be valid if we read past it (e.g. gzip with isize and crc32). So for those overlapping chunks must never happen.

For LZMA and LZW there is no way we can make these handlers find the actual end offset if the next chunk starts with data that is legit to the decompressor. Overlaps can and will happen. It's just the nature of these compression formats.

We could just get rid of the overlapping chunk but I don't like the idea, and I kinda like the fact that we can find the real end offset of a compressed chunk by using the next valid chunk start address (seriously it's great).

What we could do to limit the domino effect of such a change would be to check whether the handler that created the overlapping chunk is one of our known "bad boys" (unix compress or LZMA). If it's one of those, we re-calculate the offset like I'm doing now. What do you think ?

kukovecz: Ok, I understand this is not the perfect way to do it, but if there is no anything better, of course I am happy with your idea. Limiting this for only the "bad bays" is a great addition, I would love that. And also to see some tests for this function.

Thanks for the improvement :+1:

QKaiser:
We concluded that fixing the handler rather than modifying unblob core is a better solution :) 

The LZMA fix is at #247 , I'm closing this one.

PR #236, opened 2022-02-04
pool: remove `join` call on close

vlaci:
The join in `close` is unsound: We either call `process_until_done` and let
it do its job, then the `is_empty` condition returning False is
equivalent to a `join` that would not block or we either didn't call
`process_until_done` or it failed with an exception, in which case the
queue never gets emptied and this line will block forever.

Instead of terminate, we place a sentinel object to the input queue so
that all processes have a chance to wake up and exit. In order to not
wait for arbitrary amount of time for the workers, we empty the input
queue from the main thread. In order to guarantee that this operation
is sound, we disallow placing subsequent tasks from other threads and
processes.

This change also means that the currently running task will always
finish.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,17 +12,17 @@ repos:
       - id: check-yaml
       - id: check-added-large-files
 
-  - repo: https://github.com/PyCQA/flake8
-    rev: 4.0.1
-    hooks:
-      - id: flake8
-        name: Check flake8
-
   - repo: https://github.com/psf/black
     rev: 21.12b0
     hooks:
       - id: black
         name: Check black
+
+  - repo: https://github.com/PyCQA/flake8
+    rev: 4.0.1
+    hooks:
+      - id: flake8
+        name: Check flake8
 
   - repo: https://github.com/PyCQA/isort
     rev: 5.10.1
--- a/unblob/pool.py
+++ b/unblob/pool.py
@@ -1,6 +1,8 @@
 import abc
 import multiprocessing as mp
+import queue
 import sys
+import threading
 from multiprocessing.queues import JoinableQueue
 from typing import Any, Callable, Union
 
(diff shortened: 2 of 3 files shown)


kukovecz (approved):

kukovecz (approved):

.pre-commit-config.yaml
```diff
@@ -12,18 +12,18 @@ repos:
...
     hooks:
       - id: black
         name: Check black
 
+  - repo: https://github.com/PyCQA/flake8
+    rev: 4.0.1
+    hooks:
+      - id: flake8
+        name: Check flake8
+
```
QKaiser, line 26: Seems like you just moved it from one place to another ? Not sure this should be part of the commit.

kukovecz: It usually worth to checking the commits for more info: https://github.com/IoT-Inspector/unblob/pull/236/commits

He did it directly to execute black prior to flake8.

unblob/pool.py
```diff
@@ -38,10 +40,12 @@ def is_empty(self) -> bool:
             return self._unfinished_tasks._semlock._is_zero()  # type: ignore
 
 
+_SENTINEL = ...
```
kissgyorgy, line 43: It's quite unusual to use Ellipsis for a sentinel object, because it's a singleton object like True and False. It also indicates unfinished code, for example in a function body.
A more common way is to make a unique value not used anywhere else in the codebase, which can be simply achieved by instantiating an empty object.
```
>>> a = ...
>>> b = ...
>>> a is b
True
>>> c = object()
>>> d = object()
>>> c is d
False
>>>
```

kukovecz: Neat!  I will fix this in https://github.com/IoT-Inspector/unblob/pull/227, I need to fix some issue in the Pool anyways.

Issue #237, opened 2022-02-04
Squashfs padding is off


vlaci:
Found in interesting case, before #229 was merged, [this file](https://downloads.openwrt.org/releases/21.02.0/targets/mvebu/cortexa9/openwrt-21.02.0-mvebu-cortexa9-linksys_wrt3200acm-squashfs-sysupgrade.bin) was identified as a squashfs, and produced a chunk like this: `3023360-7266816.squashfs_v4_extract`. After that, it produced a smaller TAR file: `0-7270400.tar_extract` producing truncated extracted contents.

martonilles:
This isn't actually a tar issue. Tar is working fine. The same file sizes are created when using tar command line.

This a problem of the squashfs extractor as in this case the sqashfs is 1k padded not 4k padded.
There is no standard how squashfs should be padded, in the kernel it depends on actually the kernel settings:
https://github.com/torvalds/linux/blob/5bfc75d92efd494db37f5c4c173d3639d4772966/fs/squashfs/Kconfig#L183

Userspace tools usually use 4k padding as that is compatible with both 1k and 4k padding, but there could be 1k padded squashfs files out there:
https://dr-emann.github.io/squashfs/squashfs.html

For our end calculation using 4k could easily result in an overflow. Unfortunately there is no easy way to detect this as the bytes_used is no indication in this matter. We might be able to see if 4k would overflow the file and just failback to 1k in those cases or set it to 1k always.

As originally files were 1k padded it might be OK to always use 1k as padding, bottom line we would miss a 3k block at the end as an unknown chunk which is probably 0 padded anyhow.
https://github.com/torvalds/linux/commit/7657cacf478940b995c2c73fdff981c13cc62c5c

martonilles:
Interestingly it is even possible to create filesystem that is not padded at all (however current mksquashfs can only do 4k padding, though other tools might be different)

https://github.com/plougher/squashfs-tools/blob/23996602b0f7db407914aa9479fc41cb12b4c732/USAGE#L245
https://github.com/plougher/squashfs-tools/blob/62b7f5a411767409d2ec782a92ed632099b422d5/ACKNOWLEDGEMENTS#L101

PR #238, opened 2022-02-04
Fixed squashfs padding (#237)

martonilles:
Squashfs files can be 1k padded not just 4k padded.
There is no standard how squashfs should be padded, in the kernel it
depends on actually the kernel settings:
https://github.com/torvalds/linux/blob/5bfc75d92efd494db37f5c4c173d3639d4772966/fs/squashfs/Kconfig#L183
https://github.com/torvalds/linux/commit/7657cacf478940b995c2c73fdff981c13cc62c5ctorvalds/linux@7657cac

Userspace tools usually use 4k padding as that is compatible with both
1k and 4k padding, but there could be 1k padded squashfs files out there:
https://dr-emann.github.io/squashfs/squashfs.html

For our end calculation using 4k could easily result in an overflow.
Unfortunately there is no easy way to detect this as the bytes_used is no
indication in this matter. We might be able to see if 4k would overflow
the file and just failback to 1k in those cases or set it to 1k always.

As originally files were 1k padded it might be OK to always use 1k as
padding, bottom line we would miss a 3k block at the end as an unknown
chunk which is probably 0 padded anyhow.

Interestingly it is even possible to create filesystem that is not
padded at all (however current mksquashfs can only do 4k padding, though
other tools might be different)
https://github.com/plougher/squashfs-tools/blob/23996602b0f7db407914aa9479fc41cb12b4c732/USAGE#L245
https://github.com/plougher/squashfs-tools/blob/62b7f5a411767409d2ec782a92ed632099b422d5/ACKNOWLEDGEMENTS#L101

### DIFF
--- a/unblob/handlers/filesystem/squashfs.py
+++ b/unblob/handlers/filesystem/squashfs.py
@@ -8,7 +8,7 @@ from ...models import StructHandler, ValidChunk
 
 logger = get_logger()
 
-PAD_SIZE = 4_096
+PAD_SIZE = 1_024
 BIG_ENDIAN_MAGIC = 0x73_71_73_68
 
 


QKaiser (commented):
I don't like the idea of emitting unknown chunk because we want to use the lower bound of a handler padding (which could be zero in the `-nopad` case).

A solution to support different kinds of padding (no padding, 1K, 4K, etc) would be to seek to the end of the squashfs content (using `header.bytes_used`) and then call `read_until_past(file, b"\x00")` to read up to the end of our padding.

The only thing we have to do then is to make sure the null bytes padding is at most 4096 bytes
and page aligned so we don't overlap with other chunks that may start with null bytes. Can be done with `round_down(observed_pad_size, 4096)`.

I would also create integration test files for each squashfs format with no padding, 1K padding, and 4K padding in order to cover every possibility and validate the code.

vlaci:
This how to figure out what padding should a chunk actually use is an interesting recurring question. It may make sense in the future to figure out a generic approach to that. E.g. the chunk returned from calcuilate_chunk could encode the unpadded size and an upper bound on padding based on the format it represents, and then the padding could be resolved in a separate phase when all the chunks are gathered. This sounds like a similar thing for me to #234 where we'd need to adjust the calculated chunk size anyway....

QKaiser:
> This how to figure out what padding should a chunk actually use is an interesting recurring question. It may make sense in the future to figure out a generic approach to that. E.g. the chunk returned from calcuilate_chunk could encode the unpadded size and an upper bound on padding based on the format it represents, and then the padding could be resolved in a separate phase when all the chunks are gathered. This sounds like a similar thing for me to #234 where we'd need to adjust the calculated chunk size anyway....

I'm strongly in favor of keeping the padding calculation within the handlers. I don't like the idea of telling potential contributors "your format handler should identify the end offset minus the padding (whatever it may be in size or content) and we'll take it from here".

martonilles:
Agree with keeping the logic within the handler as there could be other constrains besides lower/upped boundaries. I would rather add some helpers to it is easy to implement reoccurring padding calculations in handlers. Overlapping chunks in my opinion are the result of wrong or incomplete chunk handling and we should rather fix the chunk handler then try to come up with a general magic to fix those. (This is like putting a band aid on the problem) So far all the problems I have seen (at have a limited view) could be addressed by fixing the handler logic. Anytime we have overlapping chunks that signals a broken chunk detection.

 I will look at this padding and try to come up with some proposal.

tests/handlers/filesystem/test_squashfs.py
```diff
@@ -0,0 +1,184 @@
...
+00000110  00 00 00 00 00 00 00 00  00 00 01 02 00 16 78 da  |..............x.|
+00000120  63 60 00 83 06 08 c5 90  c0 80 0a 14 a0 b4 03 00  |c`..............|
+00000130  1b 68 01 41 00 00 00 00  00 00 01 1c 00 00 00 00  |.h.A............|
+00000140
+"""
+)
+
+
+def pad_contents(contents: bytes, pad_len: int):
+    return contents + b"\0" * (pad_len - len(contents))
```
QKaiser, line 89: This only works if contents length is lower than pad_len. I would also use the term `alignment` rather than `pad_len`.

I would suggest something like this:

```suggestion
def pad_contents(contents: bytes, alignment: int):
    content_size = len(contents)
    padded_size = round_up(content_size, alignment)
    return contents + b"\0" * (padded_size - size)
```

tests/handlers/filesystem/test_squashfs.py
```diff
@@ -0,0 +1,184 @@
...
+)
+@pytest.mark.parametrize(
+    "start",
+    (
+        pytest.param(b"", id="zero_start"),
+        pytest.param(b"A" * 128, id="non_zero_start"),
+    ),
+)
+@pytest.mark.parametrize(
+    "pad_len",
```
QKaiser, line 108: I'd prefer 'alignment' or 'pad_align'.

tests/handlers/filesystem/test_squashfs.py
```diff
@@ -0,0 +1,184 @@
...
+    if pad_len is None:
+        pad_len = len(contents)
+    complete_content = pad_contents(contents, pad_len)
+
+    chunk = handler_class().calculate_chunk(
+        io.BytesIO(start + complete_content + extra), start_offset
+    )
+
+    assert chunk.start_offset == start_offset
+    assert chunk.end_offset == start_offset + pad_len
```
QKaiser, line 153: If you agree with the comments above, this could work:

```suggestion
def test_squashfs_chunk_is_detected(
    contents: bytes, handler_class, start: bytes, alignment: Optional[int], extra: bytes
):
    start_offset = len(start)
    if alignment is None:
        alignment = 1
    complete_content = pad_contents(contents, alignment)

    chunk = handler_class().calculate_chunk(
        io.BytesIO(start + complete_content + extra), start_offset
    )
    padded_size = round_up(len(contents), alignment)
    assert chunk.start_offset == start_offset
    assert chunk.end_offset == start_offset + padded_size
```

QKaiser (changes requested):
Some nitpicking about padding length versus padding alignment and some optimizations when reading from the file.

unblob/handlers/filesystem/squashfs.py
```diff
@@ -17,6 +17,33 @@ class _SquashFSBase(StructHandler):
...
+        except EOFError:
+            return None
+
+        end_of_data_offset = start_offset + header.bytes_used
+        file.seek(end_of_data_offset)
+        padding = file.read(round_up(header.bytes_used, max(PAD_SIZES)))
+
+        for pad_size in sorted(PAD_SIZES, reverse=True):
+            size = round_up(header.bytes_used, pad_size)
+            padding_length = size - header.bytes_used
```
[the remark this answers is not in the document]
martonilles: I was stupid, obviously the right solution here is:
padding = file.read(round_up(header.bytes_used, max(PAD_SIZES)) - header.bytes_used)

this way we read the max padding (for 4k) and do not need any subsequent seek or read

unblob/handlers/filesystem/squashfs.py
```diff
@@ -17,6 +17,33 @@ class _SquashFSBase(StructHandler):
...
+    def calculate_chunk(
+        self, file: io.BufferedIOBase, start_offset: int
+    ) -> Optional[ValidChunk]:
+        file.seek(start_offset)
+        endian = get_endian(file, BIG_ENDIAN_MAGIC)
+
+        try:
+            header = self.parse_header(file, endian)
+        except EOFError:
+            return None
```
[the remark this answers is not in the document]
martonilles: good point

QKaiser (approved):

PR #240, opened 2022-02-07
Yara rules optimization

QKaiser:
Improved yara rules for the following handlers:

- unix compress
- LZMA
- CPIO binary
- GZIP
- LZH

These rules were either fixed (GZIP operating system field acceptable values), enhanced for speed improvements (LZH), or improved by offloading some fields validation to the yara engine in order to limit the amount of false positives.

The following checks were offloaded to yara:

- GZIP (validate modification time)
- CPIO (valide c_mode field)
- LZMA (validate dictionary size and uncompressed length fields)
- unix compress (validate the first LZW code)

### DIFF
--- a/unblob/handlers/archive/cpio.py
+++ b/unblob/handlers/archive/cpio.py
@@ -93,7 +93,34 @@ class BinaryHandler(_CPIOHandlerBase):
             $cpio_binary_magic= { c7 71 } // (default, bin, hpbin)
 
         condition:
-            $cpio_binary_magic
+            $cpio_binary_magic and
+            /**
+                check that c_mode file type is valid
+                C_ISBLK		060000
+                C_ISCHR		020000
+                C_ISDIR		040000
+                C_ISFIFO	010000
+                C_ISSOCK	0140000
+                C_ISLNK		0120000
+                C_ISCTG		0110000
+                C_ISREG		0100000
+            */
+            for any i in (0o1, 0o2, 0o4, 0o6, 0o10, 0o11, 0o12, 0o14) : (((uint16(@cpio_binary_magic + 6) & 0o770000) >> 0o14) == i) and
+
+            /**
+                check that c_mode sticky bits is valid
+                C_NONE      000000
+                C_ISUID		004000
+                C_ISGID		002000
+                C_ISVTX		001000
+            */
+            for any j in (0o0, 0o1, 0o2, 0o4): (((uint16(@cpio_binary_magic + 6) & 0o7000) >> 0o11) == j) and
+
+            /**
+                check that c_mode permissions are valid.
+                should be between 0o000 (0x0) (---------) and 0o777 (rwxrwxrwx)
+            */
+            ((uint16(@cpio_binary_magic + 6) & 0o777) <= 0o777)
     """
 
     C_DEFINITIONS = r"""
--- a/unblob/handlers/compression/compress.py
+++ b/unblob/handlers/compression/compress.py
@@ -44,11 +44,21 @@ class UnixCompressHandler(StructHandler):
     NAME = "compress"
 
     YARA_RULE = r"""
+        // reference: https://fuchsia.googlesource.com/third_party/wuffs/+/HEAD/std/lzw/README.md
         strings:
-            // magic followed by valid flags bytes (!(flag & 0x60) && (9 <= (flag & 0x1F) <= 16))
-            $compress_magic = { 1f 9d ( 09 | 0A | 0B | 0C | 0D | 0E | 0F | 10 | 89 | 8A | 8B | 8C | 8D | 8E | 8F | 90 ) }
+            $compress_magic = { 1f 9d }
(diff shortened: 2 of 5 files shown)


QKaiser:
Let's work on this first round of optimizations !

unblob/handlers/archive/cpio.py
```diff
@@ -93,7 +93,34 @@ class BinaryHandler(_CPIOHandlerBase):
...
+                C_ISGID		002000 (0x2)
+                C_ISVTX		001000 (0x1)
+            */
+            for any j in (0x0, 0x1, 0x2, 0x4): (((uint16(@cpio_binary_magic + 6) & 0xe00) >> 0x9) == j) and
+
+            /**
+                check that c_mode permissions are valid.
+                should be between 0o000 (0x0) (---------) and 0o777 (0x1ff) (rwxrwxrwx)
+            */
+            ((uint16(@cpio_binary_magic + 6) & 0x777) <= 0x1ff)
```
vlaci, line 123: It seems like [Yara supports](https://github.com/VirusTotal/yara/blob/1dab22e9f1e9fbb3e5464be19fd0d247ea583065/libyara/lexer.l#L562) octal literals natively. It is funny, that it isn't mentioned at all in the documentation though...

unblob/handlers/compression/gzip.py
```diff
@@ -38,12 +40,16 @@ class GZIPHandler(Handler):
...
-        $gzip_magic = /\x1f\x8b\x08[\x00-\x1f][\x00-\xff]{4}[\x00-\x04][\x00-\x0c\xff]/
+        // eXtra FLags (2 or 4 per RFC1952 2.3.1)
+        // Operating System (0-13, or 255 per RFC1952 2.3.1)
+        $gzip_magic = {{ 1F 8B 08 (00 | 01 | 02 | 03 | 04 | 05 | 06 | 07 | 08 | 09 | 0A | 0B | 0C | 0D | 0E | 0F | 10 | 11 | 12 | 13 | 14 | 15 | 16 | 17 | 18 | 19 | 1A | 1B | 1C | 1D | 1E) [4] (02 | 04) (00 | 01 | 02 | 03 | 04 | 05 | 06 | 07 | 08 | 09 | 0A | 0B | 0C | 0D | FF) }}
     condition:
-        $gzip_magic
-    """
+        $gzip_magic and
+        // timestamp is before 2039
+        uint32(@gzip_magic+4) < {timestamp}
```
vlaci, line 49: Uhm, the comment doesn't agree with the code here. Is there a real benefit by restricting the timestamp to not be in the future? What happens with time zones?

unblob/handlers/compression/compress.py
```diff
@@ -48,7 +48,9 @@ class UnixCompressHandler(StructHandler):
             // magic followed by valid flags bytes (!(flag & 0x60) && (9 <= (flag & 0x1F) <= 16))
             $compress_magic = { 1f 9d ( 09 | 0A | 0B | 0C | 0D | 0E | 0F | 10 | 89 | 8A | 8B | 8C | 8D | 8E | 8F | 90 ) }
         condition:
-            $compress_magic
+            $compress_magic and
+            // first code must be a literal
```
vlaci, line 52: A bit more context would be helpful here, what does it mean that it must be a literal?

unblob/handlers/compression/lzma.py
```diff
@@ -17,7 +17,14 @@ class LZMAHandler(Handler):
         strings:
             $lzma_magic = { 5d 00 00 ( 00 | 01 | 04 | 08 | 10 | 20 | 40 | 80) ( 00 | 01 | 02 | 04 | 08 ) }
         condition:
-            $lzma_magic
+            $lzma_magic and
```
vlaci, line 20: It is easier to figure out then in the case of unix compress where can I look into the spec but it would be nice if it was referred here as well.

unblob/handlers/compression/gzip.py
QKaiser: I've thought about it and just removed the timestamp check, the gzip_magic should be more than sufficient. I've seen gzip files with timestamp in the future so better not check that field.

unblob/handlers/archive/cpio.py
QKaiser: Oh wow. Moved to octal notation, which is way easier to understand in this context. I mentioned the link you provided in the commit message so future readers know where this comes from.

unblob/handlers/compression/compress.py
QKaiser: Added link to the compress format reference and explained the different kinds of codes in LZW.

unblob/handlers/compression/lzma.py
QKaiser: Added LZMA file format reference and link to specific sections of that reference for each validation we do.

vlaci (approved):

Issue #147, opened 2022-01-04
Implementation of `_get_extract_command()` is very fragile and environment-dependent


vlaci:
On my machine with Python 3.9.9, `test_show_external_dependencies_exists` fails with this message
```
    def test_get_extract_command(self):
>       assert self.DummyHandler._get_extract_command() == "testcommand"
E       AssertionError: assert ('testcommand', 'with', 'some', '-test', 'arguments') == 'testcommand'
E        +  where ('testcommand', 'with', 'some', '-test', 'arguments') = <bound method Handler._get_extract_command of <class 'test_models.TestHandler.DummyHandler'>>()
E        +    where <bound method Handler._get_extract_command of <class 'test_models.TestHandler.DummyHandler'>> = <class 'test_models.TestHandler.DummyHandler'>._get_extract_command
E        +      where <class 'test_models.TestHandler.DummyHandler'> = <test_models.TestHandler object at 0x7fd5db630a00>.DummyHandler
```

QKaiser:
I may be missing something related to abstract base classes here, but why are we using this convoluted way of returning the first value of our command array ?

We could just call it and get the first argument out.

```diff
diff --git a/unblob/models.py b/unblob/models.py
index 8bf344d..a58033c 100644
--- a/unblob/models.py
+++ b/unblob/models.py
@@ -124,7 +124,7 @@ class Handler(abc.ABC):
     @classmethod
     def _get_extract_command(cls) -> str:
         """Returns which (usually 3rd party CLI) command is used for extraction."""
-        return cls.make_extract_command.__code__.co_consts[1]
+        return cls.make_extract_command("", "")[0]
```

kukovecz:
This will be handled in the scope of #166

PR #241, opened 2022-02-07
Support handlers without extracting

### LINKED ISSUE #166: Support handlers without extraction methods
We plan on writing handlers for executable formats such as PE or ELF. It's useful to carve them out of larger chunks but there is no extraction or decompression method required given that the carved out file is all there is to it.

Unblob should support handlers with a `make_extract_command` function that is either undefined or returns an empty list. Whichever solution is best.

kukovecz:
Fixes #166, #147

### DIFF
--- a/unblob/models.py
+++ b/unblob/models.py
@@ -127,9 +127,13 @@ class Handler(abc.ABC):
         return []
 
     @classmethod
-    def _get_extract_command(cls) -> str:
+    def _get_extract_command(cls) -> Optional[str]:
         """Returns which (usually 3rd party CLI) command is used for extraction."""
-        return cls.make_extract_command.__code__.co_consts[1]
+        command = cls.make_extract_command("", "")
+        if not command:
+            return None
+
+        return command[0]
 
 
 class StructHandler(Handler):
--- a/unblob/models.py
+++ b/unblob/models.py
@@ -119,9 +119,12 @@ class Handler(abc.ABC):
         """Calculate the Chunk offsets from the Blob and the file type headers."""
 
     @staticmethod
-    @abc.abstractmethod
     def make_extract_command(inpath: str, outdir: str) -> List[str]:
-        """Make the extract command with the external tool, which can be passed for subprocess.run."""
+        """
+        Make the extract command with the external tool, which can be passed for subprocess.run.
+        Returns an empty list if the handler is not supposed to perform extractions.
+        """
+        return []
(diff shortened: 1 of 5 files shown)


unblob/models.py
```diff
@@ -119,9 +119,12 @@ def calculate_chunk(
         """Calculate the Chunk offsets from the Blob and the file type headers."""
 
     @staticmethod
-    @abc.abstractmethod
     def make_extract_command(inpath: str, outdir: str) -> List[str]:
-        """Make the extract command with the external tool, which can be passed for subprocess.run."""
+        """
+        Make the extract command with the external tool, which can be passed for subprocess.run.
+        Return with an empty list of the handler does not supposed to perform extraction.
```
QKaiser, line 125: I would use "Returns an empty list if the handler is not supposed to perform extractions."

unblob/processing.py
```diff
@@ -158,13 +159,21 @@ def _process_regular_file(self, task: Task, size: int, result: TaskResult):
...
-                new_dir = extract_with_command(
-                    extract_dir, carved_valid_path, chunk.handler, result
-                )
+                inpath, outdir = get_extract_paths(extract_dir, carved_valid_path)
+                command = chunk.handler.make_extract_command(str(inpath), str(outdir))
+
+                if not command:
+                    logger.debug(
+                        "No need to extract, as this handler does not have extractor",
+                        handler=chunk.handler.NAME,
```
QKaiser, line 168: I would increase the verbosity of this log to 2.

QKaiser (commented):
Except for small changes like verbose level and docstrings, this looks good to me. Maybe just  wait for @vlaci or @kissgyorgy to take a look prior to merging tho :)

unblob/models.py
kukovecz: Fixed, thanks!

unblob/processing.py
kukovecz: The verbosity thingy is in a pull request right now, not part of the main branch - but I like the way you are so familiar with it early on! :smile: 

I leave this conversation unresolved and either fix it here if the verbosity stuff merged first, or fix it there if this is merged first.

QKaiser: Let's do it once verbosity branch gets merged :)

QKaiser (approved):

vlaci (approved):

PR #244, opened 2022-02-08
profiling: write high level profiling data of the extraction process

vlaci:
To try the visualization without checking out this branch, here is a [sample file (zip)](https://github.com/IoT-Inspector/unblob/files/8364912/speescope_sample.zip). Extract it, then upload it to https://speedscope.app

Questions:
- [x] **Q**: Where should we put the profile files? It makes sense to put them next to the extract folder or also into the extract folder.
  **A**: As there can be multiple input files, but we could write a single profile for the whole run, so it makes sense to expect the output location as an argument.

### DIFF
--- a/tests/test_cleanup.py
+++ b/tests/test_cleanup.py
@@ -11,6 +11,7 @@ import pytest
 
 from unblob.models import Handler, Handlers, ValidChunk
 from unblob.processing import ExtractionConfig, process_file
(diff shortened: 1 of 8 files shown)


vlaci:
There is still some tests to write. I thought about creating a sample list of events and assert on the contents of the generated json. Do you have better ideas?

PR #248, opened 2022-02-08
Fix ZIP handler end offset calculation.

QKaiser:
This is a complete rewrite of the ZIP handler due to multiple issues that have been identified during testing.

**Python's built-in zipfile library is not reliable for our use.**

This is due to the fact that zipfile seeks to the very end of the file and then search backwards for the first occurence of central directory.

This is problematic because if we have a file fed to unblob that is made of two concatenated ZIP files (with or without arbitrary content in between), the end offset identified by zipfile would be the one of the second zip chunk within the overall file, which is wrong in our case.

**Simply finding the end of central directory marker is not enough**

We previously searched for the EOCD marker and consider our first match as the EOCD offset of the ZIP chunk we're currently investigating.

This is not sufficient because ZIP compression may not fully scramble the data of the files it contains. This means that if you have a ZIP within a ZIP, our first EOCD marker match will be the one of the ZIP within the ZIP chunk we're currently trying to find the end offset of.

To overcome this, we iterate over all EOCD marker offsets within the file and check for each of them whether or not their `offset_of_cd` value points to a valid magic for a central directory record header. If there is no match, that means our EOCD marker is not the right one.

This commit also get rid of the 'find_first' function as it turned into dead code and vulture won't let me commit if I leave it there.

### DIFF
--- a/unblob/handlers/archive/zip.py
+++ b/unblob/handlers/archive/zip.py
@@ -1,18 +1,17 @@
 import io
-import zipfile
-from typing import List, Optional, Tuple
+import struct
+from typing import List, Optional
 
+from dissect.cstruct import Instance
 from structlog import get_logger
 
-from ...file_utils import InvalidInputFormat, find_first
+from ...file_utils import InvalidInputFormat, iterate_patterns
 from ...models import StructHandler, ValidChunk
 
 logger = get_logger()
 
-
-MAXIMUM_VERSION = 0xFF
-EOCD_RECORD_HEADER = b"\x50\x4b\x05\x06"
 ENCRYPTED_FLAG = 0b0001
+EOCD_RECORD_HEADER = 0x6054B50
 
 
 class ZIPHandler(StructHandler):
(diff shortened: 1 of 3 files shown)


unblob/handlers/archive/zip.py
```diff
@@ -58,54 +65,49 @@ class ZIPHandler(StructHandler):
     """
     HEADER_STRUCT = "local_file_header_t"
```
martonilles, line 66: cd_file_header_t

unblob/handlers/archive/zip.py
```diff
@@ -58,54 +65,49 @@ class ZIPHandler(StructHandler):
...
-        if header.version_needed_to_extract > MAXIMUM_VERSION:
-            return
-
-        end_of_zip = self._calculate_zipfile_end(file, start_offset)
-        file.seek(start_offset)
+        has_encrypted_files = False
+        zip_end = -1
+        file.seek(start_offset, io.SEEK_SET)
+
+        for offset in iterate_patterns(file, struct.pack("<I", EOCD_RECORD_HEADER)):
```
martonilles, line 90: shell we handle offset == -1 case?

unblob/handlers/archive/zip.py
```diff
@@ -58,54 +65,49 @@ class ZIPHandler(StructHandler):
...
+            file.seek(offset)
+            end_of_central_directory = self.cparser_le.end_of_central_directory_t(file)
+            central_directory_offset = (
+                start_offset + end_of_central_directory.offset_of_cd
+            )
+            file.seek(central_directory_offset)
+            magic = convert_int32(file.read(4), Endian.LITTLE)
+            if magic == CD_RECORD_HEADER:
```
martonilles, line 98: we can check if offset == end_of_central_directory.offset_of_cd + end_of_central_directory.central_directory_size

unblob/handlers/archive/zip.py
```diff
@@ -58,54 +65,49 @@ class ZIPHandler(StructHandler):
...
 
-        is_valid, has_encrypted_files = self.check_file(file)
+        if zip_end == -1:
+            raise InvalidInputFormat("Missing EOCD record header in ZIP chunk.")
 
-        if not is_valid:
-            raise InvalidInputFormat("Invalid ZIP header.")
+        file.seek(zip_end)
+        self.cparser_le.end_of_central_directory_t(file)
+        end_of_zip = file.tell()
```
martonilles, line 110: maybe do the above check here?

unblob/handlers/archive/zip.py
QKaiser: Brilliant ! Love the fact that we don't need to read with this solution. I used your solution, but fixed it a bit as we need to take the start_offset of our chunk into account.

unblob/handlers/archive/zip.py
QKaiser: I switched to `end_of_central_directory_t` as it's the header content that's the most interesting to someone running unblob in debug mode (with colored struct being printed out when we call `self.parse_header`).

unblob/handlers/archive/zip.py
QKaiser: Not entirely sure of what you mean here.

vlaci (approved):

Issue #242, opened 2022-02-08
Extraction recursion depth calculation


vlaci:
Currently recursing subdirectories counts in the depth limit set for extraction. I find this a bit counterintuitive and would only make actual file extractions count in the depth limit. What do you think?

PR #249, opened 2022-02-09
process_file: don't take directories into account for the depth limit calculation

vlaci:
Resolves: #242


QKaiser (approved):
As I said during the shared review, this is way more intuitive now.

PR #254, opened 2022-02-16
Nix packaging

vlaci:
Nix derivation for unblob and all external dependencies. Got bitten by https://github.com/NixOS/nixpkgs/issues/117208 increasing the output size by 1.4GiB (!), I have cut down the bloat by rebuilding `gnustep-base`

Steps to try out is documented in the README https://github.com/IoT-Inspector/unblob/tree/nix#using-nix

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,51 @@ Help on usage:
 docker run --rm --pull always ghcr.io/iot-inspector/unblob:latest --help
 ```
 
+### Using Nix
+
+Unblob can be built and run using [Nix](https://nixos.org). The Nix
+derivation installs all 3rd party dependencies:
+
+1. [Install and configure Nix](https://nixos.org/download.html)
+
+    a. Optional: enable the experimental features so that you don't
+       need to pass `--extra-experimental-features "nix-command
+       flakes"` to `nix` command invocations:
+
+        ```console
+        $ cat > ~/.config/nix/nix.conf <<EOF
+        experimental-features = nix-command flakes
+        EOF
+        ```
+
+    b. Optional: use pre-built binaries from GitHub using [cachix](https://app.cachix.org/cache/unblob):
+
+        ```console
+        nix-env -iA cachix -f https://cachix.org/api/v1/install
+        cachix use unblob
+        ```
+
+2. Install unblob
+
+    ```console
+    $ nix profile install github:IoT-Inspector/unblob
+    $ unblob --show-external-dependencies
+    The following executables found installed, which are needed by unblob:
+        7z                          ✓
+        jefferson                   ✓
+        lz4                         ✓
+        lziprecover                 ✓
+        lzop                        ✓
+        simg2img                    ✓
+        tar                         ✓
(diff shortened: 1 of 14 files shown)


vlaci:
Until this PR is merged, all commands in the README should be using the link `github:IoT-Inspector/unblob/nix` so they use the correct branch.

README.md
```diff
@@ -24,6 +24,51 @@ Help on usage:
...
+        lzop                        ✓
+        simg2img                    ✓
+        tar                         ✓
+        ubireader_extract_files     ✓
+        ubireader_extract_images    ✓
+        unar                        ✓
+        unromfs                     ✓
+        unsquashfs                  ✓
+        yaffshiv                    ✓
+    ```
```
kukovecz, line 70: Do we want to include this output in the readme? Hopefully this will be expanding as we will have new extractors.

README.md
```diff
@@ -24,6 +24,51 @@ Help on usage:
...
+       need to pass `--extra-experimental-features "nix-command
+       flakes"` them to `nix` command invocations:
+
+        ```console
+        $ cat > ~/.config/nix/nix.conf <<EOF
+        experimental-features = nix-command flakes
+        EOF
+        ```
+
+    b. Optional, use pre-built binaries from GitHub using [cachix](https://app.cachix.org/cache/unblob):
```
kukovecz, line 44: We use `Optionally ....` above and `Optional, ....` here. We should make them looks like the same.

Maybe add a `[OPTIONAL]` prefix?

default.nix
```diff
@@ -0,0 +1,95 @@
...
+
+      jefferson = super.jefferson.overridePythonAttrs (_: {
+        propagatedBuildInputs = [
+          # Use the _same_ version as unblob
+          self.cstruct
+        ];
+      });
+
+      yara-python = super.yara-python.overridePythonAttrs (_: {
+        # Use _our_ patched version of yara
```
kukovecz, line 54: Don't know where to put this, but general question / concerns about patching: 

* Is this a common practice to patch dependencies "locally" like this? 
* Do you consider this as a temporary work-around, so we should get rid of this and update the dependencies as soon as we can?

nix/unar-light/default.nix
```diff
@@ -0,0 +1,5 @@
+{ unar, gnustep-light }:
```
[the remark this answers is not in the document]
vlaci: Yeah, I got rid of using this but forgot to remove from history

README.md
vlaci: I wanted to emphasis that this is a self-contained software and all required tools are installed as well.

default.nix
vlaci: Yes, given a new fixed release is available from these tools, we can get rid of these specific patches. On the other hand custom gnustep will be with us for quite some time I guess.

README.md
vlaci: I am open to ideas, how to emphasis this better.

kukovecz (approved):

PR #260, opened 2022-02-18
Disable log colors for non-tty stdouts

### LINKED ISSUE #256: Option to disable colored logs/outputs
When I pipe the logs into a less/file the colored output makes it impossible to read later, also makes processing it by script much harder.
We should either detect if the output is a terminal and only use colored logs or add a command line option to disable it.

Debugging logs from a file is really a pain now

kukovecz:
Fixes #256

### DIFF
--- a/unblob/logging.py
+++ b/unblob/logging.py
@@ -93,7 +93,7 @@ def configure_logger(verbosity_level: int, extract_root: Path):
         structlog.processors.UnicodeDecoder(),
         structlog.processors.StackInfoRenderer(),
         structlog.processors.format_exc_info,
-        structlog.dev.ConsoleRenderer(),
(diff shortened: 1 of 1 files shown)


unblob/logging.py
```diff
@@ -93,7 +93,7 @@ def configure_logger(verbosity_level: int, extract_root: Path):
         structlog.processors.UnicodeDecoder(),
         structlog.processors.StackInfoRenderer(),
         structlog.processors.format_exc_info,
-        structlog.dev.ConsoleRenderer(),
+        structlog.dev.ConsoleRenderer(colors=sys.stdout.isatty()),
```
vlaci, line 96: Bonus points to anyone who can figure out how structlog decides to use stdout. Been loking into a few places before I found out :)

vlaci (approved):

PR #267, opened 2022-02-23
Fix LZMA handler yara rule (properties field)

QKaiser:
Our yara rule was too restrictive in only matching on LZMA streams with default properties field. We changed the rule to accept any valid properties field, considering its definition from section 1.1.1 available at https://svn.python.org/projects/external/xz-5.0.3/doc/lzma-file-format.txt

### DIFF
--- a/unblob/handlers/compression/lzma.py
+++ b/unblob/handlers/compression/lzma.py
@@ -15,7 +15,16 @@ class LZMAHandler(Handler):
 
     YARA_RULE = r"""
         strings:
-            $lzma_magic = { 5d 00 00 ( 00 | 01 | 04 | 08 | 10 | 20 | 40 | 80) ( 00 | 01 | 02 | 04 | 08 ) }
+            $lzma_magic = {
+                // pre-computed valid properties bytes
+                (
+                    51 | 5A | 5B | 5C | 5D | 5E | 63 | 64 | 65 | 66 | 6C | 6D | 6E | 75 | 76 | 7E |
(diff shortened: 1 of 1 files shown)


unblob/handlers/compression/lzma.py
```diff
@@ -15,7 +15,8 @@ class LZMAHandler(Handler):
 
     YARA_RULE = r"""
         strings:
-            $lzma_magic = { 5d 00 00 ( 00 | 01 | 04 | 08 | 10 | 20 | 40 | 80) ( 00 | 01 | 02 | 04 | 08 ) }
+            // pre-computed valid properties bytes
+            $lzma_magic = { ( 51 | 5A | 5B | 5C | 5D | 5E | 63 | 64 | 65 | 66 | 6C | 6D | 6E | 75 | 76 | 7E | 87 | 88 | 89 | 8A | 8B | 90 | 91 | 92 | 93 | 99 | 9A | 9B | A2 | A3 | AB | B4 | B5 | B6 | B7 | B8 | BD | BE | BF | C0 | C6 | C7 | C8 | CF | D0 | D8 ) 00 00 ( 00 | 01 | 04 | 08 | 10 | 20 | 40 | 80) ( 00 | 01 | 02 | 04 | 08 ) }
```
vlaci, line 19: Can't we break this line up somehow? Or name the match groups?

QKaiser: yes we can :)

martonilles (approved):

Issue #96, opened 2021-12-06
Support out-of-tree extensions for handling additional file formats


vlaci:
_As a security researcher, I need a quick way to add support additional extractors, so that I can analyze currently unsupported firmware images._

An extension is just a Python file stored in a well-known and-or configurable location that should be picked-up without any hassle.

martonilles:
I would also consider how to make it easier to provide detection rules and extraction code. Currently it is a bit of extra work to attach an extractor also in many cases detection and extraction might share additional data (already detected by the handler) or code to parse/process some format.

martonilles:
- Might not need priorities between handlers (yara callbacks could also make priorities less useful)
- Could create pirorities based on handler type (eg: fs, compression, etc.)

PR #273, opened 2022-02-24
Out of tree handlers

martonilles:
#96

### DIFF
--- a/tests/extractors/test_command.py
+++ b/tests/extractors/test_command.py
@@ -1,0 +1,72 @@
+from pathlib import Path, PosixPath
+from unittest import mock
+
+import pytest
+
+from unblob.extractors import Command
+from unblob.extractors.command import InvalidCommandTemplate
+from unblob.models import TaskResult
+from unblob.report import ExtractCommandFailedReport, ExtractorDependencyNotFoundReport
+
+
+def test_command_templating():
+    command = Command("{infile}", "{outdir}", "{inpath},{outdir}")
+    cmdline = command._make_extract_command(
+        Path("inputdir") / "input.file", Path("output")
+    )
+
+    assert cmdline == ["input", "output", "inputdir/input.file,output"]
+
+
+@pytest.mark.parametrize(
+    "template",
+    (
+        "{no_such_placeholder}",
+        "{malformed",
+    ),
+)
+def test_command_templating_with_invalid_substitution(template):
+    command = Command(template)
+
+    with pytest.raises(InvalidCommandTemplate, match=template):
+        command._make_extract_command(Path("input"), Path("output"))
+
+
+def test_command_execution(tmpdir: Path):
+    outdir = PosixPath(tmpdir)
+    command = Command("sh", "-c", "> {outdir}/created")
+    res = TaskResult()
+    command.extract(Path("foo"), outdir, res)
+    assert (outdir / Path("created")).is_file()
+
+    assert res.reports == []
+
+
+def test_command_execution_failure(tmpdir: Path):
+    outdir = PosixPath(tmpdir)
+    command = Command("sh", "-c", ">&1 echo -n stdout; >&2 echo -n stderr; false")
+
+    res = TaskResult()
+    command.extract(Path("input"), outdir, res)
+    assert res.reports == [
+        ExtractCommandFailedReport(
+            handler=None,
+            command=mock.ANY,
+            stdout=b"stdout",
+            stderr=b"stderr",
+            exit_code=1,
+        )
+    ]
+
+
+def test_command_not_found(tmpdir: Path):
+    outdir = PosixPath(tmpdir)
+    command = Command("this-command-should-not-exist-in-any-system")
+
+    res = TaskResult()
+    command.extract(Path("input"), outdir, res)
+    assert res.reports == [
+        ExtractorDependencyNotFoundReport(
+            handler=None, dependencies=["this-command-should-not-exist-in-any-system"]
+        )
+    ]
--- a/tests/plugins/test_plugins.py
+++ b/tests/plugins/test_plugins.py
@@ -1,0 +1,35 @@
+from pathlib import Path
+
+import pytest
+
+from unblob.hookspecs import hookspec
+from unblob.plugins import UnblobPluginManager
+
+ASSETS_PATH = Path(__file__).parent / "__assets__"
+
+
+class TestHookSpecs:
+    @hookspec
+    def hook_callback():
+        pass
+
+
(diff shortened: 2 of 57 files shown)


QKaiser:
Is there a way we can make pytest load these out-of-tree handlers and look for integration tests for them in a specific directory ?

tests/extractors/test_command.py
```diff
@@ -0,0 +1,70 @@
...
+
+import pytest
+
+from unblob.extractors import Command
+from unblob.extractors.command import InvalidCommandTemplate
+from unblob.models import TaskResult
+from unblob.report import ExtractCommandFailedReport, ExtractorDependencyNotFoundReport
+
+
+def test_command_templating():
```
martonilles, line 12: add test for `infile` template

unblob/handlers/archive/dmg.py
```diff
@@ -82,10 +88,3 @@ def calculate_chunk(
             start_offset=start_offset - header.XMLLength - header.DataForkLength,
             end_offset=start_offset + len(header),
         )
-
-    @staticmethod
-    def make_extract_command(inpath: str, outdir: str) -> List[str]:
-        # This directory is here for simulating hardlinks and other metadata
```
martonilles, line 88: I would keep the comment and move to the EXTRACTOR definition

unblob/handlers/archive/zip.py
```diff
@@ -111,8 +114,3 @@ def calculate_chunk(
             end_offset=file.tell(),
             is_encrypted=has_encrypted_files,
         )
-
-    @staticmethod
-    def make_extract_command(inpath: str, outdir: str) -> List[str]:
-        # empty password with -p will make sure the command will not hang
```
martonilles, line 117: keep comment

unblob/cli.py
```diff
@@ -23,16 +25,46 @@ def show_external_dependencies(
...
+
+    dependencies = get_dependencies(handlers)
     text = pretty_format_dependencies(dependencies)
     exit_code = 0 if all(dep.is_installed for dep in dependencies) else 1
 
     click.echo(text)
     ctx.exit(code=exit_code)
 
 
+def add_handlers_from_extensions(
```
martonilles, line 43: add_handlers_from_plugins

unblob/cli.py
```diff
@@ -23,16 +25,46 @@ def show_external_dependencies(
...
+    if not plugins:
+        return handlers
+
+    logger.info("Loaded plugins", plugins=plugins)
+
+    extra_handlers = list(itertools.chain(*plugin_manager.hook.unblob_register_handlers()))  # type: ignore
+
+    logger.debug("Loaded handlers from plugins", handlers=extra_handlers)
+
+    return handlers.with_prepended(extra_handlers)
```
martonilles, line 63: do we always want to add extra handlers at first prio.
This can be useful for overriding builtin ones, but still wondering. Probably wont hurt.

martonilles:
We can also delete `unblob/extensions.py`

unblob/cli.py
vlaci: That seemed like a good default. A possible way to make it more flexible is to pass the Handlers object to the hook callback instead of requiring it to return a list of handlers. That way it would be forward-compatible with any future API extensions.

unblob/models.py
```diff
@@ -112,29 +152,14 @@ class Handler(abc.ABC):
     # (e.g. tar magic is in the middle of the header)
     YARA_MATCH_OFFSET: int = 0
 
+    EXTRACTOR: Extractor = NullExtractor()
```
kukovecz, line 155: I think we should have some (hopefully integration?) test on a handler, which does not override the `EXTRACTOR` and uses `NullExtractor`, should not we?

kukovecz (commented):
I left a few really small comments here and there, but overall looks nice, wonderful magic done with pluggy!

unblob/processing.py
```diff
@@ -161,24 +174,36 @@ def _process_regular_file(self, task: Task, size: int, result: TaskResult):
...
+            fix_extracted_directory(outdir, result)
+
+            result.add_new_task(
+                Task(
+                    root=self._extract_root,
+                    path=outdir,
+                    depth=task.depth + 1,
                 )
+            )
+        except Exception as exc:
```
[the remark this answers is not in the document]
vlaci: The idea is whatever happens here, we want to continue with subsequent chunks. I have no strong opinion about covering `fix_extracted_directory` with this block though.

unblob/plugins.py
```diff
@@ -0,0 +1,71 @@
...
+        Multiple files
+          A directory containing one or more Python files or packages.
+          Files are loaded from the directory root.  Additional
+          sub-directories containing Python packages (directories with
+          ``__init__.py``) are loaded recursively.
+        """
+        logger.debug("Importing plugin modules", path=path)
+
+        if path.is_file():
+            # (module_name, path)
```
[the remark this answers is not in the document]
vlaci: I've almost created an attrs class for that but found it overkill. I don't know what further elaboration would be helpful here.

tests/plugins/test_plugins.py
```diff
@@ -0,0 +1,35 @@
...
+    [
+        pytest.param(path, id=path.stem)
+        for path in ASSETS_PATH.iterdir()
+        if path.stem != "__pycache__"
+    ],
+)
+def test_plugin(plugin_manager, plugin_path):
+    plugin_manager.import_path(plugin_path)
+
+    assert set(plugin_manager.hook.hook_callback()) == set(["It Works", "It Works Too"])  # type: ignore
```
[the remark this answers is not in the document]
vlaci: Fair point

unblob/cli.py
```diff
@@ -23,16 +25,46 @@ def show_external_dependencies(
...
 
+def add_handlers_from_plugins(
+    plugin_manager: UnblobPluginManager,
+    handlers: Handlers,
+    plugin_path: Optional[Path],
+):
+    if plugin_path:
+        plugin_manager.import_path(plugin_path)
+
+    plugin_manager.load_setuptools_entrypoints("unblob_v1")
```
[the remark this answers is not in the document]
vlaci: This is the version of the plugin api. If we were to introduce a breaking change, we'd need to change this as well. It may make sense to move this to a const somewhere.

unblob/models.py
vlaci: Frankly, I am inclined to remove this from this PR and introduce when needed.

kukovecz (approved):

PR #278, opened 2022-03-01
Replace extfs filesystem handler extractor.

### LINKED ISSUE #277: 7z support of extfs is insufficient
I found valid samples of ExtFS filesystems (ext2, ext3) that cannot be extracted by 7z. We need to find a better extractor for those systems.

I successfully tested debug fs, so a potential fix would be:

```diff
@staticmethod
def make_extract_command(inpath: str, outdir: str) -> List[str]:
-        return ["7z", "x", "-y", inpath, f"-o{outdir}"]
+        return ["debugfs", inpath, "-R", f"rdump / {outdir}"]
```

QKaiser:
We replaced 7z by debugfs given that 7z cannot extract all the extfs samples we have in our corpus.

Fix #277

### DIFF
--- a/.github/actions/setup-dependencies/action.yml
+++ b/.github/actions/setup-dependencies/action.yml
@@ -7,7 +7,7 @@ runs:
   using: "composite"
   steps:
     - name: Install 3rd party from apt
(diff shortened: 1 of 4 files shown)


vlaci:
Kudos for touching the nix stuff :+1:

PR #283, opened 2022-03-02
Revert to 7zip v16.02 from 7zip v21.07.

### LINKED ISSUE #258: Binary name requirement of 7-Zip is misleading/inconsistent
We use the command `7z` which refers to the `p7zip` port of 7-Zip but we require the original which is either named `7zz` or `7zzs`. We also rename the binary ourselves when we are installing it to docker. It'd be better if we could pickup either `7zz` or `7zzs` as well.

QKaiser:
We were relying on 7zip version 21.07 from the official repo due to the lack of support for ext4 filesystems in 7zip version 16.02 provided by the p7zip-full package.

We replaced the extfs extractor by debugfs since 7f77c7d, which means we can revert back to using p7zip-full to simplify our build tasks and make life easier on end users who would like to test without performing manual 'copy/install'.

Fix #258

### DIFF
--- a/.github/actions/setup-dependencies/action.yml
+++ b/.github/actions/setup-dependencies/action.yml
@@ -7,15 +7,7 @@ runs:
   using: "composite"
   steps:
     - name: Install 3rd party from apt
-      run: sudo apt install e2fsprogs unar zlib1g-dev liblzo2-dev lzop lziprecover img2simg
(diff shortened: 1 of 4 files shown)


vlaci:
Also changed in nix derivation

vlaci (approved):

PR #284, opened 2022-03-02
Add support for non-standard SquashFS filesystems.

QKaiser:
We added support for the following non-standard formats:

- DD-WRT signature (hsqt)
- Broadcom signature (shsq), used by Netgear, Cisco, D-Link, ...
- non-standard signature (zlqs)
- AVM squashfs version 4 big-endian

We created integration test cases for each of those new formats. We also added a dedicated unit test to SquashFS v4 big endian to test_squashfs.py

These changes rely on the presence of our fork of sasquatch with both sasquatch and sasquatch-v4be utils being installed (available since e193711).


vlaci (approved):

vlaci:
> God damnit, I forgot to pull your changes @vlaci before rebasing and pushing back. Can you rebase your local branch of file-format-non-standard-squashfs with origin/main and force push back ?

Done. Also, consider using `--force-with-lease` for push-es. That way, even if you accidentally forgot to rebase, you can checkout the original revision shown by git push when it returns.

PR #286, opened 2022-03-03
poetry: refresh lockfile

vlaci:
It was outdated.

Generated with `poetry lock --no-update`

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -69,7 +69,7 @@ python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*"
 
 [[package]]
 name = "coverage"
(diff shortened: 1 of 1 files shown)


vlaci:
Something is not okay around jefferson. @QKaiser could you take a look?

QKaiser:
> Something is not okay around jefferson. @QKaiser could you take a look?

The issue comes from LZO decompression that was added to jefferson, which means the expected output is not the same (it never was right in the first place).

Interestingly, the tests are passing on my development machine but not in Docker. Probably due to LZO bindings but I don't see why exactly yet. Investigating this in https://github.com/IoT-Inspector/unblob/pull/287

vlaci:
Resolved in #287

PR #288, opened 2022-03-04
Check lockfile

vlaci:
There will be a `poetry lock --check` command whenever Poetry 1.2.0 is released. Until then, we can still check if locking produces an idempotent result.

Workflow is only ran when `pyproject.toml` or `poetry.lock` changes so it won't cause any overhead for most PRs

### DIFF
--- a/.github/workflows/poetry.yml
+++ b/.github/workflows/poetry.yml
@@ -1,0 +1,24 @@
+name: Poetry
+
+on:
+  pull_request:
+    paths:
+      - 'pyproject.toml'
+      - 'poetry.lock'
+
+jobs:
+  check_poetry:
+    name: Check poetry.lock consistency
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -34,6 +34,7 @@ pyright = "^0.0.12"
 pre-commit = "^2.15.0"
 pytest-cov = "^3.0.0"
 setuptools-rust = "^1.1.2"
+pluggy = "1.0.0"
```
QKaiser, line 37: Is this a leftover from the out-of-tree handlers branch ?

vlaci: Nope, check the commit message:
>  DO NOT MERGE: test poetry.lock consitency checker

QKaiser: Aaaah I see :)

QKaiser (approved):

PR #290, opened 2022-03-09
Optimize chunk extraction

### LINKED ISSUE #275: carve_chunk_to_file use sendfile instead of a read/wrtite loop
read write loop is slow and resource intensive, we could use os.sendfile

low-prio

e3krisztian:
fixes #275

### DIFF
--- a/tests/test_extractor.py
+++ b/tests/test_extractor.py
@@ -20,21 +20,29 @@ class TestCarveUnknownChunks:
 
     def test_one_chunk(self, tmp_path: Path):
         content = b"test file"
-        test_file = io.BytesIO(content)
+        infile = tmp_path / "infile"
+        infile.write_bytes(content)
         chunk = UnknownChunk(0, len(content))
-        carve_unknown_chunks(tmp_path, test_file, [chunk])
+
+        with infile.open("rb") as test_file:
+            carve_unknown_chunks(tmp_path, test_file, [chunk])
+
         written_path = tmp_path / "0-9.unknown"
-        assert list(tmp_path.iterdir()) == [written_path]
+        assert set(tmp_path.iterdir()) == {infile, written_path}
         assert written_path.read_bytes() == content
 
     def test_multiple_chunks(self, tmp_path: Path):
         content = b"test file"
-        test_file = io.BytesIO(content)
+        infile = tmp_path / "infile"
+        infile.write_bytes(content)
         chunks = [UnknownChunk(0, 4), UnknownChunk(4, 9)]
-        carve_unknown_chunks(tmp_path, test_file, chunks)
+
+        with infile.open("rb") as test_file:
+            carve_unknown_chunks(tmp_path, test_file, chunks)
+
         written_path1 = tmp_path / "0-4.unknown"
         written_path2 = tmp_path / "4-9.unknown"
-        assert sorted(tmp_path.iterdir()) == [written_path1, written_path2]
+        assert set(tmp_path.iterdir()) == {infile, written_path1, written_path2}
         assert written_path1.read_bytes() == content[:4]
         assert written_path2.read_bytes() == content[4:]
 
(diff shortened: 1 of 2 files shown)


tests/test_extractor.py
```diff
@@ -20,21 +20,29 @@ def test_no_chunks(self, tmp_path: Path):
 
     def test_one_chunk(self, tmp_path: Path):
         content = b"test file"
-        test_file = io.BytesIO(content)
+        infile = tmp_path / "infile"
```
kissgyorgy, line 23: When you change implementation, but the feature is the same, tests should not be changed at all, which would show that the features we put there in the first place still work.

tests/test_extractor.py
```diff
@@ -20,21 +20,29 @@ def test_no_chunks(self, tmp_path: Path):
...
+
         written_path = tmp_path / "0-9.unknown"
-        assert list(tmp_path.iterdir()) == [written_path]
+        assert set(tmp_path.iterdir()) == {infile, written_path}
         assert written_path.read_bytes() == content
 
     def test_multiple_chunks(self, tmp_path: Path):
         content = b"test file"
-        test_file = io.BytesIO(content)
+        infile = tmp_path / "infile"
```
kissgyorgy, line 36: Same as above, tests should not be touched, when the feature doesn't change.

kissgyorgy (changes requested):
The commit message doesn't say anything about WHY the interface has been broken. Please follow our developer guide for commit messages!

The reason using `io.BufferedIOBase` everywhere is not by accident. We wanted to make the codebase very flexible, because the files could come from anywhere. For example, it would be possible to download a file with a HTTP request and pass the response to the `carve_chunk_to_file` function, not having to save the file to disk. OR what's even more important don't have to create actual files for tests and use `io.BytesIO` instead, to make the test suite faster (at least the unit tests).

I don't see the *intent* of breaking this (the commit message doesn't reflect this decision and the type hints has not been changed), I think you just broke it by accident. I think we should keep this flexibility, so `carve_chunk_to_file` should handle `io.BufferedIOBase` still, not just ordinary file types (`typing.BinaryIO`).  
We can break this interface, but it has to be discussed and the reason has to be documented if we chose to do so.

vlaci:
I just wanted to chime in saying that I would be impressed if yara could match on e.g a BytesIO object so that I wouldn't be overly optimistic about having these possibilities at all.

unblob/extractor.py
```diff
@@ -31,8 +30,7 @@ def make_extract_dir(root: Path, path: Path, extract_root: Path) -> Path:
 def carve_chunk_to_file(carve_path: Path, file: io.BufferedIOBase, chunk: Chunk):
     """Extract valid chunk to a file, which we then pass to another tool to extract it."""
     with carve_path.open("wb") as f:
-        for data in iterate_file(file, chunk.start_offset, chunk.size):
-            f.write(data)
+        os.sendfile(f.fileno(), file.fileno(), chunk.start_offset, chunk.size)
```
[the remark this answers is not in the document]
vlaci: That's odd,
```
>>> import io
>>> io.BufferedIOBase
<class 'io.BufferedIOBase'>
```

The method will raise an `io.UnsupportedOperation` exception when there is no underlying file object.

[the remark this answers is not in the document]
e3krisztian: Yes, indeed this was the reason, the interface was changed to support only objects with `.fileno()`.

The proper (both optimization and keep working without `fileno`) implementation would check if there is a `fileno` and if not, go with the original implementation.

If we want to support `BytesIO`, I do not think this optimization is worth it: we will have a more complex code and marginal (if any) improvement in some cases.

tests/test_extractor.py
e3krisztian: agreed.

e3krisztian:
Discussed in person, this optimization is not worth the changes.

PR #293, opened 2022-03-11
Attempt to fix nix caching 

vlaci:
Currently only the dependencies cached, not the actual build result. My theory is that it is due to the repository being dirty during workflow runs.


vlaci:
Proof, that caching works: https://github.com/IoT-Inspector/unblob/runs/5512120355?check_suite_focus=true#step:7:1 The build job didn't build anything :)

kukovecz (approved):

PR #295, opened 2022-03-11
Create codeql-analysis.yml

martonilles:


qkaiser:
Updated the Github Actions versions, reviewed and improved configuration, created a dedicated CodeQL configuration file.

All the results are available to maintainer under the Security tab. Only informational finding, with some false positives that we can manually tag in Github UI.

kukovecz (approved):

PR #298, opened 2022-03-15
Cleanup carved files

### LINKED ISSUE #261: Cleanup intermediate & carved files
As part of the extraction process we carve out part of a file or sometimes the whole file to be extracted further. Once the extraction is done, we leave the carved files in place, though this could significantly increase the output size and also generate some garbage, that the user does not need.

### Carved out files
- all successfully extracted chunks could be deleted
- we need to add a flag/option to keep extracted chunks
- keep all unknown chunk
- keep all chunks with null-extractor
- keep all chunks where extraction failed
  - to decide if extraction was successful the extractor would need to return the success state
  - for command extraction we should be able to monitor exit codes and list all success exit codes
  - maybe in the future we can also compare meta-data with extracted files to determine success
 - maybe later we can also delete unknown chunks if they are just padding (we would need padding detection for that)

### Intermediate files
Intermediate files are the one which are non-carved out content but does not have additional information, eg:
- tgz file which is first compressed into a tar file and tar files are extracted. In that case the tar file is an intermediate file. 

For the timebeing we can keep these intermediate files, maybe later we can add a logic to cleanup these as part of the extraction or in post-processing.

e3krisztian:
Fixes #261

It is not a backwards compatible change: doing the cleanup is the default, the old behavior has to be explicitly requested with `--keep-extracted-chunks` or its short version `-k`.

This is a functionally complete implementation, but it needs refactors:

- propagation of failures from `Command`s are not done with exceptions
- `_process_regular_file` fails complexity check (commit is done with skipping the pre-commit hook), so it requires breaking up. Unfortunately it looks like, it was already broken up and extracted from many times, so a bit of rethinking and introduction of new concepts/classes might be necessary

### DIFF
--- a/tests/extractors/test_command.py
+++ b/tests/extractors/test_command.py
@@ -5,7 +5,7 @@ import pytest
 
 from unblob.extractors import Command
 from unblob.extractors.command import InvalidCommandTemplate
-from unblob.models import ExtractError, TaskResult
+from unblob.models import ExtractError
 from unblob.report import ExtractCommandFailedReport, ExtractorDependencyNotFoundReport
 
 
@@ -35,40 +35,42 @@ def test_command_templating_with_invalid_substitution(template):
 def test_command_execution(tmpdir: Path):
     outdir = PosixPath(tmpdir)
     command = Command("sh", "-c", "> {outdir}/created")
-    res = TaskResult()
-    command.extract(Path("foo"), outdir, res)
-    assert (outdir / Path("created")).is_file()
 
-    assert res.reports == []
+    command.extract(Path("foo"), outdir)
+
+    assert (outdir / Path("created")).is_file()
 
 
 def test_command_execution_failure(tmpdir: Path):
     outdir = PosixPath(tmpdir)
     command = Command("sh", "-c", ">&1 echo -n stdout; >&2 echo -n stderr; false")
 
-    res = TaskResult()
-    with pytest.raises(ExtractError):
-        command.extract(Path("input"), outdir, res)
-    assert res.reports == [
-        ExtractCommandFailedReport(
-            handler=None,
-            command=mock.ANY,
-            stdout=b"stdout",
-            stderr=b"stderr",
-            exit_code=1,
-        )
-    ]
+    try:
+        command.extract(Path("input"), outdir)
+        pytest.fail("ExtractError not raised")
+    except ExtractError as e:
+        assert list(e.reports) == [
+            ExtractCommandFailedReport(
+                handler=None,
+                command=mock.ANY,
+                stdout=b"stdout",
+                stderr=b"stderr",
+                exit_code=1,
+            )
+        ]
 
 
 def test_command_not_found(tmpdir: Path):
     outdir = PosixPath(tmpdir)
     command = Command("this-command-should-not-exist-in-any-system")
 
-    res = TaskResult()
-    with pytest.raises(ExtractError):
-        command.extract(Path("input"), outdir, res)
-    assert res.reports == [
-        ExtractorDependencyNotFoundReport(
-            handler=None, dependencies=["this-command-should-not-exist-in-any-system"]
-        )
-    ]
+    try:
+        command.extract(Path("input"), outdir)
+        pytest.fail("ExtractError not raised")
+    except ExtractError as e:
+        assert list(e.reports) == [
+            ExtractorDependencyNotFoundReport(
+                handler=None,
+                dependencies=["this-command-should-not-exist-in-any-system"],
+            )
+        ]
--- a/tests/extractors/test_command.py
+++ b/tests/extractors/test_command.py
@@ -5,7 +5,7 @@ import pytest
 
 from unblob.extractors import Command
 from unblob.extractors.command import InvalidCommandTemplate
-from unblob.models import TaskResult
+from unblob.models import ExtractError, TaskResult
 from unblob.report import ExtractCommandFailedReport, ExtractorDependencyNotFoundReport
 
 
@@ -47,7 +47,8 @@ def test_command_execution_failure(tmpdir: Path):
     command = Command("sh", "-c", ">&1 echo -n stdout; >&2 echo -n stderr; false")
 
     res = TaskResult()
-    command.extract(Path("input"), outdir, res)
+    with pytest.raises(ExtractError):
+        command.extract(Path("input"), outdir, res)
     assert res.reports == [
         ExtractCommandFailedReport(
             handler=None,
@@ -64,7 +65,8 @@ def test_command_not_found(tmpdir: Path):
     command = Command("this-command-should-not-exist-in-any-system")
 
     res = TaskResult()
-    command.extract(Path("input"), outdir, res)
+    with pytest.raises(ExtractError):
+        command.extract(Path("input"), outdir, res)
     assert res.reports == [
         ExtractorDependencyNotFoundReport(
             handler=None, dependencies=["this-command-should-not-exist-in-any-system"]
--- a/unblob/extractors/command.py
+++ b/unblob/extractors/command.py
@@ -5,7 +5,7 @@ from typing import List
 
 from structlog import get_logger
 
-from unblob.models import ExtractError, Extractor, TaskResult
+from unblob.models import ExtractError, Extractor
 from unblob.report import ExtractCommandFailedReport, ExtractorDependencyNotFoundReport
 
 logger = get_logger()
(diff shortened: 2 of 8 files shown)


unblob/processing.py
```diff
@@ -187,26 +198,27 @@ def _extract_chunk(
...
+
+            fix_extracted_directory(outdir, result)
+
+            result.add_new_task(
+                Task(
+                    root=self._extract_root,
+                    path=outdir,
+                    depth=task.depth + 1,
+                )
+            )
```
martonilles, line 232: if extraction failed, shell we cleanup extracted files and process them? probably yes, just wondering about this

unblob/extractors/command.py
```diff
@@ -28,7 +28,8 @@ def extract(self, inpath: Path, outdir: Path, task_result: TaskResult):
                 stdout=subprocess.PIPE,
                 stderr=subprocess.PIPE,
             )
-            if res.returncode != 0:
+            success = res.returncode == 0
```
martonilles, line 31: CommandExtractor should be parameterized as in some cases even non zero exit codes can be ok.
We should audit current CommandExtractors what is accepted as well.

unblob/processing.py
```diff
@@ -187,26 +198,27 @@ def _extract_chunk(
...
         result: TaskResult,
-    ):
-        # We only extract every blob once, it's a mistake to extract the same blog again
-        outdir.mkdir(parents=True, exist_ok=False)
+    ) -> bool:
+        # We only extract every blob once, it's a mistake to extract the same blob again
         try:
-            extractor.extract(inpath, outdir, result)
+            outdir.mkdir(parents=True, exist_ok=False)
+            extracted = extractor.extract(inpath, outdir, result)
```
martonilles: should we log here if extraction failed, or the extractors can provide a more accurate log?

unblob/extractors/command.py
e3krisztian: This line do not change in the final version, but yes, it is a valid future parameter.

unblob/processing.py
```diff
@@ -52,14 +53,15 @@ def process_file(
...
     )
 
-    processor = Processor(
-        extract_root,
-        max_depth,
-        entropy_depth,
-        entropy_plot,
-        handlers,
-        keep_extracted_chunks,
+    parameters = Parameters(
```
vlaci, line 56: I'd consider lifting this `(Processing)Parameters` config object upper one level, so that `process_file` expects this as an argument. You can also add the defaults values to it.

tests/extractors/test_command.py
```diff
@@ -36,7 +36,7 @@ def test_command_execution(tmpdir: Path):
     outdir = PosixPath(tmpdir)
     command = Command("sh", "-c", "> {outdir}/created")
     res = TaskResult()
-    assert command.extract(Path("foo"), outdir, res)
+    command.extract(Path("foo"), outdir, res)
```
vlaci, line 39: I'd merge these changes back to the previous commit introducing the boolean return value

unblob/processing.py
```diff
@@ -187,26 +198,27 @@ def _extract_chunk(
...
-        # We only extract every blob once, it's a mistake to extract the same blog again
-        outdir.mkdir(parents=True, exist_ok=False)
+    ) -> bool:
+        # We only extract every blob once, it's a mistake to extract the same blob again
         try:
-            extractor.extract(inpath, outdir, result)
+            outdir.mkdir(parents=True, exist_ok=False)
+            extracted = extractor.extract(inpath, outdir, result)
+
+            fix_extracted_directory(outdir, result)
```
vlaci: We want to call `fix_extracted_directory` even in case of extraction errors.

unblob/processing.py
```diff
@@ -187,23 +187,8 @@ def _process_chunks(
...
-
-            inpath, outdir = get_extract_paths(extract_dir, carved_valid_path)
-
-            extractor = chunk.handler.EXTRACTOR
-            if extractor is not None:
-                extracted = self._extract_chunk(inpath, outdir, extractor)
-                if extracted and not self.parameters.keep_extracted_chunks:
-                    inpath.unlink()
+            carved_path = carve_valid_chunk(extract_dir, file, chunk)
+            self._extract_chunk(extract_dir, carved_path, chunk)
```
vlaci, line 191: I'd put unlinking the input file around these lines so that the lifteteime of the input file is easier to track: it gets created here and gets removed here as well.

vlaci (changes requested):

unblob/processing.py
e3krisztian: I have previously thought about moving the call `carve_valid_chunk` inside `_extract_chunk` instead, then the lifetime would be trackable there (it is carved there, removed there). I definitely do not want to complicate `_process_chunks` with file removal logic - it is not that abstraction level: it is about and managing the extraction directory and extracting all of the chunks.

unblob/processing.py
e3krisztian: This is a valid improvement, but maybe better done in another PR.

unblob/models.py
```diff
@@ -161,6 +180,15 @@ def calculate_chunk(
     ) -> Optional[ValidChunk]:
         """Calculate the Chunk offsets from the Blob and the file type headers."""
 
+    def extract(self, inpath: Path, outdir: Path, task_result: TaskResult):
+        if self.EXTRACTOR is None:
+            raise CanNotExtractWithoutExtractor
```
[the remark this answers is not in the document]
e3krisztian: Well, I would argue, that it is borderline error to have no extractor.  As the same behavior could be also implemented with an `Extractor` subclass, that does nothing.

When we have `None` as extractor, no extraction is done, the chunk file is expected to be left behind as is.

The side effect for raising an `ExtractError` and derivatives is that they mark the input as problematic, thus the extracted chunk will not be removed - as is intended for cases when extractor is `None`.

unblob/models.py
```diff
@@ -123,20 +134,28 @@ def reports(self):
...
         """Returns the external command dependencies."""
         return []
 
     @abc.abstractmethod
-    def extract(self, inpath: Path, outdir: Path, task_result: TaskResult) -> bool:
+    def extract(self, inpath: Path, outdir: Path, task_result: TaskResult):
         """Extract the carved out chunk.
 
-        Returns True for success.
+        Raises ExtractError on failure.
```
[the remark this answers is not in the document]
e3krisztian: I think this is part of the interface, so if they do not raise this error, they are not compliant with the interface.

When this exception is not raised, the [partially] extracted input file will be removed, as we can not determine that it was a failed extraction and thus the file is potentially needed for further analysis.

unblob/processing.py
e3krisztian: Finally it will not moved out to `_process_chunk`, but outside the `try` block - as it was previously

unblob/processing.py
e3krisztian: Agree, and it was reverted

unblob/processing.py
```diff
@@ -136,76 +143,99 @@ def _process_task(self, result: TaskResult, task: Task):
...
+                continue
+
+            inpath, outdir = get_extract_paths(extract_dir, carved_valid_path)
 
-                inpath, outdir = get_extract_paths(extract_dir, carved_valid_path)
+            extractor = chunk.handler.EXTRACTOR
+            if extractor is not None:
+                extracted = self._extract_chunk(inpath, outdir, extractor)
+                if extracted and not self.parameters.keep_extracted_chunks:
+                    inpath.unlink()
```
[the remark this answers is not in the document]
e3krisztian: A `logger.debug()` line was added before removing the file

unblob/models.py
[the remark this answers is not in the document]
vlaci: The difference between this and any other possible exceptions is only in error reporting, this type indicates an error that has been reported by the extractor, other errors will cause an error to be reported downstream.

unblob/processing.py
```diff
@@ -136,76 +143,91 @@ def _process_task(self, result: TaskResult, task: Task):
...
+
         try:
-            extractor.extract(inpath, outdir, result)
+            chunk.extract(inpath, outdir, self.result)
+            if not self.parameters.keep_extracted_chunks:
+                logger.debug("Removing extracted chunk", path=inpath)
+                inpath.unlink()
+
+        except ExtractError:
+            # already reported & logged by extractor
```
vlaci, line 213: Another indication that this needs a better name but I am out of ideas.

Another possible approach: have this error accept an argument that will be added to the TaskResult at this level. I guess it would make the expected and unexpected error handling more streamlined: 
- Got an ExtractError(reports)? Good, let's append it to the TaskResult
- Got any other (unexpected) error? Not a problem, let's convert it to an UnknownErrorReport

e3krisztian: implemented :)

unblob/processing.py
e3krisztian: As discussed, it would also be better named as `ExtractConfig`, it will be in a follow-up PR

vlaci (approved):

PR #300, opened 2022-03-17
Introduce different exit codes for the CLI

### LINKED ISSUE #299: Return different exit codes based on received reports
unblob currently returns 0 if it does not receive any reports, and 1 otherwise.

Reports include:

* **MaliciousSymlinkRemoved** - an error when malicious symlinks have been removed from disk
* **ExtractorDependencyNotFoundReport** - an error when the dependency of an extractor doesn't exist
* **ExtractCommandFailedReport** - an error when failed to run the extraction command
* **UnknownError** - an exception raised during file processing
* **CalculateChunkExceptionReport** - an exception raised during calculate_chunk execution

This represents a large array of exception and we should agree first on a list of exceptions that we are willing to accept. To me, we must accept the following exceptions:

* **MaliciousSymlinkRemoved** - it will happen all the time and is not a problem in itself
* **ExtractCommandFailedReport** - it will happen all the time on weird misidentified chunks is not a problem in itself

And we must stop if we encounter those:

* **ExtractorDependencyNotFoundReport** - things will obviously go wrong down the line
* **UnknownError** - something bad happened and we should know about it
* **CalculateChunkExceptionReport** - something bad happened and we should know about it

In the case of a vendor firmware, the processing stopped because unblob returned an exit code of '1' because it couldn't extract an ARJ archive false positive.

I'd recommend assigning an exit code to each Report class, with increasing values:

1. MaliciousSymlinkRemoved
1. ExtractCommandFailedReport
1. CalculateChunkExceptionReport
1. UnknownError
1. ExtractorDependencyNotFoundReport

The exit code values do not need to be different for each report, we could also do something like:

```python
class ReturnCode(Enum):
  OK: 0
  STILL_WORKING: 1
  REAL_BAD: 2
  CANNOT_EVEN_RUN: 3
```

And assign an enum value to our report classes, so that we can add new report types later on if needed.

kukovecz:
Until now we had 2 exit codes:
 - 0 if everything went well
 - 1 otherwise

However we want to introduce a new "warning" level, which means there
were some issues during execution, however those issues are not
considered "fatal". We should inform the user about them, but the
possible further processing can go on without any possible issues.

For this, we introduce an `return_code` field for each `Report`, and we
exit the CLI with the most severe `return_code`.

Fixes #299

### DIFF
--- a/unblob/cli.py
+++ b/unblob/cli.py
@@ -7,7 +7,7 @@ import click
 from structlog import get_logger
 
 from unblob.plugins import UnblobPluginManager
-from unblob.report import Report
+from unblob.report import Report, Severity
 
 from .cli_options import verbosity_option
 from .dependencies import get_dependencies, pretty_format_dependencies
@@ -180,6 +180,20 @@ def cli(
 cli.context_class = UnblobContext
 
 
+def get_exit_code_from_reports(reports: List[Report]) -> int:
+    severity_to_exit_code = [
+        (Severity.ERROR, 1),
+        (Severity.WARNING, 0),
+    ]
+    severities = {report.severity for report in reports}
+
+    for severity, exit_code in severity_to_exit_code:
+        if severity in severities:
+            return exit_code
+
+    return 0
+
+
 def main():
     try:
         # Click argument parsing
@@ -200,8 +214,7 @@ def main():
         logger.exception("Unhandled exception during unblob")
         sys.exit(1)
 
-    exit_code = 0 if not reports else 1
-    sys.exit(exit_code)
+    sys.exit(get_exit_code_from_reports(reports))
 
 
 if __name__ == "__main__":
--- a/unblob/report.py
+++ b/unblob/report.py
@@ -1,11 +1,21 @@
+from enum import Enum
 from typing import List, Optional
 
 import attr
 
 
+class Severity(Enum):
+    """Represents possible problems encountered during execution"""
+
+    ERROR = "ERROR"
+    WARNING = "WARNING"
+
+
 @attr.define(kw_only=True)
 class Report:
     """A common base class for different reports"""
+
+    severity: Severity
 
     # Stored in `str` rather than `Handler`, because the pickle picks ups structs from `C_DEFINITIONS`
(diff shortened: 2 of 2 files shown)


unblob/report.py
```diff
@@ -1,23 +1,42 @@
...
 import attr
 
 
+class ReturnCode(IntEnum):
+    """Represents possible problems encountered during execution. Severity is increasing in value."""
+
+    OK = 0
+    WARNING = 1
+    WRONG_PARAMETER = 2
+    FATAL = 3
```
vlaci, line 13: See discussion in #299. Shouldn't merge until we figure that out :)

vlaci (changes requested):

unblob/cli.py
```diff
@@ -185,6 +185,20 @@ def cli(
...
 
 
+def get_exit_code_from_reports(reports: List[Report]) -> int:
+    exit_code_map = {
+        Severity.ERROR: 1,
+        Severity.WARNING: 0,
+    }
+    severities = {report.severity for report in reports}
+
+    for severity, exit_code in exit_code_map.items():
```
e3krisztian, line 195: `exit_code_map` is a map, but its only usage is iterating over it.
I have also grown up with unordered dicts, where looping over items of dicts is ambiguous.
I know under py3 dicts are ordered, but still it reads wrong.
```py
exit_code_map = [
        (Severity.ERROR, 1),
        (Severity.WARNING, 0),
    ]
```
is better.

e3krisztian: An alternative implementation could be: take the set of severities, take the largest, map that value to exit code.

vlaci:
After some rumination I don't have further grudges about this :)

unblob/cli.py
kukovecz: Dicts are designed to keep insertion order from python3.7. Since this project is python3.8+, I think this is fine.

However I did change the suggested for `exit_code_map` ( I renamed it to `severity_to_exit_code` since it become a `list` ...), if you think this is easier to understand / more clear what is happening, this is fine by me.

vlaci (approved):

Issue #262, opened 2022-02-22
Ability to stop processing certain files


martonilles:
It would be important for the user to be able to stop processing at the “right” level, extracting too much details would just result in garbage, too little might not extract an important part of the firmware. Ideally we want to guess the right level otherwise it would require lot of automated work.

Unfortunately just specifying depth is not enough as different firmware could contain different level of packaging.

We want to be able to stop processing based on the file types, especially the following types should not be processed further as most likely that would result in some false-positive detection:
- media files (pictures, video content)
- office documents
- executable (ELF binaries & libraries)

We still want to extract other packaged contents:
- tar.gz/bz/etc.
- cab
- jar
- apk
- deb
- etc.

Preferably it should be configurable what file types to stop processing with a default specified above.

We can use file-magic to determine the file type. (We just want to detect if the whole file is a given type so file magic is probably a good enough option for us.
- polyglot files could be a problem (eg: gif/rar or firmware images starting with a logo.png, but we can take care of those once see more examples)

martonilles:
- libmagic is for skipping from processing
- record magic/mime into metadata
  -  make sure that "deepest" depth has metadata recorded even for files that are not processed
  - we can also run libmagic detection on carved out files as well, even though those are not processed, just extracted

PR #301, opened 2022-03-18
Skip processing

martonilles:
closes: #262

### DIFF
--- a/unblob/processing.py
+++ b/unblob/processing.py
@@ -129,11 +129,7 @@ class Processor:
             log.debug("Ignoring file based on magic")
             return
 
-        filetask = _FileTask(self._config, task, size, result)
-        filetask.process()
-        # ensure that the root extraction directory is created even for empty extractions
-        if task.depth == 0:
-            filetask.extract_dir.mkdir(parents=True, exist_ok=True)
+        _FileTask(self._config, task, size, result).process()
 
     def _should_skip_magic(self, task: Task):
         detect = magic.detect_from_filename(task.path)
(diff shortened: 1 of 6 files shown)


unblob/processing.py
```diff
@@ -146,6 +149,18 @@ def __init__(
...
+            [detect.name.startswith(pattern) for pattern in self.config.skip_magic]
+        )
+        if skip:
+            logger.debug(
+                "Skip file from processing based on file-magic",
+                magic=detect.name,
+                path=self.task.path,
+            )
+            return
+
```
vlaci, line 163: For me it'd make sense to extract this into a `_should_skip` method inside `Processor` and call from the `if..elif..` chain in `process_task`. That way all the logic determining if we want to process a file would be in the same place.

unblob/processing.py
```diff
@@ -146,6 +149,18 @@ def __init__(
     def process(self):
         logger.debug("Processing file", path=self.task.path, size=self.size)
 
+        detect = magic.detect_from_filename(self.task.path)
+        skip = any(
+            [detect.name.startswith(pattern) for pattern in self.config.skip_magic]
+        )
```
vlaci, line 155: You can get rid of the list comprehension and make the `any` call short-circuiting that way:
```suggestion
        skip = any(
            detect.name.startswith(pattern) for pattern in self.config.skip_magic
        )
```

```
>>> def is_true():
...     print("It is True!")
...     return True
... 
>>> any([is_true() for _ in range(3)])
It is True!
It is True!
It is True!
True
>>> any(is_true() for _ in range(3))
It is True!
True
>>> 
```

vlaci (changes requested):

vlaci:
Very nice!

unblob/processing.py
martonilles: done

unblob/processing.py
martonilles: good point :)

vlaci (approved):

PR #302, opened 2022-03-18
Reusable integration tests

vlaci:
The idea behind this change is to make it able to reuse the
integration tests in external handlers like this:

    @gather_integration_tests(TEST_DATA_PATH)
    def test_all_handlers(input_dir, output_dir, tmp_path):
        handlers = Handlers([(MyHandler,)])
        all_reports = process_file(
            path=input_dir,
            extract_root=tmp_path,
            entropy_depth=0,
            handlers=handlers
        )

        check_output_is_the_same(output_dir, tmp_path)
        check_reports(all_reports)

### DIFF
--- a/unblob/plugins.py
+++ b/unblob/plugins.py
@@ -1,4 +1,5 @@
 import importlib.util
+import itertools
 import sys
 from pathlib import Path
 from types import ModuleType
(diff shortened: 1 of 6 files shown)


unblob/plugins.py
```diff
@@ -79,3 +81,25 @@ def _import_modules(
             modules.append(mod)
 
         return modules
+
+    def add_handlers_to(
```
martonilles, line 85: The `add_handlers_to` name is a bit ambiguous for me as, maybe something like `extend_handlers_from_plugins` would describe this better?

PR #304, opened 2022-03-22
test_handlers: make integration test parametrization more obvious

vlaci:
`@gather_integration_tests` calling `@pytest.mark.parametrize` under
the hood was confusing, now it is turned inside out.


e3krisztian (approved):
Looks good - thanks! :)

PR #306, opened 2022-03-22
Refactor Parameters -> ExtractionConfig, process_file

e3krisztian:
Convert many parameters of `process_file` into `ExtractionConfig` and thus simplify its signature to
```py
def process_file(config: ExtractionConfig, path: Path) -> List[Report]:
```

----

With this change, we are few steps from this interface:
```py
Unblob(**extraction_parameters).process_file(path)
```

### DIFF
--- a/unblob/processing.py
+++ b/unblob/processing.py
@@ -34,6 +34,16 @@ DEFAULT_DEPTH = 10
 DEFAULT_PROCESS_NUM = multiprocessing.cpu_count()
 
 
+@attr.define
+class ExtractionConfig:
+    extract_root: Path
+    max_depth: int
+    entropy_depth: int
+    entropy_plot: bool
+    handlers: Handlers
+    keep_extracted_chunks: bool
+
+
 @terminate_gracefully
 def process_file(
     path: Path,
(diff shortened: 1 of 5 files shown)


unblob/processing.py
```diff
@@ -34,17 +34,19 @@
 DEFAULT_PROCESS_NUM = multiprocessing.cpu_count()
 
 
+@attr.define(kw_only=True)
+class ExtractionConfig:
```
kukovecz, line 38: Since this is used from multiple places, I would put this class definition be in the `models.py`

kukovecz (approved):
I like these changes, makes calling the whole thing so much smoother.

unblob/processing.py
vlaci: I was unsure about suggesting this as well. From one hand it would match better with the existing code structure, on the other hand I find this `models` module containing a bunch of unrelated datastructures a bit odd.

kukovecz: Yeah, me too. I was just going with the consistency.... On the other hand, I have already approved this when I wrote this comment, so it can be merged without this modification :smile: That's my way of saying "just a small suggestion, but don't care that much".

e3krisztian: Thanks for the review, I was thinking about moving to models as well (consistency), I merge it as it is now, we can move to models in another PR if that is a better place - later.

Issue #291, opened 2022-03-11
romfs checksum calculation fails


martonilles:
In `0c2c48ae220a1b04` firmware

```
2022-03-10 23:41.32 [error    ] Unknown error happened         exec_info=TypeError('can only concatenate str (not "error") to str') pid=8
Traceback (most recent call last):
  File "/home/unblob/.local/lib/python3.8/site-packages/unblob/finder.py", line 85, in search_chunks_by_priority
    chunk = handler.calculate_chunk(limited_reader, real_offset)
  File "/home/unblob/.local/lib/python3.8/site-packages/unblob/handlers/filesystem/romfs.py", line 64, in calculate_chunk
    if not valid_checksum(file.read(512)):
  File "/home/unblob/.local/lib/python3.8/site-packages/unblob/handlers/filesystem/romfs.py", line 24, in valid_checksum
    total += struct.unpack(">L", content[i : (i + 4)])[0]  # noqa: E203
struct.error: unpack requires a buffer of 4 bytes
```

PR #307, opened 2022-03-23
Make unromfs an Extractor subclass instead of a standalone command

vlaci:
After this change, the `extractors` package will be more or less meaningless to keep. I'd remove it in an upcoming PR.


pyproject.toml
```diff
@@ -41,7 +41,6 @@ profile = "black"
 
 [tool.poetry.scripts]
 unblob = "unblob.cli:main"
-unromfs = "unblob.extractors.unromfs:cli"
```
kukovecz, line 44: Just a question: Is it okay for us (=the user) that the `unromfs` command will no longer be available?

kukovecz (approved):
I just put one question there, otherwise looks good!

unblob/handlers/filesystem/romfs.py
```diff
@@ -35,6 +67,291 @@ def get_string(file: io.BufferedIOBase) -> bytes:
...
+            header.dump_fs()
+
+
+def valid_checksum(content: bytes) -> int:
+    """Compute the RomFS checksum of content."""
+    total = 0
+    for i in range(0, len(content), 4):
+        total += struct.unpack(">L", content[i : (i + 4)])[0]  # noqa: E203
+        total %= MAX_UINT32
+    return total == 0
```
QKaiser, line 352: This could reuse the romfs_checksum function with:

```suggestion
def valid_checksum(content: bytes) -> int:
    """Compute the RomFS checksum of content."""
    return romfs_checksum(content) == 0
```

unblob/handlers/filesystem/romfs.py
```diff
@@ -17,16 +21,44 @@
...
+    FIFO = 7
+
+
+def romfs_checksum(content: bytes) -> int:
+    """Apply a RomFS checksum and returns whether it's valid or not."""
     total = 0
+
+    # unalign content will lead to unpacking errors down the line
+    if len(content) % 4 != 0:
+        return -1
```
QKaiser, line 51: Pretty sure this will fix #291

pyproject.toml
QKaiser: Totally ok with this right now. If someone wants to extract a romfs file they can run unblob on it with depth set to 1 or 2. Check my other comments before merging.

unblob/handlers/filesystem/romfs.py
vlaci: I've added some further refactors. Thanks for the tip

PR #308, opened 2022-03-23
arj: extended yara rule to verify header size and header checksum verification

martonilles:
We implement similar logic to 7zip ARJ detection to verify if the chunk
is indeed arj. Invalid header length is easy to detect and header checksum
is easy to calculate, though both can rule out arj looking garbage.

### DIFF
--- a/unblob/handlers/archive/arj.py
+++ b/unblob/handlers/archive/arj.py
@@ -1,20 +1,31 @@
+import binascii
 import io
 from typing import Optional
 
 from structlog import get_logger
 
 from ...extractors import Command
+from ...file_utils import Endian, convert_int32
 from ...models import StructHandler, ValidChunk
 
 logger = get_logger()
+
+# CPP/7zip/Archive/ArjHandler.cpp IsArc_Arj()
+MIN_BLOCK_SIZE = 30
+MAX_BLOCK_SIZE = 2600
+BASIC_HEADER_SIZE = 4
 
 
 class ARJError(Exception):
     pass
 
 
-class ARJNullFile(ARJError):
-    """Zero-sized ARJ."""
+class InvalidARJSize(ARJError):
+    """Invalid size fields in ARJ header"""
+
+
+class ARJChecksumError(ARJError):
+    """Main ARJ header checksum missmatch"""
 
 
 class ARJExtendedHeader(ARJError):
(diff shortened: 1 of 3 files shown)


unblob/handlers/archive/arj.py
```diff
@@ -28,7 +32,12 @@ class ARJHandler(StructHandler):
         strings:
             $arj_magic = { 60 EA [5] 0? [2] 0? }
         condition:
-            $arj_magic
+            $arj_magic and
```
QKaiser, line 35: I think we can do even more checks in Yara. If you take a look at http://www.opennet.ru/docs/formats/arj.txt, we can validate:

- host_os: `uint8(@arj_magic+offset) >= 0 and uint8(@arj_magic+offset) <= 9`
- file_type: `uint8(@arj_magic+offset) >= 0 and uint8(@arj_magic+offset) <= 4`

unblob/handlers/archive/arj.py
```diff
@@ -28,7 +32,12 @@ class ARJHandler(StructHandler):
...
+            // CPP/7zip/Archive/ArjHandler.cpp IsArc_Arj()
+            // basic_header.size must be between 30 and 2600
+            uint16(@arj_magic+2) > 30 and uint16(@arj_magic+2) < 2600 and
+            // arj_header.first_hdr_size is less then basic_header.size
+            uint16(@arj_magic+2) > uint8(@arj_magic+3)
```
QKaiser, line 40: Shouldn't it be `uint16(@arj_magic+2) > uint8(@arj_magic+4)` ?

If `basic_header` is made of two uint16, that means we need to skip 4 bytes to reach `first_hdr_size`.

QKaiser (changes requested):

unblob/handlers/archive/arj.py
```diff
@@ -1,20 +1,24 @@
+import binascii
 import io
 from typing import Optional
 
 from structlog import get_logger
 
 from ...extractors import Command
+from ...file_utils import convert_int32, Endian
```
QKaiser, line 8: It's changed by isort, check your pre-commit hook.

QKaiser (changes requested):

unblob/handlers/archive/arj.py
martonilles: I wanted to use the same logic as 7zip uses for verification. It does not limit neither HostOS nor FileType fileds, so I would keep it more permissive. Length & checksum checks should be enough to distinguish between garbage and normal ARJ chunk.

unblob/handlers/archive/arj.py
martonilles: Moved to chunk logic (see #312 )

QKaiser (commented):
Code is good, if you got time I'd create a test file (`tests/handlers/archive/test_arj.py`) similar in form to `tests/handlers/filesystem/test_fat.py` to test that the following conditions trigger an ARJError:

- block size too small
- block size too large
- header size smaller than first_hdr_size
- invalid CRC

QKaiser (approved):

PR #311, opened 2022-03-24
Fix XZ overall_size calculation.

### LINKED ISSUE #310: XZ handler padding miscalculation
The idea of the XZ handler is to identify the end of a chunk by finding its footer with a string search. Then it validates that this chunk is right by parsing the XZ records to calculate the `overall_size` value. If both values correspond, we have a valid chunk.

There is an error in the `overall_size` calculation given that we increment by unpadded size of each record while we should increment by the padded size of each record.

QKaiser:
The idea of the XZ handler is to identify the end of a chunk by finding its footer with a string search. Then it validates that this chunk is right by parsing the XZ records to calculate the overall_size value. If both values correspond, we have a valid chunk.

There was an error in the overall_size calculation given that we incremented by unpadded size of each record while we should increment by the padded size of each record.

This commit fix this issue by incrementing by padded sizes.

Fix #310

### DIFF
--- a/tests/integration/compression/xz/__input__/multiblock.padded.xz
+++ b/tests/integration/compression/xz/__input__/multiblock.padded.xz
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:0856a2ca40f1b89f1f6b1e7d35b45d68eb8cd71132c9f77ea2a46e1348d8436c
+size 2168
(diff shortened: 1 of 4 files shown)


martonilles:
Patch looks good, I would probably add either a unit test or an integration test example for this.

QKaiser:
> Patch looks good, I would probably add either a unit test or an integration test example for this.

An integration test file triggering the mis-calculation in the old code has been created with the following command line:
    
```
xz -z --block-list=1021 -f /tmp/test.bin
```
  
This creates an XZ stream with small blocks of odd size, forcing the creation of multiple padded blocks.

martonilles (approved):

PR #313, opened 2022-03-24
Update YARA & throw away YARA patch

### LINKED ISSUE #97: YARA Python segmentation fault
If you download the latest ubuntu LTS, unblob will crash with segmentation fault.

e3krisztian:
Reverts d2205f20019a307e80db0f1cf40fd390507e7fd1

Resolves #97


QKaiser (approved):
Verified and checked with the image that initially led us to report #97

PR #317, opened 2022-03-25
Fix LZO to handle missing checksum and multi-block files

martonilles:
We did not handle cases where the uncompressed size of the data was over the default 256k block size. In those cases the content is compressed in multiple chunks, with each chunk having a special chunk header (similar to the previous size crc header) describing the given chunk. The last chunk has 0 as uncompressed size.

Checksums are also optional in chunk headers depending on the main header flags.

### DIFF
--- a/unblob/handlers/compression/lzo.py
+++ b/unblob/handlers/compression/lzo.py
@@ -5,12 +5,28 @@ from structlog import get_logger
 
 from unblob.extractors import Command
 
+from ...file_utils import Endian, convert_int32
 from ...models import StructHandler, ValidChunk
 
 logger = get_logger()
 
+CHECKSUM_LENGTH = 4
 
-F_H_FILTER = 0x00000800
+# Header flags defined in lzop (http://www.lzop.org/) source in src/conf.h
+F_ADLER32_D = 0x00000001  # noqa
+F_ADLER32_C = 0x00000002  # noqa
+F_STDIN = 0x00000004  # noqa
+F_STDOUT = 0x00000008  # noqa
+F_NAME_DEFAULT = 0x00000010  # noqa
+F_DOSISH = 0x00000020  # noqa
+F_H_EXTRA_FIELD = 0x00000040  # noqa
+F_H_GMTDIFF = 0x00000080  # noqa
+F_CRC32_D = 0x00000100  # noqa
+F_CRC32_C = 0x00000200  # noqa
+F_MULTIPART = 0x00000400  # noqa
+F_H_FILTER = 0x00000800  # noqa
+F_H_CRC32 = 0x00001000  # noqa
+F_H_PATH = 0x00002000  # noqa
 
 
 class LZOHandler(StructHandler):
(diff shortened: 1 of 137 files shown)


unblob/handlers/compression/lzo.py
```diff
@@ -5,12 +5,27 @@
...
+F_NAME_DEFAULT = 0x00000010  # noqa
+F_DOSISH = 0x00000020  # noqa
+F_H_EXTRA_FIELD = 0x00000040  # noqa
+F_H_GMTDIFF = 0x00000080  # noqa
+F_CRC32_D = 0x00000100  # noqa
+F_CRC32_C = 0x00000200  # noqa
+F_MULTIPART = 0x00000400  # noqa
+F_H_FILTER = 0x00000800  # noqa
+F_H_CRC32 = 0x00001000  # noqa
+F_H_PATH = 0x00002000  # noqa
```
QKaiser, line 28: Do we really need the whole list of flags ? I'd keep the flags that we actually checks and remove the `noqa` comment.

unblob/handlers/compression/lzo.py
```diff
@@ -86,13 +95,22 @@ def calculate_chunk(
 
         logger.debug("LZO header parsed", header=header, _verbosity=3)
 
-        size_crc_header = self.cparser_be.lzo_size_crc_t(file)
-        logger.debug("CRC header parsed", header=size_crc_header, _verbosity=3)
+        while True:
```
QKaiser, line 98: Just an idea to make it look less "infine-loopy":

```python
uncompressed_size = convert_int32(file.read(4), endian=Endian.BIG)
while uncompressed_size != 0:
   ...
end_offset = file.tell()
```

unblob/handlers/compression/lzo.py
e3krisztian: For future maintenance I would prefer either
- having all the structure & flag definitions
- a link to the spec

unblob/handlers/compression/lzo.py
e3krisztian: I guess this is the core change to handle multi-block, and multi-block is the reason we can not tell when we need to stop.

vlaci: I think the `uncompressed_size = convert_int32(...)` bit would still be there in the loop body as well and it'd just remove the need to for the `if` block at L101. I am in favor of this change as it'd make the exit condition more obvious.

unblob/handlers/compression/lzo.py
martonilles: Agree with @e3krisztian, I think it is better to list all flags for completeness. It makes understanding the file format easier.

There is unfortunately no specification of format, but added a comment on where to find the definition in the source.

unblob/handlers/compression/lzo.py
martonilles: fixed loop to make exit more explicit

QKaiser (approved):

PR #318, opened 2022-03-26
Convert YARA conditions into handler checks

### LINKED ISSUE #312: Fix YARA rules in handlers for multi-occurrences
When YARA rule contains references to string positions using the `@` notation, we always calculate the first string occurrences as `@string` actually refers to the 1st string like this: `@string[1]`. We need to move these checks into the handler calculate chunk logic. YARA conditions are evaluated on the whole file, not string match by match.

eg: LZMA handler
```
    YARA_RULE = r"""
        strings:
            $lzma_magic = {
                // pre-computed valid properties bytes
                (
                    51 | 5A | 5B | 5C | 5D | 5E | 63 | 64 | 65 | 66 | 6C | 6D | 6E | 75 | 76 | 7E |
                    87 | 88 | 89 | 8A | 8B | 90 | 91 | 92 | 93 | 99 | 9A | 9B | A2 | A3 | AB | B4 |
                    B5 | B6 | B7 | B8 | BD | BE | BF | C0 | C6 | C7 | C8 | CF | D0 | D8
                )
                // dictionary size
                00 00 ( 00 | 01 | 04 | 08 | 10 | 20 | 40 | 80) ( 00 | 01 | 02 | 04 | 08 )
            }
        condition:
            // LZMA file format: https://svn.python.org/projects/external/xz-5.0.3/doc/lzma-file-format.txt
            $lzma_magic and (#lzma_magic == 1 ....
            // dictionary size is non-zero (section 1.1.2 of format definition)
            uint32(@lzma_magic + 1) > 0 and
```
Here the `uint32(@lzma_magic + 1)` check would look at always the first `$lzma_magic` occurrence, so in case the first string in a file matches all other would match resulting in false-positives, while if the first does not match no subsequent string would result in a match, hence not detecting a valid chunk.

The following handlers rely on this construct and should be edited:

- [x] archive.cpio
- [x] compression.compress
- [x] compression.lzh
- [x] compression.lzma

martonilles:
closes #312

### DIFF
--- a/unblob/handlers/archive/cpio.py
+++ b/unblob/handlers/archive/cpio.py
@@ -12,6 +12,33 @@ logger = get_logger()
 CPIO_TRAILER_NAME = b"TRAILER!!!"
 MAX_LINUX_PATH_LENGTH = 0x1000
 
+C_ISBLK = 0o60000
+C_ISCHR = 0o20000
+C_ISDIR = 0o40000
+C_ISFIFO = 0o10000
+C_ISSOCK = 0o140000
+C_ISLNK = 0o120000
+C_ISCTG = 0o110000
+C_ISREG = 0o100000
+
+C_FILE_TYPES = (
+    C_ISBLK,
+    C_ISCHR,
+    C_ISDIR,
+    C_ISFIFO,
+    C_ISSOCK,
+    C_ISLNK,
+    C_ISCTG,
+    C_ISREG,
+)
+
+C_NONE = 0o00000
+C_ISUID = 0o04000
+C_ISGID = 0o02000
+C_ISVTX = 0o01000
+
+C_STICKY_BITS = (C_NONE, C_ISUID, C_ISGID, C_ISVTX)
+
 
 class _CPIOHandlerBase(StructHandler):
     """A common base for all CPIO formats
@@ -53,6 +80,16 @@ class _CPIOHandlerBase(StructHandler):
                     offset += self._pad_content(header, c_filesize, c_namesize)
                     break
 
+            c_mode = self._calculate_mode(header)
+
+            file_type = c_mode & 0o770000
+            sticky_bit = c_mode & 0o7000
+
+            # heuristics 3: check mode field
+            is_valid = file_type in C_FILE_TYPES and sticky_bit in C_STICKY_BITS
+            if not is_valid:
+                return
+
             file.seek(c_filesize, io.SEEK_CUR)
             # Rounding up the total of the header size, and the c_filesize, again. Because
             # some CPIO implementations don't align the first chunk, but do align the 2nd.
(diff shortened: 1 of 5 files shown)


unblob/handlers/archive/cpio.py
```diff
@@ -53,6 +80,20 @@ def calculate_chunk(
...
+            sticky_bit = c_mode & 0o7000
+            permission = c_mode & 0o777
+
+            # heuristics 3: check mode field
+            if not (
+                any([file_type == ft for ft in C_FILE_TYPES])
+                and any([sticky_bit == sb for sb in C_STICKY_BITS])
+                and permission <= 0o777
+            ):
+                return
```
e3krisztian, line 95: Maybe I miss something, but I think this fragment could be simplified to:

```suggestion
            is_valid = file_type in C_FILE_TYPES and sticky_bit in C_STICKY_BITS
            if not is_valid:
                return
```

Note, that permission related condition is not there, as I do not see how the permission check could ever fail.

unblob/handlers/compression/compress.py
```diff
@@ -107,7 +97,13 @@ def unlzw(  # noqa: C901
 
         header = self.parse_header(file, Endian.LITTLE)
 
+        if header.flags & 0x60:
+            raise InvalidInputFormat("Flag & 0x60")
+
         max_ = header.flags & 0x1F
+        if max_ < 9 or max_ > 16:
```
e3krisztian, line 104:
```suggestion
        if not (9 <= max_ <= 16):
```

unblob/handlers/archive/cpio.py
```diff
@@ -53,6 +80,20 @@ def calculate_chunk(
...
+
```
e3krisztian, line 96: Previously these checks had run only for `BinaryHandler` - is it also a fix for other CPIO Handlers?

martonilles: yes, all of them should be checked, it was a miss to check only for Binary only. Commit message also mentions this.

vlaci:
It is really hard to understand the intent behind these changes based on the issue description and the commit messages don't help me either. It took me a while to figure out that the underlying issue is that we thought that the  `@` (index-of) operator works on the current match but it isn't true. Instead, it works on the list of all of the matches and it defaults to the 1st match if unspecified.

The original intent behind the smarter Yara rules was performance as having many false positives to process was quite expensive. Have you looked into the impact of this change?

martonilles:
> It is really hard to understand the intent behind these changes based on the issue description and the commit messages don't help me either. It took me a while to figure out that the underlying issue is that we thought that the `@` (index-of) operator works on the current match but it isn't true. Instead, it works on the list of all of the matches and it defaults to the 1st match if unspecified.
> 
> The original intent behind the smarter Yara rules was performance as having many false positives to process was quite expensive. Have you looked into the impact of this change?

Added a more detailed explanation into the commit messages.

The performance impact is valid, though accuracy is more important. We will address performance perspective in #320

vlaci (approved):

PR #319, opened 2022-03-28
Add force overwrite extract dir option

kukovecz:
With this config, if the extraction dir exist, it will automatically
deleted before the processing, thus overridden with the new results.

### DIFF
--- a/unblob/cli.py
+++ b/unblob/cli.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 import sys
 from pathlib import Path
-from typing import Iterable, List, Optional, Tuple
+from typing import Iterable, List, Optional
 
 import click
 from structlog import get_logger
@@ -153,7 +153,7 @@ class UnblobContext(click.Context):
     expose_value=False,
 )
 def cli(
-    files: Tuple[Path],
+    files: List[Path],
     extract_root: Path,
     depth: int,
     entropy_depth: int,
--- a/unblob/cli.py
+++ b/unblob/cli.py
@@ -18,7 +18,7 @@ from .processing import (
     DEFAULT_PROCESS_NUM,
     DEFAULT_SKIP_MAGIC,
     ExtractionConfig,
-    process_file,
+    process_files,
 )
 
 logger = get_logger()
@@ -183,10 +183,7 @@ def cli(
     )
 
     logger.info("Start processing files", count=noformat(len(files)))
-    all_reports = []
-    for path in files:
-        report = process_file(config, path)
-        all_reports.extend(report)
+    all_reports = process_files(config, *files)
     return all_reports
 
 
--- a/unblob/cli.py
+++ b/unblob/cli.py
@@ -91,6 +91,13 @@ class UnblobContext(click.Context):
     help="Extract the files to this directory. Will be created if doesn't exist.",
 )
 @click.option(
+    "-f",
+    "--force",
+    is_flag=True,
+    show_default=True,
+    help="Force extraction removing previously extracted files.",
+)
+@click.option(
     "-d",
     "--depth",
     default=DEFAULT_DEPTH,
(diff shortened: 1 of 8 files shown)


unblob/cli.py
```diff
@@ -133,6 +133,13 @@ def __init__(
     show_default=True,
     help="Keep extracted chunks",
 )
+@click.option(
+    "-f",
+    "--force-override-extract-dir",
```
vlaci, line 138: Have you meant "overwrite"?

vlaci (changes requested):

unblob/cli.py
kukovecz: Ohhhhh, you are right. I made it wrong everywhere.... thanks for bringing this up.

kukovecz:
> Also, I'd write a proper error message when the extraction directory already exist and fail early instead of crashing mid-processing

Totally makes sense! At first I ignored it, since I thought there is an `ErrorReport` from it and we should deal with it in a different scope. However there is not a report and it totally makes sense to handle it beforehand. I have modified the code to check this and fail the CLI if this happens - making it a little bit more "nice", allowing empty directories to be given.

unblob/processing.py
```diff
@@ -55,6 +61,13 @@ def process_file(config: ExtractionConfig, path: Path) -> List[Report]:
         depth=0,
     )
 
+    if config.extract_root.exists() and list(config.extract_root.iterdir()):
+        if config.force_overwrite_extract_dir:
+            logger.debug("Overwriting extract root directory")
+            rmtree(config.extract_root)
+        else:
+            raise ExtractDirNotEmptyException
+
```
e3krisztian, line 70: I think it is enough to check if the directory exists or not, I do not think we need to check for if it is empty: it will be [re-]created anyway.

If the above is agreed upon, the exception `ExtractDirNotEmptyException` should also be ranamed to `ExtractDirAlreadyExistsError` (`Error` is the usual suffix for Python exceptions).

kukovecz: I don't know how the end user is expecting to use this tool. In my mind, it is a perfectly acceptable usecase to have an empty directory created and use that for extraction directory. In that case, it would be annoying to have an error to delete that empty dir, so the tool could re-create it...

But on the other hand, your solution sounds simpler.

@QKaiser / @vlaci, what do you think, how should it be? (I know this is not a big issue, but I would rather decide it here in the scope of this pull request than modifying this later on)

vlaci: I don't think, we can expect the user to figure out our directory naming scheme and create output directories ahead of time.

QKaiser: I'm with @vlaci on this one.

kukovecz: Of course, but this option is only for the extract root dir (`-e, --extract-dir DIRECTORY` parameter), not for every subsequent weird-named sub-directory :sweat_smile:

vlaci: I am confused, as extract root dir doesn't need to be empty

vlaci:
> > Also, I'd write a proper error message when the extraction directory already exist and fail early instead of crashing mid-processing
> 
> Totally makes sense! At first I ignored it, since I thought there is an `ErrorReport` from it and we should deal with it in a different scope. However there is not a report and it totally makes sense to handle it beforehand. I have modified the code to check this and fail the CLI if this happens - making it a little bit more "nice", allowing empty directories to be given.

I think it is still suboptimal: if more than one input files are passed on the command line, it will process the first, then possibly error on the second.

vlaci:
I am giving this PR another attempt

vlaci:
A possible improvement that I am thinking about is pre-creating all the `Tasks`  at the beginning so that multiple files can get processed in parallel immediately.

Also I am thinking about how to pre-generate the `_FileTasks` so that the extraction directory calculation can remain it's responsibility, but it is a tough one.

kukovecz:
I like this approach very much - detect problem as early as possible and possibly die BEFORE starting the whole processing. 

> A possible improvement that I am thinking about is pre-creating all the `Tasks` at the beginning so that multiple files can get processed in parallel immediately.

This sounds really cool. I think we can have an issue dedicated to this feature, where we can discuss how feasible it is. The only downside what I can think at this moment that our logs would be much more harder to understand, they would overlap more.

PR #321, opened 2022-03-30
pytest: do not collect tests from known non-test paths

vlaci:
Inspiration is taken from .gitignore. See the following URL for
details:
  https://docs.pytest.org/en/latest/example/pythoncollection.html?highlight=discovery#changing-directory-recursion


vlaci:
I have checked [`pytest-gitignore`](https://github.com/tgs/pytest-gitignore/blob/7dc7087f16dbc467435e08f7faeac64d7ee0f0a1/pytest_gitignore.py#L21) but I don't think that it is a particularly good solution

QKaiser (approved):
LGTM.

PR #323, opened 2022-03-30
Eliminate whole file chunks

### LINKED ISSUE #303: Do not carve out chunks when the chunk covers the whole file
If the chunk equals the whole file we can omit carving out that (basically copying) the whole file as any extractor can work on the original file instead. As we would be also cleaning up the carves after extraction there is no benefit of doing so. This could save some disk IO and also some disk space during extraction.

Further this would also help with handlers without extractor as they wont extract chunk if a handler matches the whole file.

Similar discussion from #82 #23 
---
Questions:
- how the directory structure should look like in this case? Should we skip 1 directory level in this case or we want the directory levels to be the same, regardless of this optimization?

e3krisztian:
resolves #303


martonilles:
We have had a conversation about this and the current branch removes also intermediate extracted files which should be kept. We just want to skip the carving, but not the extraction part.

Probably it would make sense to separate more explicitly the carve and extract phases.

unblob/cli.py
```diff
@@ -131,7 +131,7 @@ def __init__(
     "keep_extracted_chunks",
     is_flag=True,
     show_default=True,
-    help="Keep extracted chunks",
+    help="Keep extracted carved files around",
```
vlaci, line 134: Shouldn't we rename the option as well? E.g. `--keep-carved-files`

tests/test_cleanup.py
```diff
@@ -17,16 +16,20 @@
...
     z.close()
     return bio.getvalue()
 
 
-ZIP_BYTES = _zip_bytes()
+ZIP_BYTES = b"prefix to force carving the zip archive " + wrapzip(
+    "content.txt", _ZIP_CONTENT
+)
 DAMAGED_ZIP_BYTES = ZIP_BYTES.replace(_ZIP_CONTENT, _DAMAGED_ZIP_CONTENT)
+assert ZIP_BYTES != DAMAGED_ZIP_BYTES
```
vlaci, line 32: This looks like an independent change from the topic of this commit. It is no big deal to have a few extra changes like that in a PR but I'd like you to at least put it into separate git commits. This would make reviewing changes much easier.

tests/test_cleanup.py
```diff
@@ -92,10 +95,7 @@ class _HandlerWithNullExtractor(Handler):
     """
 
     def calculate_chunk(self, file: io.BufferedIOBase, start_offset: int) -> ValidChunk:
-        pos = file.tell()
-        length = file.seek(0, os.SEEK_END)
-        file.seek(pos, os.SEEK_SET)
-        return ValidChunk(start_offset=0, end_offset=length)
+        return ValidChunk(start_offset=start_offset, end_offset=start_offset + 1)
```
vlaci, line 98: Same, as above

unblob/processing.py
```diff
@@ -42,6 +36,8 @@ class ExtractionConfig:
     max_depth: int = DEFAULT_DEPTH
     process_num: int = DEFAULT_PROCESS_NUM
     keep_extracted_chunks: bool = False
+    eliminate_whole_file_chunks: bool = True
```
vlaci: I guess, we should reverse the meaning of this flag to match with the flag above. E.g. `keep_whole_file_chunks = False`

e3krisztian: this is a temporary flag, needed for the tests only - it is reverted in the next commit, after the test data has changed (which is a huge change, did not want to mix up with real code changes).

tests/test_cleanup.py
e3krisztian: Actually it is a required change for the tests to continue to work, as the old version would result in a whole file chunk, that is exactly the scope of this PR.

unblob/cli.py
e3krisztian: I had been thinking about it as well. We have a 3 phrases, that are differing in a nuanced way:
- carved file
- extracted file
- extracted chunk

As far as I understand "extracted chunk" means "carved file".

I will remove this commit and make a new PR that will change both the option and the internal attribute names as well.

e3krisztian: Commit has been removed.

unblob/extractor.py
```diff
@@ -15,21 +15,13 @@
...
-    extract_name = relative_path.name + APPEND_NAME
-    extract_dir = extract_root / relative_path.with_name(extract_name)
-    extract_dir.mkdir(parents=True, exist_ok=True)
-    logger.debug("Created extraction dir", path=extract_dir)
-    return extract_dir.expanduser().resolve()
-
-
 def carve_chunk_to_file(carve_path: Path, file: io.BufferedIOBase, chunk: Chunk):
     """Extract valid chunk to a file, which we then pass to another tool to extract it."""
+    assert carve_path.is_absolute()
```
martonilles, line 20: I would prefer not to have assert in a production code, that could result in DoS attacks, rather we should log it and handle it

unblob/processing.py
```diff
@@ -42,6 +36,8 @@ class ExtractionConfig:
...
+    extract_suffix: str = "_extract"
```
martonilles, line 39: do we want to make this configurable?

unblob/processing.py
```diff
@@ -127,7 +123,11 @@ def _process_task(self, result: TaskResult, task: Task):
             log.debug("Ignoring empty file")
             return
 
-        _FileTask(self._config, task, size, result).process()
+        filetask = _FileTask(self._config, task, size, result)
+        filetask.process()
+        # ensure that the root extraction directory is created even for empty extractions
+        if task.depth == 0:
+            filetask.extract_dir.mkdir(parents=True, exist_ok=True)
```
martonilles, line 129: if we can not extract anything from the file, then do we really need that empty directory?

unblob/processing.py
```diff
@@ -143,6 +143,17 @@ def __init__(
         self.size = size
         self.result = result
 
+        self.extract_dir = self._get_extract_dir()
+
+        assert self.extract_dir.is_absolute()
```
martonilles, line 147: no assert here either, check it and handle problem if exists

unblob/extractor.py
```diff
@@ -139,4 +124,4 @@ def carve_valid_chunk(
     carve_path = extract_dir / filename
     logger.info("Extracting valid chunk", path=carve_path, chunk=chunk)
     carve_chunk_to_file(carve_path, file, chunk)
-    return carve_path
+    return carve_path.resolve()
```
martonilles, line 127: why resolve it here? The carved file is created in the previous line, why do any resolve after?

also `carve_chunk_to_file` requires absolute path, so most likely it is already resolved

unblob/processing.py
```diff
@@ -143,6 +143,17 @@ def __init__(
...
+
+    def _get_extract_dir(self) -> Path:
+        """Extraction dir under root with the name of path."""
+        relative_path = self.task.path.relative_to(self.task.root)
+        extract_name = relative_path.name + self.config.extract_suffix
```
martonilles, line 152: just wondering, but do not we want to distinguish between _carve and _extract directory?

martonilles (changes requested):

unblob/extractor.py
vlaci: Better, test it. We have the code in place so that it cannot be triggered currently.

unblob/processing.py
e3krisztian: I would prefer them to be different, but would make the change in another PR.

unblob/extractor.py
e3krisztian: yes, indeed, on line 139 `extract_dir` is used, which should be an absolute path, so `resolve` is not needed.

actually it should not really matter if we are using absolute or relative paths, as long as the input ones are pointing to the input file and the extraction root can be written to/created.

unblob/processing.py
e3krisztian: I have checked it again, and previously in a few cases there were no output, and no extract directory was created. I still feel, that it is confusing and it is better to create the directory.

unblob/processing.py
e3krisztian: not necessarily, but the constant for suffix does not really hurt either at this location.

I could also think of it being configurable - in case of file name collision during extraction, which we do not handle at the moment.

unblob/extractor.py
e3krisztian: it is mostly for documentation - poor man's annotation for a pre-condition.

I do not expect it to ever fail, but if it does, we better know about it anyway.

unblob/processing.py
e3krisztian: it is for documentation only - below we can already verify what the `extract_dir` is, but I prefer to treat even helper functions as black boxes

unblob/processing.py
```diff
@@ -36,7 +36,6 @@ class ExtractionConfig:
     max_depth: int = DEFAULT_DEPTH
     process_num: int = DEFAULT_PROCESS_NUM
     keep_extracted_chunks: bool = False
-    eliminate_whole_file_chunks: bool = True
```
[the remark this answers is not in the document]
e3krisztian: done

tests/test_handlers.py
```diff
@@ -34,7 +34,6 @@ def test_all_handlers(input_dir: Path, output_dir: Path, tmp_path: Path):
         extract_root=tmp_path,
         entropy_depth=0,
         keep_extracted_chunks=True,
-        eliminate_whole_file_chunks=False,
```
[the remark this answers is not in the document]
e3krisztian: done

unblob/extractor.py
e3krisztian: removed

unblob/processing.py
e3krisztian: removed

unblob/extractor.py
e3krisztian: removed.

unblob/processing.py
e3krisztian: #326

martonilles (approved):

PR #324, opened 2022-04-04
Update Nix and ubi_reader for python-lzo

r4bbit-r4:
The package requirement for lzo was misnamed in the original ubi_reader.
Nix had to be updated as well to use the same python-lzo as unblob.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -48,6 +48,13 @@ let
           self.python-lzo
         ];
       });
+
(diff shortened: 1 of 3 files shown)


QKaiser (changes requested):
You need to update the poetry lock file based on your pyproject definition and fix the nix file because there is a recursive call.

QKaiser (approved):

PR #328, opened 2022-04-06
Stop emitting a ValidChunk for every bzip2 stream.

### LINKED ISSUE #316: Stop emitting a ValidChunk for every bzip2 stream.
We have a sample which is simply directories and files packed into a tar.bz2. 

This means that there are multiple concatenated bzip2 streams, each of them containing a tar archive. This leads the extraction directory to look like this:

- 2752647-2925267.bzip2_extract/2752647-2925267_extract/5088-899072.tar_extract/etc/ssl/certs/TWCA_Global_Root_CA.pem
- 31919792-32267698.bzip2_extract/31919792-32267698_extract/724992-898560.tar_extract/usr/share/alsa

While it ideally should look like:

- A-B.bzip2_extract/C-D.tar_extract/etc/ssl/certs/TWCA_Global_Root_CA.pem
- A-B.bzip2_extract/C-D.tar_extract/usr/share/alsa

The only way we can do that is to emit a single ValidChunk covering the whole set of bzip2 streams rather than one chunk per stream.

QKaiser:
Regardless of what is in concatenated bzip streams, we should unpack it into a single file. Unblob will pickup the chunks anyway when parsing the output of bzip decompression. It's faster and more efficient.

Fix #316

### DIFF
--- a/unblob/handlers/compression/bzip2.py
+++ b/unblob/handlers/compression/bzip2.py
@@ -13,7 +13,8 @@ logger = get_logger()
 
 BLOCK_HEADER = 0x0000_3141_5926_5359
 BLOCK_ENDMARK = 0x0000_1772_4538_5090
-
+# BCD-encoded digits of BLOCK_HEADER
+COMPRESSED_MAGIC = b"1AY&SY\x86\xc6"
 COMPRESSED_MAGIC_LENGTH = 6 * 8
 
 BLOCK_ENDMARK_SHIFTED = BLOCK_ENDMARK << COMPRESSED_MAGIC_LENGTH
@@ -52,8 +53,16 @@ class BZip2Handler(StructHandler):
         self, file: io.BufferedIOBase, start_offset: int
     ) -> Optional[ValidChunk]:
 
-        self.parse_header(file, Endian.BIG)
-        end_block_offset = self.bzip2_recover(file, start_offset)
+        header = self.parse_header(file, Endian.BIG)
+        end_block_offset = -1
+        while header.compressed_magic == COMPRESSED_MAGIC:
+            current_offset = file.tell() - len(header)
+            end_block_offset = self.bzip2_recover(file, current_offset)
+            try:
+                file.seek(end_block_offset + FOOTER_SIZE, io.SEEK_SET)
+                header = self.parse_header(file, Endian.BIG)
+            except EOFError:
+                break
 
         return ValidChunk(
             start_offset=start_offset,
--- a/tests/integration/compression/bzip2/__input__/collated.bzip2
+++ b/tests/integration/compression/bzip2/__input__/collated.bzip2
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:882c591e5eddf406213eea5b0235ea08e8e73e0507d3491f453f3c286c17de65
+size 5090
--- a/tests/integration/compression/bzip2/__input__/collated.headpad.bzip2
+++ b/tests/integration/compression/bzip2/__input__/collated.headpad.bzip2
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:7f320bd0a2dd87b00d9ba78a778747c4094fbfa6aaff108d061ddd00074fa8b9
+size 5107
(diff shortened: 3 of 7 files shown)


unblob/handlers/compression/bzip2.py
```diff
@@ -52,8 +52,16 @@ def calculate_chunk(
         self, file: io.BufferedIOBase, start_offset: int
     ) -> Optional[ValidChunk]:
 
-        self.parse_header(file, Endian.BIG)
-        end_block_offset = self.bzip2_recover(file, start_offset)
+        header = self.parse_header(file, Endian.BIG)
+        end_block_offset = -1
+        while header.compressed_magic == b"1AY&SY\x86\xc6":
```
martonilles, line 57: I would make it a CONSTANT and explain the value

https://stackoverflow.com/questions/18262703/bzip2-block-header-1aysy

martonilles, line 57: I would make it a CONSTANT and explain the value

https://stackoverflow.com/questions/18262703/bzip2-block-header-1aysy

martonilles (changes requested):

martonilles:
+1 I would also add a test file as well

unblob/handlers/compression/bzip2.py
```diff
@@ -52,8 +52,16 @@ def calculate_chunk(
...
+            current_offset = file.tell() - len(header)
```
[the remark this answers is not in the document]
QKaiser: How could it be off ? We receive the file with cursor set to `start_offset`, we then call `parse_header` which will read `len(header)` bytes from the file and move the cursor doing so. At this point, `file.tell()` is right after the header, so if we want the start offset of the bzip stream we subtract `len(header)`.

I did some testing and it works:

```
echo 'AAAAAAAAA' > /tmp/out.bzip
dd if=tests/integration/compression/bzip2/__input__/lorem.txt.bz2 >> /tmp/out.bzip
dd if=tests/integration/compression/bzip2/__input__/lorem.txt.bz2 >> /tmp/out.bzip
poetry run unblob -v /tmp/out.bzip 
--snip--
2022-04-07 15:50.39 [warning  ] Found unknown Chunks           chunks=[0x0-0xa] pid=1356921
2022-04-07 15:50.39 [info     ] Extracting unknown chunk       chunk=0x0-0xa path=out.bzip_extract/0-10.unknown pid=1356921
2022-04-07 15:50.39 [debug    ] Carving chunk                  path=out.bzip_extract/0-10.unknown pid=1356921
2022-04-07 15:50.39 [debug    ] Calculating entropy for file   path=out.bzip_extract/0-10.unknown pid=1356921 size=0xa
2022-04-07 15:50.39 [debug    ] Entropy calculated             highest=5.86 lowest=5.86 mean=5.86 pid=1356921
2022-04-07 15:50.39 [info     ] Extracting valid chunk         chunk=0xa-0x13ec path=out.bzip_extract/10-5100.bzip2 pid=1356921
2022-04-07 15:50.39 [debug    ] Carving chunk                  path=out.bzip_extract/10-5100.bzip2 pid=1356921
2022-04-07 15:50.39 [debug    ] Running extract command        command=7z x -y /home/quentin/unblob/out.bzip_extract/10-5100.bzip2 -o/home/quentin/unblob/out.bzip_extract/10-5100.bzip2_extract pid=1356921
2022-04-07 15:50.39 [debug    ] Removing extracted chunk       path=out.bzip_extract/10-5100.bzip2 pid=1356921
2022-04-07 15:50.39 [debug    ] Found directory                path=out.bzip_extract/10-5100.bzip2_extract pid=1356923
2022-04-07 15:50.39 [debug    ] Processing file                path=out.bzip_extract/10-5100.bzip2_extract/10-5100 pid=1356922 size=0x3340
--snip--
```

unblob/handlers/compression/bzip2.py
QKaiser: Fixed.

QKaiser: Double fixed.

QKaiser:
> +1 I would also add a test file as well

Added test cases for:
- two subsequent bzip streams (collated.bzip)
- two subsequent bzip streams with prior padding (collated.headpad.bzip)

PR #329, opened 2022-04-06
Migrate YARA based pattern matching to hyperscan

### LINKED ISSUE #320: Speed-up yara result evaluation and handler chunk calculation
Currently we are processing yara results in multiple passes:
- do yara search in multiple steps based on priority
- processes yara matches (rule match) on each priority level
- process string matches in each yara match by starting offset priority
If a given string match start offset is withing an already identified chunk (from previous priorities/matches/strings) we skip handler check as that would result in an inner chunk which would be discarded later.

However when processing a large tar file (or any other chunk where the inner content is "visible") the other rules could match as well. (eg: a tar file with many lzma, zip, elf etc. file within the tar) This would result a high number of match that would be later ignored, but would take a long time to process prior these are ignored. This is especially problematic as in many case chunks cover the whole file so once the chunk is identified any further processing at that depth is meaningless.

We could implement a few optimizations:
- change the current priority based processing (either prioritize these container or prioritize handlers based on their accuracy/confidence)
- order string result processing by offset across all matches (not just within one match) and maybe combine it with reworked priorities
- stop processing any further result if a chunk covers the whole file (special case optimization, though this is a very common special case)

We either need to rework the YARA match handling logic or find another pattern matching strategy (especially as we can not use yara conditions in rules #312). (Or implement different processing logic in YARA, but probably it does not worth the effort) Unfortunately we can not use YARA callbacks as those are reporting on yara rules (also in the order how yara rules are defined, not the string match order), not just yara string matches which would be something we need.

For alternative matching we should look at hyperscan which is an optimized PCRE matching library (https://github.com/intel/hyperscan & https://python-hyperscan.readthedocs.io/en/latest/)

martonilles:
closes #320


martonilles:
We would need some nix magic to get hyperscan working as well.

unblob/parser.py
```diff
@@ -0,0 +1,87 @@
...
+
+    item: LITERAL           -> literal
+        | WILDCARD          -> wildcard
+        | FIRSTNIBLE        -> first_nible
+        | SECONDNIBLE       -> second_nible
+        | JUMP              -> jump
+        | RANGE_JUMP        -> range_jump
+        | alternative
+
+    alternative: "(" item (ALTERNATIVE_SEPARATOR item+)+ ")"
```
e3krisztian, line 29: The first alternative can not be a sequence of e.g. `LITERAL`s - why do we have this limitation?

For consistency, I would also prefer `"|"` here instead of `ALTERNATIVE_SEPARATOR`, as we already have `"("` and `")"`.

tests/test_finder.py
```diff
@@ -1,65 +1,164 @@
...
+    def calculate_chunk(self, file, start_offset: int):
+        raise EOFError()
 
-    assert len(results) == 2
-    result1, result2 = results
 
-    assert result1.handler is handler1
-    assert result1.match.strings == [(0, "$handler1_magic", b"!<")]
+class TestHandlerInvalid(Handler):
+    NAME = "handlerEOF"
```
e3krisztian, line 51: `handlerInvalid`

unblob/finder.py
```diff
@@ -2,166 +2,202 @@
...
+    handler_map: Dict[int, Handler]
+    file: File
+    file_size: int
+    all_chunks: List
+    task_result: TaskResult
+
+
+class ScanContinue(Flag):
+    Continue = False
+    Terminate = True
```
e3krisztian, line 33: I feel the enum should be named `ScanTerminate` instead, as it is a surprising definition: the value of `ScanContinue.Continue` is `False`.

e3krisztian: Maybe another option would be to prefix with `_HyperScan` to make it
- module private
- explain, that it is used by Hyperscan only.

unblob/handlers/compression/lzh.py
```diff
@@ -14,25 +14,22 @@ class LZHHandler(StructHandler):
...
+        Regex(r"-lz4-"),
+        Regex(r"-lh1-"),
+        Regex(r"-lh2-"),
+        Regex(r"-lh3-"),
+        Regex(r"-lh4-"),
+        Regex(r"-lh5-"),
+        Regex(r"-lh6-"),
+        Regex(r"-lh7-"),
+        Regex(r"-lh8-"),
+        Regex(r"-lhd-"),
```
e3krisztian, line 29: HyperScan might internally produce a similar output:
```py
Regex(r"""
    -lzs- |
    -lz4- |
    -lh0- |
    -lh1- |
    -lh2- |
    -lh3- |
    -lh4- |
    -lh5- |
    -lh6- |
    -lh7- |
    -lh8- |
    -lhd-
    """, re.VERBOSE)
```

unblob/handlers/filesystem/fat.py
```diff
@@ -134,6 +136,11 @@ def valid_header(self, header: Instance) -> bool:
         return True
 
     def calculate_chunk(self, file: File, start_offset: int) -> Optional[ValidChunk]:
+        signature_start = start_offset + 510
+        signature_end = start_offset + 512
+        if file[signature_start:signature_end] != b"\x55\xaa":
+            raise InvalidInputFormat("Signature missmatch")
```
e3krisztian, line 142: typo: mismatch

tests/test_parser.py
```diff
@@ -0,0 +1,72 @@
...
+def test_simple_convert(hex_string, expected_regex):
+    regex = hexstring2regex(hex_string)
+
+    assert expected_regex == regex
+
+
+def test_single_line_comment():
+    regex = hexstring2regex(
+        """
+  // inital comment
```
e3krisztian, line 53: I would have preferred `#` for comment start - this is a DSL embedded in Python code, so my preference is to have the same construct than the host language if possible.

e3krisztian:
I like this implementation, it is very clean, with only minor cosmetic problems. The code has definitely changed for the better! :)

vlaci:
I have observed an extreme performance degradation in `test_handlers.py`. Interestingly the bottleneck is somehow the multiprocessing code. 

I have tried profiling with py-spy but yet to dive deeper into the traces but maybe pickling became too expensive of some reason.

Checked using
```
py-spy record --format speedscope -o profile.sc -s -n -F -- py.test tests/test_handlers.py -k test_all_handlers
```
(trace can be visualized on https://speescope.app)

unblob/finder.py
```diff
@@ -2,166 +2,202 @@
...
+    handlers: Tuple[Type[Handler], ...]
+) -> Tuple[hyperscan.Database, Dict]:
+    db = hyperscan.Database(mode=hyperscan.HS_MODE_VECTORED)
+    handler_map = dict()
+
+    pattern_id = 0
+    patterns = []
+    for handler in handlers:
+        h = handler()
+        for pattern in handler.PATTERNS:
```
kukovecz, line 184: This is a little bit confusing, I would:

1. rename the `handler` variable to `handler_class`
2. rename the `h` variable to `handler`
3. keep the `handler.PATTERNS` as `handler.PATTERNS`. 

At this point, we are going through the handler classes, creating a handler object from it and use that object to deal with stuff. I think with these changes, it would be more easier to understand what is going on.

unblob/finder.py
```diff
@@ -2,166 +2,202 @@
...
+
+class ScanContinue(Flag):
+    Continue = False
+    Terminate = True
+
+
+def _calculate_chunk(
+    handler: Handler, file: File, real_offset, task_result: TaskResult
+) -> Optional[ValidChunk]:
+    # limited_reader = LimitedStartReader(file, real_offset)
```
kukovecz, line 39: Please delete unused / out-of-date stuff instead of just commenting them.

tests/test_finder.py
```diff
@@ -1,65 +1,164 @@
...
+
+        def calculate_chunk(self, file, start_offset: int):
+            pass
+
+    with pytest.raises(ValueError):
+        build_hyperscan_database(tuple([TestHandlerA, TestHandlerB, InvalidHandler]))
+
+
+@pytest.mark.parametrize(
+    "content, expected_chunks",
```
kukovecz, line 115: The `expected_chunks` in these parameters are in-fact, just tuples with numbers. Can't they be `Chunk`s (or `ValidChunk`?)?

kukovecz (commented):
Small comments here and there, but overall looks like a great improvement!

unblob/parser.py
martonilles: You are right about the first, that was a typo, added a testcase and fixed it.

`"|"` would be swallowed by the parser, while `ALTERNATIVE_SEPARATOR` is a terminal and hence part of the parser result, which can be used by the Transformer visitor to group items together.

tests/test_finder.py
martonilles: fixed

unblob/handlers/archive/ar.py
```diff
@@ -18,21 +18,18 @@
...
+        )
+    ]
 
     EXTRACTOR = Command("7z", "x", "-y", "{inpath}", "-o{outdir}")
 
-    def calculate_chunk(
-        self, file: io.BufferedIOBase, start_offset: int
-    ) -> Optional[ValidChunk]:
-
-        ar = arpy.Archive(fileobj=file)  # type: ignore
```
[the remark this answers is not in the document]
martonilles: it is there indeed, but github viewer did not pick it up
https://github.com/IoT-Inspector/unblob/pull/329/commits/327e7d975b837d11885fb5448370971942cca6d9

unblob/handlers/compression/lzh.py
martonilles: indeed it will be the something similar, I would just keep it for easier readability

unblob/handlers/filesystem/fat.py
martonilles: done

pyproject.toml
```diff
@@ -27,6 +27,7 @@ plotext = "^4.1.5"
 pluggy = "^1.0.0"
 file-magic = "^0.4.0"
 hyperscan = "^0.2.0"
+lark = "^1.1.2"
```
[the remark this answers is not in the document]
martonilles: done

unblob/models.py
```diff
@@ -154,14 +141,29 @@ def extract(self, inpath: Path, outdir: Path):
         """
 
 
+class Pattern(str):
+    def as_regex(self) -> bytes:
+        raise NotImplementedError
+
+
+class Hexstring(Pattern):
```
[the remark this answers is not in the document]
martonilles: added a detailed documentation and links.

max jumpsize is a hyperscan bug, trying to reproduce it and probably report upstream

unblob/finder.py
```diff
@@ -2,166 +2,202 @@
...
+import attr
+import hyperscan
 from structlog import get_logger
 
-from .file_utils import InvalidInputFormat, LimitedStartReader
+from .file_utils import InvalidInputFormat
 from .handlers import Handlers
 from .logging import noformat
-from .models import Handler, TaskResult, ValidChunk, YaraMatchResult
+from .models import File, Handler, TaskResult, ValidChunk
```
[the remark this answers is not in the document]
martonilles: indeed models.py is bit awkward and we should clean it up a bit.

Puting `File` in models.py wold not work due to circular references to `file_utils`. Maybe we should move some of the file_utils functions referencing `File` to be methods of `File`, let's see what pattern would be easier to use.

unblob/finder.py
martonilles: done

unblob/finder.py
martonilles: done

tests/test_finder.py
martonilles: done

martonilles:
> I have observed an extreme performance degradation in `test_handlers.py`. Interestingly the bottleneck is somehow the multiprocessing code.
> 
> I have tried profiling with py-spy but yet to dive deeper into the traces but maybe pickling became too expensive of some reason.
> 
> Checked using
> 
> ```
> py-spy record --format speedscope -o profile.sc -s -n -F -- py.test tests/test_handlers.py -k test_all_handlers
> ```
> 
> (trace can be visualized on https://speescope.app)

Basically there are 3 expensive things we do and expensive:
- Lark.parse() (biggest slow down)
- hyperscan database compile
- file magic lookup

Adjusting the parser to use the following to `parser="lalr", lexer="basic",` eliminated the performance difference, though hyperscan.Database compile is still there, but that seems insignificant compare to the Lark parsing.

unblob/models.py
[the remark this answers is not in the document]
martonilles: done

unblob/parser.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def start(self, s):
+        return "".join(s).encode()
+
+
+def hexstring2regex(hexastr):
+    try:
+        parsed = _hex_string_parser.parse(hexastr)
+    except UnexpectedCharacters as e:
+        raise ValueError(str(e)) from e
```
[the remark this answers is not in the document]
martonilles: done

kukovecz (approved):

unblob/parser.py
```diff
@@ -0,0 +1,95 @@
...
+        return f"({alternatives})"
+
+    def item(self, s):
+        return s[0]
+
+    def start(self, s):
+        return "".join(s).encode()
+
+
+class InvalidHexString(Exception):
```
e3krisztian, line 85: I have seen the changes due to this specialized exception, ideally this should be a subclass of `ValueError`.

e3krisztian (approved):
I think this is an significant improvement, ideally it would be nice to have further minor polish, but I think those can be made later in follow up PRs.

This is good as it is now! :)

unblob/parser.py
martonilles: done

tests/test_parser.py
martonilles: ok, resolving it now

PR #330, opened 2022-04-06
Work around structlog's ConsoleRenderer failure

### LINKED ISSUE #292: Exception during logging an exception
In #291 the exception is not handled properly:

```
2022-03-10 23:41.32 [error    ] Unknown error happened         exec_info=TypeError('can only concatenate str (not "error") to str') pid=8
  File "/home/unblob/.local/lib/python3.8/site-packages/unblob/finder.py", line 85, in search_chunks_by_priority
  File "/home/unblob/.local/lib/python3.8/site-packages/unblob/handlers/filesystem/romfs.py", line 64, in calculate_chunk
    if not valid_checksum(file.read(512)):
  File "/home/unblob/.local/lib/python3.8/site-packages/unblob/handlers/filesystem/romfs.py", line 24, in valid_checksum
    total += struct.unpack(">L", content[i : (i + 4)])[0]  # noqa: E203
struct.error: unpack requires a buffer of 4 bytes

During handling of the above exception, another exception occurred:

  File "/home/unblob/.local/lib/python3.8/site-packages/unblob/processing.py", line 94, in process_task
  File "/home/unblob/.local/lib/python3.8/site-packages/unblob/processing.py", line 139, in _process_task
    self._process_regular_file(task, size, result)
  File "/home/unblob/.local/lib/python3.8/site-packages/unblob/processing.py", line 144, in _process_regular_file
    all_chunks = search_chunks_by_priority(
  File "/home/unblob/.local/lib/python3.8/site-packages/unblob/finder.py", line 109, in search_chunks_by_priority
    logger.error(
  File "/home/unblob/.local/lib/python3.8/site-packages/structlog/_log_levels.py", line 124, in meth
    return self._proxy_to_logger(name, event, **kw)
  File "/home/unblob/.local/lib/python3.8/site-packages/structlog/_base.py", line 203, in _proxy_to_logger
    args, kw = self._process_event(method_name, event, event_kw)
  File "/home/unblob/.local/lib/python3.8/site-packages/structlog/_base.py", line 160, in _process_event
    event_dict = proc(self._logger, method_name, event_dict)
  File "/home/unblob/.local/lib/python3.8/site-packages/structlog/dev.py", line 425, in __call__
    sio.write("\n" + exc)
TypeError: can only concatenate str (not "error") to str
```

e3krisztian:
resolves #292


unblob/logging.py
```diff
@@ -66,6 +66,31 @@ def add_pid_to_log_message(
...
+    This is necessary a hackish solution, that depends on structlog.processors.format_exc_info
+    to do its work after it.
+    """
+    try:
+        exc = event_dict["exception"]
+        if isinstance(exc, Exception):
+            event_dict["exc_info"] = exc
+    except KeyError:
+        pass
+    return event_dict
```
kukovecz, line 91: This eventually duplicates the value stored under the `exception` to the `exc_info`.
I would *move* it rather then *copy* it. 
e.g. in this case, I think we don't need the original `exception` key (let structlog handle that).

unblob/logging.py
```diff
@@ -66,6 +66,31 @@ def add_pid_to_log_message(
...
+    ConsoleRenderer has "special" requirements for keys, e.g. raises an Exception if there
+    is an "exception" key in the passed in event_dict, and its value is an Exception instance.
+
+    Counterintuitively ConsoleReader requires the exception's value to be a string!
+
+    We have some code that builds a dict for logging that includes this "exception" key:
+    e.g. logging report.UnknownError and its derivatives.
+
+    This is necessary a hackish solution, that depends on structlog.processors.format_exc_info
+    to do its work after it.
```
kukovecz, line 83: Trying to clarify this a little bit:

I can see why are you calling this a "hackish" solution, but in a way it is aligned with the standard library logging. When you log an exception in an except block, under the hood an error is logged with the `exc_info` filed with the exception object: https://github.com/python/cpython/blob/c12ba6b2ff7908c8970b978f149d51ecd3fb195c/Lib/logging/__init__.py#L1521.  Structlog' `ConsoleRenderer` relies on this behavior by getting the `exc_info` and making a pretty string from it and putting it to the `exception` key.

The "problem" with our code is that we call the exception object `exception` rather than `exc_info`, so the pure exception object will be the value of the `exception` key and not the pretty string. We could simple rename the `exception` attribute of the `UnknownError` to `exc_info`, but that maybe would make our code uglier, so I think this (include this into the log processor chain) is the correct way of solving this problem.

unblob/logging.py
e3krisztian: The functionality is there, as `structlog.processors.format_exc_info` will replace the `exception` key if `exc_info` is present.

You are right however, that by moving it it becomes more evident - will move it, this is the proper solution - thanks!

unblob/logging.py
vlaci: Another probing question: Why do we need to store exception objects in the first place? Isn't it enough to format the backtrace and store that? An exception has references top all of the associated frame objects and transitively to all variables that were in scope in each of the affected frames. I can see a  possibility of having some hard to detect bugs by keeping exceptions until the end of the program lifetime.

unblob/logging.py
e3krisztian: workaround is different: model change

PR #334, opened 2022-04-07
sevenzip: check header checksum to eliminate false-positive matches

martonilles:


QKaiser (approved):
The `[-START_HEADER_SIZE:]` array construct confused me at first but it's all good :)

PR #335, opened 2022-04-07
lzh: check header length to eliminate some false-positive

martonilles:
7z checks also for minimum length as well. Maybe we can also check some sizes
for sanity as well.


QKaiser:
I guess we could validate the `header_checksum` too.

martonilles:
> I guess we could validate the `header_checksum` too.

actually header_checksum are part of the header size for level 2 archives, so this would work only for level 0 & 1.
There is also a checksum for level 0 & 1 archives for the content, but that would require uncompromising the whole content

My goal was to replicate the same checks that 7z does to quickly verify if the archive is valid or not.

QKaiser (approved):

PR #336, opened 2022-04-08
Cpio padding fix

martonilles:
There are two paddings in CPIO: one for the header + name and an other for
    the filesize. We calculated one for the header and one for name + filesize
    which is incorrect, though worked in some cases when things aligned well.
    
    Old ascii format uses no padding (paddig=1) instead of 2 bytes.


QKaiser (approved):
Nice find, although I feel really dumb as the guy behind this handler now :)

PR #340, opened 2022-04-12
Remove handler priority

### LINKED ISSUE #338: Remove priorities from chunk search with handler
We do not need priorities with hyperscan, but rather search for patterns from all handlers at the same time.

martonilles:
closes #338


unblob/models.py
```diff
@@ -243,20 +243,15 @@ def parse_header(self, file: File, endian=Endian.LITTLE):
 
 
 class Handlers:
```
vlaci, line 245: At this point this could be just a plain list. We would loose the immutability aspects but I am not sure having it worth the extra complexity

martonilles: done. kept Handlers as a type but it is only a tuple definition

vlaci (approved):

PR #343, opened 2022-04-13
Fix bzip2 compressed magic handling & use hyperscan instead of iterbits to search for end of stream magic

martonilles:
There is a stream header and a block header in bzip2. Stream header includes the gile magic (BZ), the version and the 100k_blocksize.
    
Block header includes the magic, a crc and a randomised bit flag.

Properly separated stream header, footer and block header handling.

Instead of iterating bit by bit which is super slow, we pre-calculate each possible stream end magic bytes (thanks Florian for the inspiration) and try to match on those using hyperscan.

### DIFF
--- a/tests/handlers/compression/test_bzip2.py
+++ b/tests/handlers/compression/test_bzip2.py
@@ -3,8 +3,17 @@ import pytest
 from unblob.file_utils import File, InvalidInputFormat
 from unblob.handlers.compression.bzip2 import BZip2Handler
 
-BLOCK_HEADER = b"\x31\x41\x59\x26\x53\x59"
-BLOCK_ENDMARK = b"\x17\x72\x45\x38\x50\x90"
+STREAM_MAGIC = b"BZ"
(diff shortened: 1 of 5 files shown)


QKaiser:
It does not look compatible with this recent change: https://github.com/IoT-Inspector/unblob/pull/328/files

QKaiser:
> It does not look compatible with this recent change: #328

I mean in terms of emitting one chunk for multiple concatenated bzip2 streams, not on correctness :)

martonilles:
> > It does not look compatible with this recent change: #328
> 
> I mean in terms of emitting one chunk for multiple concatenated bzip2 streams, not on correctness :)

Well, actually it does emit one chunk for conctenated bzip2 streams and there are also both integration and unit tests on this. ;)

QKaiser (approved):
@martonilles walked me through the changes, I checked the branch and looked through the test cases. To me it's good to go.

PR #346, opened 2022-04-16
lzma: ignore very small truncated streams

martonilles:
LZMA can pick up many false positives by considering valid LZMA chunks cases
when the stream is terminated early. In case the actual read data is less
then 10% of the expected uncompressed size we just consider the chunk
invalid. (Maybe we can even increase the ratio, but wanted to start at a
very safe point.)


QKaiser (approved):

QKaiser:
Let's start with 10% and see if we need to increase based on usage.

PR #347, opened 2022-04-16
Extend magic skip list

### LINKED ISSUE #339: add magic file to the exclude file types
magic compiled database has many false-positive match, probably we should skip it from processing.

file type:
`magic binary file for file(1) cmd (version 14) (little endian)`

martonilles:
Extend the default list of skipped files by their magic type
    
    Certain files are not worth trying to extract further as there is usually
    nothing that can be extracted, but due to their file type could lead to
    many false detection and failed extraction.

close #339

### DIFF
--- a/unblob/processing.py
+++ b/unblob/processing.py
@@ -28,7 +28,17 @@ logger = get_logger()
 
 DEFAULT_DEPTH = 10
 DEFAULT_PROCESS_NUM = multiprocessing.cpu_count()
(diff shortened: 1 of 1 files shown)


QKaiser (commented):
I'm ok with the new list, let's just wait for #340 to land first. I guess a comment explaining why we skip those file types wouldn't hurt.

QKaiser (approved):

PR #354, opened 2022-04-22
Add a copy_to_file utility function to file_utils.

QKaiser:
Extractors tend to copy content from the input file into separate
output files by seeking to specific offset and reading a set amount of
bytes. This is a behavior we observed multiple times in our internal
handlers.

Depending on the format and what's packed in it, these lengths can be
quite large and could be a problem in terms of RAM usage.

The 'copy_to_file' function include everything (opening the output file,
seeking to the right offset, reading bytes), while limiting memory usage
by using our iterate_file Iterator that reads 4KB at a time.

### DIFF
--- a/unblob/file_utils.py
+++ b/unblob/file_utils.py
@@ -234,6 +234,12 @@ def iterate_file(
         yield data
 
 
+def copy_to_file(input_file: File, output_path: Path, offset: int, length: int):
+    with output_path.open("wb") as f:
+        for line in iterate_file(input_file, offset, length):
+            f.write(line)
+
+
 class StructParser:
     """Wrapper for dissect.cstruct to handle different endianness parsing dynamically."""
 
--- a/vulture_whitelist.py
+++ b/vulture_whitelist.py
@@ -2,7 +2,7 @@ import sys
 
 import unblob.plugins
 from unblob import cli
-from unblob.file_utils import File, iterbits
+from unblob.file_utils import File, copy_to_file, iterbits
 from unblob.parser import _HexStringToRegex
 
 _HexStringToRegex.literal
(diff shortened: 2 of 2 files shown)


unblob/file_utils.py
```diff
@@ -234,6 +234,12 @@ def iterate_file(
         yield data
 
 
+def copy_to_file(input_file: File, output_path: Path, offset: int, length: int):
```
e3krisztian, line 237: There is a `carve_chunk_to_file` in `extractor.py` that does something like this, except it takes `Chunk` instead of `offset` and `length` and most importantly creates the parent directory if it is missing!

`copy_to_file` is fine in `file_utils`, but then the `carve_chunk_to_file` should go.

unblob/file_utils.py
```diff
@@ -234,6 +234,12 @@ def iterate_file(
...
+    with output_path.open("wb") as f:
+        for line in iterate_file(input_file, offset, length):
```
e3krisztian, line 239: a data block/chunk is read, not a `line` (which implies text and line terminator).

e3krisztian (changes requested):

unblob/file_utils.py
```diff
@@ -234,6 +234,12 @@ def iterate_file(
         yield data
 
 
+def copy_to_file(input_file: File, output_path: Path, offset: int, length: int):
+    with output_path.open("wb") as f:
```
martonilles, line 238: I would also add a check somehow to see if the output path is within the extraction root, this is something we should check always in extractors to avoid any path traversal issue

QKaiser:
Closing this pull request as we decided to follow https://github.com/IoT-Inspector/unblob/pull/354#discussion_r857398726

PR #357, opened 2022-04-28
Basic metadata

### LINKED ISSUE #327: Basic metadata on chunks & extracted files
As limited meta-data report, we want to report on chunks & extracted files to represent the layers of a firmware image.
We need to represent the relationship between files & chunks, but do not need details like file permission/owner, additional details from chunks etc. Those can be added later in #16.

Our goal is to export the following as a first milestone:

### FilesystemObject
- parent (Chunk, null for root)
- children (list of Chunk, could be zero)
- path
- type (File, Directory, Device, Symlink etc.)
- size
- file magic (if available)

### Chunk
- parent (File)
- children (list of Files, could be zero)
- start/end offset
- length
- type (handler)

Include any extraction/processing error.
Store the result in a json (or similar machine readable file, which can be later extended with additional meta-data)

e3krisztian:
resolves #327


martonilles:
Generally I like the PR a few missing/inconsistent things:
- unknown chunks are missing from the report
- it is very difficult to link Tasks to the Chunks they originate from (I can probably figure out from the path, but a more explicit reference would be better)
- I would also assign an ID to each task, so linking subtasks and tasks would be much easier, instead of just using the path again
- Error reports should be assigned to the chunk/file somehow, now it is difficult to see which chunk or file generated the report, usually it is possible to figure out from the error text the path and find the corresponding chunk/file, but explicit linking would be better (eg: ExtractCommandFailedReport, but could be anything other error report as well)
- `--json-report` currently overwrites the file if it already exists, I would only do it if `-f` is configured, otherwise I would error out
- in the Magic report I would store the magic_mime as well (though this could go to a separate PR)
- in the StatReport for symlinks I would store the symlink target as well (this could be a separate PR as well)

Nice job!

martonilles (changes requested):
Generally I like the PR a few missing/inconsistent things:
- unknown chunks are missing from the report
- it is very difficult to link Tasks to the Chunks they originate from (I can probably figure out from the path, but a more explicit reference would be better)
- I would also assign an ID to each task, so linking subtasks and tasks would be much easier, instead of just using the path again
- Error reports should be assigned to the chunk/file somehow, now it is difficult to see which chunk or file generated the report, usually it is possible to figure out from the error text the path and find the corresponding chunk/file, but explicit linking would be better (eg: ExtractCommandFailedReport, but could be anything other error report as well)
- `--json-report` currently overwrites the file if it already exists, I would only do it if -f is configured, otherwise I would error out
- in the Magic report I would store the magic_mime as well (though this could go to a separate PR)
- in the StatReport for symlinks I would store the symlink target as well (this could be a separate PR as well)

Nice job!

QKaiser:
In full agreement with vlaci here. A single `--report` option is sufficient, we can add output format specification later if users actually request it, and if so we should emit a single type of report per run (JSON, CSV, XML).

tests/test_report.py
```diff
@@ -0,0 +1,68 @@
...
+            "-m",
+            "unblob.cli",
+            "--extract-dir",
+            str(extract_root),
+            "--report",
+            str(report_file),
+            str(input_file),
+        ]
+        print(f"command=\n{' '.join(cmd)}")
+        result = subprocess.run(cmd, executable=sys.executable)
```
vlaci, line 46: I didn't expect this test to execute the unblob cli in a subprocess. Can you elaborate? Not even `test_cli` does that.

unblob/models.py
```diff
@@ -116,6 +118,7 @@ def add_subtask(self, task: Task):
 
 @attr.define
 class ProcessResult:
+    # TODO: rethink the type of ProcessResult.results - is it a simple list?
```
vlaci, line 144: This TODO is not resolved in this PR. If we want to track this question, you should create an issue

unblob/models.py
```diff
@@ -131,6 +134,25 @@ def update(self, other: "ProcessResult"):
     def register(self, result: TaskResult):
         self.results[result.task] = result
 
+    def to_json(self, indent="  "):
+        return json.dumps(list(self.results.values()), cls=_JSONEncoder, indent=indent)
+
+
+class _JSONEncoder(json.JSONEncoder):
```
vlaci, line 175: I know, we had discussed IRL about using `attr.asdict`-s features to do the type conversion. I don't recall exactly why we didn't go down that route. If we could create a Python representation of the json, which can be converted to string without a custom encoder would help testing the feature tremendously.

tests/test_report.py
e3krisztian: At the time this test was written the logic to write the report was part of the `cli.py`, so that must have been called, however I do not remember the reason anymore, it is definitely no longer valid anymore.

unblob/models.py
e3krisztian: I remember you said something about it, but most probably the channel was bad/noisy and I did not understand your proposal.

unblob/processing.py
```diff
@@ -113,21 +131,62 @@ def process_result(pool, result):
...
+        return True
+
+    if report_file.exists() and not config.force_extract:
+        logger.error("Report file exists and --force not specified.", path=report_file)
+        return False
+
+    try:
+        if report_file.exists():
+            logger.info("Removing existing report file", path=report_file)
+        report_file.write_text("")
```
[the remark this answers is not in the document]
e3krisztian: This function is responsible to ensure (as much as possible), that the requested report can be written. This `write_text` is an (imperfect) attempt to test for that - it is removed immediately after being written to.

unblob/models.py
e3krisztian: TODO was resolved (type has been changed).

tests/test_report.py
```diff
@@ -0,0 +1,68 @@
...
+            stderr=b"stdout is pretty strange ;)",
+            exit_code=1,
+        )
+
+        task_result.add_report(report)
+        json_text = ProcessResult(results={task_result.task: task_result}).to_json()
+
+        # output must be a valid json string
+        assert isinstance(json_text, str)
+        json.loads(json_text)
```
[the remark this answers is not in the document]
e3krisztian: Some extra tests are added to unblob/test_reports.py.

It is not trivial if this was worth: they are big, not so easy to understand and fragile. We will see.

tests/test_report.py
```diff
@@ -0,0 +1,543 @@
...
+                        "end_offset": 256,
+                        "extraction_reports": [
+                            {
+                                "__typename__": "ExtractCommandFailedReport",
+                                "command": "dump all bytes",
+                                "exit_code": 1,
+                                "severity": "WARNING",
+                                "stderr": "stdout is pretty strange ;)",
+                                "stdout": (
+                                    "b'\\x00\\x01\\x02\\x03\\x04\\x05\\x06\\x07"
```
vlaci, line 183: This looks wrong, the binary contents seems to be double-encoded here.

vlaci (commented):
I have yet to fully understand the last two tests. Also, the last two fail one on my machine as well (the output order is different)

tests/test_report.py
e3krisztian: This is a regression test, that was triggered by a failing command.

Although the output looks non-trivial/funny, there is no double-encoding here - it is the python `repr` of a byte-string.

-----

The original content (stdout) is a byte-array, which means that it needs to be converted somehow to a readable representation for JSON, whenever possible it is decoded as UTF-8 and the string representation is what we expect, however in this case we really have a binary output, so we default to the Python `repr`.

vlaci (approved):

martonilles (approved):

Issue #355, opened 2022-04-26
CPIO portable ASCII end offset miscalculation.


QKaiser:
The CPIO portable ASCII handler identified a chunk, but the end offset is way past the actual end, which leads to the non-discovery of DTB at offset 0x208.

The chunk in question:

```
00000000  30 37 30 37 30 31 30 30  30 30 30 32 44 31 30 30  |070701000002D100|
00000010  30 30 34 31 45 44 30 30  30 30 30 30 30 30 30 30  |0041ED0000000000|
00000020  30 30 30 30 30 30 30 30  30 30 30 30 30 32 36 31  |0000000000000261|
00000030  30 33 30 36 38 34 30 30  30 30 30 30 30 30 30 30  |0306840000000000|
00000040  30 30 30 30 30 33 30 30  30 30 30 30 30 31 30 30  |0000030000000100|
00000050  30 30 30 30 30 30 30 30  30 30 30 30 30 30 30 30  |0000000000000000|
00000060  30 30 30 30 30 34 30 30  30 30 30 30 30 30 64 65  |00000400000000de|
00000070  76 00 00 00 30 37 30 37  30 31 30 30 30 30 30 32  |v...070701000002|
00000080  44 32 30 30 30 30 32 31  38 30 30 30 30 30 30 30  |D200002180000000|
00000090  30 30 30 30 30 30 30 30  30 30 30 30 30 30 30 30  |0000000000000000|
000000a0  30 31 36 31 30 33 30 36  38 34 30 30 30 30 30 30  |0161030684000000|
000000b0  30 30 30 30 30 30 30 30  30 33 30 30 30 30 30 30  |0000000003000000|
000000c0  30 31 30 30 30 30 30 30  30 35 30 30 30 30 30 30  |0100000005000000|
000000d0  30 31 30 30 30 30 30 30  30 43 30 30 30 30 30 30  |010000000C000000|
000000e0  30 30 64 65 76 2f 63 6f  6e 73 6f 6c 65 00 00 00  |00dev/console...|
000000f0  30 37 30 37 30 31 30 30  30 30 30 32 44 33 30 30  |070701000002D300|
00000100  30 30 34 31 43 30 30 30  30 30 30 30 30 30 30 30  |0041C00000000000|
00000110  30 30 30 30 30 30 30 30  30 30 30 30 30 32 36 31  |0000000000000261|
[...]

PR #361, opened 2022-05-04
Fix CPIO file padding calculation.

QKaiser:
We previously made the mistake of expecting every CPIO chunk to have a file padding, but we identified CPIO chunks in our corpus without this file padding.

We added a _file_pad class function that will read potential file padding bytes. If these are all null bytes, that means we have padding, otherwise we don't.

The observed sample has been added to integration tests.

Related to #355


unblob/handlers/archive/cpio.py
```diff
@@ -100,15 +100,22 @@ def calculate_chunk(self, file: File, start_offset: int) -> Optional[ValidChunk]
...
             start_offset=start_offset,
-            end_offset=offset,
+            end_offset=end_offset,
         )
 
+    @classmethod
+    def _pad_file(cls, file: File, end_offset: int) -> int:
+        """CPIO archives can have a 512 bytes block padding at the end."""
+        file.seek(end_offset, io.SEEK_SET)
+        padded_end_offset = round_up(end_offset, cls._FILE_PAD_ALIGN)
```
e3krisztian, line 113: I think this works for most of the inputs, as the input is most probably aligned to block boundaries.  However when the `start_offset` is not aligned to block boundary, all of the block based calculations need to take this into account, otherwise the end result would be aligned to block boundary of the input file, not to block boundary when the chunk is extracted.

Take a look at `OffsetFile` introduced in 462aa6c by @martonilles (beware, that there is a confusing amount of highly related file-like objects in `calculate_chunk` there: `file`, `offset_file`, `ar.file`).

QKaiser: I followed your advice and used `OffsetFile`. I also removed unnecessary piece of code from the handler that were there due to our mis-handling of unaligned content. I also added an integration test file that starts at offset 1 to verify that we can handle alignment properly.

QKaiser: Fixed the end offset calculation as discussed.

e3krisztian (approved):

vlaci:
Please, in the future squash fixup! commits before merging (maybe we could come up with a check for that).

PR #363, opened 2022-05-05
Complete rewrite of XZ handler.

### LINKED ISSUE #362: Stop emitting a ValidChunk for every xz stream
Very similar to https://github.com/IoT-Inspector/unblob/issues/316

We should return a single ValidChunk holding all the concatenated XZ streams rather than one chunk per stream. Otherwise it leads to situation such as this:

```
./73479421-74088645.xz_extract
./74088645-74843029.xz_extract
./74843029-74891825.xz_extract
./74891825-75341173.xz_extract
./75341173-75532341.xz_extract
./75532341-75878861.xz_extract
./75878861-76788105.xz_extract
./76788105-77570353.xz_extract
./77570353-78222665.xz_extract
./78222665-79330565.xz_extract
```

QKaiser:
In order to emit a single ValidChunk for concatenated XZ compression streams, we converted the handler to use Hyperscan like we did with bzip2 handler (see c97007735f985ecf3fbd0563f1102485dc2a5846).

The idea is that for a given match, we use Hyperscan to look for end of stream markers after our current start offset. If we have a match, we attempt to decompress our candidate chunk using unblob's file iterator (iterate_file) to limit memory consumption. If decompression is successful, we adjust start and end offsets of our context and instruct Hyperscan to continue looking for other end of stream marker matches. Otherwise, we stop the search and return.

Closes #362


unblob/handlers/compression/xz.py
```diff
@@ -1,110 +1,102 @@
...
+) -> bool:
 
-            index_size = 0
-            index_indicator = convert_int8(file.read(1), Endian.LITTLE)
-            # index indicator must be 0, per xz standard
-            if index_indicator != 0:
-                continue
+    end_offset = offset + 4
+    if end_offset < context.start_offset:
+        return True
```
martonilles, line 60: why stop matching if match is before the start_offset, shouldn't we continue matching and just ignore this?

unblob/handlers/compression/xz.py
```diff
@@ -1,110 +1,102 @@
...
 
-            index_size += 1
+    context.file.seek(context.start_offset, io.SEEK_SET)
+    decompressor = lzma.LZMADecompressor(format=lzma.FORMAT_XZ)
```
martonilles, line 63: why do we want to actually decompress the content, wouldn't it be easier just to verify that the size is correct from the Index field?

unblob/handlers/compression/xz.py
```diff
@@ -1,110 +1,102 @@
...
+    try:
+        size = end_offset - context.start_offset
+        for data in iterate_file(context.file, context.start_offset, size):
+            decompressor.decompress(data)
+        context.end_block_offset = end_offset
+        context.start_offset = context.file.tell()
```
martonilles, line 69: if we already matched on this footer, it is unlikely we would match something before, so no need to set start_offset

unblob/handlers/compression/xz.py
```diff
@@ -1,110 +1,102 @@
...
+        return False
```
martonilles, line 70: before continue matching shouldn't we check if there is a new stream starting after this stream footer (potentially after a stream padding?)

unblob/handlers/compression/xz.py
```diff
@@ -1,110 +1,102 @@
...
-MAGIC_BYTES_SIZE = 6
-BACKWARD_SIZE_LEN = 4
-CRC32_LEN = 4
-FLAG_LEN = 2
+STREAM_END_MAGIC_PATTERNS = [
+    Regex(r"\x00\x00YZ"),  # None
+    Regex(r"\x00\x01YZ"),  # CRC32
+    Regex(r"\x00\x04YZ"),  # CRC64
+    Regex(r"\x00\x0AYZ"),  # SHA-256
+]
```
martonilles, line 24: probably not, but was wondering if we should only match on flags that were present in the stream header? Or maybe at least check if the flag is the same on a match?

martonilles (changes requested):
Maybe also add a test file for a multi-stream file?

unblob/handlers/compression/xz.py
QKaiser: Fixed.

unblob/handlers/compression/xz.py
QKaiser: The code now checks if a new stream starts after the matched stream footer before continuing. Padding alignment between two subsequent streams is also validated.

unblob/handlers/compression/xz.py
QKaiser: I kept the same search rule so we don't generate a hyperscan db on every XZ magic match. I do however validate that concatenated streams have the same flags, tested by `test_xz_calculate_chunk_unique_stream` and `test_xz_calculate_chunk_unique_stream_padding`.

unblob/handlers/compression/xz.py
QKaiser: `context.start_offset` is used when validating that the calculated size of the XZ stream corresponds to the obtained end_offset - `context.start_offset`. If we don't update the context start offset, this calculation will be off.

Example: when we match on `XZ_END_2` without setting `context.start_offset` to `XS_START_2` on the first match on `XZ_END_1`, `calculated_size != XZ_END_2 - XZ_START_1`. If we do update `context.start_offset`, then `calculated_size = XZ_END_2 - XZ_START_2`.

````
XZ_START_1
XZ_END_1
XZ_START_2
XZ_END_2
```

martonilles (changes requested):
I would also add a few unit tests at least to cover the stream continuation and padding logic
Also an integration test example for a multi-stream file would be nice

unblob/handlers/compression/xz.py
```diff
@@ -40,71 +88,138 @@ def read_multibyte_int(file: File) -> Tuple[int, int]:
...
+        index_size += size
+
+        blocks_size += round_up(unpadded_size, XZ_PADDING)
+
+    index_size += CRC32_LEN
+
+    overall_size = round_up(
+        (STREAM_HEADER_LEN + blocks_size + index_size + STREAM_FOOTER_LEN),
+        XZ_PADDING,
+    )
```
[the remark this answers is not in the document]
QKaiser: Moved into `get_stream_size` function.

unblob/handlers/compression/xz.py
```diff
@@ -40,71 +88,138 @@ def read_multibyte_int(file: File) -> Tuple[int, int]:
...
+        return -1
+    next_stream_flag = convert_int16(flag, Endian.BIG)
+    logger.error(
+        "_validate_block_header",
+        magic=magic,
+        stream_magic=STREAM_START_MAGIC,
+        stream_flag=stream_flag,
+        next_stream_flag=next_stream_flag,
+    )
+    return magic == STREAM_START_MAGIC and stream_flag == next_stream_flag
```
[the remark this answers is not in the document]
QKaiser: So I looked into it and unxz does not care about the flags, so we should not check for next stream flags here, or flags correspondence for that matter.

unblob/handlers/compression/xz.py
```diff
@@ -40,71 +88,138 @@ def read_multibyte_int(file: File) -> Tuple[int, int]:
     return size, mbi
 
 
+def _validate_block_header(file: File, stream_flag: int):
```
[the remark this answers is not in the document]
QKaiser: Fixed wording.

unblob/handlers/compression/xz.py
```diff
@@ -40,71 +88,138 @@ def read_multibyte_int(file: File) -> Tuple[int, int]:
...
+    end_offset = offset + FLAG_LEN + EOS_MAGIC_LEN
+    if end_offset < context.start_offset:
+        return False
+
+    context.file.seek(offset, io.SEEK_SET)
+
+    # get the flag and check that it matches the context flag
+    stream_flag = convert_int16(context.file.read(2), Endian.BIG)
+    if stream_flag != context.stream_flag:
+        return True
```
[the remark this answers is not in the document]
QKaiser: That is an excellent question. I think we should continue scanning to be on the safe side. Fixed.

unblob/handlers/compression/xz.py
```diff
@@ -40,71 +88,138 @@ def read_multibyte_int(file: File) -> Tuple[int, int]:
...
+            return ValidChunk(
+                start_offset=start_offset, end_offset=context.end_block_offset
             )
 
-            # if the identified chunk size is equal to the size calculated from
-            # the Index's records, that means we matched on the right footer and
-            # we can return
-            if (end_offset - start_offset) == overall_size:
-                return ValidChunk(start_offset=start_offset, end_offset=end_offset)
+        raise InvalidInputFormat("No valid xz compression block was detected")
```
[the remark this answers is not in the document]
QKaiser: Fixed wording.

unblob/handlers/compression/xz.py
```diff
@@ -1,34 +1,82 @@
...
+
+
+hyperscan_stream_end_magic_db = build_stream_end_scan_db(STREAM_END_MAGIC_PATTERNS)
+
+
+@attr.define
+class XZSearchContext:
+    start_offset: int
+    file: File
+    end_block_offset: int
```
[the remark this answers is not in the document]
QKaiser: Fixed wording.

unblob/handlers/compression/xz.py
```diff
@@ -40,71 +88,138 @@ def read_multibyte_int(file: File) -> Tuple[int, int]:
...
+            context.start_offset = end_padding_offset
+            return False
+    return True
+
+
 class XZHandler(Handler):
     NAME = "xz"
 
-    PATTERNS = [HexString("FD 37 7A 58 5A 00")]
+    PATTERNS = [Regex(r"\xFD\x37\x7A\x58\x5A\x00")]
```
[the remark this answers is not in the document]
QKaiser: Fixed.

QKaiser:
@martonilles fixed the code following your remarks. I had forgotten to add the test file to my commit :) Let me know what you think. I'll squash the commits when approved.

tests/handlers/compression/test_xz.py
```diff
@@ -0,0 +1,704 @@
...
+        ),
+        pytest.param(
+            XZ_SHA256_CONTENT,
+            0,
+            SHA256_STREAM_FLAG,
+            id="stream_flag_crc_sha256",
+        ),
+    ),
+)
+def test_xz_stream_flag(content: bytes, start_offset: int, stream_flag: int):
```
martonilles: start_offset is always 0, do we need to test other values or do we need to specify it here?

maybe add negative tests as well?

tests/handlers/compression/test_xz.py
```diff
@@ -0,0 +1,704 @@
...
+        ),
+        pytest.param(
+            XZ_SHA256_CONTENT + NULL_PADDING + XZ_SHA256_CONTENT + NULL_PADDING,
+            0,
+            len(XZ_SHA256_CONTENT + NULL_PADDING + XZ_SHA256_CONTENT + NULL_PADDING),
+            id="sha256_concat_pad2",
+        ),
+    ),
+)
+def test_xz_calculate_chunk(content: bytes, start_offset: int, expected_length: int):
```
martonilles, line 277: - start_offset is always 0
- expected length is always the length of the content

do we want to specify those or test anything that have other values?

tests/handlers/compression/test_xz.py
```diff
@@ -0,0 +1,704 @@
...
+
+    for first_chunk, second_chunk in itertools.combinations(formats, 2):
+        handler = XZHandler()
+        fake_file = File.from_bytes(first_chunk + second_chunk)
+        chunk = handler.calculate_chunk(fake_file, 0)
+        assert chunk is not None
+        assert chunk.end_offset == len(first_chunk)
+
+
+def test_xz_calculate_chunk_unique_stream_padding():
```
martonilles, line 633: `test_xz_calculate_chunk_unique_stream` is the same use parameterize

martonilles (changes requested):

tests/handlers/compression/test_xz.py
QKaiser: I don't know what happened, there was a mismatch between my local version and this one. Anyways, this test was removed because we no longer verify stream flags.

tests/handlers/compression/test_xz.py
QKaiser: Again, mismatch. The start offset is indeed always 0 but the expected length can differ because I have test cases with bad padding.

tests/handlers/compression/test_xz.py
QKaiser: Fixed together.

tests/handlers/compression/test_xz.py
```diff
@@ -0,0 +1,704 @@
...
+        ),
+        pytest.param(
+            XZ_SHA256_CONTENT + NULL_PADDING + XZ_SHA256_CONTENT + NULL_PADDING,
+            0,
+            len(XZ_SHA256_CONTENT + NULL_PADDING + XZ_SHA256_CONTENT + NULL_PADDING),
+            id="sha256_concat_pad2",
+        ),
+    ),
+)
+def test_xz_calculate_chunk_prefix_suffix(
```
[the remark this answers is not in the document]
QKaiser: Fixed together.

tests/handlers/compression/test_xz.py
```diff
@@ -0,0 +1,704 @@
...
+        ),
+        pytest.param(
+            XZ_SHA256_CONTENT + NULL_PADDING + XZ_SHA256_CONTENT + NULL_PADDING,
+            0,
+            len(XZ_SHA256_CONTENT + NULL_PADDING + XZ_SHA256_CONTENT + NULL_PADDING),
+            id="sha256_concat_pad2",
+        ),
+    ),
+)
+def test_xz_calculate_chunk_suffix(
```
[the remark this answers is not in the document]
QKaiser: Fixed together.

tests/handlers/compression/test_xz.py
```diff
@@ -0,0 +1,704 @@
...
+        ),
+        pytest.param(
+            XZ_SHA256_CONTENT + NULL_PADDING + XZ_SHA256_CONTENT + NULL_PADDING,
+            0,
+            len(XZ_SHA256_CONTENT + NULL_PADDING + XZ_SHA256_CONTENT + NULL_PADDING),
+            id="sha256_concat_pad2",
+        ),
+    ),
+)
+def test_xz_calculate_chunk_prefix(
```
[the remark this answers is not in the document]
QKaiser: Fixed together.

tests/handlers/compression/test_xz.py
QKaiser: Fixed together.

tests/handlers/compression/test_xz.py
QKaiser: Fixed together.

QKaiser:
@martonilles no obvious timing differences in pytest runtime (compared with previously merged branches).

martonilles (approved):

PR #364, opened 2022-05-05
Fix handling of LZ4 legacy frames.

### LINKED ISSUE #360: lz4 compressed linux kernel chunk detection fails
Some old linux kernels (ARM 4.X) are using lz4 legacy compression. See https://manpages.debian.org/testing/lz4/unlz4.1.en.html (-l option) for details. However there is no end magic after the compressed content, but could be detected as except the last block all blocks are 8MB uncompressed.

The vmlinx-to-elf project (https://github.com/marin-m/vmlinux-to-elf/issues/27) had a similar bug and has a work-around as well:
https://github.com/marin-m/vmlinux-to-elf/blob/master/vmlinux_to_elf/utils/lz4_legacy.py

QKaiser:
There is no end magic after the compressed content, but we can detect it given that except for the last block, all blocks are 8MB uncompressed.

Inspired by vmlinux-to-elf (see
https://github.com/marin-m/vmlinux-to-elf/blob/master/vmlinux_to_elf/utils/lz4_legacy.py).

Resolves #360

### DIFF
--- a/unblob/handlers/compression/lz4.py
+++ b/unblob/handlers/compression/lz4.py
@@ -4,6 +4,7 @@ LZ4 frame format definition: https://github.com/lz4/lz4/blob/dev/doc/lz4_Frame_f
 import io
 from typing import Optional
 
+from lz4.block import decompress
 from structlog import get_logger
 
 from unblob.extractors import Command
@@ -28,6 +29,7 @@ BLOCK_SIZE_LEN = (
     FRAME_SIZE_LEN
 ) = BLOCK_CHECKSUM_LEN = CONTENT_CHECKSUM_LEN = MAGIC_LEN = DICTID_LEN = 4
 FLG_LEN = BD_LEN = HC_LEN = 1
+MAX_LEGACY_BLOCK_SIZE = 8 * 1024 * 1024  # 8 MB
 
 
 class FLG:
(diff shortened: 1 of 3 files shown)


unblob/handlers/compression/lz4.py
```diff
@@ -89,7 +91,11 @@ def calculate_chunk(self, file: File, start_offset: int) -> Optional[ValidChunk]
                 file.seek(-4, io.SEEK_CUR)
                 break
 
-            file.seek(block_compressed_size, io.SEEK_CUR)
+            compressed_block = file.read(block_compressed_size)
+            uncompressed_block = decompress(compressed_block, MAX_LEGACY_BLOCK_SIZE)
+
+            if len(uncompressed_block) < MAX_LEGACY_BLOCK_SIZE:
+                break
```
martonilles, line 98: I would just add a comment with a reference on this change

martonilles:
Please also squash the fixup patch into the original one

unblob/handlers/compression/lz4.py
QKaiser: Done :)

QKaiser:
> Please also squash the fixup patch into the original one

Done !

martonilles (approved):

PR #365, opened 2022-05-06
Move AR archive extractor from 7zip to unar.

QKaiser:
7-zip made the conscious decision of hiding information from debian archives (.deb, ar format).

See:
- https://sourceforge.net/p/sevenzip/bugs/1472/
- https://sourceforge.net/p/sevenzip/bugs/1130/

In order to be complete in our extraction process, we moved to unar to extract ar archive as it won't be hiding files and extract the whole archive.

Integration tests have been adjusted to reflect new output names due to our move to unar.


vlaci:
Wouldn't we need this switch?
> -no-directory (-D)                 Never create a containing directory for the contents of the unpacked archive.

QKaiser:
> Wouldn't we need this switch?
> 
> > -no-directory (-D)                 Never create a containing directory for the contents of the unpacked archive.

Definitely makes things simpler in the output tree :) Added.

I'll open another merge request where I use this switch for rar, stuffit, and arc.

vlaci (approved):

PR #370, opened 2022-05-06
Elf handler

martonilles:


martonilles:
> cmake is required by the build process of lief, nix package has to be adapted.

Working on that... ;)

vlaci (approved):

QKaiser:
🚀

PR #371, opened 2022-05-09
elf: fixed kernel initramfs padding calculation

martonilles:


vlaci:
Could you figure out at the end what those 4 bytes are?

martonilles:
> Could you figure out at the end what those 4 bytes are?

It is actually not extra bytes, but an 8 byte alignment with padding of the size variable where the initramfs precedes the size, but that is placed only on a 4 byte alignment. Depending on where the initramfs is placed our padding calculation was broken.

QKaiser (approved):
Took me a while to figure everything out, looks good to me based on what I see in the linker script.

PR #375, opened 2022-05-13
jffs2: check header CRC to eliminate false-positve matches

martonilles:


QKaiser (approved):

tests/handlers/filesystem/test_jffs2.py
```diff
@@ -57,6 +59,10 @@ def get_valid_jffs2_old_be_header():
     )
 
 
+def calculate_crc(header: Instance):
```
kissgyorgy, line 62: I would make this the other way around: put this into the jffs2 handler, and call it from this test instead.

unblob/handlers/filesystem/jffs2.py
```diff
@@ -64,6 +65,15 @@ def guess_endian(self, file: File) -> Endian:
         return endian
 
     def valid_header(self, header: Instance, node_start_offset: int, eof: int) -> bool:
+        header_crc = (binascii.crc32(header.dumps()[:-4], -1) ^ -1) & 0xFFFFFFFF
```
kissgyorgy, line 68: This is very complex and duplicated, see my comment above.

tests/handlers/filesystem/test_jffs2.py
```diff
@@ -57,6 +59,10 @@ def get_valid_jffs2_old_be_header():
...
+    return (binascii.crc32(header.dumps()[:-4], -1) ^ -1) & 0xFFFFFFFF
```
kissgyorgy, line 63: Why we are dropping the last 4 bytes? Naming that number would be more readable.

PR #376, opened 2022-05-16
file-magic related override moved to upstream

vlaci:
https://github.com/nix-community/poetry2nix/pull/636


vlaci:
Ehh, I am not sure it's worth the effort, as it needs two override steps...

PR #380, opened 2022-05-23
elf: fix kernel initramfs extraction padding calculation

martonilles:
As the initramfs starts on a 4 byte offset and the size is put on an 8byte
offset it is not too well defined on where the start offset is, as we can
exactly calculate it from just the size and the size offset.

#define INIT_RAM_FS                                                     \
        . = ALIGN(4);                                                   \
        __initramfs_start = .;                                          \
        KEEP(*(.init.ramfs))                                            \
        . = ALIGN(8);                                                   \
        KEEP(*(.init.ramfs.info))

To calculate the padding we try to look at the max padding size where the
padding is the alignment before the ALIGN(8), which should be 0 padded.

Unfortunatelly sometime the init.ramfs itself has some zero padding, so
we use the calculated padding +4 bytes for the cases when there is an extra
4 bytes but use maximum padding up till 0 bytes are detected. This could
still result in misscalculation, but only if the 4 byte alignments overlap
with 0bytes at the end of valid initramfs.

### DIFF
--- a/unblob/handlers/executable/elf.py
+++ b/unblob/handlers/executable/elf.py
@@ -73,7 +73,14 @@ class ELFKernelExtractor(Extractor):
                     endian=endian,
                 )
 
-            padding = round_up(initramfs_size + 4, 8) - initramfs_size - 4
+            max_padding = 0
+            while (
+                file[initramfs_size_offset - max_padding - 1] == 0 and max_padding < 8
+            ):
+                max_padding += 1
+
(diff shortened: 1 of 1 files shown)


QKaiser:
> This could still result in misscalculation, but only if the 4 byte alignments overlap with 0bytes at the end of valid initramfs.

How frequent is it ? Did you observe samples leading to miscalculation ? Any way we can fully get rid of miscalculations ?

QKaiser:
Let's keep it this way for now and we can get back to symbol parsing when we have larger coverage of kernel archs.

PR #390, opened 2022-06-07
elf: consider cases when program segments are located at the end of the file

### LINKED ISSUE #388: ELF chunks are miscalculated on OpenWRT
Elf files from [this firmware](https://downloads.openwrt.org/releases/21.02.3/targets/mvebu/cortexa9/openwrt-21.02.3-mvebu-cortexa9-linksys_wrt3200acm-squashfs-sysupgrade.bin) gets sliced up to very small elf and the remainding unknown chunks.

e.g:

```
$ ls usr/sbin/dropbear_extract
0-276.elf32  0-276.elf32_extract  276-176611.unknown
```

martonilles:
Program segments can be placed at the end of the file, up till now we did not
calculate the program segment end hence ELF files where program segments
were placed at the end of the file were miss calculated. This could happen
when there are no sections in the ELF file, which is unusual though possible.

closes #388

### DIFF
--- a/tests/integration/executable/elf/elf32/__input__/sample_32_prg_end.elf
+++ b/tests/integration/executable/elf/elf32/__input__/sample_32_prg_end.elf
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:5755d4f41982bdaae3edb565813ffe193db1d0576d77d4170b3b0082d04dda37
+size 36917
--- a/unblob/handlers/executable/elf.py
+++ b/unblob/handlers/executable/elf.py
@@ -95,6 +95,7 @@ class _ELFBase(StructHandler):
 
     EXTRACTOR = ELFKernelExtractor()
     SECTION_HEADER_STRUCT = "elf_shdr_t"
+    PROGRAM_HEADER_STRUCT = "elf_phdr_t"
 
     @staticmethod
     def _check_field(field, value):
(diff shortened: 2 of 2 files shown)


vlaci:
It seams to work well but I have some observations:
- Some logs escape to stderr from [lief](https://github.com/lief-project/LIEF/blob/f8c711dd940ba2c1dbf8cb803e7df2d19bec7853/src/ELF/Parser.tcc#L70). Having a bunch of these lines: `The current binary doesn't have a section header`
- Because ELFKernelExtractor is set for ELF handlers, an empty extract directory always gets created, even for whole-file elf binaries

e3krisztian:
I can confirm, that the other firmware linked do not have the 1-3 bytes unknown with this change.

martonilles:
> It seams to work well but I have some observations:
> 
> * Some logs escape to stderr from [lief](https://github.com/lief-project/LIEF/blob/f8c711dd940ba2c1dbf8cb803e7df2d19bec7853/src/ELF/Parser.tcc#L70). Having a bunch of these lines: `The current binary doesn't have a section header`
Yes, that is unfortunatelly hardcoded in LIEF, we can file a bug report, but not sure we should/could do much more on this.

> * Because ELFKernelExtractor is set for ELF handlers, an empty extract directory always gets created, even for whole-file elf binaries

Yes, that is true for any handler which matches a whole file. I would open a ticket as that would require significant rework on how extraction directories are created.

QKaiser (approved):
I'm ok with these changes since @vlaci took care of creating tickets for LIEF and extractor directories issues.

PR #393, opened 2022-06-08
unblob: renaming leftover references to IoT-Inspector

vlaci:
One functional breakage beside references to the company name right
now is the container registry which makes publishing images break.

### DIFF
--- a/.github/workflows/build-publish-image.yml
+++ b/.github/workflows/build-publish-image.yml
@@ -9,7 +9,7 @@ on:
       - '**'
 
 env:
-  DOCKER_IMAGE: ghcr.io/iot-inspector/unblob:latest
+  DOCKER_IMAGE: ghcr.io/onekey-sec/unblob:latest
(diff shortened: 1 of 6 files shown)


.github/workflows/build-publish-image.yml
```diff
@@ -9,7 +9,7 @@ on:
       - '**'
 
 env:
-  DOCKER_IMAGE: ghcr.io/iot-inspector/unblob:latest
+  DOCKER_IMAGE: ghcr.io/onekey-sec/unblob:latest
```
QKaiser, line 12: Is it ready to be merged ? Does this registry exist ?

vlaci: Yes, it is the registry we inherited through renaming the organization https://github.com/onekey-sec/unblob/pkgs/container/unblob

QKaiser (approved):

PR #395, opened 2022-06-09
Add ZSTD compression format

QKaiser:
Support is provided by the zstandard library recommended by ZSTD format maintainers (see https://python-zstandard.readthedocs.io/en/latest/decompressor.html).
    
The implementation is similar to the one we got for gzip, except we did not have to monkeypatch it this time.
    
More information about ZSTD can be found at https://github.com/facebook/zstd/blob/dev/doc/zstd_compression_format.md

### DIFF
--- a/unblob/handlers/compression/zstd.py
+++ b/unblob/handlers/compression/zstd.py
@@ -1,0 +1,84 @@
+import io
+from typing import Optional
+
+from structlog import get_logger
+
+from unblob.extractors import Command
+
+from ...file_utils import Endian, InvalidInputFormat, convert_int8
+from ...models import File, Handler, HexString, ValidChunk
+
+logger = get_logger()
+
+MAGIC_LEN = 4
+BLOCK_HEADER_LEN = 3
+RAW_BLOCK = 0
+RLE_BLOCK = 1
+COMPRESSED_BLOCK = 2
+DICT_ID_FIELDSIZE_MAP = [0, 1, 2, 4]
+FRAME_CONTENT_FIELDSIZE_MAP = [0, 2, 4, 8]
+
+
+class ZSTDHandler(Handler):
+    NAME = "zstd"
+
+    PATTERNS = [HexString("28 B5 2F FD")]
+
+    EXTRACTOR = Command("zstd", "-d", "{inpath}", "-o", "{outdir}/{infile}")
+
+    def get_frame_header_size(self, frame_header_descriptor: int) -> int:
+        single_segment = (frame_header_descriptor >> 5 & 1) & 0b1
+        dictionary_id = frame_header_descriptor >> 0 & 0b11
+        frame_content_size = (frame_header_descriptor >> 6) & 0b1
+        return (
+            int(not single_segment)
+            + DICT_ID_FIELDSIZE_MAP[dictionary_id]
+            + FRAME_CONTENT_FIELDSIZE_MAP[frame_content_size]
+            + (single_segment and not frame_content_size)
+        )
+
+    def calculate_chunk(self, file: File, start_offset: int) -> Optional[ValidChunk]:
+
+        file.seek(start_offset, io.SEEK_SET)
+        file.seek(MAGIC_LEN, io.SEEK_CUR)
+
+        frame_header_descriptor = convert_int8(file.read(1), Endian.LITTLE)
+        frame_header_size = self.get_frame_header_size(frame_header_descriptor)
+
+        content_checksum_flag = frame_header_descriptor >> 2 & 1
+        if content_checksum_flag:
+            content_checksum_size = 4
+        else:
+            content_checksum_size = 0
+
+        unused_bit = frame_header_descriptor >> 4 & 1
+        reserved_bit = frame_header_descriptor >> 3 & 1
+
+        # these values MUST be zero per the standard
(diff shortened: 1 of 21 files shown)


unblob/handlers/compression/zstd.py
```diff
@@ -0,0 +1,43 @@
...
+import zstandard
+from structlog import get_logger
+
+from unblob.extractors import Command
+
+from ...file_utils import InvalidInputFormat
+from ...models import File, Handler, HexString, ValidChunk
+
+logger = get_logger()
+MIN_READ_RATIO = 0.1
```
martonilles, line 14: we do not need this

unblob/handlers/compression/zstd.py
```diff
@@ -0,0 +1,43 @@
...
+
+    EXTRACTOR = Command("zstd", "-d", "{inpath}", "-o", "{outdir}/{infile}")
+
+    def calculate_chunk(self, file: File, start_offset: int) -> Optional[ValidChunk]:
+
+        file.seek(start_offset, io.SEEK_SET)
+        decompressor = zstandard.ZstdDecompressor()
+        chunks = 0
+        try:
+            for _ in decompressor.read_to_iter(
```
martonilles, line 30: why do we want to decompress the content to check the size?

we can just parse the frame header and iterate over the blocks, where the block size is clearly defined also the last block is signaled.

unblob/handlers/compression/zstd.py
```diff
@@ -0,0 +1,43 @@
...
+
+    PATTERNS = [HexString("28 B5 2F FD")]
+
+    EXTRACTOR = Command("zstd", "-d", "{inpath}", "-o", "{outdir}/{infile}")
+
+    def calculate_chunk(self, file: File, start_offset: int) -> Optional[ValidChunk]:
+
+        file.seek(start_offset, io.SEEK_SET)
+        decompressor = zstandard.ZstdDecompressor()
+        chunks = 0
```
martonilles, line 28: we would be able to verify the frame header to make sure we do not pick up false-positive matches based on the magic.

martonilles (changes requested):
I would consider not decompressing the content in the handler, rather just iterate over the block, which should be significantly faster.

unblob/handlers/compression/zstd.py
QKaiser: Rewrote the whole thing as discussed.

unblob/handlers/compression/zstd.py
QKaiser: Rewrote the whole thing as discussed.

unblob/handlers/compression/zstd.py
```diff
@@ -0,0 +1,84 @@
...
+
+logger = get_logger()
+
+MAGIC_LEN = 4
+BLOCK_HEADER_LEN = 3
+RAW_BLOCK = 0
+RLE_BLOCK = 1
+COMPRESSED_BLOCK = 2
+dict_id_fieldsize_map = [0, 1, 2, 4]
+frame_content_fieldsize_map = [0, 2, 4, 8]
```
martonilles, line 19: constants should be uppercase

unblob/handlers/compression/zstd.py
```diff
@@ -0,0 +1,84 @@
...
+    NAME = "zstd"
+
+    PATTERNS = [HexString("28 B5 2F FD")]
+
+    EXTRACTOR = Command("zstd", "-d", "{inpath}", "-o", "{outdir}/{infile}")
+
+    def get_frame_header_size(self, frame_header_descriptor: int) -> int:
+        single_segment = (frame_header_descriptor >> 5 & 1) & 0xFF_FF_FF_FF
+        dictionary_id = frame_header_descriptor >> 0 & 0b11
+        frame_content_size = (frame_header_descriptor >> 6) & 0xFF_FF_FF_FF
```
martonilles, line 32: it is only 2 bits, why the 0xffffffff mask?

unblob/handlers/compression/zstd.py
```diff
@@ -0,0 +1,84 @@
...
+
+class ZSTDHandler(Handler):
+    NAME = "zstd"
+
+    PATTERNS = [HexString("28 B5 2F FD")]
+
+    EXTRACTOR = Command("zstd", "-d", "{inpath}", "-o", "{outdir}/{infile}")
+
+    def get_frame_header_size(self, frame_header_descriptor: int) -> int:
+        single_segment = (frame_header_descriptor >> 5 & 1) & 0xFF_FF_FF_FF
```
martonilles, line 30: 1 bit value, should be 1bit mask not 0xffffffff

martonilles (changes requested):

unblob/handlers/compression/zstd.py
QKaiser: Fixed.

unblob/handlers/compression/zstd.py
QKaiser: Fixed.

unblob/handlers/compression/zstd.py
QKaiser: Fixed.

martonilles (approved):

PR #396, opened 2022-06-09
Explicitly set LIEF logging level.

### LINKED ISSUE #391: Logs from LIEF escape to stderr
While testing #390 I've noticed that sdome logs escape to stderr from [lief](https://github.com/lief-project/LIEF/blob/f8c711dd940ba2c1dbf8cb803e7df2d19bec7853/src/ELF/Parser.tcc#L70). Having a bunch of these lines: `The current binary doesn't have a section header`

It looks like, there is an ability to [control logging from python](https://lief-project.github.io/doc/latest/api/python/index.html#logging), that we need to tinker with.

QKaiser:
We don't want informational logs from LIEF polluting unblob logs, but we do want to be informed when something really wrong happened so instead of completely disabling LIEF logging, we set the level to CRITICAL.

Fixes #391

### DIFF
--- a/unblob/handlers/executable/elf.py
+++ b/unblob/handlers/executable/elf.py
@@ -16,6 +16,8 @@ from unblob.file_utils import (
 )
 from unblob.models import ExtractError, Extractor, HexString, StructHandler, ValidChunk
 
+lief.logging.set_level(lief.logging.LOGGING_LEVEL.ERROR)
+
 
 class NullExtract(ExtractError):
     pass


vlaci:
I don't see any logs defined with critical severity in lief so this change effectively disables logging too.

QKaiser:
Set to ERROR level and edited the commit message accordingly. Double checked that log pollution does not happen with ERROR level.

vlaci:
If we later find that we'd need these logs in the python logger, it doesn't seem that hard to improve LIEF's Python binding to swap the logger to a wrapper to Python's logging library.

PR #397, opened 2022-06-09
Improve cramfs CRC check

### LINKED ISSUE #394: Disable cramfs CRC check
Although it's a nice way to get rid of false positives, vendors (*cough* Cisco) don't care about CRCs and ship cramfs with broken checksums.

The signature check we do at https://github.com/onekey-sec/unblob/blob/main/unblob/handlers/filesystem/cramfs.py#L46 should be sufficient for false positive reduction.

QKaiser:
We observed a sample from Cisco where the CRC value was stored in a different endianness. This is non standard but we decided to support it anyway.
    
We also added a check for old cramfs format. If it's an old format, the CRC check always returns true because it's not supported.
    
More information about CRC validation and old format: https://github.com/secularbird/cramfs/blob/master/cramfsck.c

Closes #394

### DIFF
--- a/unblob/handlers/filesystem/cramfs.py
+++ b/unblob/handlers/filesystem/cramfs.py
@@ -1,4 +1,5 @@
 import binascii
+import struct
 from typing import Optional
 
 from dissect.cstruct import Instance
@@ -8,7 +9,12 @@ from unblob.extractors import Command
 from ...file_utils import Endian, convert_int32, get_endian
 from ...models import File, HexString, StructHandler, ValidChunk
 
+CRAMFS_FLAG_FSID_VERSION_2 = 0x00000001
 BIG_ENDIAN_MAGIC = 0x28_CD_3D_45
+
+
+def swap_int32(i):
+    return struct.unpack("<I", struct.pack(">I", i))[0]
 
 
 class CramFSHandler(StructHandler):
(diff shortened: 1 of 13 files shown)


martonilles:
can we somehow identify those firmware and just disable in those cases?

QKaiser:
> can we somehow identify those firmware and just disable in those cases?

We actually need to keep the crc check, it was an endianness issue: `computed_crc=0x3af2f56e header_crc=0x6ef5f23a`

martonilles (changes requested):
would it be possible to add cases for this in the integration tests?
- swapped crc
- omitted crc

unblob/handlers/filesystem/cramfs.py
```diff
@@ -8,9 +9,14 @@
 from ...file_utils import Endian, convert_int32, get_endian
 from ...models import File, HexString, StructHandler, ValidChunk
 
+CRAMFS_FLAG_FSID_VERSION_2 = 0x00000001
 BIG_ENDIAN_MAGIC = 0x28_CD_3D_45
 
 
+def swap32(i):
```
vlaci, line 16: `swap_int32` may be a name that better aligns with the other stuff we have in file_utils

QKaiser: sounds good, done.

QKaiser:
> would it be possible to add cases for this in the integration tests?
> 
> * swapped crc
> * omitted crc

Created test cases for both situations.

martonilles (approved):

PR #399, opened 2022-06-17
Change from file-magic to python-magic, use MAGIC_CONTINUE

e3krisztian:
MAGIC_CONTINUE is more transparently supported with python-magic.
Also file-magic's last release was in 2018,
while there are recent changes on github.

### DIFF
--- a/unblob/report.py
+++ b/unblob/report.py
@@ -140,6 +140,18 @@ class StatReport(Report):
         )
 
 
+# libmagic helpers
+# file magic uses a rule-set to guess the file type, however as rules are added they could
+# shadow each other. File magic uses rule priorities to determine which is the best matching
+# rule, however this could shadow other valid matches as well, which could eventually break
+# any further processing that depends on magic.
+# By enabling keep_going (which eventually enables MAGIC_CONTINUE) all matching patterns
+# will be included in the magic string at the cost of being a bit slower, but increasing
+# accuracy by no shadowing rules.
+get_magic = magic.Magic(keep_going=True).from_file
+get_mime_type = magic.Magic(mime=True).from_file
+
+
 @attr.define(kw_only=True)
 class FileMagicReport(Report):
     magic: str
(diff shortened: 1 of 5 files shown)


unblob/report.py
```diff
@@ -147,8 +147,9 @@ class FileMagicReport(Report):
 
     @classmethod
     def from_path(cls, path: Path):
-        detected = magic.detect_from_filename(path)
-        return cls(magic=detected.name, mime_type=detected.mime_type)
+        mime_type = magic.Magic(mime=True).from_file(path)
+        full = magic.Magic(keep_going=True).from_file(path)
```
martonilles, line 151: I would not initialize Magic every time we need to lookup as that is very expensive.

I would rather create 2 Magic and just reuse it. Even the library has a cache to do so.

for mime based one without keep_going we can use the builtin `from_file`

e3krisztian: At first the code was using `magic.from_file`, but was changed as you saw it, because it is more explicit.

Will look into caching `Magic()` objects.

e3krisztian:
In Nix pulling in `file-magic` pulled also `libmagic`, somehow it does not happen with `python-magic` :(

vlaci:
> In Nix pulling in `file-magic` pulled also `libmagic`, somehow it does not happen with `python-magic` :(

We did hard work to achieve that https://github.com/onekey-sec/unblob/blob/main/default.nix#L66

martonilles:
I would extend the commit message explaining why we need neeed `MAGIC_CONTINUE` or `keep_going`:

file magic uses a rule-set to guess the file type, however as rules are added they could shadow each other. File magic uses rule priorities to determine which is the best matching rule, however this could shadow other valid matches as well, which could eventually break any further processing that depends on magic.
By enabling keep_going (which eventually enables MAGIC_CONTINUE) all matching patterns will be included in the magic string at the cost of being a bit slower, but increasing accuracy by no shadowing rules.

martonilles (approved):

PR #401, opened 2022-06-22
Config option for custom magic file

kissgyorgy:
unblob creates metadata about files, which includes magic and mime type
using libmagic. When using unlob as a library, the user might want to
control which exact magic file to use for fixed results for these 2 fields.

python-magic supports this with the magic_file init argument.


kissgyorgy:
Never mind, this is the simplest solution I can come up with and will work for us. We generate the compiled magic file during build.

e3krisztian (approved):

PR #408, opened 2022-07-07
Nix improvements

vlaci:
- A `nix-shell` will create an editable environment from which all the imports are properly resolved inside the source tree. Previously it required a `PYTHONPATH=.` magic.
- I'd like to have `mkPoetryApp` helper extracted some time. This PR contains some work to make it little more general and not specific to unblob.
- Also extending `mkPoetryApp` to be able to override its attributes from downstream Nix derivations, e.g. to use a different Python version.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -96,8 +96,6 @@ let
       rust.cargo
       rust.rustc
     ];
-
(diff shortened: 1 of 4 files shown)


kissgyorgy (approved):
We reviewed it together

Issue #422, opened 2022-08-17
Format Support - ZLIB compression


QKaiser:
We want to add support for ZLIB compression format.

Documentation: https://www.rfcreader.com/#rfc1950

Samples can be generated with `zlib-flate -compress < /tmp/hello.txt > /tmp/compressed.txt`.

PR #423, opened 2022-08-17
Add zlib format support

QKaiser:
See #422

### DIFF
--- a/docs/formats.md
+++ b/docs/formats.md
@@ -55,19 +55,19 @@ unblob supports more than 30 formats. You can see their code in
 
 For compression formats, metadata cannot be preserved, as this information in most cases is not stored in the format.
 
-| Format        | Handler                                     | Extractor                       |
-| ------------- | ------------------------------------------- | ------------------------------- |
-| bzip2         | [compression/bzip2.py][bzip2-handler]       | [`7z`][bzip2-extractor]         |
-| UNIX compress | [compression/compress.py][compress-handler] | [`7z`][compress-extractor]      |
-| gzip          | [compression/gzip.py][gzip-handler]         | [`7z`][gzip-extractor]          |
-| LZ4           | [compression/lz4.py][lz4-handler]           | [`lz4`][lz4-extractor]          |
-| LZH           | [compression/lzh.py][lzh-handler]           | [`7z`][lzh-extractor]           |
-| LZIP          | [compression/lzip.py][lzip-handler]         | [`lziprecover`][lzip-extractor] |
-| LZMA          | [compression/lzma.py][lzma-handler]         | [`7z`][lzma-extractor]          |
-| LZO           | [compression/lzo.py][lzo-handler]           | [`lzop`][lzo-extractor]         |
-| XZ            | [compression/xz.py][xz-handler]             | [`7z`][xz-extractor]            |
-| ZLIB          | [compression/zlib.py][zlib-handler]         | [`ZlibExtractor` custom code][zlib-extractor]        |
-| ZSTD          | [compression/zstd.py][zstd-handler]         | [`zstd`][zstd-extractor]        |
+| Format        | Handler                                     | Extractor                                     |
+| ------------- | ------------------------------------------- | --------------------------------------------- |
+| bzip2         | [compression/bzip2.py][bzip2-handler]       | [`7z`][bzip2-extractor]                       |
+| UNIX compress | [compression/compress.py][compress-handler] | [`7z`][compress-extractor]                    |
+| gzip          | [compression/gzip.py][gzip-handler]         | [`7z`][gzip-extractor]                        |
+| LZ4           | [compression/lz4.py][lz4-handler]           | [`lz4`][lz4-extractor]                        |
+| LZH           | [compression/lzh.py][lzh-handler]           | [`7z`][lzh-extractor]                         |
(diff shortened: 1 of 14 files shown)


QKaiser:
False positive zlib content is identified in squashfs integration tests samples. I need to find a way for false positive reduction.

vlaci:
Yeah, these 2 byte patterns are well, not very specific :)

vlaci:
> > Yeah, these 2 byte patterns are well, not very specific :)
> 
> It's weird because it's a true positive (metadata in squashfs being compressed with zlib), but I don't understand why the squashfsv3 handler does not have priority over zlib in that the zlib chunk is within a larger squashfs chunk.

You can check what happens, when you disable other handlers. It is possible that the squashfs chunk is discarded for some reason, for example because the file is truncated.

unblob/handlers/compression/zlib.py
```diff
@@ -0,0 +1,48 @@
...
+        with File.from_path(inpath) as f:
+            while not decompressor.eof:
+                outpath.write_bytes(decompressor.decompress(f.read(DEFAULT_BUFSIZE)))
+
+
+class ZlibHandler(Handler):
+    NAME = "zlib"
+
+    PATTERNS = [
+        HexString("78 9c"),  # default compression
```
kissgyorgy, line 26: What about 78 01?

QKaiser: nice catch, fixed.

vlaci:
I've been thinking that would it make sense for the time being to detect zlib only at the begining of files? Any legitimate file can contain a bunch zlib compressed or zlib like data streams which may or may not make sense to extract and detect. Even if we wouldn't have the current issue it colliding with squashfs.

QKaiser:
This would not be sufficient for some samples that we have where the firmware is a zlib stream in the middle of another file.

Another way would be to check that the zlib chunk is bigger than some arbitrary value and emit an InvalidInput error otherwise.

QKaiser:
The integration test failure was due to a bug in our squashfs ddwrt handler. I fixed it in https://github.com/onekey-sec/unblob/pull/426. As for zlib streams in multi-media files we can use the default skip list. 

I ran this branch against a bunch of firmwares, the only file type we should add to the skip list is .woff font files.

kissgyorgy (changes requested):
Also please update the documentation too!

QKaiser:
Fixed the docs, but still had a problem with DMG archives. Given that they can be made of multiple zlib streams, without a custom header but a custom footer. The fix I wrote checks if the file contains a DMG magic and returns early. This may be generalized in the future if similar cases arise.

docs/formats.md
```diff
@@ -66,6 +66,7 @@ For compression formats, metadata cannot be preserved, as this information in mo
 | LZMA          | [compression/lzma.py][lzma-handler]         | [`7z`][lzma-extractor]          |
 | LZO           | [compression/lzo.py][lzo-handler]           | [`lzop`][lzo-extractor]         |
 | XZ            | [compression/xz.py][xz-handler]             | [`7z`][xz-extractor]            |
+| ZLIB          | [compression/zlib.py][zlib-handler]         | [`ZlibExtractor` custom code]zlib-extractor]        |
```
kissgyorgy, line 69: There is a missing `[` so it's not rendered correctly.  
You can spot these kinds of errors by using `mkdocs serve` and looking at the generated documentation.

unblob/handlers/compression/zlib.py
```diff
@@ -0,0 +1,59 @@
...
+        HexString("78 5e"),  # compressed
+    ]
+
+    EXTRACTOR = ZlibExtractor()
+
+    def calculate_chunk(self, file: File, start_offset: int) -> Optional[ValidChunk]:
+
+        eof = file.size()
+        for pattern in DMGHandler.PATTERNS:
+            if re.search(pattern.as_regex(), file[eof - 512 : eof]):  # noqa: E203
```
kissgyorgy, line 41: You can just say "the last 512 bytes" without using the size of the file.
```suggestion
            if re.search(pattern.as_regex(), file[-512:]):
```

docs/formats.md
```diff
@@ -86,6 +87,8 @@ For compression formats, metadata cannot be preserved, as this information in mo
 [lzo-extractor]: https://github.com/onekey-sec/unblob/blob/3008039881a0434deb75962e7999b7e35aca8271/unblob/handlers/compression/lzo.py#L78
 [xz-handler]: https://github.com/onekey-sec/unblob/blob/main/unblob/handlers/compression/xz.py
 [xz-extractor]: https://github.com/onekey-sec/unblob/blob/3008039881a0434deb75962e7999b7e35aca8271/unblob/handlers/compression/xz.py#L173
+[zlib-handler]: https://github.com/onekey-sec/unblob/blob/main/unblob/handlers/compression/zlib.py
```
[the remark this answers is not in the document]
QKaiser: Done.

unblob/handlers/compression/zlib.py
QKaiser: 👍

docs/formats.md
QKaiser: good to know ! fixed.

kissgyorgy (approved):

PR #426, opened 2022-08-20
Fix squashfs v3 ddwrt big endian handler

QKaiser:
The big endian pattern was actually little-endian. Fixed the order and integration tests accordingly.

Can't believe it went through in the first place😿


unblob/handlers/filesystem/squashfs.py
```diff
@@ -118,7 +118,7 @@ class SquashFSv3DDWRTHandler(SquashFSv3Handler):
             // 00000000  74 71 73 68 21 02 00 00  00 00 00 00 00 00 00 00  |tqsh!...........|
             // 00000010  00 00 00 00 00 00 00 00  50 02 00 03 00 00 00 00  |........P.......|
```
kissgyorgy, line 119: The example also doesn't add up, there is only 23 bytes before 00 03, not 24

QKaiser: That's right. I fixed the example by running hexdump on actual integration test files. Thanks.

kissgyorgy (approved):

PR #428, opened 2022-08-23
feat(handler/cpio): add support for truncated CPIO archives

QKaiser:
We recently analyzed a sample containing a truncated CPIO archive, which led unblob to discard it.

Given that 7z is capable of extracting truncated CPIO archives, we now handle the EOFError that happens when unblob tries to parse the next CPIO header on a truncated file and cleanly break out of the parsing loop.

I have integration tests files locally, but due to 7z returning a non-zero status code on truncated CPIO archive extraction our integration test suite fails due to `AssertionError: Unexpected error reports`. I'll include them if we can find an elegant solution, otherwise we keep the CPIO integration tests samples this way.

### DIFF
--- a/unblob/testing.py
+++ b/unblob/testing.py
@@ -9,6 +9,7 @@ from unblob.finder import build_hyperscan_database
 from unblob.logging import configure_logger
 from unblob.models import ProcessResult
 from unblob.processing import ExtractionConfig
+from unblob.report import ExtractCommandFailedReport
 
 
 @pytest.fixture(scope="session", autouse=True)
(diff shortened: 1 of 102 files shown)


e3krisztian (approved):

unblob/testing.py
```diff
@@ -80,5 +81,11 @@ def check_output_is_the_same(reference_dir: Path, extract_dir: Path):
...
 def check_result(reports: ProcessResult):
     __tracebackhide__ = True
-
-    assert reports.errors == [], "Unexpected error reports"
+    # filter out error reports about truncated integration test files
+    errors = [
+        error
+        for error in reports.errors
+        if not isinstance(error, ExtractCommandFailedReport)
+        or error.stderr != b"\nERRORS:\nUnexpected end of archive\n\n"
```
kissgyorgy, line 89: Did you meant to skip every instance of `ExtractCommandFailedReport` regardless of the error message? Maybe I misunderstood something, but didn't we agree only ignoring this one very specific message and nothing else?

```suggestion
        if not (
            isinstance(error, ExtractCommandFailedReport)
            and error.stderr == b"\nERRORS:\nUnexpected end of archive\n\n"
        )
```
I'm fine skipping all of these, just wanted to clarify if it was the intention.

QKaiser: Indeed. Fixed.

e3krisztian: FYI, `(not A) or (not B)` is the same as `not (A and B)`, i.e. it was not a fix: nothing changed.

PR #430, opened 2022-08-31
Ignore special files

### LINKED ISSUE #429: Ignore special files
unblob is crashing on pipes, so we should ignore those (we are already ignoring symlinks, zero-sized files) and other special files (device files, etc) and look for other possible errors.

QKaiser:
Fix #429

A change has been applied so that we only process files and directories.

A naming convention has been defined for special files:
- `name.socket` for socket files
- `name.fifo` for fifo files
- `name.symlink` for symink files

This can be used by integration test files so that they can contain special files without triggering false alarms on diff. Diff would report errors like `File /A/fifo0 is a fifo while file /B/fifo0 is a fifo`, which is useless.

Thanks to this change, I could add a squashfs filesystem sample that contains the following special files:
    
- block device: not extracted because we're not running as root
- character device: not extracted because we're not running as root
- fifo device: extracted
- symlink: extracted
- socket: extracted
    
The fifo, symlink, and socket files are not processed by unblob, otherwise it would hang (e.g. by waiting for data on the fifo)., demonstrating that we have a proper fix.
    
Note that these special files cannot be tracked by git, so they're not present in the output directory.

### DIFF
--- a/tests/integration/filesystem/squashfs/squashfs_v4_le/__input__/squashfs_v4.specfile.bin
+++ b/tests/integration/filesystem/squashfs/squashfs_v4_le/__input__/squashfs_v4.specfile.bin
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:d27f2e4baf57df961b9aa7298ac390a54fd0d2c904bf1d4baaee49cbdd0a93f1
+size 4096
--- a/tests/integration/filesystem/squashfs/squashfs_v4_le/__output__/squashfs_v4.specfile.bin_extract/file.symlink
+++ b/tests/integration/filesystem/squashfs/squashfs_v4_le/__output__/squashfs_v4.specfile.bin_extract/file.symlink
@@ -1,0 +1,1 @@
+hello.txt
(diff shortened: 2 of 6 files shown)


kissgyorgy (approved):
Neat!

QKaiser:
Ok lots of things do not make sense in this build.

**poetry.lock inconsistency**

Fails on GH (https://github.com/onekey-sec/unblob/runs/8113030821?check_suite_focus=true#step:4:10) but works locally:

```
poetry lock --no-update
git diff --exit-code -- poetry.lock
echo $?
0
```

Which probably leads to the poetry install to fail, and therefore the build too.

vlaci:
> Ok lots of things do not make sense in this build.
> 
> **poetry.lock inconsistency**
> 
> Fails on GH (https://github.com/onekey-sec/unblob/runs/8113030821?check_suite_focus=true#step:4:10) but works locally:
> 
> ```
> poetry lock --no-update
> git diff --exit-code -- poetry.lock
> echo $?
> 0
> ```
> 
> Which probably leads to the poetry install to fail, and therefore the build too.

I had the same issue in #425. I've given up figuring it out, just committed the change with the adjusted diff from the output of the checker. @kissgyorgy ran into a similar issue on his machine: when we both issued a `poetry lock`, we got the same lock file but its contents were ordered differently. I am okay giving up this check for the time being, as it causes more harm than good.

vlaci:
This seems to work for me:
```diff
diff --git a/pyproject.toml b/pyproject.toml
index 70c5455..1ad1d66 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -66,5 +66,5 @@ exclude = ["unblob/_py/", "unblob/testing.py"]
 exclude = ["build"]
 
 [build-system]
-requires = ["poetry-core>=1.0.0", "setuptools>=60.9.3"]
+requires = ["poetry-core>=1.0.0", "setuptools>=60.9.3", "setuptools-rust==1.5.1"]
 build-backend = "poetry.core.masonry.api"
```

PR #432, opened 2022-09-01
Add hash report

kukovecz:
The aim of this PR is to extend the reporting with a new report type, called  `HashReport`.

HashReport is a new Report type, which is reporting hashes (md5, sha1, sha256 at the moment, but it is easily extendible) from a non-special, non-empty file.

### DIFF
--- a/tests/test_report.py
+++ b/tests/test_report.py
@@ -13,6 +13,7 @@ from unblob.report import (
     ChunkReport,
     ExtractCommandFailedReport,
     FileMagicReport,
+    HashReport,
     StatReport,
     UnknownChunkReport,
 )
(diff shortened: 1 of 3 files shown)


kissgyorgy (approved):
Looks very good!
I would make a small comment about why Wrapping the filename in `ZipInfo` solves the problem; because of `date_time=(1980, 1, 1, 0, 0, 0)`.

kukovecz:
> Looks very good! I would make a small comment about why Wrapping the filename in `ZipInfo` solves the problem; because of `date_time=(1980, 1, 1, 0, 0, 0)`.

I have created a separate commit for it, I hope that is enough rather than commenting ti to every use. Maybe I should modify the commit message to make it even clearer?

PR #433, opened 2022-09-05
feat(processing): Add `HashReport` for empty files

kukovecz:
`HashReport` was introduced in eb2f74f1046675cb3191e65e7a0aa56bd903dfc1.
There we tried to optimize it to skip calculations for empty files, but
honestly that is not a big overhead. This small patch to make the
code and the usage of reports cleaner, so we will have a `HashReport`
for every file, like `FileMagicReport`.


kissgyorgy (approved):

kissgyorgy:
It's more consistent with file magic and mime types too.

PR #436, opened 2022-09-15
fix(handler/tar): replace 7zip with python tarfile module.

### LINKED ISSUE #435: symlinks in tar files are not extracted to symlinks
7zip apparently just creates a file with the symlink target, which is inaccurate:
- https://sourceforge.net/p/sevenzip/bugs/1302/
- https://sourceforge.net/p/p7zip/bugs/137/
- https://sourceforge.net/p/sevenzip/bugs/1188/

QKaiser:
Given that 7zip just creates a text file with the symlink target path in it, which is inaccurate, we had to switch to another extractor.

The decision was made to move to Python3's built-in tarfile module given that's what we already use to calculate the chunk offsets.

tarfile module, like the tar and unar cli is unsafe when it comes to symlink based traversal. This is no longer a problem since 97230c2 given that unblob fixes link based traversal once the extraction has been performed.

Fixes #435

### DIFF
--- a/unblob/handlers/archive/tar.py
+++ b/unblob/handlers/archive/tar.py
@@ -102,9 +102,7 @@ class TarHandler(StructHandler):
     """
     HEADER_STRUCT = "posix_header_t"
 
-    EXTRACTOR = Command(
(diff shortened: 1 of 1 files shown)


vlaci (approved):
Let's see how it behaves on the long run.

PR #440, opened 2022-09-21
fix(handlers): improve ZIP handler to support ZIP64.

### LINKED ISSUE #439: Add support for ZIP64 extension in ZIP handler
We should add support for ZIP64 archive.

See https://pkware.cachefly.net/webdocs/APPNOTE/APPNOTE-6.3.1.TXT

QKaiser:
Documentation about ZIP64:
https://pkware.cachefly.net/webdocs/APPNOTE/APPNOTE-6.3.1.TXT

zip64 sample was generated with `zip -fz apple.zip *txt`, the `-fz` switch force ZIP64 usage.

Resolve #439

### DIFF
--- a/unblob/handlers/archive/zip.py
+++ b/unblob/handlers/archive/zip.py
@@ -13,6 +13,8 @@ logger = get_logger()
 
 ENCRYPTED_FLAG = 0b0001
 EOCD_RECORD_HEADER = 0x6054B50
+ZIP64_EOCD_SIGNATURE = 0x06064B50
+ZIP64_EOCD_LOCATOR_HEADER = 0x07064B50
 
 
 class ZIPHandler(StructHandler):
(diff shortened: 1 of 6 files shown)


unblob/handlers/archive/zip.py
```diff
@@ -83,15 +112,53 @@ def calculate_chunk(self, file: File, start_offset: int) -> Optional[ValidChunk]
...
+                        if zip64_eocd.signature != ZIP64_EOCD_SIGNATURE:
+                            raise InvalidInputFormat(
+                                "Missing ZIP64 EOCD header record header in ZIP chunk."
+                            )
+                        end_of_central_directory = zip64_eocd
+                        found_eocd = True
+                        break
+
+                if found_eocd:
+                    break
```
kukovecz, line 147: Just a small thing:

This whole zip64 finding and parsing could be in a new, separate function, making the core logic simpler, like:

```suggestion
            if self.is_zip64_eocd(end_of_central_directory):
                end_of_central_directory = self._parse_zip64(file, start_offset, offset)
                break
```

kukovecz (approved):

unblob/handlers/archive/zip.py
QKaiser: Good point. Done.

PR #444, opened 2022-09-27
fix(squashfs): fix handling of SquashFS v4 big-endian image built by AVM

### LINKED ISSUE #443: wrong extraction command for AVM SquashFS v4 big-endian
When faced with a SquashFS v4 big-endian filesystem, unblob will attempt to run the following command:

```sh
sasquatch-v4be -be -no-exit-code -f -d ${outdir} ${inpath}
```

This triggers an exception:

```
Reading a different endian SQUASHFS filesystem on ${inpath}
read_id_table: failed to read id table block
FATAL ERROR: File system corruption detected
```

This is due to the presence of the '-be' operator, as it works without it:

```
sasquatch-v4be -no-exit-code -d ${outdir} -f ${inpath}
SquashFS version [1024.0] / inode count [-2111373312] suggests a SquashFS image of a different endianess
Trying to decompress using default xz decompressor...
Successfully decompressed with default xz decompressor
Parallel unsquashfs: Using 1 processor
9586 inodes (10853 blocks) to write
--snip--
created 8896 files
created 528 directories
created 689 symlinks
created 0 devices
created 0 fifos
created 0 sockets
```

QKaiser:
AVM has a very interesting way of interpreting standards so we have to use a different extractor.

We identify whether an chunk is from AVM by taking advantage of a tweak they do, documented at https://raw.githubusercontent.com/Freetz/freetz/master/tools/make/squashfs4-host-be/AVM-BE-format.txt:

> super-block field 'mkfs_time' contains (instead of the filesystem creation time) the same value as the super-block field 'bytes_used'

Depending on the AVM-ness, we use a different Command extractor to do the job.

Resolves #443

### DIFF
--- a/unblob/handlers/filesystem/squashfs.py
+++ b/unblob/handlers/filesystem/squashfs.py
@@ -1,11 +1,13 @@
-from typing import Optional
+from pathlib import Path
+from typing import List, Optional
 
+from dissect.cstruct import Instance
 from structlog import get_logger
 
 from unblob.extractors import Command
 
-from ...file_utils import get_endian, round_up
-from ...models import File, HexString, StructHandler, ValidChunk
+from ...file_utils import Endian, StructParser, get_endian, round_up
+from ...models import Extractor, File, HexString, StructHandler, ValidChunk
 
 logger = get_logger()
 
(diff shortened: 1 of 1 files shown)


unblob/handlers/filesystem/squashfs.py
```diff
@@ -216,6 +218,40 @@ class SquashFSv4LEHandler(_SquashFSBase):
     HEADER_STRUCT = "squashfs4_super_block_t"
 
 
+class SquashFSv4BEExtractor(Extractor):
```
kukovecz, line 221: When we define a new `Extractor`, we should define the `get_dependencies` function if that `Extractor` is depending on a specific command / tool. The `Command` extractor used before returns with the `command` by default.

You can check that `sasquatch-v4be` disappeared from this branch from the output of `unblob --show-external-dependencies`.

You can do something like this:


```suggestion
class SquashFSv4BEExtractor(Extractor):

    EXECUTIBLE = "sasquatch-v4be"

    def get_dependencies(self) -> List[str]:
        return [self.EXECUTIBLE]
     
```

unblob/handlers/filesystem/squashfs.py
```diff
@@ -216,6 +218,40 @@ class SquashFSv4LEHandler(_SquashFSBase):
...
+            else:
+                extractor = Command(
+                    "sasquatch-v4be",
+                    "-be",
+                    "-no-exit-code",
+                    "-f",
+                    "-d",
+                    "{outdir}",
+                    "{inpath}",
+                )
```
kukovecz, line 251: I would suggest a small refactor here. In this way we can be sure that they are always called with the same arguments:

Note, this is just optional, your solution is also correct.

```suggestion
        commands_args = []

        if not self.is_avm(header):
            commands_args.append("-be")

        commands_args.extend(
            [
                "-no-exit-code",
                "-f",
                "-d",
                "{outdir}",
                "{inpath}",
            ]
        )

        extractor =  Command(self.EXECUTIBLE, *commands_args)
```

kukovecz (commented):
Aside from small technical details, it looks very promising!

unblob/handlers/filesystem/squashfs.py
```diff
@@ -216,6 +218,40 @@ class SquashFSv4LEHandler(_SquashFSBase):
...
+        # see https://raw.githubusercontent.com/Freetz/freetz/master/tools/make/squashfs4-host-be/AVM-BE-format.txt
+        return header.bytes_used == header.mkfs_time
+
+    def extract(self, inpath: Path, outdir: Path):
+        struct_parser = StructParser(SquashFSv4LEHandler.C_DEFINITIONS)
+
+        with File.from_path(inpath) as f:
+            header = struct_parser.parse(
+                SquashFSv4LEHandler.HEADER_STRUCT, f, Endian.BIG
+            )
```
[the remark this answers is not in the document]
QKaiser: Fixed.

unblob/handlers/filesystem/squashfs.py
QKaiser: Good catch ! Fixed.

unblob/handlers/filesystem/squashfs.py
QKaiser: I prefer your design :) Fixed.

kukovecz (approved):

PR #449, opened 2022-09-28
Stop carving ELF from signed kernel modules

QKaiser:
Consider the PKCS7 signature as part of the ELF file when encountering signed kernel modules.

Unblob previously split signed kernel modules in two separate components: the ELF file, and the PKCS7 signature appended at the end.

While this behavior is perfectly right from a purely analytic standpoint, it causes problems as unblob creates a twisted view of what is actually on a file system by splitting each and every signed kernel module into their own extraction directory.

We decided to consider the PKCS7 signature as part of the ELF file. To do so, whenever a relocatable ELF file is observed we:
- verify that it is followed by an ASN1 sequence
- parse the ASN1 sequence length
- set the end offset to the ELF end offset + ASN1 sequence length + kernel module signature end marker (`~Module signature appended~`) length

Integration tests file built from a kernel module hello world sample and signed using every possible hash (SHA1, SHA256, SHA384, SHA512) were also added to validate our changes.

### DIFF
--- a/unblob/handlers/executable/elf.py
+++ b/unblob/handlers/executable/elf.py
@@ -4,6 +4,7 @@ from typing import Optional
 
 import lief
 from dissect.cstruct import Instance
+from structlog import get_logger
 
 from unblob.extractor import carve_chunk_to_file
 from unblob.file_utils import (
@@ -12,11 +13,17 @@ from unblob.file_utils import (
     convert_int8,
     convert_int32,
     convert_int64,
+    iterate_patterns,
     round_up,
 )
 from unblob.models import ExtractError, Extractor, HexString, StructHandler, ValidChunk
 
 lief.logging.set_level(lief.logging.LOGGING_LEVEL.ERROR)
+
+logger = get_logger()
+
+KERNEL_MODULE_SIGNATURE_INFO_LEN = 12
+KERNEL_MODULE_SIGNATURE_FOOTER = b"~Module signature appended~\n"
 
 
 class NullExtract(ExtractError):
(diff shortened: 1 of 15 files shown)


unblob/handlers/executable/elf.py
```diff
@@ -18,6 +19,49 @@
 
 lief.logging.set_level(lief.logging.LOGGING_LEVEL.ERROR)
 
+logger = get_logger()
+
+# length is encoded in at most 2 bytes starting after the 2nd byte - 30 82 ?? ??
+ASN1_PREFIX_LEN = 4
+ASN1_SEQUENCE_START = b"\x30"
+KERNEL_MODULE_SIGNATURE_PADDING_LEN = 12
```
martonilles, line 27: this would be actually the signature info :+1: 
```
struct module_signature {
        uint8_t         algo;           /* Public-key crypto algorithm [0] */
        uint8_t         hash;           /* Digest algorithm [0] */
        uint8_t         id_type;        /* Key identifier type [PKEY_ID_PKCS7] */
        uint8_t         signer_len;     /* Length of signer's name [0] */
        uint8_t         key_id_len;     /* Length of key identifier [0] */
        uint8_t         __pad[3];
        uint32_t        sig_len;        /* Length of signature data */
};
```

I would maybe call this `KERNEL_MODULE_SIGNATURE_INFO_LEN`

unblob/handlers/executable/elf.py
```diff
@@ -190,13 +234,39 @@ def get_end_offset(
...
         file.seek(start_offset, io.SEEK_SET)
         header = self.parse_header(file, endian)
         if not self.is_valid_header(header):
             return
         end_offset = self.get_end_offset(file, start_offset, header, endian)
+
+        if self.is_signed_kernel(file, header, end_offset):
+            length_bytes = file[end_offset : end_offset + ASN1_PREFIX_LEN]  # noqa: E203
+            signature_length = get_asn1_sequence_length(length_bytes)
+            if signature_length:
```
martonilles, line 254: do we want to do any signature sanity check? It is not very likely, but if the kernel module has `0x30` byte after it's end we most likely would pick it up as a signature.

We would be able to check the signature.sig_len value which should contain the signature len.

Maybe a better alternative would be to search for the `KERNEL_MODULE_SIGNATURE_FOOTER` after the end of the ELF file and check the signature info to see if the size would fit. This way we could avoid the asn1 length parsing as well if we want

martonilles (changes requested):

unblob/handlers/executable/elf.py
QKaiser: This was my initial idea and it makes even more sense knowing about `module_signature`.

We could:

1. search for `KERNEL_MODULE_SIGNATURE_FOOTER`
2. if found, seek to the start of `module_signature`
3. parse `module_signature` and check that `signature.sig_len` corresponds to the end offset of our signature footer minus the end offset of current ELF file.

This way we can drop the ASN1 sequence parsing.

QKaiser: Just pushed my changes, let me know what you think.

unblob/handlers/executable/elf.py
```diff
@@ -18,6 +19,49 @@
...
+
+# length is encoded in at most 2 bytes starting after the 2nd byte - 30 82 ?? ??
+ASN1_PREFIX_LEN = 4
+ASN1_SEQUENCE_START = b"\x30"
+KERNEL_MODULE_SIGNATURE_PADDING_LEN = 12
+KERNEL_MODULE_SIGNATURE_FOOTER_LEN = 28
+KERNEL_MODULE_SIGNATURE_FOOTER = b"~Module signature appended~\n"
+
+
+def get_asn1_sequence_length(asn_sequence_beginning: bytes) -> Optional[int]:
```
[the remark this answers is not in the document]
QKaiser: Removed ASN1 parsing in the latest revision.

unblob/handlers/executable/elf.py
QKaiser: Added `module_signature` parsing in the latest revision.

unblob/handlers/executable/elf.py
martonilles: Added a few test cases for non-happy paths and simplified the code to use `iterate_patterns`

martonilles (approved):

PR #455, opened 2022-10-04
fix(elf/kernel/initramfs) extraction when initrd for kernel is disabled

e3krisztian:


unblob/handlers/executable/elf.py
```diff
@@ -93,11 +93,21 @@ def extract_initramfs(self, elf, inpath, outdir):
...
-                outdir.joinpath("initramfs"),
-                file,
-                ValidChunk(start_offset=initramfs_start, end_offset=initramfs_end),
-            )
+            # initramfs can be turned off (https://www.linux.com/training-tutorials/kernel-newbie-corner-initrd-and-initramfs-whats/)
+            # in which case the above calculations most probably end up with bogus chunk offsets
+            if (
+                init_data.file_offset
+                <= initramfs_start
+                <= initramfs_end
```
QKaiser, line 101: With this condition, `initramfs_start` and `initramfs_end` can be equal, which would lead to an `InvalidInputFormat` being triggered.

QKaiser (changes requested):

QKaiser:
Other than the small condition change, I really like this fix. It would be great if we could have integration test files for `initrd`, but I'm not sure it's worth the trouble compiling kernels for different architectures and moving them to LFS. 

It's a big investment in terms of time and space that we may want to do later on.

What do you think @martonilles ?

martonilles:
> Other than the small condition change, I really like this fix. It would be great if we could have integration test files for `initrd`, but I'm not sure it's worth the trouble compiling kernels for different architectures and moving them to LFS.
> 
> It's a big investment in terms of time and space that we may want to do later on.
> 
> What do you think @martonilles ?

That would be awsome to have kernels with/without initramfs and different architecture and packaging. Though I would just open a ticket for that and not block this ticket.

unblob/handlers/executable/elf.py
e3krisztian: Modified to `<`, so we are slightly more strict about it.

e3krisztian:
I wanted to add test files as well, however could not make one trivially.
I have tried to compile and test one without initrd (`make tinyconfig; make`) for x64, but that one did not cause the problem we faced, so it might indeed be arch specific/highly dependent on the data at the end of the `.init.data` section.

QKaiser (approved):

Issue #459, opened 2022-10-15
python tarfile module is vulnerable to path traversal


QKaiser:
Yet another episode in our quest to find a proper TAR extractor that does not get in the f* way.

I had a quick chat with someone at Hexacon who mentioned that tarfile module is vulnerable to path traversal (not only through symlinks), did a quick search and ... [yeah](https://davidhamann.de/2022/09/23/python-tarfile-vulnerability/).

I already have a fix that use a similar trick to what we did with the gzip handler.

PR #460, opened 2022-10-15
Resolve path traversal and unhandled permission error in tar handler

### LINKED ISSUE #456: tar extractor returns non-zero exit code when attempting to create device files
Python3 builtin tar extractor does not pay attention to the current privilege level and will try to create device files even if we don't have elevated privileges:

```
2022-10-03 07:44.28 [error    ] 2022-10-03 07:44.28 [error    ] Extract command failed         command=python3 -m tarfile -e /results/extraction/firmware_extract/40-392835312.bzip2_extract/40-392835312_extract/system_2.0.0.0.tbz2_extract/system_2.0.0.0.tar /results/extraction/firmware_extract/40-392835312.bzip2_extract/40-392835312_extract/system_2.0.0.0.tbz2_extract/system_2.0.0.0.tar_extract exit_code=0x1 pid=15 severity=<Severity.WARNING: 'WARNING'> stderr=Traceback (most recent call last):
  File "/usr/local/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/local/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/usr/local/lib/python3.10/tarfile.py", line 2614, in <module>
    main()
  File "/usr/local/lib/python3.10/tarfile.py", line 2576, in main
    tf.extractall(path=curdir)
  File "/usr/local/lib/python3.10/tarfile.py", line 2059, in extractall
    self.extract(tarinfo, path, set_attrs=not tarinfo.isdir(),
  File "/usr/local/lib/python3.10/tarfile.py", line 2100, in extract
    self._extract_member(tarinfo, os.path.join(path, tarinfo.name),
  File "/usr/local/lib/python3.10/tarfile.py", line 2179, in _extract_member
    self.makedev(tarinfo, targetpath)
  File "/usr/local/lib/python3.10/tarfile.py", line 2252, in makedev
    os.mknod(targetpath, mode,
PermissionError: [Errno 1] Operation not permitted
```

I recommend we keep the same module but call it from unblob while catching `PermissionError` exceptions. This way we don't have this verbose log that could make users think something went wrong.

I don't have the motivation to report this to tarfile maintainers.

QKaiser:
Fixed path traversal in python's builtin tarfile module and added a privilege check when tarfile attempts to create device files.

This fixes #456 and #459

Brought to you by the excellent upstream connection on Thalys high speed train.

More info about traversal: https://davidhamann.de/2022/09/23/python-tarfile-vulnerability/

### DIFF
--- a/unblob/handlers/archive/_safe_tarfile.py
+++ b/unblob/handlers/archive/_safe_tarfile.py
@@ -1,0 +1,29 @@
+import os
+from pathlib import Path
+from tarfile import TarFile
+
+from structlog import get_logger
+
+from unblob.extractor import is_safe_path
+
+logger = get_logger()
+
+RUNNING_AS_ROOT = os.getuid() == 0
+
+
+class SafeTarFile(TarFile):
+    def extract(self, member, path="", set_attrs=True, *, numeric_owner=False):
+        path_as_path = Path(str(path))
+        member_name_path = Path(str(member.name))
+
+        if not RUNNING_AS_ROOT and (member.ischr() or member.isblk()):
+            logger.warn(
+                "missing elevated permissions, skipping block and character device creation",
+                path=member_name_path,
+            )
+            return
+        if not is_safe_path(path_as_path, member_name_path):
+            logger.warn("traversal attempt", path=member_name_path)
+            return
+
+        super().extract(member, path, set_attrs, numeric_owner=numeric_owner)
--- a/tests/integration/archive/tar/__input__/traversal.tar
+++ b/tests/integration/archive/tar/__input__/traversal.tar
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:0c0558a0d64c93f706175b7a38158a2909c07bf174ed54572e9f31896ddb20a6
+size 260
--- a/tests/integration/archive/tar/__output__/traversal.tar_extract/traversal
+++ b/tests/integration/archive/tar/__output__/traversal.tar_extract/traversal
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:9d4ab5759508af8d62e4686b236f896210a4145a146ab399797993dcddcc9b22
+size 10240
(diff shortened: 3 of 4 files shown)


QKaiser:
Note: modifications to `extract` and `_extract_member` are preceded by a `MOD:` comment. The code was copied in place otherwise, and a the code complexity check had to be disabled.

unblob/handlers/archive/_safe_tarfile.py
```diff
@@ -0,0 +1,101 @@
...
+
+from unblob.extractor import is_safe_path
+
+logger = get_logger()
+
+
+class SafeTarFile(TarFile):
+    def extract(  # noqa: C901
+        self, member, path="", set_attrs=True, *, numeric_owner=False
+    ):
```
vlaci, line 15: The changes in this function can be implemented with minor copy-paste if you reuse the original method via `super().extract(...)`.

unblob/handlers/archive/_safe_tarfile.py
```diff
@@ -0,0 +1,101 @@
...
+        except OSError:
+            if self.errorlevel is not None and self.errorlevel > 0:
+                raise
+        except ExtractError:
+            if self.errorlevel is not None and self.errorlevel > 1:
+                raise
+
+    def _extract_member(  # noqa: C901
+        self, tarinfo, targetpath, set_attrs=True, numeric_owner=False
+    ):
```
vlaci, line 63: It may be a bit more tricky but the same applies here. The changes can be put in the overriden `extract` method (or checked post-fact in an exception handler for `PermissonError`.

Also, it'd make sense to cache if we are root, to spare a syscall for each special file.

vlaci (changes requested):
I am against copy-pasting from the standard library at all cost. It is not only a maintenance burden but introduces potential licensing complications as well.

I am okay fixing the standard library behavior from overriden methods while the boundary between our and the standard library code is preserved. If it is not possible I'd go back to just bundle a newer 7z version to unblob which doesn't have issues with symlinks.

vlaci (approved):

PR #464, opened 2022-10-28
Untangle (refactor) processing of a file

e3krisztian:
The goal is to produce a simpler to reason about report output, where all of the output files appear at top level.

It is achieved by breaking up the long pipe of processing a file into a series of much smaller tasks.

----

How to look/review it: it is a big code change, however it is mostly moving code around, and making some necessary structural changes, so please do not mention smaller, stylistic things, as the problem might have been already there, and the intention is not to clean up everything.

Overall I feel it a step forward and even a couple of smaller issues were fixed as well.

----

Status:
- tests run
- produces identical result to `main` (except for some empty `_extract` directories, and when run with `depth` limit reached)
- code review input, and further polishing work is required (commits, and probably some further cleanup).
- results in more use of multiprocessing, with a potential overhead (it is small, see below a performance comparison on a real FW)

Known issues/work still to do:
- `Handler` class has lost its `abc.ABC` base so as to allow `Handler` to be a singleton - needs discussion
- zip handler needs some refactor work (see also `ValidChunk.is_encrypted` below)
- `fixers.py` might not be the best name, it was created by extracting the functions from `processing.py` and all the functions start with `fix_` there. (maybe they belong to `file_utils.py`?)
- `depth` has changed in some cases, resulting in potentially shallower results compared to `unblob`-ing with `main` (needs discussion)
- `*Chunk` objects should not leak to tasks, as they are simply start-end offsets
- `ValidChunk.is_encrypted` is not used anymore, but calculating it takes time at chunk discovery time (they might even be calculated several times for the same chunk, if an archive is embedded in another file)
- `get_magic` and `get_mime_type` is now on `ExtractionConfig`, but they could be top-level/static function on their only user, the `ClassifierTask` class
- `ChunkReport` is not used anymore (except in tests)
- testing against more firmwares to exclude regression

----

Performance comparison on a real firmware:
```sh
$ du -hs $fw
163M    $fw
$ time unblob $fw --report main.report -f -d 100
...
real    1m16,511s
user    9m1,543s
sys     0m17,518s
$ git switch untangle
$ time unblob $fw --report untangle.report -f -d 100
...
real    1m22,435s
user    9m17,219s
sys     0m16,947s
$ wc main.report untangle.report
  2066851   4294827  95982807 main.report
  2015680   4033785  83462641 untangle.report
  4082531   8328612 179445448 total
```
The difference is small enough, and on the `untangle` run I even got a temperature sensor warning, so the run might have been partially throttled at some parts, though I think it was just a warning without throttling, and the differences are real.

----

A couple of smaller fixes were made, mostly type related (e.g. one, that caused editor marks is the `math` module).


martonilles:
Closing this PR as obsolete

PR #466, opened 2022-11-07
Zip chunk discovery performance fix

e3krisztian:
Eliminate loops for chunk discovery.

Zip handler used the initial local file header signature to find zip chunks. Once the local file header was found the end of central directory signature was looked for, then the distance between the two was used to validate the match.

This is not optimal, as there are one local file signature for every file in a zip file, resulting in a huge amount of signature matches and also a lot of failed chunk calculation. Worse still, for every match there are potentially many end of central directory headers to check against.

The new method directly looks only at the end of central directory headers, and from that calculates the first local file header (= start of the chunk).

Changing how zip file calculation works surfaced another problem: the handling of overlapping chunks.

When overlapping chunks have been found, the whole file have been rejected (in most cases silently, but in rare cases with a visible exception) by an `InvalidInputFormat` exception raised from `calculate_unknown_chunks`.

This has been changed, which also results in extraction of overlapping chunks - there is a chance, that one of them will be the real one, producing usable output.

### DIFF
--- a/tests/test_finder.py
+++ b/tests/test_finder.py
@@ -57,6 +57,40 @@ class TestHandlerExc(Handler):
         raise Exception("Error")
 
 
+class TestHandlerForward(Handler):
+    """Matches stuff between parens: (whatever)
+
+    Matches the initial ( and finds the next ) as chunk end.
+    """
+
+    NAME = "handlerEXC"
+    PATTERNS = [Regex("[(]")]
+
+    def calculate_chunk(self, file, start_offset: int):
+        end_offset = file.find(b")", start_offset)
+        assert end_offset >= 0
+        return ValidChunk(start_offset, end_offset + 1)
+
+
+class TestHandlerBackward(Handler):
+    """Matches stuff between square brackets: [whatever]
+
+    Matches the closing ] and finds the previous [ as chunk start.
+
+    This is not a common strategy, but can be a valid/more efficient strategy in some cases
+    (e.g. ZIP files).
+    """
+
+    NAME = "handlerEXC"
+    PATTERNS = [Regex("]")]
+
+    def calculate_chunk(self, file, start_offset: int):
+        end_offset = start_offset
+        start_offset = file.rfind(b"[", 0, end_offset)
+        assert start_offset >= 0
+        return ValidChunk(start_offset, end_offset + 1)
+
+
 def test_build_hyperscan_database():
     db, handler_map = build_hyperscan_database((TestHandlerA, TestHandlerB))
     matches = []
@@ -104,18 +138,12 @@ def test_invalid_hexstring_pattern_raises():
     [
         pytest.param(b"00A23450", [ValidChunk(2, 7)], id="single-chunk"),
         pytest.param(
-            b"0BB34A678900", [ValidChunk(0, 10)], id="chunk-with-relative-match-offset"
+            b"0BB34X678900", [ValidChunk(0, 10)], id="chunk-with-relative-match-offset"
         ),
         pytest.param(
             b"A23450BB3456789",
             [ValidChunk(0, 5), ValidChunk(5, 15)],
             id="multiple-chunk",
-        ),
-        pytest.param(b"0BC34A67890", [ValidChunk(0, 10)], id="inner-chunk-ignored"),
-        pytest.param(
-            b"0BC34A67890A2345",
-            [ValidChunk(0, 10), ValidChunk(11, 16)],
-            id="inner-chunk-ignored-scan-continues",
         ),
         pytest.param(b"A23450BB34", [ValidChunk(0, 5)], id="overflowing-chunk-ignored"),
         pytest.param(
@@ -138,6 +166,56 @@ def test_invalid_hexstring_pattern_raises():
         pytest.param(
             b"EXCA2345", [ValidChunk(3, 8)], id="exception-ignored-scan-continues"
         ),
+        pytest.param(
+            b"(.....[overlap]......)",
+            # ^            [^ - this pattern match is optimized away]
+            # 0123456789012345678901
+            [ValidChunk(0, 22)],
+            id="internal-chunk-whole-file-match-optimization",
+        ),
+        pytest.param(
+            b"[.....(overlap)......]",
+            #       ^              ^ backward matched second, the internal chunk is already there
+            # 0123456789012345678901
+            [ValidChunk(0, 22), ValidChunk(6, 15)],
+            id="internal-chunk-whole-file-match-optimization-does-not-work-for-backward-match",
+        ),
+        pytest.param(
+            b"[zip(..]zlib)",
+            #     ^  ^
+            # 0123456789012
+            [ValidChunk(0, 8), ValidChunk(4, 13)],
+            id="overlapping-chunks-backward-first",
+        ),
+        pytest.param(
+            b"(zlib[..)zip]",
+            # ^           ^
+            # 01234567890123
+            [ValidChunk(0, 9), ValidChunk(5, 13)],
+            id="overlapping-chunks-backward-second",
+        ),
+        pytest.param(
+            b"((AAA(((..).",
+            # ^ ^
+            # 01234567890123456789
+            [ValidChunk(0, 11), ValidChunk(2, 7)],
+            id="internal-matches-of-same-handler-are-optimized-away",
+            # An extra at the end is needed, because whole file chunks are
+            # stopping the pattern match.
+            #
+            # This optimization is intended for containers, like tar, cpio, ...
+            # where we have many file headers and the same - well determined -
+            # end for all of the file headers.
+            #
+            # Unfortunately it is possible only for forward handlers,
+            # where the pattern is at the beginning of the structure, not for
+            # something like zip, which we can better match from the end.
+            #
+            # It is also somewhat questionable if we want it generally,
+            # as e.g. for HandlerA it will result matching only a single A chunk
+            # while in fact there are 3 overlapping matches here (2-7, 3-8, 4-9),
+            # but we just drop the last 2.
+        ),
     ],
 )
 def test_search_chunks(content, expected_chunks, task_result):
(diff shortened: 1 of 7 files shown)


unblob/handlers/archive/zip.py
```diff
@@ -107,70 +121,79 @@ def is_zip64_eocd(end_of_central_directory: Instance):
...
+        # start_offset = start of EOCD record
+        # EOCD = end of central directory
+        eocd_offset = start_offset
+        file.seek(eocd_offset, io.SEEK_SET)
+        # from here calculate the beginning of the zip file (needs to interpret the CD offsets)
+        # REASON: EOCD SIGNATURE is much rarer when there are many embedded zip files in a blob
+        # while there are a lot of LOCAL FILE HEADER SIGNATUREs
+
+        end_of_central_directory = self.parse_header(file)
+        assert end_of_central_directory.signature == EOCD_SIGNATURE
```
QKaiser, line 162: This will always be true since we're matching on `EOCD_SIGNATURE`.

unblob/handlers/archive/zip.py
```diff
@@ -107,70 +121,79 @@ def is_zip64_eocd(end_of_central_directory: Instance):
...
+        zip64_eocd = self.cparser_le.zip64_end_of_central_directory_t(file)
+        if zip64_eocd.signature != ZIP64_EOCD_SIGNATURE:
+            raise InvalidInputFormat("Zip64 EOCD not found")
+        return zip64_eocd_offset, zip64_eocd
 
     def calculate_chunk(self, file: File, start_offset: int) -> Optional[ValidChunk]:
+        # start_offset = start of EOCD record
+        # EOCD = end of central directory
+        eocd_offset = start_offset
+        file.seek(eocd_offset, io.SEEK_SET)
```
QKaiser, line 156: Do we really need to seek ? I think we're already on that position when `calculate_chunk` is called.

unblob/handlers/archive/zip.py
e3krisztian: Yes, I am aware, that it is always true (now), however it is a bit implicit and that's why it is an assert: previously we matched against something else.

It is not strictly needed, indeed, I have added this assert to keep alive the constant `EOCD_SIGNATURE` (vulture complained for it).

It is harmless, and maybe more of a documentation than a noise, so I would not remove it.

unblob/handlers/archive/zip.py
e3krisztian: You are right, it will be removed.

e3krisztian:
Overlapping chunks cause performance issues with at least tar and cpio, for which every start of new file header is a valid start of tar/cpio sub-chunk, thus for them we would get the same number of chunks as the files contained, however calculating the chunks are expensive, depend on the number of files contained and the file size, so for these handlers, the chunk calculation algorithm changes from O(n) to O(n**2) where n is the number of files in the archive.

We need to prevent the sub-chunk calculations for these handlers, for this we might use `HyperscanMatchContext` and exclude matches for handlers inside existing chunks of the same handler.

e3krisztian:
Added back enough pattern-match-skipping to cover tar, cpio, and containers in general: matches are skipped if they are inside a chunk of the same handler.
This optimization also eliminated most of zlib false positives, as zlib chunks are also overlapping with themselves a lot.
Also fixed and cleaned up `calculate_unknown_chunks` to work properly with overlapped chunks.

unblob/processing.py
```diff
@@ -422,34 +423,24 @@ def calculate_unknown_chunks(
     chunks: List[ValidChunk], file_size: int
 ) -> List[UnknownChunk]:
     """Calculate the empty gaps between chunks."""
-    if not chunks or file_size == 0:
+    # we do not want whole file unknown chunks (!?)
```
QKaiser, line 426: sentence is unclear. Maybe something like "carving out an unknown chunk covering the entire file would lead to an infinite loop" ?

QKaiser (commented):
> Some further work is still needed, as by fixing this optimization,
a lot of new false zlib chunks are extracted, producing a lot of noise
in the output.
False zlib chunks are due to zlib having an exceptionally short pattern
(only 2 bytes), and an implicit stream terminator.

Is it something you plan on working in this branch ? Do you already know how to thix this ? Can it be fixed actually ?

I really like what you did in `test_finder.py`, it provides a nice framework to think about what we actually want out of overlapping chunks.

unblob/processing.py
e3krisztian: I did not think about the infinite loop here, as carving would happen somewhere else, this condition should be there to prevent the infinite loop.

I have copied this comment, not verbatim, but documenting a former decision - I remember adding the `(!?)` as I do not fully agree with/understand the decision. The original comment was here:

https://github.com/onekey-sec/unblob/blob/main/unblob/processing.py#L324

```py
# we don't consider whole files as unknown chunks, but we still want to
# calculate entropy for whole files which produced no valid chunks
```

I have removed this comment, as it was not here before, and just induces confusion.

e3krisztian:
> > Some further work is still needed, as by fixing this optimization,
> > a lot of new false zlib chunks are extracted, producing a lot of noise
> > in the output.
> > False zlib chunks are due to zlib having an exceptionally short pattern
> > (only 2 bytes), and an implicit stream terminator.
> 
> Is it something you plan on working in this branch ? Do you already know how to thix this ? Can it be fixed actually ?

Short answer: no, I do not plan on working on it here, it is OK-ish now.

Longer version:
With partially going back on extracting all overlapping chunks, and continuing to ignore the second of overlapping chunk pairs *iff* the chunk's handler is the same as the second matches' handler solves a lots of the problems with zlib for my test set, as zlib overlapped a lot with itself :smile:.  It is also a relatively safe cleanup, as a more aggressive one was there before.

I have tested the last version only on a couple of images, and the zlib outputs depend a lot on the input, but it looks OK for me.

Possible future development could remove zlib matches overlapping with any other chunk to still further reduce the false zlib junk, however it is also a question of how eager we want to go with this - it could also occasionally remove valid zlib chunks as well. I have not seen any valid zlib chunks so far, so I might be to fast to dismiss them.

unblob/finder.py
```diff
@@ -82,9 +82,15 @@ def _hyperscan_match(
...
 
+    # Do not overlap with chunks of the same handler.
+    # This is mostly valid for containers, so should be configurable for each handler.
     # Skip chunk calculation if this would start inside another one,
     # similar to remove_inner_chunks, but before we even begin calculating.
-    if any(chunk.contains_offset(real_offset) for chunk in context.all_chunks):
+    if any(
+        chunk.contains_offset(real_offset)
+        for chunk in context.all_chunks
+        if chunk.handler == handler
```
QKaiser, line 92: Future readers: outside of the ZIP handler, the most important change of this MR is here.

QKaiser:
@e3krisztian Regarding "false" zlib, an optimisation could be to disregard any zlib chunk shorter than some arbitrary length. Could you send me the sample that generates this zlib noise ?

QKaiser (approved):
Verified locally with some samples. Observable performance for ZIP chunks. Next step is to take care of zlib noise.

tests/test_finder.py
```diff
@@ -57,6 +57,40 @@ def calculate_chunk(self, file, start_offset: int):
...
+        assert end_offset >= 0
+        return ValidChunk(start_offset, end_offset + 1)
+
+
+class TestHandlerBackward(Handler):
+    """Matches stuff between square brackets: [whatever]
+
+    Matches the closing ] and finds the previous [ as chunk start.
+
+    This is not a common strategy, but can be a valid/more efficient strategy in some cases
```
martonilles, line 80: While this could be a valid strategy, this would fundamentally change how we think about chunk matching. So far all handlers are working in a start->end direction and we use this to optimize our chunk finding. Changing it would have a severe performance impact on how unblob works.

unblob/finder.py
```diff
@@ -114,9 +120,9 @@ def _hyperscan_match(
     logger.debug("Found valid chunk", chunk=chunk, handler=handler.NAME, _verbosity=2)
     context.all_chunks.append(chunk)
 
-    # Terminate scan if we match till the end of the file
-    if chunk.end_offset == context.file_size:
-        logger.debug("Chunk covers till end of the file", chunk=chunk)
+    # Terminate scan if the whole file is matched
+    if chunk.start_offset == 0 and chunk.end_offset == context.file_size:
```
martonilles, line 124: This is also a heavy performance optiomization to match till the end of the file. It is a common pattern to have some prefix in a file followed by the content we pick up. Once we identify that the chunk covers till the end of the file, we can just bail out searching.

tests/test_processing.py
```diff
@@ -101,6 +106,18 @@ def test_remove_inner_chunks(
             20,
             [UnknownChunk(0x5, 0x8), UnknownChunk(0xA, 0xF)],
         ),
+        pytest.param(
```
martonilles, line 109: not sure we want to solve the case when there are overlapping chunks. This is a problem in the chunk handler and by handling overlapping chunks, we would hid potential handler bugs.

martonilles (changes requested):
I am not sure that changing fundamentally unblob chunk logic handling is a good way to fix this particular zip problem, considering the performance penalties (given I understand correctly the changes). Maybe we should just optimize the zip handling instead of changing unblob in general just for the sake of improving zip. Let's talk about this.

QKaiser:
> I am not sure that changing fundamentally unblob chunk logic handling is a good way to fix this particular zip problem, considering the performance penalties (given I understand correctly the changes). Maybe we should just optimize the zip handling instead of changing unblob in general just for the sake of improving zip. Let's talk about this.

Agreed. Probably better to identify the root cause of the miscalculation with the sample at hand within the handlers themselves (ZIP, zlib) than changing unblob core. The problem is not that we have overlapping chunks, it's that we emit them in the first place.

martonilles:
Closing this PR as obsolete

PR #468, opened 2022-11-14
fix tar handler: truncated tar files were not processed correctly

e3krisztian:
Truncated tar files had a performance impact due to an `EOFError` not being caught and thus the tar file not being recognized, and this got repeated for every file in the truncated tar file.


e3krisztian:
`File.seek` no longer raises `EOFError`, but a new `SeekError` was introduced, which is derived from `ValueError` to maintain `seek`'s expected behavior.

----

Commits are not nice, they show development history,  before merging the commits should be squashed with the last commit message.

QKaiser (approved):

Issue #477, opened 2022-11-30
unblob cannot scan files larger than 4GB


QKaiser:
Due to the following limitation of Hyperscan (see https://github.com/intel/hyperscan/issues/118):

> hs_scan, hs_scan_vector and hs_scan_stream use an unsigned int parameter for providing the length of the buffer to scan. This effectively limits the length to 4GB-1. [0]

Unblob is not able to identify patterns within files larger than 4GB in size. This is problematic for our larger samples.

A naive but somewhat effective approach would be to switch to streaming mode with Hyperscan.

```diff
--- a/unblob/finder.py
+++ b/unblob/finder.py
@@ -146,16 +146,15 @@ def search_chunks(  # noqa: C901
     )
 
+    CHUNK_SIZE = (100 * 1024 * 1024)  # 100MB or whatever arbitrary value
 
     try:
-        hyperscan_db.scan(
-            [
-                file[i : i + CHUNK_SIZE]  # noqa: E203
-                for i in range(0, file.size(), CHUNK_SIZE)
-            ],
-            match_event_handler=_hyperscan_match,
-            context=hyperscan_context,
-        )
+        with hyperscan_db.stream(match_event_handler=_hyperscan_match, context=hyperscan_context) as stream:
+            for i in range(0, file.size(), CHUNK_SIZE):
+                logger.info("scanning range", start=i, end=i+CHUNK_SIZE)
+                stream.scan(
+                    file[i : i + CHUNK_SIZE], # noqa: E203
+                )
     except hyperscan.error as e:
         if e.args and e.args[0] == f"error code {hyperscan.HS_SCAN_TERMINATED}":
             logger.debug(
@@ -178,7 +177,9 @@ def search_chunks(  # noqa: C901
 
 @lru_cache
 def build_hyperscan_database(handlers: Handlers) -> Tuple[hyperscan.Database, Dict]:
-    db = hyperscan.Database(mode=hyperscan.HS_MODE_VECTORED)
+    db = hyperscan.Database(
+        mode=(hyperscan.HS_MODE_STREAM | hyperscan.HS_MODE_SOM_HORIZON_LARGE)
+    )
     handler_map = dict()
```

This would require a bit of work to adapt our test suite and making sure we keep track of the current chunk start offset in the scanning context, but it's entirely doable.

[...]

PR #480, opened 2022-12-06
Lift 4GB file size limit by moving to streaming mode

QKaiser:
This merge request replace Hyperscan scans in vectored mode with scans in streaming mode in order to lift the 4GB size limit documented in https://github.com/onekey-sec/unblob/issues/477. I'm using `DEFAULT_BUFSIZE` from `file_utils` as the streaming scan increment.

This change brings some speed improvement for files with a single valid chunk covering the whole file, since we return earlier from `search_chunks`. Here's the time difference when scanning a 4GB ExtFS file:

```
(branch:477-lift-4gb-limit)
time poetry run unblob -d 1 -e /out -f sample_4GB.extfs
2022-12-06 09:41.19 [info     ] Start processing file          file=swissphone/1.img pid=16109
poetry run unblob -d 1 -e /out -f sample_4GB.extfs  46,55s user 57,72s system 26% cpu 6:35,66 total
```

```
(branch:main)
time poetry run unblob -d 1 -e /out -f sample_4GB.extfs
2022-12-06 09:49.47 [info     ] Start processing file          file=swissphone/1.img pid=16308
poetry run unblob -d 1 -e /out -f sample_4GB.extfs  58,86s user 65,66s system 27% cpu 7:35,64 total
```

Note: I also extended the `File` class to support `.size()` calls on file objects created from byte arrays.


e3krisztian:
I am confused with the description - had it also become faster with moving to the streaming mode?

QKaiser:
> I am confused with the description - had it also become faster with moving to the streaming mode?

Yes, but specifically for files that have a valid chunk covering the whole size.

In vectored mode, hyperscan would scan the whole file content and unblob would spend time discarding callbacks for matches within the already identified valid chunk.

In streaming mode, the time spent discarding callbacks for matches within the already identified valid chunk is reduced because we scan less content (65535 bytes to be exact), therefore leading to less callbacks to discard.

unblob/finder.py
```diff
@@ -139,11 +139,15 @@ def search_chunks(  # noqa: C901
...
-            return all_chunks
+        for i in range(0, file.size(), DEFAULT_BUFSIZE):
+            if (
+                scanner.scan(file[i : i + DEFAULT_BUFSIZE])  # noqa: E203
+                == Scan.Terminate
+            ):
+                logger.debug(
+                    "Scanning terminated as chunk matches till end of file",
+                )
+                return all_chunks
```
e3krisztian, line 150: When removing the logging (which was already logged on line 111 above) and instead of returning immediately just exiting the loop like this:
```suggestion
        for i in range(0, file.size(), DEFAULT_BUFSIZE):
            if (
                scanner.scan(file[i : i + DEFAULT_BUFSIZE])  # noqa: E203
                == Scan.Terminate
            ):
                break
```

We ended up with a non-trivial code chunk, that has been repeated twice more in other files (`xz` and `bzip2` handlers).

I propose to extract this common functionality to a separate function in `file_utils`:

```py
def stream_scan(scanner, file: File):
    for i in range(0, file.size(), DEFAULT_BUFSIZE):
        if (
            scanner.scan(file[i : i + DEFAULT_BUFSIZE])  # noqa: E203
            == Scan.Terminate
        ):
            break
```

(Maybe also add a type annotation for `scanner` and a doc-string)

e3krisztian (changes requested):

e3krisztian (approved):

PR #484, opened 2022-12-16
Fix lzop extraction 

QKaiser:
lzop decompression command refuse to decompress files without a `.lzo` suffix unless the `-f|--force` command line switch is provided twice

### DIFF
--- a/unblob/handlers/compression/lzo.py
+++ b/unblob/handlers/compression/lzo.py
@@ -75,7 +75,7 @@ class LZOHandler(StructHandler):
     """
     HEADER_STRUCT = "lzo_header"
 
-    EXTRACTOR = Command("lzop", "-d", "-f", "-N", "-p{outdir}", "{inpath}")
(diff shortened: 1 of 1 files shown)


unblob/handlers/compression/lzo.py
```diff
@@ -75,7 +75,7 @@ class LZOHandler(StructHandler):
     """
     HEADER_STRUCT = "lzo_header"
 
-    EXTRACTOR = Command("lzop", "-d", "-f", "-N", "-p{outdir}", "{inpath}")
```
martonilles, line 78: just wondering how this could happen? the carved chunk will always have `.lzo` suffix. Or was it a whole file chunk without the proper suffix?

QKaiser: Namely a file within an archive named 'ramdisk'.

martonilles: Should have read the original issue first ... :(

Issue #486, opened 2022-12-20
Optional "carving-only" Parameter


dorpvom:
Hi,

I was looking into integrating unblob in [FACT](https://github.com/fkie-cad/FACT_core) (more precisely [fact_extractor](https://github.com/fkie-cad/fact_extractor)) and needed unblob to work as a carver for this. Meaning, to ignore all the nice unpacker integrations you offer and maintain, and just extract the binary of e.g. file system, from the container blob.

I [implemented this](https://github.com/onekey-sec/unblob/compare/main...dorpvom:unblob:allow-carving-only) using a simple conditional triggered by a command line option.

So my question is: Is that interesting to you, i.e. should I make an effort to open a PR for this and possibly write test cases?

Cheers!

QKaiser:
Hi ! Your proposal makes sense and limited in scope so I don't think we should refuse it.

We'll discuss it internally and I'll get back to you with more details. In the meantime you can open a merge request so that we can review it.

martonilles:
@dorpvom I am just wondering what is the reason you do not need the extraction? In case a multi-layered blob you would need to call unblob recursively anyhow. Or you just do not need the extraction?

Otherwise thanks for the contribution and please go ahead an open a PR.
Quick glance on your branch:
- I would probably return early from the function when exraction skipped, probably doing `fix_extracted_directory`
- also this would imply using depth=1

dorpvom:
Thanks for the feedback.

@martonilles The reason for skipping extraction for me specifically is that I want to use my own extraction harness for now and simply intend to replace good-old-binwalk. I do think it has some other use cases as well but that's it for me.

I'll look into the branch to incorporate your feedback and will then open a PR.

PR #487, opened 2023-01-03
Allow carving only

dorpvom:
- Added command line option
- Applied early exit before extraction (@martonilles)

### DIFF
--- a/unblob/cli.py
+++ b/unblob/cli.py
@@ -152,6 +152,14 @@ class UnblobContext(click.Context):
     help="File to store metadata generated during the extraction process (in JSON format).",
 )
 @click.option(
+    "-s",
+    "--skip_extraction",
+    "skip_extraction",
+    is_flag=True,
+    show_default=True,
+    help="Only carve chunks and skip further extraction",
+)
+@click.option(
     "-k",
     "--keep-extracted-chunks",
     "keep_extracted_chunks",
@@ -176,6 +184,7 @@ def cli(
     depth: int,
     entropy_depth: int,
     skip_magic: Iterable[str],
+    skip_extraction: bool,
     keep_extracted_chunks: bool,
     handlers: Handlers,
     plugins_path: Optional[Path],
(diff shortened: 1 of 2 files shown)


dorpvom:
Regarding depth:

Yes, using this obviously enforces depth = 1. I would add a logger.debug for this as it is trivial when understanding the option. It would also be possible to make it a warning if depth is not the default, and one can assume it was changed on purpose.
I would probably put that into the `cli` method in _unblob/cli.py_, but I'm happy for suggestions.

vlaci:
Thanks for sending this PR!
I see that you already had some discussion with the guys in #486 about this feature. I am sure they'll look at it as soon as they are able.

QKaiser:
@dorpvom could you clean up your commit history and provide a single commit ? I see changes within the same code blocks in `processing.py` from both a87578963ef25095d2bce0ba4247c7dedb0e1a74 and f5a94902baddcabc8b8a5ecaab26c8882dfaba35. Also d658210f1e7f17cf6f9e45125005335fd355e7ee should not be there.

We recommend using [semantic commit messages](https://gist.github.com/joshbuchea/6f47e86d2510bce28f8e7f42ae84c716), yours could be `feat: add carve-only capability with corresponding command line switch (-s)`. Do not hesitate to provide a detailed explanation of what you did within the commit message body, this is how we document things.

dorpvom:
> @dorpvom could you clean up your commit history and provide a single commit ? I see changes within the same code blocks in `processing.py` from both [a875789](https://github.com/onekey-sec/unblob/commit/a87578963ef25095d2bce0ba4247c7dedb0e1a74) and [f5a9490](https://github.com/onekey-sec/unblob/commit/f5a94902baddcabc8b8a5ecaab26c8882dfaba35). Also [d658210](https://github.com/onekey-sec/unblob/commit/d658210f1e7f17cf6f9e45125005335fd355e7ee) should not be there.
> 
> We recommend using [semantic commit messages](https://gist.github.com/joshbuchea/6f47e86d2510bce28f8e7f42ae84c716), yours could be `feat: add carve-only capability with corresponding command line switch (-s)`. Do not hesitate to provide a detailed explanation of what you did within the commit message body, this is how we document things.

@QKaiser So I'm not usually in bed with rebasing and force-pushing but I guess I got this history where you want it. Any additional requests?

QKaiser (approved):

PR #488, opened 2023-01-04
jffs2: ignore CRC for inaccurate node types

ffontaine:
Ignore CRC for inaccurate node types (e.g. c001 or c002) to fix extraction of JFFS2 from two different flash dumps of two different vendors

Signed-off-by: Fabrice Fontaine <fabrice.fontaine@orange.com>


vlaci:
Thank you for your contribution!
It looks sound to me. Someone with more in-depth knowledge of the intricacies of this format will take a look soon.

QKaiser (approved):

QKaiser:
This makes perfect sense, merging now. @ffontaine if you happen to have a sample that you can share with us, I'll happily turn it into a reduced test case in our integration tests files.

ffontaine:
Thanks for merging this PR @QKaiser, here is a JFFS2 sample: [jffs2_extract.zip](https://github.com/onekey-sec/unblob/files/10351955/jffs2_extract.zip)

Issue #392, opened 2022-06-08
Empty *_extract directory is created for whole-file ELF binaries


vlaci:
While testing #390 I found that a bunch of unneded directories are created for ELF executables.

Because ELFKernelExtractor is set for ELF handlers, an empty extract directory always gets created, even for whole-file elf binaries.

This results in a very cluttered output, e.g:

```
❯ find ! -type l -ls
  1155940      0 drwxrwxr-x  25 vlaci    users        1280 jún  8 10:17 .
  1156377      0 drwxrwxr-x   2 vlaci    users          40 jún  8 10:17 ./dnsmasq_extract
  1156376      0 drwxrwxr-x   2 vlaci    users          40 jún  8 10:17 ./dropbear_extract
  1156369      0 drwxrwxr-x   2 vlaci    users          40 jún  8 10:12 ./fw_printenv_extract
  1156368      0 drwxrwxr-x   2 vlaci    users          40 jún  8 10:12 ./iw_extract
  1156367      0 drwxrwxr-x   2 vlaci    users          40 jún  8 10:12 ./odhcp6c_extract
  1156366      0 drwxrwxr-x   2 vlaci    users          40 jún  8 10:12 ./odhcpd_extract
  1156365      0 drwxrwxr-x   2 vlaci    users          40 jún  8 10:12 ./pppd_extract
  1156364      0 drwxrwxr-x   2 vlaci    users          40 jún  8 10:12 ./px5g_extract
  1156363      0 drwxrwxr-x   2 vlaci    users          40 jún  8 10:12 ./ubiattach_extract
  1156362      0 drwxrwxr-x   2 vlaci    users          40 jún  8 10:12 ./ubiblock_extract
  1156361      0 drwxrwxr-x   2 vlaci    users          40 jún  8 10:12 ./ubicrc32_extract
  1156360      0 drwxrwxr-x   2 vlaci    users          40 jún  8 10:12 ./ubidetach_extract
  1156359      0 drwxrwxr-x   2 vlaci    users          40 jún  8 10:12 ./ubiformat_extract
  1156358      0 drwxrwxr-x   2 vlaci    users          40 jún  8 10:12 ./ubimkvol_extract
  1156357      0 drwxrwxr-x   2 vlaci    users          40 jún  8 10:12 ./ubinfo_extract
  1156356      0 drwxrwxr-x   2 vlaci    users          40 jún  8 10:12 ./ubinize_extract
[...]

PR #491, opened 2023-01-16
delete empty _extract directories when an ExtractError is raised.

QKaiser:
Related to #392 

When a handler raises a `NullExtract` exception, we are left with an empty `something_extract` directory, which is useless. More generally, when a `ExtractError` is raised we may be left with such an empty directory if the extraction process did not produce any kind of results.

Whenever such an exception is raised, we delete the dangling directory if it still exists and does not contain anything. We only delete empty directories because an `ExtractError` exception could still be raised for an extraction run that generated partial results.

### DIFF
--- a/unblob/processing.py
+++ b/unblob/processing.py
@@ -389,7 +389,7 @@ class _FileTask:
 
         except ExtractError as e:
             extraction_reports.extend(e.reports)
-
+            delete_empty_extract_dir(extract_dir)
         except Exception as exc:
             logger.exception("Unknown error happened while extracting chunk")
             extraction_reports.append(UnknownError(exception=exc))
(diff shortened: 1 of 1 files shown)


unblob/processing.py
```diff
@@ -409,6 +409,12 @@ def _extract_chunk(self, file, chunk: ValidChunk):
             )
 
 
+def delete_empty_extract_dir(extract_dir: Path):
+    # delete empty extract directories
+    if extract_dir.exists() and next(extract_dir.iterdir(), None) is None:
```
martonilles, line 414: easier is:

`if extract_dir.exists() and not any(extract_dir.iterdir()):`

or just use `rmdir` and handle `OSError: [Errno 39] Directory not empty` error

QKaiser: Used your construct and removed the superfluous comment.

kukovecz (approved):

PR #496, opened 2023-01-23
Fix documentation deployment

kissgyorgy:
More dependencies has been added so pipx run mkdocs is not enouch anymore, all other dependencies need to be installed.
I created an optional dependency group `docs` which can be selectively installed from the documentation job.


pyproject.toml
```diff
@@ -38,6 +38,12 @@ pyright = "^0.0.12"
 pre-commit = "^2.15.0"
 pytest-cov = "^3.0.0"
 setuptools-rust = "^1.1.2"
+
```
QKaiser, line 41: You probably need to use `[tool.poetry.group.dev.dependencies]` above to make our use of group consistent (and make the Nix build pass).

QKaiser:
```diff
diff --git a/pyproject.toml b/pyproject.toml
index 0567094..f8bfb64 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,8 +31,10 @@ lark = "^1.1.2"
 lz4 = "^4.0.0"
 lief = "^0.12.3"
 
+[tool.poetry.group.dev]
+optional = true
 
-[tool.poetry.dev-dependencies]
+[tool.poetry.group.dev.dependencies]
 pytest = "^6.2.4"
 pyright = "^0.0.12"
 pre-commit = "^2.15.0"
```

kissgyorgy: Good catch, moved it.

kissgyorgy:
Can you trigger the documentation build? It's not set to trigger on Pull Requests, but I think it will work now.

QKaiser (approved):

Issue #353, opened 2022-04-21
Define public API surface


vlaci:
Unblob is converted to a py.typed library in #352. After that change we are also able to define library-private and public modules and types. We should expose the functionality needed for external handlers and for running unblob as library and make the related code public.

https://github.com/microsoft/pyright/blob/main/docs/typed-libraries.md#library-interface

kissgyorgy:
It is better and easier to start with the absolute minimal public surface and open up later, because once we make something public, we essentially can't "take it back", because people can start relying on those. Also it's easier to refactor internal implementation when we don't expose too much.

Therefore I suggest to expose *only* the main entry point `process_files` and the related objects (`ExtractionConfig` and `Report`), and later we can open other functions as necessary.

Issue #243, opened 2022-02-08
CLI arguments are inconsistent


vlaci:
Now almost all command line arguments have a short and a long name. Most short names are undecipherable or misleading without reading through the help. Also, few argument names leak implementation details. We shouldn't expose short switches for more obscure functionality at all and the ones we are exposing should be as intuitive as possible.

I think we should double think about the names of these switches before 1.0:

- `-n`: It means dry-run in most command line programs and it may make sense to have a dry-run switch in the future, but right now it is configuring entropy calculation.
- `-e`: Output location can be configured either via positional argument or `-o` many times. As we are supporting arbitrary input files, we shouldn't use a positional argument though.
- `--process-num`/`-p`: There is no generally established name here. Some times `-j` is used to denote "jobs", or `-p` to indicate parallelism. I think we shouldn't indicate on our public API that we are using processes for parallelism. (It is also true for the actual `process_file` function too) 
- `--show-external-dependencies` this is an odd one we agreed about in the past that it doesn't really seem right and may make sense to use a subcommand instead but we didn't want to introduce them just for this functionality. If we are to use subcommands elsewhere, we should reconsider this as well

QKaiser:
Can you show us your ideal `--help` output ? :)

I think the `-e` probably comes from our usage of binwalk in the past, I'd gladly change it to `-o / --output-dir` which would fall in line with the CLI of most of our extractors (7zip, jefferson, ...). Note: if we do so, we also have to change that option in unromfs.

We can decide to reserve `-n` for future use (dry-run switch) and move the entropy calculation switch to something more meaningful. The symbol for entropy is S, so maybe use `-s / --entropy-depth` ?

I don't have opinions on `--show-external-dependencies` at the moment. I'm not against it.

Regarding `-p`:

> I think we shouldn't indicate on our public API that we are using processes for parallelism. (It is also true for the actual process_file function too)

You would keep the option and hide it from the --help section ? Or just hide it from the general documentation (README, wiki) ? In terms of options both `-j` and `-p` work for me.

vlaci:
> Can you show us your ideal `--help` output ? :)
> 
> I think the `-e` probably comes from our usage of binwalk in the past, I'd gladly change it to `-o / --output-dir` which would fall in line with the CLI of most of our extractors (7zip, jefferson, ...). Note: if we do so, we also have to change that option in unromfs.

Interestingly, binwalk's use of `-e` is different from ours, they have `-C`˙for the output directory.

> 
[...]

PR #498, opened 2023-01-25
Pypi release

### LINKED ISSUE #493: Binary Installation
Am I missing that? Any package managers? apt-get/brew/pip?

kissgyorgy:
I made the GitHub Action to release the Python package to PyPI, the same as we did for 
https://github.com/onekey-sec/python-client.
WIP, because you need to come up with a versioning scheme before this can be used.
I think the `POETRY_PYPI_TOKEN_PYPI` secret is org level, but I don't remember, need to be checked.

Fixes #493

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,20 @@
+name: PyPI Release
+
+on:
+  release:
+    types: [created]
+
+jobs:
+  release:
+    runs-on: ubuntu-20.04
+
+    steps:
+      - name: Checkout source code
+        uses: actions/checkout@v2
+
+      - name: Setup dependencies
+        uses: ./.github/actions/setup-dependencies
+
+      - run: poetry publish --build
+        env:
+          POETRY_PYPI_TOKEN_PYPI: ${{secrets.POETRY_PYPI_TOKEN_PYPI}}
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -5,6 +5,37 @@ hide:
 
 # Installation
 
+Unblob consists of two main parts:
+
+- unblob, the Python package (with optional Rust modules).
+- extractor command line tools like `7zip`, `unar`, etc. (See [Extractors](./extractors.md) for explanation.)
+
+All of these need to be installed to make unblob fully functional.  
+Depending the packaging solution you choose, you might need to
+install external extractors manually.
+
+## Python package
+
+unblob can be installed (without the extractors) from PyPI (Python Package Index).  
+This might be the easiest method, depending on whether you have Python 3 installed already.
+
+1.  First, install the Python package:
+
+        python3 -m pip install --user unblob
+
+    This will install the `unblob` script in `~/.local/bin`. You can put that
+    directory in your `PATH` environment variable, or call it directly.
+
+    !!! Warning
+
+        System-wide installation (with `sudo`) is not recommended, because it can potentially break your system.
+
+2.  Make sure to [install extractors](#install-extractors).
+
+3.  Check that everything works correctly:
+
+        unblob --show-external-dependencies
+
 ## Docker image
 
 unblob can be used right away from a `docker` image: `ghcr.io/onekey-sec/unblob:latest`,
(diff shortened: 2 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,20 @@
+name: PyPI Release
+
+on:
```
QKaiser, line 3: My understanding is that it would only publish when we explicitly click on "release" within the Github interface. I would add another trigger: when a tag gets pushed on `main` with a specific tag format (e.g. `v.*.*.*`).

This would enforce the use of tags, which are super helpful when tracking bugs / regressions between versions.

What do you think ?

.github/workflows/release.yml
```diff
@@ -0,0 +1,20 @@
...
+    steps:
+      - name: Checkout source code
+        uses: actions/checkout@v2
+
+      - name: Setup dependencies
+        uses: ./.github/actions/setup-dependencies
+
+      - run: poetry publish --build
+        env:
+          POETRY_PYPI_TOKEN_PYPI: ${{secrets.POETRY_PYPI_TOKEN_PYPI}}
```
QKaiser, line 20: This secret is not org level so we need to either set it at org level or put it into this repo. I'll keep you posted.

QKaiser:
On versioning scheme, I think we should use the same scheme than the one we used with python-client (`major.minor.patch`). I'd start with `0.1.1` since we're currently on `0.1.0` (and always were before).

.github/workflows/release.yml
martonilles: This:

```
on:
  release:
    types: [created]
```

is only triggered when we have a full release (done manually)

kissgyorgy: Tags are enforced by GitHub. When you create a release, it will automatically create a git tag for it, or if you specify an existing git tag, it will assign the release to that tag, so I don't think there is a big difference on what exactly triggers the release. With this trigger, it is possible to push the tag beforehand, try out the tagged build and only make a release if everything works. It's even possible to make Draft releases, which you might not want to publish to PyPI (e.g. alpha, beta releases).

kissgyorgy:
I asked about the versioning scheme, because it is a somewhat complex question.
Here are a couple of questions:
- Do we consider unblob stable at this point? from what point? As a library or a CLI?
- What does the version mean? API stability? Command line interface stability? Extract folder strucutre? JSON metadata output?
- How often will you break the contract?
- Different version schemes communicate different things to users. How do you want to cummunicate the changes or the overall state of the project? https://calver.org/ and https://semver.org/ are very different 
- When do you want to bump version numbers? Every release? Only when API changes? Only when CLI interface changes?

I think it makes sense to consider CalVer, which would essentially mean "this project is constantly improved, the release version means nothing and we don't guarantee any kind of stability between versions". Maye this would manage user expectations better?

.github/workflows/release.yml
QKaiser: @martonilles is on it. Please mark this thread as resolved when done :)

martonilles: Added POETRY_PYPI_TOKEN_PYPI secrets that can be used to publish to pypi

martonilles:
Agree, I would use preferably calver, like this: YYYY.MM.DD or YYYY.MM.MICRO  otherwise in semver, I would go for a 1.1.0 release now as I think we are stable for a while.

martonilles (approved):

kissgyorgy:
I added a documentation in the install section on how to install from PyPI.

docs/installation.md
```diff
@@ -5,6 +5,37 @@ hide:
 
 # Installation
 
+Unblob consists of two main parts:
```
QKaiser, line 8: Documentation looks good to me.

QKaiser (approved):

kissgyorgy:
> I would go for a 1.1.0 release now as I think we are stable for a while.

There are 3 milestones, none of them are closed: https://github.com/onekey-sec/unblob/milestones
At least we should solve #353 and #243 before we should call it 1.0, so people can rely on the behavior of the CLI and the stability of the API (as a library).

Or we could use calver immediately with no guarantees and delete the milestones 😄

QKaiser:
Marked as ready. Let's use calendar versioning (my preference goes to `YYYY.MM.DD`). I leave the merge, version bump, and first release to our overlord @martonilles 🙇

martonilles (approved):

PR #504, opened 2023-02-06
chore(deps): replace ubi-reader fork by official PyPi version.

QKaiser:
We cannot use git hosted dependencies if we want to publish unblob to PyPi. We therefore moved from our fork of ubi-reader to the official package of ubi-reader.

Our fork was synced with upstream except for one UI change we did to get rid of superfluous directories created by ubireader_extract_images. Since the maintainer of ubi-reader does not want to include it upstream, we moved that logic into the UBIExtractor class.

See https://github.com/jrspruitt/ubi_reader/pull/67 for more details.

### DIFF
--- a/unblob/handlers/filesystem/ubi.py
+++ b/unblob/handlers/filesystem/ubi.py
@@ -1,4 +1,6 @@
+import shutil
 import statistics
+from pathlib import Path
 from typing import Optional
 
 from structlog import get_logger
@@ -95,6 +97,18 @@ class UBIFSHandler(StructHandler):
         )
 
 
+class UBIExtractor(Command):
+    def extract(self, inpath: Path, outdir: Path):
+        super().extract(inpath, outdir)
+        # ubireader_extract_images creates a superfluous directory named
+        # after the UBI file (inpath here), so we simply move the files up
+        # and delete the remaining directory.
+        superfluous_dir_path = outdir.joinpath(inpath.name)
+        for file_path in superfluous_dir_path.iterdir():
+            shutil.move(file_path.as_posix(), outdir.as_posix())
+        shutil.rmtree(superfluous_dir_path.as_posix())
+
+
 class UBIHandler(Handler):
     NAME = "ubi"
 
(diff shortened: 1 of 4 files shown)


unblob/handlers/filesystem/ubi.py
```diff
@@ -95,14 +97,26 @@ def calculate_chunk(self, file: File, start_offset: int) -> Optional[ValidChunk]
         )
 
 
+class UBIExtractor(Command):
```
kissgyorgy, line 100: What if we don't subclass `Command`, but treat ubireader as a library and copy the minimum required code from `scripts/ubireader_extract_images` here? This would be a little bit more work, but I suspect the output would be more stable over time as unblob would control it completely. It would also result in the final structure we want instead of moving things around and deleting files.

QKaiser: It's an easy fix to get closer to a PyPi release ((it's actually an internal proposal)) so that we can spend our time working on a UBI/UBIFS extractor using unblob's API and utilities later on rather than maintaining a standalone vendored script.

kissgyorgy: So you want to completely remove `ubi-reader` as a dependency long-term?

QKaiser: > So you want to completely remove `ubi-reader` as a dependency long-term?

Probably yeah.

QKaiser: > I'm not suggesting vendoring, I'm suggesting having our own siomplified CLI to the ubi-reader library instead of using ubireader_extract_images at all.

I looked into it and there's a problem with this approach. If ubi-reader encounter an error and that `settings.error_action` is set to `exit` or that the error level is considered `fatal`, ubi-reader just call `sys.exit(1)` and unblob hangs there.

See https://github.com/jrspruitt/ubi_reader/blob/master/ubireader/debug.py#L36

The advantage of calling the CLI is that it can fail or return some error without affecting unblob's processing.

kissgyorgy: Yes I seen that too, I would just put it in extractor folder and call it as a command, but if ubi-reader will be removed, doesn't really matter.

kissgyorgy (approved):

vlaci (approved):

PR #506, opened 2023-02-08
Add support for NETGEAR CHK format

nyuware:
Add support for the NETGEAR CHK format,

### DIFF
--- a/unblob/handlers/__init__.py
+++ b/unblob/handlers/__init__.py
@@ -1,5 +1,6 @@
 from ..models import Handlers
 from .archive import ar, arc, arj, cab, cpio, dmg, rar, sevenzip, stuffit, tar, zip
+from .archive.netgear import chk
 from .compression import (
     bzip2,
     compress,
@@ -46,6 +47,7 @@ BUILTIN_HANDLERS: Handlers = (
     ubi.UBIFSHandler,
     yaffs.YAFFSHandler,
     yaffs.YAFFS2Handler,
+    chk.NetgearCHKHandler,
     sparse.SparseHandler,
     ar.ARHandler,
     arc.ARCHandler,
--- a/tests/integration/archive/netgear/chk/__input__/sample.chk
+++ b/tests/integration/archive/netgear/chk/__input__/sample.chk
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:18e9a759fc06d2c99e531188b39f4482d06a7b772393e225fd06550fe16321ee
+size 86
--- a/tests/integration/archive/netgear/chk/__output__/sample.chk_extract/kernel
+++ b/tests/integration/archive/netgear/chk/__output__/sample.chk_extract/kernel
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:ec76142ac57dd39fad43929f8b58f3cb541e4f8a4d17c9f383a6f311047068e7
+size 14
--- a/tests/integration/archive/netgear/chk/__output__/sample.chk_extract/rootfs
+++ b/tests/integration/archive/netgear/chk/__output__/sample.chk_extract/rootfs
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:1056e7a94e821ab7ae3a47e925863cf8cdf977a8577c555fa3c80c354a717470
+size 14
--- a/unblob/handlers/archive/netgear/chk.py
+++ b/unblob/handlers/archive/netgear/chk.py
@@ -1,0 +1,80 @@
+import io
+from pathlib import Path
+from typing import Optional
+
+from dissect.cstruct import Instance
+from structlog import get_logger
+
+from unblob.extractor import carve_chunk_to_file
+from unblob.file_utils import Endian, File, StructParser
+from unblob.models import Chunk, Extractor, HexString, StructHandler, ValidChunk
+
+logger = get_logger()
+
+C_DEFINITIONS = r"""
+    #define FIXED_HEADER_LEN 40
+    typedef struct chk_header {
+        uint32 magic;           /* "2A 23 24 5E" */
+        uint32 header_len;
+        uint8  reserved[8];
+        uint32 kernel_chksum;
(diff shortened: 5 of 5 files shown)


unblob/handlers/filesystem/netgearchk.py
```diff
@@ -0,0 +1,76 @@
...
+            uint32 header_len;
+            uint8  reserved[8];
+            uint32 kernel_chksum;
+            uint32 rootfs_chksum;
+            uint32 kernel_len;
+            uint32 rootfs_len;
+            uint32 image_chksum;
+            uint32 header_chksum;
+            /* char board_id[total_header_len - header_len] upto MAX_BOARD_ID_LEN */
+            /* uint32 total_header_len; */
```
QKaiser, line 25: Turns out the `header_len` is the header size *including the variable length board_id*. So you can do something like this:

 ```suggestion
            char board_id[header_len-40];
```

You can even be more explicit with something along those lines (can be defined above the struct):

```c
#define FIXED_HEADER_LEN 40
```

And use it this way:

 ```c
char board_id[header_len-FIXED_HEADER_LEN];
```

File: unblob/handlers/filesystem/netgearchk.py [hunk 2]
```diff
@@ -0,0 +1,76 @@
...
+    HEADER_STRUCT = "chk_header_t"
+    EXTRACTOR = CHKExtractor()
+
+    def calculate_chunk(self, file: File, start_offset: int) -> Optional[ValidChunk]:
+        header = self.parse_header(file, endian=Endian.BIG)
+        header_len = len(header)
+        if header_len < header.header_len:
+            board_id = file.read(header.header_len - len(header))
+        else:
+            board_id = None
```
QKaiser, line 67: You don't need this if you implement the recommendation above about the header struct.

unblob/handlers/filesystem/netgearchk.py
```diff
@@ -0,0 +1,76 @@
...
+        logger.debug("CHK header", header=header, board_id=board_id)
```
QKaiser, line 68: This line is not required since `self.parse_header` logs the header structure in verbose mode (`-vvv`).

unblob/handlers/filesystem/netgearchk.py
```diff
@@ -0,0 +1,76 @@
...
+    NAME = "chk"
+
+    PATTERNS = [HexString("2a 23 24 5e")]
+
+    C_DEFINITIONS = CHK_HEADER
+    HEADER_STRUCT = "chk_header_t"
+    EXTRACTOR = CHKExtractor()
+
+    def calculate_chunk(self, file: File, start_offset: int) -> Optional[ValidChunk]:
+        header = self.parse_header(file, endian=Endian.BIG)
```
QKaiser, line 62: We usually recommend to add a header validation function to make sure we don't end up with false positives. A simple implementation woud look like this:

```python
from dissect.cstruct import Instance
# --snip---
def is_valid_header(header: Instance) -> bool:
    if header.header_len != len(header):
        return False
     return True
```

Let's look at it together tomorrow so we expand the validation.

QKaiser (changes requested):

QKaiser:
Some information about the checksum can be found at https://github.com/Getnear/R7800/blob/master/tools/firmware-utils/src/mkchkimg.c

Most interesting part:

```
struct ngr_checksum {
	uint32_t c0;
	uint32_t c1;
};

static inline void
netgear_checksum_init (struct ngr_checksum * c)
{
	c->c0 = c->c1 = 0;
}

static inline void
netgear_checksum_add (struct ngr_checksum * c, unsigned char * buf, size_t len)
{
	size_t i;

	for (i=0; i<len; i++) {
		c->c0 += buf[i] & 0xff;
		c->c1 += c->c0;
	}
}

static inline unsigned long
netgear_checksum_fini (struct ngr_checksum * c)
{
	uint32_t b, checksum;

	b = (c->c0 & 65535) + ((c->c0 >> 16) & 65535);
	c->c0 = ((b >> 16) + b) & 65535;
	b = (c->c1 & 65535) + ((c->c1 >> 16) & 65535);
	c->c1 = ((b >> 16) + b) & 65535;
	checksum = ((c->c1 << 16) | c->c0);
	return checksum;
}
```

Not sure it makes sense to implement the checksum or not.

unblob/handlers/__init__.py
```diff
@@ -27,6 +27,7 @@
     yaffs,
 )
 from .filesystem.android import sparse
+from .filesystem.netgear import chk
```
QKaiser, line 30: I would move it to `archive` rather than `filesystem`. CHK has no concept of filesystem.

unblob/handlers/filesystem/netgearchk.py
nyuware: We set a fix header len of 40, this way we could uncomment the board_id.
char board_id[header_len - FIXED_HEADER_LEN]; /* upto MAX_BOARD_ID_LEN */

QKaiser (changes requested):

unblob/handlers/filesystem/netgearchk.py
nyuware: This part has been removed since we find the board_id length dynamically now

unblob/handlers/filesystem/netgearchk.py
nyuware: Removed the logged.debug line as it's already in the `-vvv` verbose mode

QKaiser:
@nyuware you can create an artificial CHK file and place it under `tests/integration/archive/netgear/__input__` with the extracted content in `tests/integration/archive/netgear/__output__`. Otherwise tests won't pass.

unblob/handlers/filesystem/netgearchk.py
nyuware: Implemented the `is_valid_header` check aswell as a try expect to see if the `board_id` contained any character that were not `utf-8`

unblob/handlers/__init__.py
nyuware: I changed the location of to `netgear` under the folder `chk` to mark the difference with chk and trx

QKaiser (approved):

PR #507, opened 2023-02-08
fix github actions

QKaiser:
Preventive strike so that we don't end up with a broken build when Github pulls the plug for outdated actions.
    
References:   
- https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/
- https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

Other things performed by this MR:
- add github actions and github workflows validators to pre-commit
- add missing fields to github actions (identified by pre-commit)
- upgrade pre-commit dependencies

### DIFF
--- a/.github/workflows/build-nix.yml
+++ b/.github/workflows/build-nix.yml
@@ -8,15 +8,15 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2.4.0
+    - uses: actions/checkout@v3
     - name: Setup git lfs
       uses: ./.github/actions/setup-git-lfs
 
-    - uses: cachix/install-nix-action@v15
+    - uses: cachix/install-nix-action@v19
       with:
         extra_nix_config: |
           access-tokens = github.com=${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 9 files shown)


.github/workflows/build-nix.yml
```diff
@@ -8,15 +8,15 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2.4.0
+    - uses: actions/checkout@v3
```
kukovecz, line 11: Really good job updating all of our `workflows` to use newer actions, but you missed our `actions`. In the `.github/actions` folder there are still some `.yml` files which are not updated - containing things like `actions/cache@v2`, etc and these are also tringgering the `save-state` warning we are trying to get rid off here: https://github.com/onekey-sec/unblob/actions/runs/4125774689

QKaiser: Good catch ! Will take care of it.

QKaiser: Done.

kukovecz (approved):

PR #508, opened 2023-02-08
Add support for NETGEAR TRX format

nyuware:
Add support for the NETGEAR TRX format,

### DIFF
--- a/unblob/handlers/__init__.py
+++ b/unblob/handlers/__init__.py
@@ -1,6 +1,6 @@
 from ..models import Handlers
 from .archive import ar, arc, arj, cab, cpio, dmg, rar, sevenzip, stuffit, tar, zip
-from .archive.netgear import chk
+from .archive.netgear import chk, trx
 from .compression import (
     bzip2,
     compress,
@@ -48,6 +48,8 @@ BUILTIN_HANDLERS: Handlers = (
     yaffs.YAFFSHandler,
     yaffs.YAFFS2Handler,
     chk.NetgearCHKHandler,
+    trx.NetgearTRXv1Handler,
+    trx.NetgearTRXv2Handler,
     sparse.SparseHandler,
     ar.ARHandler,
     arc.ARCHandler,
--- a/unblob/handlers/archive/netgear/trx.py
+++ b/unblob/handlers/archive/netgear/trx.py
@@ -1,0 +1,114 @@
+# import binascii
+import binascii
+import io
+from pathlib import Path
+from typing import Optional
+
+from dissect.cstruct import Instance
+from structlog import get_logger
+
+from unblob.extractor import carve_chunk_to_file
+from unblob.file_utils import Endian, File, InvalidInputFormat, StructParser
+from unblob.models import Chunk, Extractor, HexString, StructHandler, ValidChunk
+
+logger = get_logger()
+CRC_CONTENT_OFFSET = 12
+
+TRX_V1_C_DEFINITION = r"""
+    typedef struct trx_header {
+        uint32 magic;              /* "HDR0" */
+        uint32 len;                /* header size + data */
+        uint32 crc32;              /* 32-bit CRC from flag_version to end of file */
+        uint16 flags;
+        uint16 version;
+        uint32 offsets[3];         /* Offsets of partitions from start of header */
+    } trx_header_t;
+"""
+
+TRX_V2_C_DEFINITION = r"""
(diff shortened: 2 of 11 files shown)


unblob/handlers/filesystem/netgeartrx.py
```diff
@@ -0,0 +1,70 @@
...
+    NAME = "trx"
+
+    PATTERNS = [HexString("48 44 52 30")]
+
+    C_DEFINITIONS = TRX_HEADER
+    HEADER_STRUCT = "trx_header_t"
+    EXTRACTOR = TRXExtractor()
+
+    def calculate_chunk(self, file: File, start_offset: int) -> Optional[ValidChunk]:
+        header = self.parse_header(file, endian=Endian.LITTLE)
```
QKaiser, line 65: We usually recommend to add a header validation function to make sure we don't end up with false positives. A simple implementation woud look like this:

```python
from dissect.cstruct import Instance
# --snip---
def is_valid_header(header: Instance) -> bool:
    if header.len > len(header):
        return False
     return True
```

I would also look into implementing a CRC32 check since the format implements it. You can look at https://github.com/onekey-sec/unblob/blob/main/unblob/handlers/filesystem/cramfs.py#L57 for an example.

QKaiser (changes requested):

unblob/handlers/filesystem/netgeartrx.py
```diff
@@ -0,0 +1,70 @@
...
+        logger.debug("TRX header", header=header)
```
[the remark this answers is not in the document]
nyuware: The debug line has been removed.

unblob/handlers/filesystem/netgeartrx.py
nyuware: Properly identifying if the CRC32 and header fields are valid is done.

QKaiser:
Add integration test files, squash your commits into a single one and then we can move on with this one.

unblob/handlers/__init__.py
```diff
@@ -48,6 +48,8 @@
     yaffs.YAFFSHandler,
     yaffs.YAFFS2Handler,
     chk.NetgearCHKHandler,
+    trx.NetGearTRXv1Handler,
```
QKaiser, line 51: `Netgear` rather than `NetGear`

unblob/handlers/__init__.py
```diff
@@ -48,6 +48,8 @@
...
+    trx.NetGearTRXv2Handler,
```
QKaiser, line 52: `Netgear` rather than `NetGear`

unblob/handlers/archive/netgear/trx.py
```diff
@@ -0,0 +1,115 @@
...
+class NetgearTRXBase(StructHandler):
+    HEADER_STRUCT = "trx_header_t"
+
+    def calculate_chunk(self, file: File, start_offset: int) -> Optional[ValidChunk]:
+        header = self.parse_header(file, endian=Endian.LITTLE)
+
+        if not self.is_valid_header(header):
+            raise InvalidInputFormat("Invalid TRX header.")
+
+        if header.version == 1 and not self._is_crc_valid(file, start_offset, header):
```
QKaiser, line 68: Investigate why the CRC32 does not work with version 2.

nyuware: The utility used to create the sample files created wrong headers, the code worked for both versions initially.

unblob/handlers/archive/netgear/trx.py
```diff
@@ -0,0 +1,115 @@
...
+    def is_valid_header(self, header: Instance) -> bool:
+        if header.len < len(header):
+            return False
+        return True
+
+    def _is_crc_valid(self, file: File, start_offset: int, header: Instance) -> bool:
+        file.seek(start_offset + CRC_CONTENT_OFFSET)
+        content = bytearray(file.read(header.len - CRC_CONTENT_OFFSET))
+        computed_crc = (binascii.crc32(content) ^ -1) & 0xFFFFFFFF
+        logger.info("Calcul CRC", computed_crc=computed_crc, header_crc=header.crc32)
```
QKaiser, line 84: This is a leftover from debug sessions, remove it.

QKaiser (changes requested):

QKaiser (approved):

PR #510, opened 2023-02-09
fix(handlers): improve LZMA handler to discard nonsense chunks

mucoze:
We've encountered chunks that are too small to be anything of interest and they are often filled with null bytes. 
We discard these nonsense chunks by size control.

### DIFF
--- a/unblob/handlers/compression/lzma.py
+++ b/unblob/handlers/compression/lzma.py
@@ -20,6 +20,9 @@ logger = get_logger()
 # 256GB
 MAX_UNCOMPRESSED_SIZE = 256 * 1024 * 1024 * 1024
 
+# This an arbitrary value
+MIN_COMPRESSED_SIZE = 256
+
 MIN_READ_RATIO = 0.1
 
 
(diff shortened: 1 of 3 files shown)


unblob/handlers/compression/lzma.py
```diff
@@ -47,19 +50,19 @@ def calculate_chunk(self, file: File, start_offset: int) -> Optional[ValidChunk]
...
-        if not (
-            uncompressed_size == 0xFFFFFFFFFFFFFFFF
-            or uncompressed_size < MAX_UNCOMPRESSED_SIZE
+        if (
+            not (
+                uncompressed_size == 0xFFFFFFFFFFFFFFFF
+                or uncompressed_size < MAX_UNCOMPRESSED_SIZE
+            )
+            or dictionary_size == 0
+            or (dictionary_size & (dictionary_size - 1)) != 0
```
QKaiser, line 65: I would keep this as two separate conditions since it helps reasoning about them and we can put dedicated comments on top. I agree with the move of `uncompressed_size` read, but not the split.

QKaiser (changes requested):

unblob/handlers/compression/lzma.py
mucoze: Actually, flake8 forced me to do this. Otherwise, calculate_chunk() becomes too complex to be able to pass the tests.

QKaiser: When flake8 reports such thing it's usually recommended to move the logic into a separated function.

```python
def is_valid_stream(dictionnary_size: int, uncompressed_size: int) -> bool:
    # dictionary size is non-zero (section 1.1.2 of format definition)
    # dictionary size is a power of two  (section 1.1.2 of format definition)
    if dictionary_size == 0 or (dictionary_size & (dictionary_size - 1)) != 0:
        return False
    # uncompressed size is either unknown (0xFFFFFFFFFFFFFFFF) or
    # smaller than 256GB  (section 1.1.3 of format definition)
    if not (
        uncompressed_size == 0xFFFFFFFFFFFFFFFF
        or uncompressed_size < MAX_UNCOMPRESSED_SIZE
    ):
        return False
    return True
```

QKaiser (commented):
I would add integration test files that cover the following paths:
- decompressed size smaller than compressed size
- compressed size smaller than `MIN_COMPRESSED_SIZE`

QKaiser:
I thought that since https://github.com/onekey-sec/unblob/pull/491 got merged we would not see empty extract directories. It's not the case since we're only cleaning these directories upon `ExtractError` being raised.

Two solutions: clean up empty directories upon `InvalidInputFormat` or raise `ExtractError` here. I'm inclined on implementing the first solution but we would need to check the impact on other handlers.

QKaiser (approved):

PR #512, opened 2023-02-15
Fix extraction of zipped up extractions

e3krisztian:
This is a bunch of small fixes, that have been triggered by anomalies detected when unblob-ing of an accidentally zipped up previous extraction.

While the anomalies appear only in very specific inputs, that do not occur naturally, the tool should discover, handle, and report the appearing problems.

### DIFF
--- a/unblob/processing.py
+++ b/unblob/processing.py
@@ -32,7 +32,6 @@ from .report import (
     ExtractDirectoryExistsReport,
     FileMagicReport,
     HashReport,
-    Report,
     StatReport,
     UnknownError,
 )
@@ -111,17 +110,21 @@ def process_file(
     if not input_path.is_file():
         raise ValueError("input_path is not a file", input_path)
 
-    errors = prepare_extract_dir(config, input_path)
+    extract_dir = config.get_extract_dir_for(input_path)
+    if config.force_extract and extract_dir.exists():
+        logger.info("Removing extract dir", path=extract_dir)
+        shutil.rmtree(extract_dir)
+
     if not prepare_report_file(config, report_file):
         logger.error(
             "File not processed, as report could not be written", file=input_path
         )
         return ProcessResult()
 
-    if errors:
-        process_result = ProcessResult([TaskResult(task, errors)])
-    else:
-        process_result = _process_task(config, task)
+    process_result = _process_task(config, task)
+
+    # ensure that the root extraction directory is created even for empty extractions
+    extract_dir.mkdir(parents=True, exist_ok=True)
 
     if report_file:
         write_json_report(report_file, process_result)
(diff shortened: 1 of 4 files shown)


unblob/handlers/filesystem/romfs.py
```diff
@@ -284,7 +284,7 @@ def create_inode(self, extract_root: Path, inode: FileHeader):
         elif inode.type == FS_TYPE.DIRECTORY:
             output_path.mkdir(mode=inode.mode, exist_ok=True)
         elif inode.type == FS_TYPE.FILE:
-            with output_path.open("wb") as f:
+            with output_path.open("xb") as f:
```
martonilles, line 287: just wondering why change it here?

also `output_path.write_bytes()` might be easier

unblob/processing.py
```diff
@@ -375,6 +375,13 @@ def _extract_chunk(self, file, chunk: ValidChunk):
             extract_dir = self.carve_dir / (inpath.name + self.config.extract_suffix)
             carved_path = inpath
 
+        # We only extract every blob once, it's a mistake to extract the same blob again
```
martonilles, line 378: the comment is miss-leading as there could be other reason why the directory exists:
- for some reason we extract the blob
- the content actually contains a directory with the same name as we would extract, in this case we would also skip extraction.

Unfortunately we do not know which is the above case, but we can report the error

unblob/__init__.py
```diff
@@ -0,0 +1,4 @@
+# get a traceback on crashes, as well as the responsible signal
+import faulthandler
+
+faulthandler.enable()
```
martonilles, line 4: I would rather put this into some init function as this could override the faulthandler setting of the main application when unblob is used as a library at import time. Maybe somehow make it configurable like logging?

Not support important, just wondering here.

martonilles (changes requested):

unblob/handlers/filesystem/romfs.py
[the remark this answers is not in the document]
e3krisztian: No, I just thought that while there was a bug that rewrote a file that was already `mmap`-ed, let's search for similar `open(.., "w` instances, that could be a potential problem, and this was the only other one.

TBH, I was not aware of the `"x"` mode before, I was happy to write just `"w"` everywhere, and thought this ignorance is more widespread, so let's fix it.

I do not think the original had any real problem. I just was trigger happy to make sure the same problem will not emerge in another spot. That being said, I have not do any extensive testing or looking at the source much to verify the `romfs` change.

As it is not a problem that emerged, and was flagged by both you and @martonilles below, I have reverted this change.

e3krisztian: Answered above, you are right, that `write_bytes()` would be a simpler equivalent to current, but my change was not compatible with it - it would have succeeded only if the file does not exist.

Anyway, this change was reverted, it is not worth the time.

unblob/processing.py
e3krisztian: This was copied over, but was a good point.

It was a comment that made me re-think if this is the only place we need this check, and eventually led to more work.

unblob/__init__.py
e3krisztian: I mostly agree with a YAGNI feeling: I have some doubts if anyone ever needs to improve upon it.

This is not important from the fix perspective, but helped a lot to investigate the issue of zombie processes and stalled unblob, and has 0 impact on run-time otherwise. So left it here to avoid the hunt without traceback next time.

Maybe I am simply ignorant (`faulthandler` was new to me) and it can be/is used in other context, then I would postpone the refactoring work, when it is needed, alternatively I can drop this hunk.

QKaiser: How about enabling it only if verbose level is `-vvv` ?

e3krisztian: I have dropped it, it was not important, the functionality is available by prefixing the `unblob` command with `PYTHONFAULTHANDLER=1` like this:

```sh
$ PYTHONFAULTHANDLER=1 unblob firmware.bin
```

QKaiser (approved):

QKaiser:
Did a final pass, looks good to me. I really like the overall simplification brought by 766f054b54da6ef45dc92efa58d2043c81766175

martonilles (approved):

PR #513, opened 2023-02-17
YAFFS extractor

qkaiser:
Initial objective was to get rid of `yaffshiv` as it was our last git based dependency blocking us from publishing unblob to PyPi.
    
This change introduce an internal extractor for YAFFS and YAFFS2 formats, completely rewritten from scratch in order to provide proper extraction capabilities.
    
The extractor relies on `treelib` to provide tree structures in Python.
    
That structure is used to load the filesystem representation in memory and perform depth-first walks of it in order to re-create the filesystem on disk during extraction.


QKaiser:
In YAFFS and YAFFS2, each node has an object identifier stored within the spare section that follows the node first page:

```
----------------------------------------------------------------
page 0 (yaffs header of node 0)
----------------------------------------------------------------
spare 0 (yaffs spare structure of node 0)
----------------------------------------------------------------
page 1 (file content of node 0 if it's a file)
----------------------------------------------------------------
spare 1
----------------------------------------------------------------
page n
----------------------------------------------------------------
spare n
----------------------------------------------------------------
```

YAFFS and YAFFS2 store "tags" within that spare, using different structures depending on whether you're using v1 (YAFFS) or v2 (YAFFS2). This is where the object_id of a node is stored. The object_id is a unique identifier that is used to reconstruct the filesystem tree since a YAFFS node has a `parent_obj_id` referencing the `object_id` of their parent, with nodes at the root having a `parent_obj_id` set to 1.

The dirty little secret of `yaffshiv` is that it never properly parsed these tags and always considered the sparse section to be made of two unsigned 32bit integers: `chunk_id` and `obj_id`. The reason it "worked" is that yaffshiv simply considered that nodes without a valid parent node should be linked to the current root.

So if you build a YAFFS image from content like this:

```
.
├── dir
│   ├── apple.txt
│   ├── cherry.txt
```

yaffshiv will actually export it this way:

```
.
├── apple.txt
├── cherry.txt
└── dir
```

The tags structure is way more complex than just two ints, as demonstrated below:

```c
typedef struct yaffs2_packed_tags {
        uint32 seq_number;
        uint32 obj_id;
        uint32 chunk_id;
        uint32 n_bytes;
    }  yaffs2_packed_tags_t;

typedef struct yaffs1_packed_tags {
        uint32 chunk_id:20;
        uint32 serial_number:2;
        uint32 n_bytes:10;
        uint32 obj_id:18;
        uint32 ecc:12;
        uint32 deleted:1;
        uint32 unused_stuff:1;
        uint32 should_be_ff;
    } yaffs1_packed_tags_t;
```

// yes these are bitfields definition for v1 tags

The objective of this MR is therefore to move away from yaffshiv and use proper filesystem parsing and representations (this time using a tree structure).

Some limitations are still present in the current implementation (work in progress) and should be resolved before we merge it:

- [ ] heuristics to detect the presence of ECC layout are too weak
- [ ] YAFFSv1 spare tags are not handled
- [ ] YAFFSv1 integration tests files are missing
- [ ] YAFFS v2 and v1 integration tests with traversal payloads are missing

qkaiser:
Hi potential reader. This issue is not stale, I'm just making progress towards a valid YAFFS extractor that properly discards deleted chunks, get the last revision of any duplicated chunk, and overall follow YAFFS implementation guidelines.

Side note: our YAFFS2 handler never triggered because the pattern is wrong, but YAFFS handler took care of it anyway.

qkaiser:
Except for a small bug when parsing yaffs2 packed tags without ECC in big-endian, this code works with any combination of page size and spare size.

Some known limitations that will not be covered as part of this MR:

- YAFFS2 with page_size < 1024 bytes is not supported since it would require adding support for inband tags
- YAFFS2 shrink headers are not currently supported, should not bring up any issues on clean images
- YAFFS1 only supports page size of 512 bytes with spare size of 16 bytes, which is the original design

I need a fresh pair of eyes to go through everything and create a 100 comments long review :)

qkaiser:
You'll find almost everything about the format in https://yaffs.net/sites/default/files/downloads/HowYaffsWorks.pdf , some light reading of the yaffs source code may be required.

qkaiser:
To see it in operation against real-life filesystems, you can download https://dl.google.com/android/repository/sys-img/android/sysimg_mips-16_r04.zip which contains two YAFFS2 filesystems.

unblob/handlers/filesystem/yaffs/utils.py
```diff
@@ -0,0 +1,572 @@
...
+            logger.debug("creating file", file_path=out_path, _verbosity=3)
+            with out_path.open("wb") as f:
+                for chunk in self.get_file_bytes(entry):
+                    f.write(chunk)
+        elif entry.type == YAFFS_OBJECT_TYPE.SPECIAL and (os.geteuid() == 0):
+            logger.debug("creating special file", special_path=out_path, _verbosity=3)
+            os.mknod(out_path.as_posix(), entry.yst_mode, entry.yst_rdev)
+        elif entry.type == YAFFS_OBJECT_TYPE.SYMLINK:
+            logger.debug("creating symlink", file_path=out_path, _verbosity=3)
+            out_path.symlink_to(Path(entry.alias))
```
martonilles, line 565: can here be a traversal attempt?

unblob/handlers/filesystem/yaffs/utils.py
```diff
@@ -0,0 +1,572 @@
...
+        elif entry.type == YAFFS_OBJECT_TYPE.HARDLINK:
+            logger.debug("creating hardlink", file_path=out_path, _verbosity=3)
+            src_entry = self.file_entries[entry.equiv_id]
+            src_path = self.resolve_path(src_entry)
+            src_path.link_to(out_path)
```
martonilles, line 570: can here be a traversal attempt?

unblob/handlers/filesystem/yaffs/utils.py
qkaiser: Indeed, in a similar way to https://bugzilla.redhat.com/show_bug.cgi?id=CVE-2021-41072

unblob/handlers/filesystem/yaffs/utils.py
qkaiser: Indeed, in a similar way to https://bugzilla.redhat.com/show_bug.cgi?id=CVE-2021-41072

kukovecz (commented):
I am really-really not familiar with the insights of YAFFS format, so consider my review more of a developer point of view rather than format specific ...

Beside the few things I mentioned I would draw your attention towards the issues raised by pyright. Among the annoying type annotation things, there are real findings there (possibly unbound variable, double declaration).

unblob/handlers/filesystem/yaffs/utils.py
qkaiser: Fixed

unblob/handlers/filesystem/yaffs/utils.py
qkaiser: Fixed

unblob/handlers/__init__.py
```diff
@@ -31,6 +31,7 @@
     yaffs,
 )
 from .filesystem.android import sparse
+from .filesystem.yaffs import yaffs1, yaffs2
```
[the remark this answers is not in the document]
qkaiser: Fixed

unblob/handlers/filesystem/yaffs/yaffs1.py
```diff
@@ -0,0 +1,145 @@
...
+                )
+                self.insert_entry(entry)
+            else:
+                # this is a data chunk, so we add it to our object
+                self.get_entry(chunk.object_id).chunks.append(chunk)
+        self.end_offset = self.file.tell()
+
+
+class YAFFS1Extractor(Extractor):
+    BIG_ENDIAN_MAGIC = 0x00_00_00_03
```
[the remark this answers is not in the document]
qkaiser: Moved

unblob/handlers/filesystem/yaffs/yaffs2.py
```diff
@@ -0,0 +1,172 @@
...
+
+
+class YAFFS2Handler(StructHandler):
+    NAME = "yaffs2"
+
+    C_DEFINITIONS = C_DEFINITIONS
+
+    HEADER_STRUCT = "yaffs_obj_hdr_t"
+
+    # TODO: YAFFS1 and YAFFS2 signatures are the same, we need a way to differentiate between
```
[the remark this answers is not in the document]
qkaiser: Removed the `TODO` and made the patterns more explicit. It's not possible they both match on the same chunk.

unblob/handlers/filesystem/yaffs/utils.py
```diff
@@ -0,0 +1,572 @@
...
+
+        if entry.type == YAFFS_OBJECT_TYPE.DIRECTORY:
+            logger.debug("creating directory", dir_path=out_path, _verbosity=3)
+            out_path.mkdir(exist_ok=True)
+        elif entry.type == YAFFS_OBJECT_TYPE.FILE:
+            logger.debug("creating file", file_path=out_path, _verbosity=3)
+            with out_path.open("wb") as f:
+                for chunk in self.get_file_bytes(entry):
+                    f.write(chunk)
+        elif entry.type == YAFFS_OBJECT_TYPE.SPECIAL and (os.geteuid() == 0):
```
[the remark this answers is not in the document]
qkaiser: Agreed. Used your code in there :)

unblob/handlers/filesystem/yaffs/yaffs2.py
```diff
@@ -0,0 +1,172 @@
...
+        HexString(
+            "01 00 00 00 01 00 00 00 ff ff // LE, Look for YAFFS_OBJECT_TYPE_DIRECTORY with a null name"
+        ),
+        HexString("00 00 00 03 00 00 00 01 ff ff // BE"),
+        HexString("00 00 00 01 00 00 00 01 ff ff // BE"),
+    ]
+
+    BIG_ENDIAN_MAGIC = 0x00_00_00_01
+
+    EXTRACTOR = YAFFS2Extractor(DEFAULT_CONFIG)
```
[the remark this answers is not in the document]
qkaiser: That makes sense. I set it to `None`.

Note that I'm not super happy with the dynamic modification of a class attribute, but that's the only way I could envision to not do the config identification twice.

unblob/handlers/filesystem/yaffs/utils.py
```diff
@@ -0,0 +1,572 @@
...
+    parent_obj_id: int
+    sum_no_longer_used: int
+    name: str
+    alias: str
+    file_size: int
+    start_offset: int
+    chunks: List[YAFFSChunk]
+
+    @property
+    def chunks(self):
```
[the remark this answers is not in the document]
qkaiser: What do you recommend ?

qkaiser:
@vlaci Nix is not happy about the treelib dependency and I don't know what to do. Can you have a look ? Nothing urgent :)

unblob/handlers/filesystem/yaffs/utils.py
```diff
@@ -0,0 +1,574 @@
...
+        self.file_entries.show()
+        for entry in [
+            self.file_entries.get_node(node).data
+            for node in self.file_entries.expand_tree(mode=Tree.DEPTH)
+        ]:
+            if entry is None:
+                continue
+            self.extract_entry(entry, outdir)
+
+    def extract_entry(self, entry: YAFFS2Entry, outdir: Path):  # noqa: C901
```
martonilles, line 519: is entry always YAFFS2Entry, what if we extract v1?

unblob/handlers/filesystem/yaffs/utils.py
```diff
@@ -0,0 +1,574 @@
...
+                os.mknod(out_path.as_posix(), entry.yst_mode, entry.yst_rdev)
+            else:
+                logger.warning(
+                    "creating special files requires elevated privileges, skipping.",
+                    path=out_path,
+                    yst_mode=entry.yst_mode,
+                    yst_rdev=entry.yst_rdev,
+                )
+        elif entry.object_type == YaffsObjectType.SYMLINK:
+            if not is_safe_path(outdir, out_path / Path(entry.alias)):
```
martonilles, line 552: wondering now, that this would not be reported as MaliciousSymlinkRemoved report, while other extractors would

Also the above path-traversal is just logger, but not reported..

unblob/handlers/filesystem/yaffs/utils.py
```diff
@@ -0,0 +1,574 @@
...
+        for chunk in entry.get_chunks():
+            self.file.seek(
+                entry.start_offset
+                + (
+                    (chunk.chunk_id - 1)
+                    * (self.config.page_size + self.config.spare_size)
+                ),
+                io.SEEK_SET,
+            )
+            yield self.file.read(chunk.byte_count)
```
martonilles, line 507: instead of seek+read, we can return file[start:end] as a byteview, though not sure that is actually faster

unblob/handlers/filesystem/yaffs/utils.py
```diff
@@ -0,0 +1,574 @@
...
+                    "creating special file", special_path=out_path, _verbosity=3
+                )
+                os.mknod(out_path.as_posix(), entry.yst_mode, entry.yst_rdev)
+            else:
+                logger.warning(
+                    "creating special files requires elevated privileges, skipping.",
+                    path=out_path,
+                    yst_mode=entry.yst_mode,
+                    yst_rdev=entry.yst_rdev,
+                )
```
martonilles, line 550: is it v2 only?

unblob/handlers/filesystem/yaffs/utils.py
martonilles, line 519: do we want to set owner/permission for other v2 objects besides nodes?

unblob/handlers/filesystem/yaffs/utils.py
```diff
@@ -0,0 +1,574 @@
...
+    shrink_size: int
+    top_level: int
+
+
+@attr.define(kw_only=True)
+class YAFFS2Entry(YAFFSEntry):
+    chksum: int
+    yst_mode: int
+    yst_uid: int
+    yst_gid: int
```
martonilles, line 268: why are these v2 specific, when v1 also has mos of these fields?

unblob/handlers/filesystem/yaffs/utils.py
```diff
@@ -0,0 +1,574 @@
...
+
+        # Since we're scanning backwards, the most recently written - and thus current - chunk
+        # matching an obj_id:chunk_id pair will be encountered first and all subsequent matching chunks must be obsolete and treated as deleted.
+
+        # note: there is no deletion marker in YAFFS2
+
+        for _, chunks in itertools.groupby(
+            sorted(self.chunks, key=lambda chunk: chunk.chunk_id)
+        ):
+            yield max(chunks, key=lambda chunk: chunk.seq_number)
```
martonilles, line 301: seq number can loop, 11 -> 00 (not sure how often happens, but could (https://yaffs.net/documents/how-yaffs-works chapter 9)

unblob/handlers/filesystem/yaffs/yaffs2.py
```diff
@@ -0,0 +1,172 @@
...
+                # this is a data chunk, so we add it to our object
+                entry = self.get_entry(chunk.object_id)
+                # can happen during bruteforcing
+                if entry is not None:
+                    entry.chunks.append(chunk)
+                else:
+                    break
+
+        self.end_offset = self.file.tell()
+        return count
```
martonilles, line 124: why return count? (only for v2)

unblob/handlers/filesystem/yaffs/yaffs2.py
```diff
@@ -0,0 +1,172 @@
...
+        )
+
+        return YAFFS2Chunk(
+            chunk_id=yaffs2_packed_tags.chunk_id,
+            seq_number=yaffs2_packed_tags.seq_number,
+            byte_count=yaffs2_packed_tags.byte_count,
+            object_id=yaffs2_packed_tags.object_id,
+        )
+
+    def parse(self):
```
martonilles, line 48: most of parser is the same for v1 and v2, except the entry creation

martonilles, line 48: when called from the Handler, we do not need to store the entries and chunks

unblob/handlers/filesystem/yaffs/utils.py
```diff
@@ -0,0 +1,574 @@
...
+        try:
+            self.file_entries.create_node(
+                entry.object_id,
+                entry.object_id,
+                data=entry,
+                parent=entry.parent_obj_id,
+            )
+        except NodeIDAbsentError:
+            pass
+        except DuplicatedNodeIdError:
```
martonilles, line 478: wouldn't we want the latest object_id? why ignore it, instead of overwrite?

unblob/handlers/filesystem/yaffs/utils.py
```diff
@@ -0,0 +1,574 @@
...
+
+    def insert_entry(self, entry: YAFFSEntry):
+        try:
+            self.file_entries.create_node(
+                entry.object_id,
+                entry.object_id,
+                data=entry,
+                parent=entry.parent_obj_id,
+            )
+        except NodeIDAbsentError:
```
martonilles, line 476: how could this happen, why ignore it?

unblob/handlers/filesystem/yaffs/utils.py
```diff
@@ -0,0 +1,574 @@
...
+
+# 00 00 03 ff ff ff 00 00  95 aa a7 40 3f 56 aa ab
+# 00 00 03 ff ff ff 00 40  0c ff 0f 40 83 56 aa ab
+# 00 00 03 ff ff ff 00 40  00 ff f3 80 bf fc ff 03
+
+
+class YAFFSParser:
+    def __init__(self, file: File, config: Optional[YAFFSConfig] = None):
+        self.file_entries = Tree()
+        self.file_entries.create_node(ROOT, 1)
```
martonilles, line 336: isn't it v1 specific? Also using a global singleton wouldn't be a problem when multiple yffs identified in one unblob run?

unblob/handlers/filesystem/yaffs/utils.py
```diff
@@ -0,0 +1,574 @@
...
+    if header.type > 5:
+        return False
+    if header.sum_no_longer_used != 0xFFFF:
+        return False
+    return True
+
+
+@attr.define
+class YAFFSEntry:
+    object_type: int
```
martonilles, line 204: shouldn't this bea  YaffsObjectType

unblob/handlers/filesystem/yaffs/utils.py
```diff
@@ -0,0 +1,574 @@
...
+            raise InvalidInputFormat("Can't find YAFFS config through bruteforce.")
+
+        return config
+
+    def auto_detect(self) -> YAFFSConfig:
+        """Auto-detect page_size, spare_size, and ECC using known signatures."""
+        config = None
+        page_size = 0
+        for page_size in VALID_PAGE_SIZES:
+            if self.file[page_size:].startswith(SPARE_START_LITTLE_ENDIAN_ECC):
```
martonilles, line 388: `self.file[page_size:]` could be very big and this result ina  copy (I think) we should specify and end offset as wll

unblob/handlers/filesystem/yaffs/utils.py
```diff
@@ -0,0 +1,574 @@
...
+                )
+                break
+            if self.file[page_size:].startswith(SPARE_START_BIG_ENDIAN_NO_ECC):
+                config = YAFFSConfig(
+                    endianness=Endian.BIG, page_size=page_size, ecc=False, spare_size=-1
+                )
+                break
+
+        if config is None:
+            return self.bruteforce()
```
martonilles, line 416: when can bruteforce find something, when it checks the same VALID_PAGE_SIZES?

unblob/handlers/filesystem/yaffs/utils.py
```diff
@@ -0,0 +1,574 @@
...
+                + (
+                    (chunk.chunk_id - 1)
+                    * (self.config.page_size + self.config.spare_size)
+                ),
+                io.SEEK_SET,
+            )
+            yield self.file.read(chunk.byte_count)
+
+    def extract(self, outdir: Path):
+        self.file_entries.show()
```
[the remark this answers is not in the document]
qkaiser: No, it's a left over from debug sessions :)

unblob/handlers/filesystem/yaffs/utils.py
qkaiser: yeah it's faster, moved to this approach

unblob/handlers/filesystem/yaffs/utils.py
qkaiser: It can't happen, I removed that code. I also fixed my code around tree handling so that we no longer need to handle these errors. If they are raised, something is wrong and it's legit to raise an `InvalidInputFormat`.

unblob/handlers/filesystem/yaffs/utils.py
qkaiser: We need it for version 2 since they don't store an empty root node to gain space (as opposed to version 1). So we create this root node explicitly.

This has been moved to the specific `YAFFS2Parser`. I'm now using a local variable rather than a global singleton so we should be safe in terms of sharing objects.

unblob/handlers/filesystem/yaffs/utils.py
qkaiser: Indeed. Now only checking as many bytes as required.

unblob/handlers/filesystem/yaffs/utils.py
qkaiser: I got rid of the bruteforce approach. It is not required to have it.

qkaiser:
@martonilles your comments made me realize there was an overlap between YAFFSv2 and YAFFSv1 handler due to their patterns. I decided to move everything into a single yaffs handler that decides which version handler to use based on the presence of the directory object with a empty name (only present in version 1).

I cleaned up my code and re-organize the file to have constants, classes, and utility functions well ordered. Next step is to fix your last comments and see how I can limit code duplication between `YAFFSParser`s implementations.

unblob/handlers/filesystem/yaffs/utils.py
qkaiser: This is something that must absolutely be handled within the extractor. If we don't we're vulnerable to this attack:

1. create a symlink entry linking outside (`/home/extract/a -> /etc/passd`)
2. create a file entry with the same name (`/home/extract/a`)

When the file entry gets extracted, it will write out to the symlink target.

qkaiser: This is probably something we should look for in our other handlers by the way.

unblob/handlers/filesystem/yaffs/utils.py
qkaiser: YAFFSv1 does not support special files and if I set this type to `YAFFSEntry` pyright complaints that special files fields (e.g. `yst_mode`) does not exist.

Could be solved by a better class design for entries.

unblob/handlers/filesystem/yaffs/utils.py
qkaiser: TODO @qkaiser create a ticket for this

unblob/handlers/filesystem/yaffs/utils.py
qkaiser: These fields have been generalized into the `YAFFSEntry` class, they're no longer v2 specific.

unblob/handlers/filesystem/yaffs/utils.py
qkaiser: I'm now using this construct:

```
# serial is a 2 bit, this function works since there's always at most
# two chunks with the same chunk_id at any given time
yield max(chunks, key=lambda chunk: ((chunk.serial + 1) & 3))
```

unblob/handlers/filesystem/yaffs/yaffs2.py
qkaiser: Indeed. Generalized in `YAFFSParser` that rely on specialized functions such as `build_chunk` and `build_entry`.

unblob/handlers/filesystem/yaffs/utils.py
qkaiser: No because it breaks extraction if we're no longer allowed to write to those directories :)

unblob/handlers/filesystem/yaffs/yaffs2.py
qkaiser: Fixed.

qkaiser:
I need to fix the YAFFSv1 detection heurisitics because I have a YAFFS2 filesystem with an empty node at the beginning here:

```
00000000  03 00 00 00 01 00 00 00  ff ff 00 00 00 00 00 00  |................|
00000010  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
*
00000100  00 00 00 00 00 00 00 00  00 ff ff ff ed 41 00 00  |.............A..|
00000110  00 00 00 00 00 00 00 00  c8 70 01 51 3b 70 01 51  |.........p.Q;p.Q|
00000120  3b 70 01 51 ff ff ff ff  ff ff ff ff ff ff ff ff  |;p.Q............|
00000130  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
*
000001c0  ff ff ff ff ff ff ff ff  ff ff ff ff 00 00 00 00  |................|
000001d0  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
*
00000800  00 10 00 00 01 00 00 00  00 00 00 00 ff ff 00 00  |................|
00000810  30 ff ff ff 05 00 00 00  05 00 00 00 ff ff ff ff  |0...............|
00000820  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
*
00000840  03 00 00 00 01 00 00 00  ff ff 62 69 6e 00 00 00  |..........bin...|
00000850  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
*
00000940  00 00 00 00 00 00 00 00  00 ff ff ff ed 41 00 00  |.............A..|
00000950  00 00 00 00 d0 07 00 00  c8 70 01 51 71 6f 01 51  |.........p.Qqo.Q|
00000960  71 6f 01 51 ff ff ff ff  ff ff ff ff ff ff ff ff  |qo.Q............|
00000970  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
```

Probably better to look at the parse data, see if it makes sense for the fixed page size / spare size of YAFFS1.

unblob/handlers/filesystem/yaffs/utils.py
qkaiser: Done. https://github.com/onekey-sec/unblob/issues/550

martonilles (approved):

PR #514, opened 2023-02-17
Add support for Engenius encrypted firmwares.

qkaiser:
Initial work on encryption courtesy of @ryancdotorg . I reversed parts of the header structure and ported the code from `decrypt.py` available at https://github.com/ryancdotorg/enfringement.

The signature is quite unique and header validation should ensure we don't yield false positives.

Need to find a fast and memory efficient way of performing the XOR operation as the whole decrypted file is stored in memory prior to being written to the output file right now.

Not sure integration tests files are in order, but I'd recommend adding coverage for the XOR utility function.

### DIFF
--- a/unblob/handlers/archive/engeniustech/engenius.py
+++ b/unblob/handlers/archive/engeniustech/engenius.py
@@ -1,0 +1,102 @@
+from pathlib import Path
+from typing import Optional
+
+from dissect.cstruct import Instance
+from structlog import get_logger
+
+from unblob.file_utils import Endian, File, InvalidInputFormat, StructParser
+from unblob.models import Extractor, HexString, StructHandler, ValidChunk
+
+logger = get_logger()
+
+C_DEFINITIONS = r"""
+    typedef struct engenius_header {
+        uint32 unknown_1;
+        uint32 vendor_id;
+        uint32 product_id;
+        char version[20];
+        uint32 length;
+        uint32 unknown_2;
+        char checksum[16];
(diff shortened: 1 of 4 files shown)


QKaiser:
Plenty of samples available at https://www.engeniustech.com/engenius-firmware-updates.html

ryancdotorg:
I believe immediately after the 0x12345678 magic is a null terminated string (or maybe three characters) indicating what regulatory domain the firmware is for, but everything that has it is "all".

Also, per some scripts in the unpacked firmware (not that they're particularly important):

u16 @ offset 6: vendor id
u16 @ offset 10: product id
u16 @ offset 106: major version
u16 @ offset 110: minor version

mucoze (approved):

unblob/handlers/archive/engeniustech/engenius.py
```diff
@@ -0,0 +1,102 @@
...
+
+    PATTERNS = [HexString("12 34 56 78 61 6c 6c")]
+
+    C_DEFINITIONS = C_DEFINITIONS
+    HEADER_STRUCT = "engenius_header_t"
+    EXTRACTOR = EngeniusExtractor()
+    PATTERN_MATCH_OFFSET = -0x5C
+
+    def is_valid_header(self, header: Instance) -> bool:
+        if header.length <= len(header):
```
martonilles, line 85: as the header.length does not include the length of the header, not sure about this check.

PR #517, opened 2023-02-20
feat(handler): Add support for encrpted_img format

nyuware:
More information can be found here : [Decryption Keys for D-Link](https://onekey.com/blog/extracting-decryption-keys-dlink/) as well as here : [Breaking the D-Link DIR3060 Firmware Encryption](https://0x00sec.org/t/breaking-the-d-link-dir3060-firmware-encryption-recon-part-1/21943)

The firmware is encrypted in AES in CBC mode, the KEY and IV are in the code and the technical write-up on how to find them is in the first link of this PR.

The `def extract` can be improved.

For samples: [DIR-x1560](https://www.dlink.com/en/products/dir-x1560-exo-ax-ax1500-wi-fi-6-router) or [DIR-x5460](https://www.dlink.com/en/products/dir-x5460-exo-ax-ax5400-wi-fi-6-router)

This script also attempts to fix the corrupted UBI headers after the decryption of the firmware. (to be looked into)

I'm not really sure how I would create test files for this format.


QKaiser:
Please cleanup your commits. 7317393249e50dd32a2034f9f308619de468b7b1 should be dropped from here, 1e56289ae1f779dcc5fd9ecd71d109b3c5f40c33, and 3dae67a5d1568b57b5f0abb517e9d51f6f674f9b can be fixed up into 3378ee8d94732bae7078b4783887a40654031d85

QKaiser (changes requested):
Integration test file does not reflect the format. It's a plaintext UBI image.

QKaiser (approved):

PR #518, opened 2023-02-21
Keep carved out ELF chunks as files

e3krisztian:
Process carved ELF chunks in two steps

Carved chunks are generally not important, so when they are fully
extracted, they can be deleted. ELF chunks are special, as they had
in general no extraction output. This is changed to have the extraction
output the ELF file itself by overriding chunk extraction.

Whole files have a single chunk that covers the file, they are also
processed differently from other chunks, as carving is skipped.

In the end ELF chunks are processed in two steps:
- carved files are simply copied, Handler defined extraction is not done
- non-carved whole files (copies of carved ELF chunks made previously
  are whole files now) are extracted as defined by the Handler


unblob/processing.py
```diff
@@ -326,6 +327,8 @@ def process(self):
             )
             outer_chunks = remove_inner_chunks(all_chunks)
             unknown_chunks = calculate_unknown_chunks(outer_chunks, self.size)
+            patch_chunks(outer_chunks, host_file=self.task.path)
+            patch_chunks(unknown_chunks, host_file=self.task.path)
```
martonilles, line 331: just wondering why can not we do this "patching" in `_hyperscan_match` after the `_calculate_chunk` call?

We have the file and also the file size there available, we can maybe even use it to simplify the check on chunk till end of file

unblob/models.py
```diff
@@ -63,6 +65,13 @@ def size(self) -> int:
     def range_hex(self) -> str:
         return f"0x{self.start_offset:x}-0x{self.end_offset:x}"
 
+    @property
+    def is_whole_file(self):
+        assert self.host_file
+        return (
+            self.start_offset == 0 and self.end_offset == self.host_file.stat().st_size
```
martonilles, line 72: can't we use the `size()` call to get the file size? Our `File` has a size method which also works for the `from_bytes()` setup as well.

Also it might make sense to cache this, instead of calling stat() multiple times

unblob/models.py
```diff
@@ -47,6 +47,8 @@ class Chunk:
 
     id: str = attr.field(factory=new_id)
 
+    host_file: Optional[Path] = None
```
martonilles, line 50: I would just use `file` what is `host` here?

unblob/handlers/executable/elf.py
```diff
@@ -26,6 +28,27 @@
...
+    def extract(self, inpath: Path, outdir: Path):
+        # ELF file extraction is special in that in the general case no new files are extracted, thus
+        # when we want to clean up all carves to save place, carved ELF files would be deleted as well,
+        # however we want to keep carved out ELF files, as they are the interesting stuff!
+        if self.is_whole_file:
+            # whole file chunks are not carved, so we can do normal extraction
+            super().extract(inpath, outdir)
+        else:
+            # make a copy, and let the carved chunk be deleted
+            outdir.mkdir(parents=True, exist_ok=False)
```
martonilles, line 42: wondering if we can simplify it even further, considering the options:

- 1: simple ELF file, whole file -> nothing to do, we already have the ELF file
- 2: simple ELF file, not whole file -> need to copy the carve to have an ELF file
- 3: kernel ELF part, whole file -> need to extract the initramfs part, no need to do anything with the ELF file
- 4 kernel ELF part, not whole file -> need to extract the initramfs, also no need to anything with the ELF part as the kernel image might contain an ELF part but we already have the kernel image as a parent file

martonilles (changes requested):

unblob/models.py
e3krisztian: `parent_file` it will be.

unblob/processing.py
e3krisztian: done

unblob/models.py
e3krisztian: `stat` would have been called at most twice for ELF files, and only once for all the rest, so it was not a really expensive thing, but changing from `Path` to `File` made it unnecessary.

----

One drawback of using `File`, is that with `Path` the option to process chunks in parallel was there, with `File` it is not: `File` objects will not travel through process boundaries.

unblob/processing.py
e3krisztian: After it was done, I took a second look, and have decided to revert it:

The patching logic was spread over multiple places (the ones you proposed + processing's `calculate_unknown_chunks`), and even existing tests had to be modified (`test_search_chunks` + `test_calculate_unknown_chunks`).

One option would be to simplify away patching unknown chunks, but I do not like that.  I think the parent file belongs to the chunk, unknown chunks included.

----

The current patching is also easier to change if we ever want to go with `Path`-s and parallel chunk processing.

e3krisztian: https://github.com/onekey-sec/unblob/commit/af050d646929d393e6b9c8e1413581e70a8e37eb

unblob/handlers/executable/elf.py
e3krisztian: I do not get why the last item is valid, and without that one, the first 3 describes the current MR behavior.

One thing I have in mind is a disk image which is not of a known filesystem format or damaged.
While we can carve the ELF kernel, and extract the initramfs, the kernel file itself will not be kept.

e3krisztian: OK, so the kernel is not an ELF by itself, just part of it, and that part is not that important by itself, just the initramfs, so the ELF can be deleted - if needed the chunk information and the kernel image is there.

unblob/models.py
e3krisztian: File is used now.

unblob/models.py
e3krisztian: `file` is the name

unblob/handlers/executable/elf.py
```diff
@@ -43,6 +43,7 @@ def extract(self, inpath: Path, outdir: Path):
         is_kernel = (
             elf.header.file_type == lief.ELF.E_TYPE.EXECUTABLE
             and elf.has_section(KERNEL_SYMBOLS_SECTION)
+            # TODO: this is incomplete: __ksymtab is only present if "loadable module support" is enabled!!!
```
QKaiser, line 46: Why look up for `__ksymtab` (KERNEL_SYMBOLS_SECTION) when we could just look for `.init.data` (KERNEL_INIT_DATA_SECTION) ? This way we also cover cases when "loadable module support" is not enabled, and we can get rid of a superfluous if on line 63 if we do the check here.

unblob/handlers/executable/elf.py
```diff
@@ -26,6 +28,27 @@
...
+        else:
+            # make a copy, and let the carved chunk be deleted
+            outdir.mkdir(parents=True, exist_ok=False)
+            shutil.copy2(inpath, outdir)
+            # more work will be done, when outdir is picked up by processing,
+            # and the ELF file is processed as a whole file.
+            # As a performance side effect, ELF files will be searched for chunks twice.
+            # Even though the second chunk search one is short-circuited,
+            # because the ELF handler will recognize it as a whole file
+            # other handlers might burn some cycles on the file as well.
```
QKaiser, line 59: Theoretically this is what happens with any plaintext structure like extfs or tar, the container is scanned (firs scan), files are extracted, then files are scanned (second scan). Usually optimized by streaming mode and skipping candidates within a valid chunk.

unblob/processing.py
```diff
@@ -417,6 +418,12 @@ def _extract_chunk(self, file, chunk: ValidChunk):  # noqa: C901
             )
 
 
+def patch_chunks(chunks: Sequence[Chunk], file: File):
```
QKaiser, line 421: `patch_chunks` is too generic as a function name, I would use something like `assign_file_to_chunks`

unblob/handlers/executable/elf.py
e3krisztian: Well, this is different here, as ELF file plays the role of extfs/tar in your example, yet the ELF file is scanned twice, once as a chunk, then as a whole file:

- ELF chunk is recognized by scanning a file
- chunk is extracted, and processed by ELF handler, which just copies it to the extraction directory
- extraction directory is added as sub-task
- eventually the copied ELF file is scanned again without knowing it is an ELF file
- the processing of whole file non-kernel ELF files is just a no-op, so nothing is done

This is different from how an extfs chunk is handled, as that chunk is extracted in one step.
Actually this difference is the whole reason behind this MR: while throwing away all chunks, we want to keep extracted ELF chunks for further analysis.

e3krisztian:
This is the problem with the kernel initramfs extraction: we want to calculate back from the pointer of `.init.ramfs.info` the `__initramfs_start`.  The structure is embedded in the kernel like this (include/asm-generic/vmlinux.lds.h):
```
     . = ALIGN(4);                           \
     __initramfs_start = .;                      \
     KEEP(*(.init.ramfs))                        \
     . = ALIGN(8);                           \
     KEEP(*(.init.ramfs.info))
```
from `.init.ramfs.info` we have the `.init.ramfs` size.
we know that it is aligned to 8 bytes, so there could be 0-7 padding bytes.
the problem is, that the *start* of initramfs is aligned to 4 bytes, so we can have 2 valid choices for the padding, depending on the start position being also 8 aligned or not (we do not know what structures are before): 0 and 4, 1 and 5, 2 and 6, 3 and 7.

the kernels I test with has the same `initramfs.cpio.gz` embedded (initramfs_size = 366109), while the padding in one case is 3, while in the other is 7. To further complicate the issue, this initramfs also ends with a 0 byte, so the 0 bytes before the size are 4 and 8 respectively.

e3krisztian:
The offset/padding calculation ambiguity is resolved by potentially extracting a bigger than initramfs chunk. The extra is 4 bytes, which leads to an extra unknown chunk.

unblob/processing.py
e3krisztian: done

unblob/handlers/executable/elf.py
e3krisztian: good catch! changed.

qkaiser (approved):

martonilles (approved):

PR #525, opened 2023-03-01
Fix CI issues

vlaci:
- Pin nix version to work-around regression, it is introduced by Nix 14.0 https://github.com/cachix/cachix-action/issues/138
- Ignore newly reported missing import when rust binding is unavailable


vlaci:
I have no idea, why the change happened in Pyright

qkaiser (approved):

qkaiser:
Merging, but I don't like that pyright started behaving differently out of the blue.

PR #529, opened 2023-03-02
feat(handler): Add support for XIAOMI HDR1 format

nyuware:
Lots of useful information here : https://lxr.openwrt.org/source/firmware-utils/src/xiaomifw.c
Xiaomi is transitioning to HDR2, info can be found here : https://ghostbin.me/6368ed31c0332

### DIFF
--- a/unblob/handlers/archive/xiaomi/hdr.py
+++ b/unblob/handlers/archive/xiaomi/hdr.py
@@ -1,0 +1,190 @@
+import binascii
+import io
+from pathlib import Path
+from typing import Iterable, Optional, Tuple
+
+from dissect.cstruct import Instance
+from structlog import get_logger
+
+from unblob.extractor import carve_chunk_to_file, is_safe_path
+from unblob.file_utils import File, InvalidInputFormat, iterate_file, snull
+from unblob.models import (
+    Chunk,
+    Endian,
+    Extractor,
+    HexString,
+    StructHandler,
+    StructParser,
+    ValidChunk,
+)
+
+logger = get_logger()
+CRC_CONTENT_OFFSET = 12  # The CRC32 value is located after 12 byte in the header
+SIGNATURE_LEN = 272  # signature_header_t contains 4 bytes of size + 12 bytes for padding x3 + 0x100 is 256 in decimal
+BLOB_MAGIC = 0x000000BEBA  # Blob header magic
+
+# https://lxr.openwrt.org/source/firmware-utils/src/xiaomifw.c
+C_DEFINITIONS = r"""
+    struct hdr1_header {
+          char magic[4];                /* HDR1 */
+          uint32 signature_offset;
+          uint32 crc32;
+          uint16 unused;
+          uint16 device_id;             /* RA70 */
+          uint32 blob_offsets[8];
+    } hdr1_header_t;
+
+    struct hdr2_header {
+          char magic[4];                /* HDR1 */
+          uint32 signature_offset;
+          uint32 crc32;
+          uint32 unused1;
+          uint64 device_id;             /* RA70 */
+          uint64 region;                /* EU */
+          uint64 unused2[2];
+          uint32 blob_offsets[8];
+    } hdr2_header_t;
+
+    struct xiaomi_blob_header {
+          uint32 magic;                 /* 0x0000babe */
+          uint32 flash_offset;
+          uint32 size;                  /* Size of blob */
+          uint16 type;                  /* Type of blob */
+          uint16 unused;
+          char name[32];                /* Name of blob */
+    } blob_header_t;
+
+    struct xiaomi_signature_header {
+          uint32 size;
+          uint32 padding[3];
+          uint8 content[0x100];
+    } signature_header_t;
+ """
+
+
+def calculate_crc(file: File, start_offset: int, size: int) -> int:
+    digest = 0
+    for chunk in iterate_file(file, start_offset, size):
+        digest = binascii.crc32(chunk, digest)
+    return (digest ^ -1) & 0xFFFFFFFF
+
+
+def is_valid_blob_header(blob_header: Instance) -> bool:
+    if blob_header.magic == BLOB_MAGIC:
+        return False
+    if not blob_header.size:
+        return False
+    try:
+        snull(blob_header.name).decode("utf-8")
+    except UnicodeDecodeError:
+        return False
+    return True
+
+
+def is_valid_header(header: Instance) -> bool:
+    if header.signature_offset < len(header):
+        return False
+    if not header.blob_offsets[0]:
+        return False
+    return True
+
+
+class HDRExtractor(Extractor):
+    def __init__(self, header_struct: str):
+        self.header_struct = header_struct
+        self._struct_parser = StructParser(C_DEFINITIONS)
+
+    def extract(self, inpath: Path, outdir: Path):
+        with File.from_path(inpath) as file:
+            for output_path, chunk in self.parse(file):
+                if not is_safe_path(outdir, output_path):
+                    logger.warn(
+                        "Path traversal attempt, discarding.", output_path=output_path
+                    )
+                    return
+                carve_chunk_to_file(outdir.joinpath(output_path), file, chunk)
+
+    def parse(self, file: File) -> Iterable[Tuple[Path, Chunk]]:
+        header = self._struct_parser.parse(self.header_struct, file, Endian.LITTLE)
+        for offset in header.blob_offsets:
+            if not offset:
+                break
+
+            file.seek(offset, io.SEEK_SET)
+            blob_header = self._struct_parser.parse(
+                "blob_header_t", file, Endian.LITTLE
+            )
+            logger.debug("blob_header_t", blob_header_t=blob_header, _verbosity=3)
+            if not is_valid_blob_header(blob_header):
+                raise InvalidInputFormat("Invalid HDR blob header.")
+
+            # file.tell() points to right after the blob_header
+            yield (
+                (
+                    Path(snull(blob_header.name).decode("utf-8")),
+                    Chunk(
+                        start_offset=file.tell(),
+                        end_offset=file.tell() + blob_header.size,
+                    ),
+                )
+            )
+
+
+class HDRHandlerBase(StructHandler):
+    HEADER_STRUCT = "hdr1_header_t"
+
+    def calculate_chunk(self, file: File, start_offset: int) -> Optional[ValidChunk]:
+        header = self.parse_header(file, endian=Endian.LITTLE)
+
+        if not is_valid_header(header):
+            raise InvalidInputFormat("Invalid HDR header.")
+
+        end_offset = start_offset + header.signature_offset + SIGNATURE_LEN
+
+        if not self._is_crc_valid(file, header, start_offset, end_offset):
+            raise InvalidInputFormat("CRC32 does not match in HDR header.")
+
+        return ValidChunk(start_offset=start_offset, end_offset=end_offset)
+
+    def _is_crc_valid(
+        self, file: File, header: Instance, start_offset: int, end_offset: int
+    ) -> bool:
+        computed_crc = calculate_crc(
+            file,
+            start_offset=start_offset + CRC_CONTENT_OFFSET,
+            size=end_offset - start_offset + CRC_CONTENT_OFFSET,
+        )
+        return header.crc32 == computed_crc
+
+
+class HDR1Handler(HDRHandlerBase):
+    NAME = "hdr1"
+    PATTERNS = [
+        HexString(
+            """
+            //  48 44 52 31 90 32 e2 00  02 2a 5b 6a 00 00 11 00  HDR1.2...*[j....
+            //  30 00 00 00 70 02 00 00  a4 02 e0 00 00 00 00 00  0...p...........
+            48 44 52 31
+        """
+        ),
+    ]
+    C_DEFINITIONS = C_DEFINITIONS
+    HEADER_STRUCT = "hdr1_header_t"
+
+    EXTRACTOR = HDRExtractor("hdr1_header_t")
+
+
+class HDR2Handler(HDRHandlerBase):
+    NAME = "hdr2"
+    PATTERNS = [
+        HexString(
+            """
+            // 48 44 52 32 d4 02 78 02  68 54 e8 fa 00 00 00 00  HDR2..x.hT......
+            // 52 41 37 30 00 00 00 00  45 55 00 00 00 00 00 00  RA70....EU......
+            48 44 52 32
+        """
+        ),
+    ]
+    C_DEFINITIONS = C_DEFINITIONS
+    HEADER_STRUCT = "hdr2_header_t"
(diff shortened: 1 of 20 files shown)


unblob/handlers/archive/xiaomi/hdr1.py
```diff
@@ -0,0 +1,142 @@
...
+
+    PATTERNS = [HexString("48 44 52 31")]
+
+    C_DEFINITIONS = HDR1_HEADER_DEFINITION
+    HEADER_STRUCT = "hdr_header_t"
+    EXTRACTOR = HDRExtractor()
+
+    def calculate_chunk(self, file: File, start_offset: int) -> Optional[ValidChunk]:
+        header = self.parse_header(file, endian=Endian.LITTLE)
+        end_offset = start_offset + header.signature_offset + SIGNATURE_LEN
```
qkaiser, line 113: I would move the `end_offset` calculation to after the call to `is_valid_header`. This way we only compute it if it's a true positive.

qkaiser (changes requested):

unblob/handlers/archive/xiaomi/hdr1.py
```diff
@@ -0,0 +1,142 @@
...
+    Extractor,
+    HexString,
+    StructHandler,
+    StructParser,
+    ValidChunk,
+)
+
+logger = get_logger()
+CRC_CONTENT_OFFSET = 12
+SIGNATURE_LEN = 272
```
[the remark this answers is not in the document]
nyuware: #signature_header_t contains 4 bytes of size + 12 bytes for padding x3 + 0x100 is 256 in decimal

unblob/handlers/archive/xiaomi/hdr1.py
```diff
@@ -0,0 +1,142 @@
...
+          char magic[4];                /* HDR1 */
+          uint32 signature_offset;
+          uint32 crc32;
+          uint16 unused;
+          uint16 device_id;             /* RA70 */
+          uint32 blob_offsets[8];
+    } hdr_header_t;
+
+    struct xiaomi_blob_header {
+          uint32 magic;                 /* HDR1 */
```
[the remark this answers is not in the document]
nyuware: Corrected, https://lxr.openwrt.org/source/firmware-utils/src/xiaomifw.c#L439

unblob/handlers/archive/xiaomi/hdr1.py
```diff
@@ -0,0 +1,142 @@
...
+    def parse(self, file: File):
+        """Parse header + blob headers and populate self._blobs"""
+        header = self._struct_parser.parse("hdr_header_t", file, Endian.LITTLE)
+        for offset in header.blob_offsets:
+            if not offset:
+                continue
+            file.seek(offset, io.SEEK_SET)
+            blob_header = self._struct_parser.parse(
+                "blob_header_t", file, Endian.LITTLE
+            )
```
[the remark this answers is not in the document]
nyuware: Header validation has been set to the following:

- Magic must equal to a constant.
- Size of the blob header musn't be 0
- Blob name should be in utf-8.

unblob/handlers/archive/xiaomi/hdr1.py
nyuware: end_offset has been moved after the is_valid_header to avoid unnecessary calculation.

unblob/handlers/archive/xiaomi/hdr1.py
```diff
@@ -0,0 +1,167 @@
...
+                    name=snull(blob_header.name).decode("utf-8"),
+                    size=blob_header.size,
+                )
+            )
+
+    def is_valid_header(self, blob_header: Instance, file) -> bool:
+        if blob_header.magic == BLOB_MAGIC:
+            raise InvalidInputFormat("Invalid HDR blob header.")
+        if blob_header.size == 0:
+            raise ValueError
```
qkaiser, line 109:
```suggestion
        if not blob_header.size:
            return False
```

unblob/handlers/archive/xiaomi/hdr1.py
```diff
@@ -0,0 +1,167 @@
...
+                Blob(
+                    offset=file.tell(),
+                    name=snull(blob_header.name).decode("utf-8"),
+                    size=blob_header.size,
+                )
+            )
+
+    def is_valid_header(self, blob_header: Instance, file) -> bool:
+        if blob_header.magic == BLOB_MAGIC:
+            raise InvalidInputFormat("Invalid HDR blob header.")
```
qkaiser, line 107:
```suggestion
            return False
```

unblob/handlers/archive/xiaomi/hdr1.py
```diff
@@ -0,0 +1,167 @@
...
+                continue
+
+            file.seek(offset, io.SEEK_SET)
+            blob_header = self._struct_parser.parse(
+                "blob_header_t", file, Endian.LITTLE
+            )
+
+            if not self.is_valid_header(blob_header, file):
+                raise InvalidInputFormat("Invalid blob header.")
+            file.seek(-BLOB_HEADER_LEN, io.SEEK_CUR)
```
qkaiser, line 94: Remove this. If you keep it, the carved out file will contain both the blob and blob header. We only want the blob.

unblob/handlers/archive/xiaomi/hdr1.py
```diff
@@ -0,0 +1,167 @@
...
+            start_offset=start_offset + CRC_CONTENT_OFFSET,
+            size=end_offset - start_offset + CRC_CONTENT_OFFSET,
+        )
+        return header.crc32 == computed_crc
+
+
+def calculate_crc(file: File, start_offset: int, size: int) -> int:
+    file.seek(start_offset)
+    content = bytearray(file.read(size))
+    return (binascii.crc32(content) ^ -1) & 0xFFFFFFFF
```
qkaiser, line 167: Move to top, before the definition of the `Blob` class

qkaiser (changes requested):

nyuware:
The code I pushed is non-functional, it's a failed attempt at making the handler work for both HDR1 and HDR2

qkaiser:
> The code I pushed is non-functional, it's a failed attempt at making the handler work for both HDR1 and HDR2

Run pre-commit checks, rebase on `main`, and adapt your test cases. The code starts to look good :) I'll make some comments when you're done with those tasks.

unblob/handlers/archive/xiaomi/hdr.py
```diff
@@ -0,0 +1,197 @@
...
+    try:
+        snull(blob_header.name).decode("utf-8")
+    except UnicodeDecodeError:
+        return False
+    return True
+
+
+def is_valid_header(header: Instance) -> bool:
+    if header.signature_offset < len(header):
+        return True
```
qkaiser, line 86:
```suggestion
        return False
```

unblob/handlers/archive/xiaomi/hdr.py
```diff
@@ -0,0 +1,197 @@
...
+    if not header.blob_offsets[0]:
+        return True
```
qkaiser, line 88:
```suggestion
        return False
```

unblob/handlers/archive/xiaomi/hdr.py
```diff
@@ -0,0 +1,197 @@
...
+    def calculate_chunk(self, file: File, start_offset: int) -> Optional[ValidChunk]:
+        header = self.parse_header(file, endian=Endian.LITTLE)
+
+        if not is_valid_header(header):
+            raise InvalidInputFormat("Invalid HDR header.")
+
+        end_offset = start_offset + header.signature_offset + SIGNATURE_LEN
+
+        if not self._is_crc_valid(file, header, start_offset, end_offset):
+            raise InvalidInputFormat("Invalid CRC32.")
```
qkaiser, line 152:
```suggestion
            raise InvalidInputFormat("CRC32 does not match in HDR header.")
```

unblob/handlers/archive/xiaomi/hdr.py
```diff
@@ -0,0 +1,197 @@
...
+            if not offset:
+                break
+
+            file.seek(offset, io.SEEK_SET)
+            blob_header = self._struct_parser.parse(
+                "blob_header_t", file, Endian.LITTLE
+            )
+            logger.debug("blob_header_t", blob_header_t=blob_header, _verbosity=3)
+            if not is_valid_blob_header(blob_header):
+                raise InvalidInputFormat("Invalid blob header.")
```
qkaiser, line 127:
```suggestion
                raise InvalidInputFormat("Invalid HDR blob header.")
```

unblob/handlers/archive/xiaomi/hdr.py
```diff
@@ -0,0 +1,197 @@
...
+
+            # file.tell() points to right after the blob_header
+            blobs.append(
+                Blob(
+                    offset=file.tell(),
+                    name=snull(blob_header.name).decode("utf-8"),
+                    size=blob_header.size,
+                )
+            )
```
qkaiser, line 136: I looked at what we do in our internal handlers, and a good way to limit usage of custom objects is to use `Chunk` object everywhere.

We usually do this by returning a list of tuple like this `[("file_name", Chunk()), ...]`.

You can use that approach here with something like:

```suggestion
            blob_name = 
            blobs.append(
                (
                snull(blob_header.name).decode("utf-8"),
                Chunk(start_offset=file.tell(), end_offset=file.tell()+ blob.size)
            ))
```

Using this approach, you can remove the lines `92-96` and use chunks directly in the `extract` function.

unblob/handlers/archive/xiaomi/hdr.py
```diff
@@ -0,0 +1,197 @@
...
+    def extract(self, inpath: Path, outdir: Path):
+        with File.from_path(inpath) as file:
+            blobs = self.parse(file)
+            for blob in blobs:
+                chunk = Chunk(
+                    start_offset=blob.offset, end_offset=blob.offset + blob.size
+                )
+                carve_chunk_to_file(outdir.joinpath(Path(blob.name)), file, chunk)
+
+    def parse(self, file: File) -> List[Blob]:
```
qkaiser, line 113: If you apply the changes recommend below, the signature becomes:

```suggestion
    def parse(self, file: File) -> List[Tuple[str, Chunk]]:
```

unblob/handlers/archive/xiaomi/hdr.py
```diff
@@ -0,0 +1,197 @@
...
+        self._struct_parser = StructParser(C_DEFINITIONS)
+
+    def extract(self, inpath: Path, outdir: Path):
+        with File.from_path(inpath) as file:
+            blobs = self.parse(file)
+            for blob in blobs:
+                chunk = Chunk(
+                    start_offset=blob.offset, end_offset=blob.offset + blob.size
+                )
+                carve_chunk_to_file(outdir.joinpath(Path(blob.name)), file, chunk)
```
qkaiser, line 111: If you follow the recommendations below, this should be:

```suggestion
            for name, chunk in self.parse(file):
                carve_chunk_to_file(outdir.joinpath(Path(name)), file, chunk)
```

You must add a protection against path traversals using `is_safe_path` before calling `carve_chunk_to_file` too.

qkaiser (changes requested):
Left some comments to further improve the code. When it's done we can discuss how to improve performance by using iterators :)

unblob/handlers/archive/xiaomi/hdr.py
```diff
@@ -0,0 +1,192 @@
...
+
+    def extract(self, inpath: Path, outdir: Path):
+        with File.from_path(inpath) as file:
+            for name, chunk in self.parse(file):
+                if not is_safe_path(outdir, Path(name)):
+                    logger.warn(
+                        "Path traversal attempt, discarding.", output_path=Path(name)
+                    )
+                    return
+                carve_chunk_to_file(outdir.joinpath(Path(name)), file, chunk)
```
qkaiser, line 104:
```suggestion
            for output_path, chunk in self.parse(file):
                if not is_safe_path(outdir, output_path):
                    logger.warn(
                        "Path traversal attempt, discarding.", output_path=output_path
                    )
                    return
                carve_chunk_to_file(outdir.joinpath(output_path), file, chunk)
```

unblob/handlers/archive/xiaomi/hdr.py
```diff
@@ -0,0 +1,192 @@
...
+
+    def parse(self, file: File) -> List[Tuple[str, Chunk]]:
```
qkaiser, line 106:
```suggestion
    def parse(self, file: File) -> Iterable[Tuple[Path, Chunk]]:
```

unblob/handlers/archive/xiaomi/hdr.py
```diff
@@ -0,0 +1,192 @@
...
+            blobs.append(
+                (
+                    snull(blob_header.name).decode("utf-8"),
+                    Chunk(
+                        start_offset=file.tell(),
+                        end_offset=file.tell() + blob_header.size,
+                    ),
+                )
+            )
+        return blobs
```
qkaiser, line 132:
```suggestion
```

unblob/handlers/archive/xiaomi/hdr.py
```diff
@@ -0,0 +1,192 @@
...
+            # file.tell() points to right after the blob_header
+            blobs.append(
+                (
+                    snull(blob_header.name).decode("utf-8"),
+                    Chunk(
+                        start_offset=file.tell(),
+                        end_offset=file.tell() + blob_header.size,
+                    ),
+                )
+            )
```
qkaiser, line 131:
```suggestion
            yield (
                    Path(snull(blob_header.name).decode("utf-8")),
                    Chunk(
                        start_offset=file.tell(),
                        end_offset=file.tell() + blob_header.size,
                    ),
                )
```

qkaiser (commented):
Some further optimisations.

martonilles (changes requested):

unblob/handlers/archive/xiaomi/hdr.py
```diff
@@ -0,0 +1,189 @@
...
+          uint32 size;
+          uint32 padding[3];
+          uint8 content[0x100];
+    } signature_header_t;
+ """
+
+
+def calculate_crc(file: File, start_offset: int, size: int) -> int:
+    file.seek(start_offset)
+    content = bytearray(file.read(size))
```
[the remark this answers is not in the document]
qkaiser: @nyuware you can use the function `iterate_file` for that https://github.com/onekey-sec/unblob/blob/bd83f96061f0505375e69670d7b06f0e268cf298/unblob/file_utils.py#L226-L232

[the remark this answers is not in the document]
qkaiser: Look at the definition of `crc32(data, value)` to understand how you can calculate a CRC32 over multiple chunks of bytes - https://docs.python.org/3/library/binascii.html#binascii.crc32

qkaiser (approved):

PR #530, opened 2023-03-02
Remove cstruct

vlaci:
It was pulled in as a transitive dependency for jefferson. Since it
got its proper packaging treatment in https://github.com/onekey-sec/unblob/pull/505, it is not needed here.


qkaiser (approved):
💯

Issue #474, opened 2022-11-25
Add support for squashfs version 2


QKaiser:
Even though it's quite old, some firmwares in our corpus have squashfs v2 images. One example is the Asus NAS M25 firmware image.

The easiest path to support squashfsv2 is to release a dedicated binary from our sasquatch fork. I worked on https://github.com/onekey-sec/sasquatch/tree/modern-squashfs-v2.0-r2 to have a working codebase with version 2 images. What is missing is building the dedicated v2 binary.

Once it's there we can add v2 to the existing squashfs handler :)

qkaiser:
sasquatch MR to fix a bug in squashfs-tools that blocked us from extracting version 2 and 1 FS: https://github.com/onekey-sec/sasquatch/pull/13

When it's merge I'm ready with the code to make it work in unblob.

PR #532, opened 2023-03-09
feat(handler): add support for squashfsv2.

qkaiser:
We could not support old versions of squashfs due to a bug in squashfs-tools. It's now fixed in our fork so time to move here.

See #474 and https://github.com/onekey-sec/sasquatch/pull/13

### DIFF
--- a/.github/actions/setup-dependencies/action.yml
+++ b/.github/actions/setup-dependencies/action.yml
@@ -15,7 +15,7 @@ runs:
 
     - name: Install sasquatch
       run: |
-        curl -L -o sasquatch_1.0_amd64.deb https://github.com/onekey-sec/sasquatch/releases/download/sasquatch-v4.5.1-1/sasquatch_1.0_amd64.deb
+        curl -L -o sasquatch_1.0_amd64.deb https://github.com/onekey-sec/sasquatch/releases/download/sasquatch-v4.5.1-2/sasquatch_1.0_amd64.deb
         sudo dpkg -i sasquatch_1.0_amd64.deb
         rm -f sasquatch_1.0_amd64.deb
       shell: bash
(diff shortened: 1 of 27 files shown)


qkaiser:
The hunt for a version of `mksquashfs` that can build v1 and v2 samples begins.

vlaci:
> The hunt for a version of `mksquashfs` that can build v1 and v2 samples begins.

There is a bunch of old release tarballs on sourceforge https://sourceforge.net/projects/squashfs/files/squashfs/

qkaiser:
@vlaci I had the same mods locally.

I'm all for Internet archeology but if squashfs 1.3r3 was human it could drink and drive (and probably would). I'm getting segfaults in mksquashfs `get_checksum` function, if I comment it and just returns 0 it fails if you have more than one inode.

If you have a version that does not segfault, feel free to add integration test files. Otherwise I think it's okay if we limit ourselves to version 2.

qkaiser:
Tests are failing because we need to release sasquatch deb package with the v2 fix and use it here.

.github/actions/setup-dependencies/action.yml
```diff
@@ -15,7 +15,7 @@ runs:
 
     - name: Install sasquatch
       run: |
-        curl -L -o sasquatch_1.0_amd64.deb https://github.com/onekey-sec/sasquatch/releases/download/sasquatch-v4.5.1-1/sasquatch_1.0_amd64.deb
```
vlaci, line 18: This line is duplicated in `installation.md` and `Dockerfile` as well

vlaci (changes requested):

.github/actions/setup-dependencies/action.yml
qkaiser: fixed

vlaci (approved):

PR #534, opened 2023-03-16
Elf initramfs regression

martonilles:
ELF refactor broke the initramfs extraction from kernel ELF files in case the ELF is not on offset 0


qkaiser:
Looks good to me. We should have picked up the issue fixed by 20415bab65f566efd01afc326611390dab090ab1 earlier, it's the kind of offset miscalculation bug we were always on the lookout for during the early days of unblob.

I'll probably write it down in https://unblob.org/development/#common-unblob-handler-mistakes

qkaiser:
How difficult do you think it would be to synthesize a sample that demonstrates the regression fixed by this MR ?

martonilles:
I have some samples, but not artificial and indeed private ones.
We can probably craft manually an ELF file, surprised that the kernel initramfs is not picked up earlier by tests

unblob/handlers/executable/elf.py
```diff
@@ -211,11 +211,11 @@ def get_end_offset(
             start_offset + header.e_phoff + (header.e_phnum * header.e_phentsize)
         )
 
-        last_section_end = self.get_last_section_end(
+        last_section_end = start_offset + self.get_last_section_end(
             file, start_offset + header.e_shoff, header.e_shnum, endian
         )
 
-        last_program_end = self.get_last_program_end(
+        last_program_end = start_offset + self.get_last_program_end(
```
e3krisztian, line 218: With this change, every component has a `start_offset` added, so what about rewriting the function like this?
```py
    def get_end_offset(
        self, file: File, start_offset: int, header: Instance, endian
    ) -> int:
        # Usually the section header is the last, but in some cases the program headers are
        # put to the end of the file, and in some cases sections header and actual sections
        # can be also intermixed, so we need also to check the end of the last section and
        # also the last program segment.
        # We check which one is the last and use it as a file size.
        section_headers_end = (
            header.e_shoff + (header.e_shnum * header.e_shentsize)
        )
        program_headers_end = (
            header.e_phoff + (header.e_phnum * header.e_phentsize)
        )

        last_section_end = self.get_last_section_end(
            file, start_offset + header.e_shoff, header.e_shnum, endian
        )

        last_program_end = self.get_last_program_end(
            file, start_offset + header.e_phoff, header.e_phnum, endian
        )

        return start_offset + max(
            section_headers_end, program_headers_end, last_section_end, last_program_end
        )
```

martonilles: done

martonilles:
I have also added test cases for both problems

qkaiser (approved):

PR #536, opened 2023-03-17
feat(handler): Add support for INSTAR BNEG format

nyuware:
As requested, the minor and major do not change (yet), they only are 1  in little endian in both fields. 
The firmware is composed of a zip file and a romfs file.

### DIFF
--- a/unblob/handlers/archive/instar/bneg.py
+++ b/unblob/handlers/archive/instar/bneg.py
@@ -1,0 +1,97 @@
+from pathlib import Path
+from typing import Optional
+
+from dissect.cstruct import Instance
+from structlog import get_logger
+
+from unblob.extractor import carve_chunk_to_file
+from unblob.file_utils import Endian, File, InvalidInputFormat, StructParser
+from unblob.models import Chunk, Extractor, HexString, StructHandler, ValidChunk
+
+logger = get_logger()
+
+C_DEFINITIONS = r"""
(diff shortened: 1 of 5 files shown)


unblob/handlers/archive/instar/bneg.py
```diff
@@ -0,0 +1,95 @@
...
+    } bneg_header_t;
+"""
+
+
+class BNEGExtractor(Extractor):
+    def __init__(self):
+        self._struct_parser = StructParser(C_DEFINITIONS)
+
+    def extract(self, inpath: Path, outdir: Path):
+        outpath = outdir.joinpath(f"{inpath.name}")
```
e3krisztian, line 29: Using just `outdir` would simplify the output structure:
```
tests/integration/archive/instar/bneg/__output__/output.bin_extract/output.bin/part1
```
->
```
tests/integration/archive/instar/bneg/__output__/output.bin_extract/part1
```

----

Another issue is duplicating `inpath.name`. While this duplication happens elsewhere as well, it is causing some headaches, so will be fixed soon. One of the known problems is, that carved out chunks would not have a meaningful name, but `start_offset-end_offset.bneg`, which also appears in the output twice - adding noise to the output. It is better to use just a fixed name, like `bneg` or even better no extra directory at all.

e3krisztian (approved):
Thanks!

PR #537, opened 2023-03-20
feat(handler): Add support for INSTAR HD format

nyuware:
This format is just a zip file with the headers modified.
This handler attempts to modify the header for the zip handler to handle it afterwards

### DIFF
--- a/unblob/handlers/archive/instar/instar_hd.py
+++ b/unblob/handlers/archive/instar/instar_hd.py
@@ -1,0 +1,47 @@
+import mmap
+import shutil
+from pathlib import Path
+
+from structlog import get_logger
+
+from unblob.file_utils import File
+from unblob.handlers.archive.zip import ZIPHandler
+from unblob.models import Extractor, HexString
+
+logger = get_logger()
+
+LOCAL_FILE_HEADER_ZIP = b"\x50\x4b\x03\x04"
+LOCAL_FILE_HEADER_INSTAR = b"\x50\x4b\x03\x07"
+
+EOCD_ZIP = b"\x50\x4b\x05\x06"
+EOCD_INSTAR = b"\x50\x4b\x05\x09"
+
+CD_FILE_HEADER_ZIP = b"\x50\x4b\x01\x02"
+CD_FILE_HEADER_INSTAR = b"\x50\x4b\x01\x08"
+
+
+class InstarHDExtractor(Extractor):
+    def extract(self, inpath: Path, outdir: Path):
+        outfile = outdir / "instar_hd.deobfuscated"
+        shutil.copyfile(inpath, outfile)
+        with File.from_path(outfile, access=mmap.ACCESS_WRITE) as mm:
+            replacements = {
+                LOCAL_FILE_HEADER_INSTAR: LOCAL_FILE_HEADER_ZIP,
+                EOCD_INSTAR: EOCD_ZIP,
+                CD_FILE_HEADER_INSTAR: CD_FILE_HEADER_ZIP,
+            }
+            for pattern, replacement in replacements.items():
+                pos = mm.find(pattern)
+                while pos != -1:
+                    mm[pos : pos + len(pattern)] = replacement
+                    pos = mm.find(pattern, pos + len(pattern))
+
+
+class InstarHDHandler(ZIPHandler):
+    NAME = "instar_hd"
+
+    PATTERNS = [HexString("50 4b 03 07")]  # match on the modified zip header
+
+    EXTRACTOR = InstarHDExtractor()
+
+    EOCD_RECORD_HEADER = 0x9054B50
--- a/tests/integration/archive/instar/instar_hd/__input__/sample.pkg
+++ b/tests/integration/archive/instar/instar_hd/__input__/sample.pkg
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:4bd1e5307a3f1346d99039b4abc50190ce41a0f807b00cf64504c08509617a77
+size 187
--- a/tests/integration/archive/instar/instar_hd/__output__/sample.pkg_extract/instar_hd.deobfuscated
+++ b/tests/integration/archive/instar/instar_hd/__output__/sample.pkg_extract/instar_hd.deobfuscated
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:fab9135f2faaf1c9b40b8a4c052d92c13df21c1be7c5058858df68ab43e55440
+size 187
--- a/tests/integration/archive/instar/instar_hd/__output__/sample.pkg_extract/instar_hd.deobfuscated_extract/sample.txt
+++ b/tests/integration/archive/instar/instar_hd/__output__/sample.pkg_extract/instar_hd.deobfuscated_extract/sample.txt
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:e6478ea19e2122817e1986151847d3590e114ffd7c5d65d8d91cccea316c1cf2
+size 17
--- a/unblob/file_utils.py
+++ b/unblob/file_utils.py
@@ -29,9 +29,10 @@ class File(mmap.mmap):
         return m
 
     @classmethod
-    def from_path(cls, path: Path):
-        with path.open("rb") as base_file:
-            return cls(base_file.fileno(), 0, access=mmap.ACCESS_READ)
+    def from_path(cls, path: Path, access=mmap.ACCESS_READ):
+        mode = "r+b" if access == mmap.ACCESS_WRITE else "rb"
+        with path.open(mode) as base_file:
+            return cls(base_file.fileno(), 0, access=access)
 
     def seek(self, pos: int, whence: int = os.SEEK_SET) -> int:
         try:
--- a/unblob/handlers/__init__.py
+++ b/unblob/handlers/__init__.py
@@ -3,7 +3,7 @@ from .archive import ar, arc, arj, cab, cpio, dmg, rar, sevenzip, stuffit, tar, 
 from .archive.dlink import encrpted_img, shrs
 from .archive.engeniustech import engenius
 from .archive.hp import bdl, ipkg
-from .archive.instar import bneg
+from .archive.instar import bneg, instar_hd
 from .archive.netgear import chk, trx
 from .archive.qnap import qnap_nas
 from .archive.xiaomi import hdr
@@ -63,6 +63,7 @@ BUILTIN_HANDLERS: Handlers = (
     qnap_nas.QnapHandler,
     bneg.BNEGHandler,
     bdl.HPBDLHandler,
+    instar_hd.InstarHDHandler,
     ipkg.HPIPKGHandler,
     sparse.SparseHandler,
     ar.ARHandler,
--- a/unblob/handlers/archive/zip.py
+++ b/unblob/handlers/archive/zip.py
@@ -11,11 +11,6 @@ from ...models import File, HexString, StructHandler, ValidChunk
 
 logger = get_logger()
 
-ENCRYPTED_FLAG = 0b0001
-EOCD_RECORD_HEADER = 0x6054B50
-ZIP64_EOCD_SIGNATURE = 0x06064B50
-ZIP64_EOCD_LOCATOR_HEADER = 0x07064B50
-
 
 class ZIPHandler(StructHandler):
     NAME = "zip"
@@ -86,6 +81,11 @@ class ZIPHandler(StructHandler):
     # empty password with -p will make sure the command will not hang
     EXTRACTOR = Command("7z", "x", "-p", "-y", "{inpath}", "-o{outdir}")
 
+    ENCRYPTED_FLAG = 0b0001
+    EOCD_RECORD_HEADER = 0x6054B50
+    ZIP64_EOCD_SIGNATURE = 0x06064B50
+    ZIP64_EOCD_LOCATOR_HEADER = 0x07064B50
+
     def has_encrypted_files(
         self,
         file: File,
@@ -95,7 +95,7 @@ class ZIPHandler(StructHandler):
         file.seek(start_offset + end_of_central_directory.offset_of_cd, io.SEEK_SET)
         for _ in range(0, end_of_central_directory.total_entries):
             cd_header = self.cparser_le.cd_file_header_t(file)
-            if cd_header.flags & ENCRYPTED_FLAG:
+            if cd_header.flags & self.ENCRYPTED_FLAG:
                 return True
         return False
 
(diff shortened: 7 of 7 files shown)


unblob/handlers/archive/instar/hd.py
```diff
@@ -0,0 +1,105 @@
...
+
+EOCD_RECORD_HEADER = 0x9054B50
+
+
+class InstarHDExtractor(Extractor):
+    def extract(self, inpath: Path, outpath: Path):
+        with open(inpath, "rb") as file:
+            outfile = outpath.joinpath(f"{inpath.name}")
+            with open(outfile, "wb") as out_file:
+                headers = file.read()
```
qkaiser, line 21: This will load the whole file in memory, better use `iterate_on_file`.

unblob/handlers/archive/instar/hd.py
```diff
@@ -0,0 +1,105 @@
...
+
+                if b"\x50\x4b\x05\x09" in headers:
+                    headers = headers.replace(b"\x50\x4b\x05\x09", b"\x50\x4b\x05\x06")
+
+                if b"\x50\x4b\x01\x08" in headers:
+                    headers = headers.replace(b"\x50\x4b\x01\x08", b"\x50\x4b\x01\x02")
+                out_file.write(headers)
+
+
+class ZIPHandler(StructHandler):
```
martonilles, line 34: I would use a more specific name like InstarHDZip or something similar

unblob/handlers/archive/instar/hd.py
```diff
@@ -0,0 +1,105 @@
...
+                headers = file.read()
+
+                if b"\x50\x4b\x03\x07" in headers:
+                    headers = headers.replace(b"\x50\x4b\x03\x07", b"\x50\x4b\x03\x04")
+
+                if b"\x50\x4b\x05\x09" in headers:
+                    headers = headers.replace(b"\x50\x4b\x05\x09", b"\x50\x4b\x05\x06")
+
+                if b"\x50\x4b\x01\x08" in headers:
+                    headers = headers.replace(b"\x50\x4b\x01\x08", b"\x50\x4b\x01\x02")
```
qkaiser, line 30: I would move this into a dedicated function named `fix_headers_magic`. I would also remove the `if` condition (the `if` will traverse the bytearray a first time, and then if something is found it will be traversed a second time).

I would also assign the byte strings to constants so that the reader understands what you are doing. Something like:

```python
CD_FILE_HEADER_ZIP = b"\x50\x4b\x03\x04"
CD_FILE_HEADER_INSTAR = b"\x50\x4b\x03\x07"
```

qkaiser (changes requested):

unblob/handlers/archive/instar/hd.py
```diff
@@ -0,0 +1,105 @@
...
+            uint16 comment_len;
+            char zip_file_comment[comment_len];
+        } end_of_central_directory_t;
+    """
+
+    HEADER_STRUCT = "end_of_central_directory_t"
+
+    EXTRACTOR = InstarHDExtractor()
+
+    def calculate_chunk(self, file: File, start_offset: int) -> Optional[ValidChunk]:
```
martonilles, line 80: this is almost the same as the ZIP handler, can not we inherit from that handler in order to avoid code duplication?

unblob/handlers/archive/instar/hd.py
```diff
@@ -0,0 +1,105 @@
...
+
+                if b"\x50\x4b\x03\x07" in headers:
+                    headers = headers.replace(b"\x50\x4b\x03\x07", b"\x50\x4b\x03\x04")
+
+                if b"\x50\x4b\x05\x09" in headers:
+                    headers = headers.replace(b"\x50\x4b\x05\x09", b"\x50\x4b\x05\x06")
+
+                if b"\x50\x4b\x01\x08" in headers:
+                    headers = headers.replace(b"\x50\x4b\x01\x08", b"\x50\x4b\x01\x02")
+                out_file.write(headers)
```
martonilles, line 31: shouldn't we also run the zip extraction here? This is like the duplicate of the zip handler, except a few bytes are replaced in the carved out content?

martonilles (changes requested):

unblob/handlers/archive/instar/hd.py
qkaiser: Two ways of going about it:
1. we carve the chunk, our extractor converts it to classic zip, zip handler picks it up
2. we carve the chunk, our extractor converts it to classic zip **in place** and calls the zip extractor on it using a `Command`

While option 2 is more efficient, it kinda breaks the contract that a carved out chunk corresponds exactly to what was found in its parent file from start offset to end offset.

unblob/handlers/archive/instar/hd.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+class ZIPHandler(StructHandler):
+    NAME = "hd"
```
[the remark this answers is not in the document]
qkaiser: This is **not** resolved.

unblob/handlers/archive/instar/hd.py
qkaiser: This is **not** resolved.

unblob/handlers/archive/instar/hd.py
qkaiser: This is **not** resolved.

unblob/handlers/archive/instar/instar_hd.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def extract(self, inpath: Path, outdir: Path):
+        with File.from_path(inpath) as file:
+            outfile = outdir.joinpath(f"{inpath.name}.zip")
+            with open(outfile, "wb") as out_file:
+                for chunk in iterate_file(file, 0, inpath.stat().st_size):
+                    out_file.write(self.fix_headers(chunk))
+
+
+class InstarHDZipHandler(ZIPHandler):
```
qkaiser, line 37:
```suggestion
class InstarHDHandler(ZIPHandler):
```

unblob/handlers/archive/instar/instar_hd.py
```diff
@@ -0,0 +1,44 @@
...
+            .replace(EOCD_INSTAR, EOCD_ZIP)
+            .replace(CD_FILE_HEADER_INSTAR, CD_FILE_HEADER_ZIP)
+        )
+
+    def extract(self, inpath: Path, outdir: Path):
+        with File.from_path(inpath) as file:
+            outfile = outdir.joinpath(f"{inpath.name}.zip")
+            with open(outfile, "wb") as out_file:
+                for chunk in iterate_file(file, 0, inpath.stat().st_size):
+                    out_file.write(self.fix_headers(chunk))
```
qkaiser, line 34: Are ZIP headers always aligned ? Is it possible that a chunk ends in the middle of a header you want to replace ? 

paging our zip expert @e3krisztian 📟

unblob/handlers/archive/instar/instar_hd.py
```diff
@@ -0,0 +1,44 @@
...
+CD_FILE_HEADER_ZIP = b"\x50\x4b\x01\x02"
+CD_FILE_HEADER_INSTAR = b"\x50\x4b\x01\x08"
+
+
+class InstarHDExtractor(Extractor):
+    def fix_headers(self, chunk: bytes) -> bytes:
+        return (
+            chunk.replace(LOCAL_FILE_HEADER_INSTAR, LOCAL_FILE_HEADER_ZIP)
+            .replace(EOCD_INSTAR, EOCD_ZIP)
+            .replace(CD_FILE_HEADER_INSTAR, CD_FILE_HEADER_ZIP)
```
e3krisztian, line 26: Using `replace` can also cause some files not extracting: if the patterns occur in the content, not in the header - small probability, but can be triggered intentionally. E.g. zip files can store files without compression, it is sometimes useful to embed a zip file in a zip file, in this case the internal zip/INSTAR would be extracted with a crc error.

The proper solution would be to walk the zip structures starting from the EOCD record and using the central directory records to locate the local headers, and fix the magic numbers for them one by one.

unblob/handlers/archive/instar/instar_hd.py
```diff
@@ -0,0 +1,101 @@
...
+class InstarHDExtractor(Extractor):
+    def __init__(self, c_definitions: str):
+        self._struct_parser = StructParser(c_definitions)
+
+    def fix_headers(self, file: Path):
+        eocd_instar = 0x9054B50
+
+        local_file_header_zip = b"P\x4b\x03\x04"
+        cd_file_header_zip = b"P\x4b\x01\x02"
+        eocd_zip = b"P\x4b\x05\x06"
```
qkaiser, line 60: Use file level constants to store those.

unblob/handlers/archive/instar/instar_hd.py
```diff
@@ -0,0 +1,101 @@
...
+        last_offset = list(iterate_patterns(file, struct.pack("<I", eocd_instar)))[-1]
+
+        file.seek(last_offset, io.SEEK_SET)
+        file.write(eocd_zip)
+        file.seek(last_offset, io.SEEK_SET)
+        oecd_header = self._struct_parser.parse(
+            "end_of_central_directory_t", file, Endian.LITTLE
+        )
+        logger.debug("oecd_header", oecd_header=oecd_header)
+        file.seek(oecd_header.offset_of_cd)
```
qkaiser, line 71:
```suggestion
        file.seek(oecd_header.offset_of_cd, io.SEEK_SET)
```

unblob/handlers/archive/instar/instar_hd.py
```diff
@@ -0,0 +1,101 @@
...
+
+        last_offset = list(iterate_patterns(file, struct.pack("<I", eocd_instar)))[-1]
+
+        file.seek(last_offset, io.SEEK_SET)
+        file.write(eocd_zip)
+        file.seek(last_offset, io.SEEK_SET)
+        oecd_header = self._struct_parser.parse(
+            "end_of_central_directory_t", file, Endian.LITTLE
+        )
+        logger.debug("oecd_header", oecd_header=oecd_header)
```
qkaiser, line 70:
```suggestion
        logger.debug("end_of_central_directory", end_of_central_directory=eocd_header, _verbosity=3)
```

unblob/handlers/archive/instar/instar_hd.py
```diff
@@ -0,0 +1,101 @@
...
+        local_file_header_zip = b"P\x4b\x03\x04"
+        cd_file_header_zip = b"P\x4b\x01\x02"
+        eocd_zip = b"P\x4b\x05\x06"
+
+        last_offset = list(iterate_patterns(file, struct.pack("<I", eocd_instar)))[-1]
+
+        file.seek(last_offset, io.SEEK_SET)
+        file.write(eocd_zip)
+        file.seek(last_offset, io.SEEK_SET)
+        oecd_header = self._struct_parser.parse(
```
qkaiser, line 67:
```suggestion
        eocd_header = self._struct_parser.parse(
```

unblob/handlers/archive/instar/instar_hd.py
qkaiser, line 71:
```suggestion
        file.seek(eocd_header.offset_of_cd)
```

unblob/handlers/archive/instar/instar_hd.py
```diff
@@ -0,0 +1,101 @@
...
+            "end_of_central_directory_t", file, Endian.LITTLE
+        )
+        logger.debug("oecd_header", oecd_header=oecd_header)
+        file.seek(oecd_header.offset_of_cd)
+        for _ in range(oecd_header.total_entries):  # modify the CD and LF headers
```
qkaiser, line 72:
```suggestion
        for _ in range(eocd_header.total_entries):  # modify the CD and LF headers
```

qkaiser:
We discovered that obfuscated ZIP files can contain other ZIP files that are only partially obfuscated. This is due to the fact that Instar only applies obfuscation to the outer layer. Depending on the archive content and structure, magic values of inner ZIP files may or may not be compressed, which means they may or may not be obfuscated by their "match and replace" obfuscation.

This leads to scenarios where:
- all the magics of the outer ZIP are obfuscated
- EOCD and CD magics of an inner ZIP file are obfuscated but not the local file header magic of that file

The only way to solve this problem is to implement the deobfuscation in the most basic way, without any consideration for the ZIP structure, since this is how Instar performs the obfuscation in the first place.

It must be a match and replace, only applied to the outer layer.

nyuware:
Reverted back to previous working state to reflect to Quentin's comment.
The handler converts the zip EOCD, CD and file header to their initial values, note that we still might miss a header if a header is situated between two chunks of data since `iterate_file` yields chunks of 64K size.

unblob/handlers/archive/instar/instar_hd.py
```diff
@@ -0,0 +1,44 @@
...
+        return (
+            chunk.replace(LOCAL_FILE_HEADER_INSTAR, LOCAL_FILE_HEADER_ZIP)
+            .replace(EOCD_INSTAR, EOCD_ZIP)
+            .replace(CD_FILE_HEADER_INSTAR, CD_FILE_HEADER_ZIP)
+        )
+
+    def extract(self, inpath: Path, outdir: Path):
+        with File.from_path(inpath) as file:
+            outpath = outdir.joinpath(f"{inpath.name}")
+            with Path.open(outpath, "wb") as outfile:
```
qkaiser, line 32:
```suggestion
            outpath = outdir / "instar_hd.deobfuscated"
            with outpath .open("wb") as outfile:
```

qkaiser (changes requested):

qkaiser (approved):

unblob/handlers/archive/instar/instar_hd.py
```diff
@@ -0,0 +1,44 @@
...
+            .replace(EOCD_INSTAR, EOCD_ZIP)
+            .replace(CD_FILE_HEADER_INSTAR, CD_FILE_HEADER_ZIP)
+        )
+
+    def extract(self, inpath: Path, outdir: Path):
+        with File.from_path(inpath) as file:
+            outpath = outdir / "instar_hd.deobfuscated"
+            with outpath.open("wb") as outfile:
+                for chunk in iterate_file(file, 0, inpath.stat().st_size):
+                    outfile.write(self.fix_headers(chunk))
```
e3krisztian, line 34: What about 
- copying the file to `instar_hd.deobfuscated`
- `mmap` the new file
- fix headers on the mmap

this would leave out the `iterate_file` which could cause an extraction failure, as mentioned in https://github.com/onekey-sec/unblob/pull/537#issuecomment-1503165259

martonilles:
I still do not understand this. If there is a valid byte sequence equal to the magic in the file (not in the header, EOCD, etc), then it would be overwritten, and hence would break the extraction.

Also not sure what this means: `obfuscated ZIP files can contain other ZIP files that are only partially obfuscated` if there are embedded partially obfuscated zip files then those should be picked up in an other chunk detection and we should also support that partial obfuscation.

Maybe I am wrong and completely miss something. :(

qkaiser:
> then those should be picked up in an other chunk detection and we should also support that partial obfuscation.
> 

Maybe this wonderful diagram will help to reason about what I'm saying 

instar_hd drawio (1)

I don't want to spend time writing a unblob handler for each configuration:
- a match on the End Of Central Directory magic
- a match on the Central Directory magic
- a match on Local File Header magic

Because this is what happens depending on how well compressed ZIP files within the main ZIP are. Since they don't recursively apply their obfuscation layer to ZIP files.

e3krisztian (approved):

nyuware:
After discussion with @martonilles, I adapted the `file_utils.py` so we can override the access mode in the `File.from_path`, this way I could avoid code redundancy on my previous push

martonilles (dismissed):

martonilles (approved):

PR #538, opened 2023-03-21
Fix ELF extractor: do not repeat carved chunk offsets in output

e3krisztian:


martonilles:
Just wondering, could we have 2 ELF binaries embedded in 1 file, wouldn't in that case the 2 carved.elf collide?

qkaiser:
> Just wondering, could we have 2 ELF binaries embedded in 1 file, wouldn't in that case the 2 carved.elf collide?

Not really, see below for two ELF files next to each other in a bigger file with 3 bytes prefix and suffix:

```
/tmp/out/dual_elf.bin_extract/276420-276424.unknown
/tmp/out/dual_elf.bin_extract/4-138212.elf64_extract/carved.elf
/tmp/out/dual_elf.bin_extract/0-4.unknown
/tmp/out/dual_elf.bin_extract/138212-276420.elf64_extract/carved.elf
```

unblob/handlers/compression/gzip.py
```diff
@@ -60,7 +60,7 @@ class GZIPHandler(Handler):
         )
     ]
 
-    EXTRACTOR = Command("7z", "x", "-y", "{inpath}", "-o{outdir}")
+    EXTRACTOR = Command("7z", "x", "-y", "{inpath}", "-so", stdout="gzip.uncompressed")
```
qkaiser, line 63: Yeah I'm definitely against loosing the potential name information, it's a useful bit of information.

Some test I did with **very** rough edges:

```diff
diff --git a/unblob/handlers/compression/gzip.py b/unblob/handlers/compression/gzip.py
index 275d938..795fd35 100644
--- a/unblob/handlers/compression/gzip.py
+++ b/unblob/handlers/compression/gzip.py
@@ -13,6 +13,7 @@ library tries to read the next stream header.
 """
 import gzip
 import io
+import struct
 import zlib
 from typing import Optional

@@ -29,7 +30,8 @@ logger = get_logger()
 GZIP2_CRC_LEN = 4
 GZIP2_SIZE_LEN = 4
 GZIP2_FOOTER_LEN = GZIP2_CRC_LEN + GZIP2_SIZE_LEN
-
+FLAG_EXTRA = 4
+FLAG_NAME = 8

 class GZIPHandler(Handler):
     NAME = "gzip"
@@ -62,11 +64,36 @@ class GZIPHandler(Handler):

     EXTRACTOR = Command("7z", "x", "-y", "{inpath}", "-so", stdout="gzip.uncompressed")

+    def get_gzip_name(self, file: File) -> str:
+        name = b""
+        # skip magic bytes and method
+        file.read(2)
+        (method, flag, last_mtime) = struct.unpack("<BBIxx", file.read(8))
+
+        if flag & FLAG_EXTRA:
+            # Read & discard the extra field, if present
+            extra_len, = struct.unpack("<H", file.read(2))
+            file.seek(extra_len, io.SEEK_CUR)
+        if flag & FLAG_NAME:
+            # Read and discard a null-terminated string containing the filename
+            while True:
+                s = file.read(1)
+                if not s or s==b'\000':
+                    break
+                name += s
+        return name.decode('utf-8')
+
     def calculate_chunk(self, file: File, start_offset: int) -> Optional[ValidChunk]:
         fp = SingleMemberGzipReader(file)
+
         if not fp.read_header():
             return
+        file.seek(start_offset, io.SEEK_SET)
+        name = self.get_gzip_name(file)
+        if name:
+            self.EXTRACTOR = Command("7z", "x", "-y", "{inpath}", "-so", stdout=name)
+
         try:
             fp.read_until_eof()
         except (gzip.BadGzipFile, zlib.error) as e:
```

unblob/handlers/compression/zlib.py
```diff
@@ -16,11 +16,11 @@
 class ZlibExtractor(Extractor):
     def extract(self, inpath: Path, outdir: Path):
         decompressor = zlib.decompressobj()
-        outpath = outdir.joinpath(inpath.stem)
-        with File.from_path(inpath) as f:
+        outpath = outdir / "zlib.uncompressed"
+        with File.from_path(inpath) as f, outpath.open("wb") as outfile:
```
qkaiser, line 20: TIL a new construct !

unblob/handlers/compression/zlib.py
```diff
@@ -16,11 +16,11 @@
...
             content = f.read(DEFAULT_BUFSIZE)
             while content and not decompressor.eof:
-                outpath.write_bytes(decompressor.decompress(content))
+                outfile.write(decompressor.decompress(content))
```
qkaiser, line 23: Excellent catch on this bug !!!

unblob/handlers/compression/gzip.py
e3krisztian: Thanks, it is much easier to move forward with this code fragment!

However this rather belongs to a custom extractor. `Handler`s are technically singletons, so storing `EXTRACTOR` on it overwrites the previous value, which would cause different chunks being extracted with the last chunk's name.

PR #540, opened 2023-03-28
Use ruff instead of flake8 and isort

vlaci:
I've been experimenting with using the [ruff](https://beta.ruff.rs/docs/) linter for a long time. I am proposing to switch to it from flake8, isort as it is:

- supports auto-fixing many common errors, like unused imports
-  error reporting can be integrated to code editors
- very fast due to extensive caching
- has builtin support for a huge subset functionality of additional linters like [bugbear](https://pypi.org/project/flake8-bugbear/) and [bandit](https://bandit.readthedocs.io/en/latest/)

I enabled a bunch of extensions based on what seemed right to me, and also silenced very picky errors (khm, pylint's too-many-anything rules) but I am open to drop those you are not agreeing with.


martonilles (commented):
Looks good to me, did not check, but are there any rules that are not enabled? If not why?

vlaci:
> Looks good to me, did not check, but are there any rules that are not enabled? If not why?

There are a couple of dozen more rules available: https://beta.ruff.rs/docs/rules/

- Some of them doesn't apply for us (e.g flake8-django)
- would come with false positives (eradicate)
- found too opinionated (flake8-errmsg)
- or are meaningful but just didn't go through the ruleset (e.g. pydocstyle)

If we agree that this is a meaningful change, I'd be happy to go through the leftover ones. I also invite you to skim through the possibilities.

docs/development.md
```diff
@@ -432,6 +432,10 @@ We adhere to PEP8 and enforce proper formatting of source files using
 [black](https://github.com/psf/black) so you should not worry about formatting
 source code at all, `pre-commit` will take care of it.
 
+For linting we use [ruff](https://ruff.rs). Lint errors can be shown in your
+editor off choice by one of the [editor
```
qkaiser, line 436:
```
editor of choice by one of the [editor
```

unblob/handlers/archive/qnap/qnap_nas.py
```diff
@@ -20,7 +20,7 @@
 logger = get_logger()
 
 FOOTER_LEN = 74
-SECRET = "QNAPNASVERSION"
+SECRET = "QNAPNASVERSION"  # noqa: S105
```
qkaiser, line 23: lmao

qkaiser (commented):
Lots of love for `RET504`, the whole thing is really impressive and makes our code even cleaner.

kukovecz (approved):
I have been experimenting with `ruff` all morning, I have never seen it before, but I like this approach very much - a fine addition!

setup.cfg
```diff
@@ -1,6 +0,0 @@
-[flake8]
-# We start with a strict setting, this is ranked as A in radon
-max-complexity = 8
-ignore = E501,W503,E203
```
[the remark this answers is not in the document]
vlaci: At one rebase somewhere I dropped RUFF rules accidentally, which would take care of this

qkaiser:
@vlaci just so we're on the same page, are you currently going through all the ruff checkers to see their impact and adding them if they're ok ? Just to know if we should do something on our end 🤓

vlaci:
> @vlaci just so we're on the same page, are you currently going through all the ruff checkers to see their impact and adding them if they're ok ? Just to know if we should do something on our end nerd_face

I am almost there :)

vlaci:
Finished (MS apt repo doesn't like me...)

pyproject.toml
```diff
@@ -103,6 +104,9 @@ ignore-init-module-imports = true
   "T201"  # print
 ]
 
+[tool.ruff.flake8-builtins]
+builtins-ignorelist = ["id"]
```
qkaiser, line 121: Why specifically ignore this one ?

unblob/handlers/archive/qnap/qnap_nas.py
```diff
@@ -58,6 +58,7 @@ def is_valid_header(header: Instance) -> bool:
 def _hyperscan_match(
     context: QTSSearchContext, pattern_id: int, offset: int, end: int
 ) -> Scan:
+    del pattern_id, end
```
qkaiser, line 61: I'd rather use `noqa` than a `del` statement to force use the variable. This is super confusing if you don't know it's related to a linting rule.

pyproject.toml
```diff
@@ -83,6 +83,7 @@ select = [
   "S",      # flake8-bandit
   "SIM",    # flake8-simplify
   "SLF",    # flake8-self
+  "T10",    # flake8-debugger
```
qkaiser, line 86: Why enable it if it's to explicitly disable its check with `noqa` ?

vlaci: It could be helpful to prevent newly checked in code if it accidentally contains a set breakpoint.

unblob/handlers/archive/qnap/qnap_nas.py
kukovecz, line 61: A couple of these `del`s  were introduced due to [flake8-unused-arguments](https://beta.ruff.rs/docs/rules/#flake8-unused-arguments-arg) rule.

I have mixed feelings abut this. One thing is that I like this rule, it sounds a good thing to have. On the other hand,  using explicit `del` statement feels "weird". I know this is very opinionated but I can not phrase it better.

Is there anything else we can do here?  I would say just don't provide those parameters, but they must be there for a reason (hyperscan need that?)

If the code can not be changed that way, what else can we do? I found out starting the parameter name with underscore (`_pattern_id`, `_end` in this case) would also solve this problem (not trigger ruff). 

For me, it "looks" better - but as I was saying, this is very subjective so I can be convinced leaving this.

unblob/handlers/filesystem/romfs.py
```diff
@@ -64,11 +62,11 @@ def get_string(file: File) -> bytes:
     return filename.rstrip(b"\x00")
 
 
-class FileHeader(object):
+class FileHeader:
     addr: int
     next_filehdr: int
     spec_info: int
-    type: FS_TYPE
+    type_: FSType
```
kukovecz, line 69: I very much like [flake8-builtins](https://beta.ruff.rs/docs/rules/#flake8-builtins-a) - I don't like shadowing variables.

I think a fine(er) replacement for the occurred problems in this commit would be to rename the variables  in a more .... meaningful ... name than just adding a `_` suffix.

If you have the energy / creativity to do so that would be better ... otherwise, it is okay, I can sacrifice those names for the greater good :sweat_smile:

kukovecz (approved):
Two very small, really subjective comment for the latest changes.

Still great, still approve, love the new rules!

unblob/handlers/archive/qnap/qnap_nas.py
vlaci: I've got the idea from vulture https://github.com/jendrikseipp/vulture#marking-unused-variables
Unfortunately `noqa` prevents detecting changes in unused variables.

Another possibility could be to arrign it to `_`:

```suggestion
    _ = pattern_id, end
```

vlaci: Renaming parameters breaks the interface though, as it can no longer called with keyword arguments.

Let's continue at https://github.com/onekey-sec/unblob/pull/540#discussion_r1152900184

qkaiser: Let's keep it if it's industry standard. Maybe just a comment to make it explicit ?

```python
del pattern_id, end # recommended way to mark unused variables according to pylint
```

unblob/handlers/filesystem/romfs.py
qkaiser: Would using `fs_type` as variable name be acceptable to you @kukovecz ?

If so:  `sed -i s/type_/fs_type/g' unblob/handlers/filesystem/romfs.py`

pyproject.toml
qkaiser: Let's keep it this way. At least now it's documented why we did so :)

unblob/handlers/filesystem/romfs.py
kukovecz: Sure!

PR #542, opened 2023-03-29
feat(handler): Add support for HP BDL and HP IPKG format

nyuware:
Format is build like this BDL -> multiple IPKG
the header contains the offset at which the table of content is built,
the table of content has the offset and the size of each IPKG
the same pattern is applied for the IPKG's

```
+---------------------+
|      BDL header     |
+---------------------+
|  Table of content   |
+---------------------+
|     ipkg header 1   |
+---------------------+
|     ipkg data 1     |
+---------------------+
|     ipkg header n   |
+---------------------+
|     ipkg data n     |
+---------------------+
```

some information can be found about this format on [this repo](https://github.com/tylerwhall/hpbdl/blob/master/src/main.rs)

plenty of sample [here](https://support.hp.com/us-en/drivers/selfservice/hp-laserjet-enterprise-500-color-printer-m551-series/4184891), [here](https://support.hp.com/us-en/drivers/selfservice/hp-laserjet-enterprise-m406-series/22732206), [here](https://support.hp.com/us-en/drivers/selfservice/hp-color-laserjet-managed-mfp-m577-series/8362240), or [here](https://support.hp.com/us-en/drivers/selfservice/hp-color-laserjet-enterprise-m552-series/7345200/model/7345201)

### DIFF
--- a/unblob/handlers/__init__.py
+++ b/unblob/handlers/__init__.py
@@ -2,6 +2,7 @@ from ..models import Handlers
 from .archive import ar, arc, arj, cab, cpio, dmg, rar, sevenzip, stuffit, tar, zip
 from .archive.dlink import encrpted_img, shrs
 from .archive.engeniustech import engenius
+from .archive.hp import bdl
 from .archive.instar import bneg
 from .archive.netgear import chk, trx
 from .archive.qnap import qnap_nas
@@ -62,6 +63,7 @@ BUILTIN_HANDLERS: Handlers = (
     hdr.HDR2Handler,
     qnap_nas.QnapHandler,
     bneg.BNEGHandler,
+    bdl.HPBDLHandler,
     sparse.SparseHandler,
     ar.ARHandler,
     arc.ARCHandler,
--- a/unblob/handlers/__init__.py
+++ b/unblob/handlers/__init__.py
@@ -2,7 +2,7 @@ from ..models import Handlers
 from .archive import ar, arc, arj, cab, cpio, dmg, rar, sevenzip, stuffit, tar, zip
 from .archive.dlink import encrpted_img, shrs
 from .archive.engeniustech import engenius
-from .archive.hp import bdl
+from .archive.hp import bdl, ipkg
 from .archive.instar import bneg
 from .archive.netgear import chk, trx
 from .archive.qnap import qnap_nas
@@ -64,6 +64,7 @@ BUILTIN_HANDLERS: Handlers = (
     qnap_nas.QnapHandler,
     bneg.BNEGHandler,
     bdl.HPBDLHandler,
+    ipkg.HPIPKGHandler,
     sparse.SparseHandler,
     ar.ARHandler,
     arc.ARCHandler,
--- a/unblob/handlers/archive/hp/bdl.py
+++ b/unblob/handlers/archive/hp/bdl.py
@@ -1,0 +1,104 @@
+import io
+from pathlib import Path
+from typing import Optional
+
+from dissect.cstruct import Instance
+from structlog import get_logger
+
+from unblob.extractor import carve_chunk_to_file
+from unblob.file_utils import Endian, File, InvalidInputFormat, StructParser, snull
+from unblob.models import Chunk, Extractor, HexString, StructHandler, ValidChunk
+
+logger = get_logger()
+
+C_DEFINITIONS = r"""
+    typedef struct bdl_toc_entry {
+        uint64 offset;
+        uint64 size;
+    } bdl_toc_entry_t;
+
+    typedef struct bdl_header {
+        char magic[4];
+        uint16 major;
+        uint16 minor;
+        uint32 toc_offset;
+        char unknown[4];
+        uint32 toc_entries;
+        uint32 unknowns_2[3];
+        char release[256];
+        char brand[256];
+        char device_id[256];
+        char unknown_3[9];
+        char version[256];
+        char revision[256];
+    } bdl_header_t;
+"""
+
+
+def is_valid_header(header: Instance) -> bool:
+    if header.toc_offset == 0 or header.toc_entries == 0:
+        return False
+    try:
+        snull(header.release).decode("utf-8")
+        snull(header.brand).decode("utf-8")
+        snull(header.device_id).decode("utf-8")
+        snull(header.version).decode("utf-8")
+        snull(header.revision).decode("utf-8")
+    except UnicodeDecodeError:
+        return False
+    return True
+
+
+class HPBDLExtractor(Extractor):
+    def __init__(self):
+        self._struct_parser = StructParser(C_DEFINITIONS)
+
+    def extract(self, inpath: Path, outdir: Path):
+        entries = []
+        with File.from_path(inpath) as file:
+            header = self._struct_parser.parse("bdl_header_t", file, Endian.LITTLE)
+            file.seek(header.toc_offset, io.SEEK_SET)
+            for i in range(header.toc_entries):
+                entry = self._struct_parser.parse(
+                    "bdl_toc_entry_t", file, Endian.LITTLE
+                )
+                entries.append(
+                    (
+                        outdir.joinpath(outdir.joinpath(Path(f"ipkg{i:03}"))),
+                        Chunk(
+                            start_offset=entry.offset,
+                            end_offset=entry.offset + entry.size,
+                        ),
+                    )
+                )
+
+            for carve_path, chunk in entries:
+                carve_chunk_to_file(
+                    file=file,
+                    chunk=chunk,
+                    carve_path=carve_path,
+                )
+
+
+class HPBDLHandler(StructHandler):
+    NAME = "bdl"
+
+    PATTERNS = [HexString("69 62 64 6C 01 00 01 00")]
+
+    C_DEFINITIONS = C_DEFINITIONS
+    HEADER_STRUCT = "bdl_header_t"
+    EXTRACTOR = HPBDLExtractor()
+
+    def calculate_chunk(self, file: File, start_offset: int) -> Optional[ValidChunk]:
+        header = self.parse_header(file, endian=Endian.LITTLE)
+
+        if not is_valid_header(header):
+            raise InvalidInputFormat("Invalid BDL header.")
+
+        file.seek(start_offset + header.toc_offset, io.SEEK_SET)
+        end_offset = -1
+        for _ in range(header.toc_entries):
+            entry = self._struct_parser.parse("bdl_toc_entry_t", file, Endian.LITTLE)
+            end_offset = max(end_offset, start_offset + entry.offset + entry.size)
+
+        return ValidChunk(start_offset=start_offset, end_offset=end_offset)
--- a/unblob/handlers/archive/hp/ipkg.py
+++ b/unblob/handlers/archive/hp/ipkg.py
@@ -1,0 +1,112 @@
+import io
+from pathlib import Path
+from typing import Optional
+
+from dissect.cstruct import Instance
+from structlog import get_logger
+
+from unblob.extractor import carve_chunk_to_file, is_safe_path
+from unblob.file_utils import Endian, File, InvalidInputFormat, StructParser, snull
+from unblob.models import Chunk, Extractor, HexString, StructHandler, ValidChunk
+
+logger = get_logger()
+
+C_DEFINITIONS = r"""
+    typedef struct ipkg_file_entry {
+        char name[256];
+        uint64 offset;
+        uint64 size;
+        uint32 crc32;
+    } ipkg_toc_entry_t;
+
+    typedef struct ipkg_header {
+        char magic[4];
+        uint16 major;
+        uint16 minor;
+        uint32 toc_offset;
+        uint32 unknown_1;
+        uint32 toc_entries;
+        uint32 unknown_2[2];
+        uint32 always_null;
+        char file_version[256];
+        char product_name[256];
+        char ipkg_name[256];
+        char signature[256];
+    } ipkg_header_t;
+"""
+
+
+def is_valid_header(header: Instance) -> bool:
+    if header.toc_offset == 0 or header.toc_entries == 0:
+        return False
+    try:
+        snull(header.ipkg_name).decode("utf-8")
+        snull(header.file_version).decode("utf-8")
+        snull(header.product_name).decode("utf-8")
+    except UnicodeDecodeError:
+        return False
+    return True
+
+
+class HPIPKGExtractor(Extractor):
+    def __init__(self):
+        self._struct_parser = StructParser(C_DEFINITIONS)
+
+    def extract(self, inpath: Path, outdir: Path):
+        entries = []
+        with File.from_path(inpath) as file:
+            header = self._struct_parser.parse("ipkg_header_t", file, Endian.LITTLE)
+            file.seek(header.toc_offset, io.SEEK_SET)
+            for _ in range(header.toc_entries):
+                entry = self._struct_parser.parse(
+                    "ipkg_toc_entry_t", file, Endian.LITTLE
+                )
+                entry_path = Path(snull(entry.name).decode("utf-8"))
+                if entry_path.parent.name:
+                    raise InvalidInputFormat("Entry name contains directories.")
+                if not is_safe_path(outdir, entry_path):
+                    logger.warning(
+                        "Path traversal attempt, discarding.",
+                        outdir=outdir,
+                    )
+                    continue
+                entries.append(
+                    (
+                        outdir.joinpath(outdir / entry_path.name),
+                        Chunk(
+                            start_offset=entry.offset,
+                            end_offset=entry.offset + entry.size,
+                        ),
+                    )
+                )
+
+            for carve_path, chunk in entries:
+                carve_chunk_to_file(
+                    file=file,
+                    chunk=chunk,
+                    carve_path=carve_path,
+                )
+
+
+class HPIPKGHandler(StructHandler):
+    NAME = "ipkg"
+
+    PATTERNS = [HexString("69 70 6B 67 01 00 03 00")]
+
+    C_DEFINITIONS = C_DEFINITIONS
+    HEADER_STRUCT = "ipkg_header_t"
+    EXTRACTOR = HPIPKGExtractor()
+
+    def calculate_chunk(self, file: File, start_offset: int) -> Optional[ValidChunk]:
+        header = self.parse_header(file, endian=Endian.LITTLE)
+
+        if not is_valid_header(header):
+            raise InvalidInputFormat("Invalid IPKG header.")
+
+        file.seek(start_offset + header.toc_offset, io.SEEK_SET)
+        end_offset = -1
+        for _ in range(header.toc_entries):
+            entry = self._struct_parser.parse("ipkg_toc_entry_t", file, Endian.LITTLE)
+            end_offset = max(end_offset, start_offset + entry.offset + entry.size)
+
+        return ValidChunk(start_offset=start_offset, end_offset=end_offset)
--- a/tests/integration/archive/hp/bdl/__input__/sample.bdl
+++ b/tests/integration/archive/hp/bdl/__input__/sample.bdl
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:240533e4188e0af859989fca729c4a0847de3d5688906c4a22d36f843afdf271
+size 4043
--- a/tests/integration/archive/hp/bdl/__output__/sample.bdl_extract/ipkg000
+++ b/tests/integration/archive/hp/bdl/__output__/sample.bdl_extract/ipkg000
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:c99dd4e215b508395aed42ebfee1e87d64b6bb5551c165934fe1a759e40093ee
+size 1345
--- a/tests/integration/archive/hp/bdl/__output__/sample.bdl_extract/ipkg000_extract/sample.txt
+++ b/tests/integration/archive/hp/bdl/__output__/sample.bdl_extract/ipkg000_extract/sample.txt
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:e6478ea19e2122817e1986151847d3590e114ffd7c5d65d8d91cccea316c1cf2
+size 17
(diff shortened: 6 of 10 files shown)


unblob/handlers/archive/hp/bdl.py
```diff
@@ -0,0 +1,109 @@
...
+        header = self._struct_parser.parse("bdl_header_t", file, Endian.LITTLE)
+        file.seek(header.content_table_offset, io.SEEK_SET)
+
+        entries = [
+            (entry.offset, entry.size)
+            for entry in (
+                self._struct_parser.parse("bdl_toc_entry_t", file, Endian.LITTLE)
+                for _ in range(header.number_of_element)
+            )
+        ]
```
qkaiser, line 78: I feel like you could simplify this by doing everything in the `extract` function within a single loop.

unblob/handlers/archive/hp/ipkg.py
```diff
@@ -0,0 +1,111 @@
...
+
+        entries = [
+            (entry.file_name, entry.offset, entry.size)
+            for entry in (
+                self._struct_parser.parse("ipkg_file_entry_t", file, Endian.LITTLE)
+                for _ in range(header.number_of_files)
+            )
+        ]
+
+        return entries
```
qkaiser, line 82: Same feeling here, I think this can be simplified by putting everything in `extract` and using a single clear loop.

qkaiser (changes requested):
Nice work on this one. I strongly recommend you split your work in two separate commits:
- 1 for BDL
- 1 for IPKG

It will make the history easier to understand and easier to track changes.

unblob/handlers/archive/hp/ipkg.py
```diff
@@ -0,0 +1,111 @@
...
+
+logger = get_logger()
+
+
+C_DEFINITIONS = r"""
+    typedef struct ipkg_file_entry {
+        char file_name[256];
+        uint64 offset;
+        uint64 size;
+        char unknown[4];
```
qkaiser, line 20: maybe/probably a CRC32 as a `uint32` ?

nyuware: as discussed, sometimes the CRC is valid, sometimes it isn't, we decided not to validate it

unblob/handlers/__init__.py
```diff
@@ -2,6 +2,7 @@
 from .archive import ar, arc, arj, cab, cpio, dmg, rar, sevenzip, stuffit, tar, zip
 from .archive.dlink import encrpted_img, shrs
 from .archive.engeniustech import engenius
+from .archive.hp import bdl, ipkg
```
qkaiser, line 5: Should be:

```python
from .archive.hp import bdl
```

unblob/handlers/__init__.py
```diff
@@ -62,6 +63,8 @@
     hdr.HDR2Handler,
     qnap_nas.QnapHandler,
     bneg.BNEGHandler,
+    bdl.HPBDLHandler,
+    ipkg.HPIPKGHandler,
```
qkaiser, line 67: Should not be part of this commit.

unblob/handlers/archive/hp/bdl.py
```diff
@@ -0,0 +1,106 @@
...
+        char unknown_3[9];
+        char version[256];
+        char revision[256];
+    } bdl_header_t;
+"""
+
+
+def is_valid_header(header: Instance) -> bool:
+    if header.toc_offset == 0 and header.toc_entries == 0:
+        return False
```
qkaiser, line 40: Split into two ifs or do:

```python
if header.toc_offset == 0 or header.toc_entries == 0:
    return False
```

unblob/handlers/archive/hp/bdl.py
```diff
@@ -0,0 +1,106 @@
...
+        snull(header.brand).decode("utf-8")
+        snull(header.device_id).decode("utf-8")
+        snull(header.version).decode("utf-8")
+        snull(header.revision).decode("utf-8")
+    except UnicodeDecodeError:
+        return False
+    return True
+
+
+class BDLExtractor(Extractor):
```
qkaiser, line 56: I would name it `HPBDLExtractor` to be in line with the handler name.

unblob/handlers/archive/hp/bdl.py
```diff
@@ -0,0 +1,106 @@
...
+    def __init__(self, header_struct: str):
+        self.header_struct = header_struct
+        self._struct_parser = StructParser(C_DEFINITIONS)
```
qkaiser, line 59: You don't need the `header_struct` argument here.

```python
 def __init__(self):
    self._struct_parser = StructParser(C_DEFINITIONS)
```

unblob/handlers/archive/hp/bdl.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+class HPBDLHandler(StructHandler):
+    NAME = "bdl"
+
+    PATTERNS = [HexString("69 62 64 6C")]
+
+    C_DEFINITIONS = C_DEFINITIONS
+    HEADER_STRUCT = "bdl_header_t"
+    EXTRACTOR = BDLExtractor(C_DEFINITIONS)
```
qkaiser, line 87: Can be: `EXTRACTOR = HPBDLExtractor()`

unblob/handlers/archive/hp/ipkg.py
```diff
@@ -0,0 +1,104 @@
...
+        snull(header.product_name).decode("utf-8")
+    except UnicodeDecodeError:
+        return False
+    return True
+
+
+class IPKGExtractor(Extractor):
+    def __init__(self, header_struct: str):
+        self.header_struct = header_struct
+        self._struct_parser = StructParser(C_DEFINITIONS)
```
qkaiser, line 57: Same than BDL, you don't need `header_struct`.

unblob/handlers/archive/hp/ipkg.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+class HPIPKGHandler(StructHandler):
+    NAME = "ipkg"
+
+    PATTERNS = [HexString("69 70 6B 67")]
+
+    C_DEFINITIONS = C_DEFINITIONS
+    HEADER_STRUCT = "ipkg_header_t"
+    EXTRACTOR = IPKGExtractor(C_DEFINITIONS)
```
qkaiser, line 85:
```python
EXTRACTOR = IPKGExtractor()
```

unblob/handlers/archive/hp/ipkg.py
```diff
@@ -0,0 +1,104 @@
...
+    try:
+        snull(header.ipkg_name).decode("utf-8")
+        snull(header.file_version).decode("utf-8")
+        snull(header.product_name).decode("utf-8")
+    except UnicodeDecodeError:
+        return False
+    return True
+
+
+class IPKGExtractor(Extractor):
```
qkaiser, line 54: Use `HPIPKGExtractor` to be in line with the handler name.

unblob/handlers/archive/hp/ipkg.py
```diff
@@ -0,0 +1,104 @@
...
+
+    typedef struct ipkg_header {
+        char magic[4];
+        uint16 major;
+        uint16 minor;
+        uint32 toc_offset;
+        uint32 unknown_1;
+        uint32 toc_entries;
+        uint32 unknown_2[2];
+        uint32 null;
```
qkaiser, line 31: Use names that are not language keywords. You can use `always_null` for example.

unblob/handlers/archive/hp/bdl.py
```diff
@@ -0,0 +1,105 @@
...
+"""
+
+
+def is_valid_header(header: Instance) -> bool:
+    if header.toc_offset == 0 or header.toc_entries == 0:
+        return False
+    if header.major != 0x1:
+        return False
+    if header.minor != 0x1:
+        return False
```
martonilles, line 44: you can put the major/minor also in the pattern, so even less false-positive match happens

unblob/handlers/archive/hp/bdl.py
```diff
@@ -0,0 +1,105 @@
...
+
+    C_DEFINITIONS = C_DEFINITIONS
+    HEADER_STRUCT = "bdl_header_t"
+    EXTRACTOR = HPBDLExtractor()
+
+    def calculate_chunk(self, file: File, start_offset: int) -> Optional[ValidChunk]:
+        header = self.parse_header(file, endian=Endian.LITTLE)
+
+        if not is_valid_header(header):
+            raise InvalidInputFormat("Invalid BDL header.")
```
martonilles, line 92: I would also log the header in case it is invalid, that helps debugging

unblob/handlers/archive/hp/bdl.py
```diff
@@ -0,0 +1,105 @@
...
+        file.seek(start_offset + header.toc_offset, io.SEEK_SET)
+        entries = []
+        for _ in range(header.toc_entries):
+            entries.append(
+                self._struct_parser.parse("bdl_toc_entry_t", file, Endian.LITTLE)
+            )
+
+        end_offset = start_offset + max(
+            [(entry.offset + entry.size) for entry in entries]
+        )
```
martonilles, line 103: you do not need to store the entries just for the calculation, you can just calculate that end_offset. If there are many entries, it could use lot of memory

unblob/handlers/archive/hp/bdl.py
```diff
@@ -0,0 +1,105 @@
...
+                )
+                next_header_offset = file.tell()
+                carve_chunk_to_file(
+                    file=file,
+                    chunk=Chunk(
+                        start_offset=entry.offset, end_offset=entry.offset + entry.size
+                    ),
+                    carve_path=outdir.joinpath(outdir.joinpath(Path(f"ipkg{i}"))),
+                )
+                file.seek(next_header_offset)
```
martonilles, line 76: I would rather collect the start_offset, size instead of seeking back-and forth in the file

unblob/handlers/archive/hp/bdl.py
```diff
@@ -0,0 +1,105 @@
...
+                entry = self._struct_parser.parse(
+                    "bdl_toc_entry_t", file, Endian.LITTLE
+                )
+                next_header_offset = file.tell()
+                carve_chunk_to_file(
+                    file=file,
+                    chunk=Chunk(
+                        start_offset=entry.offset, end_offset=entry.offset + entry.size
+                    ),
+                    carve_path=outdir.joinpath(outdir.joinpath(Path(f"ipkg{i}"))),
```
martonilles, line 74: do we want to zero pad the file name?

unblob/handlers/archive/hp/bdl.py
```diff
@@ -0,0 +1,105 @@
+import io
```
martonilles, line 1: I would add a link to the specification or where more information is available on the format at least in the commit comment.

unblob/handlers/archive/hp/ipkg.py
```diff
@@ -0,0 +1,103 @@
...
+        char signature[256];
+    } ipkg_header_t;
+"""
+
+
+def is_valid_header(header: Instance) -> bool:
+    if header.major != 0x1:
+        return False
+    if header.minor != 0x3:
+        return False
```
martonilles, line 44: can also add version to the pattern?

Also do we want to support other/older versions?

unblob/handlers/__init__.py
```diff
@@ -2,7 +2,7 @@
 from .archive import ar, arc, arj, cab, cpio, dmg, rar, sevenzip, stuffit, tar, zip
```
martonilles, line 2: add reference to the specification in the commit

unblob/handlers/archive/hp/ipkg.py
```diff
@@ -0,0 +1,103 @@
...
+        file.seek(start_offset + header.toc_offset, io.SEEK_SET)
+        entries = []
+        for _ in range(header.toc_entries):
+            entries.append(
+                self._struct_parser.parse("ipkg_file_entry_t", file, Endian.LITTLE)
+            )
+
+        end_offset = start_offset + max(
+            [(entry.offset + entry.size) for entry in entries]
+        )
```
martonilles, line 101: same, no need to create a list to calculate the end_offset

unblob/handlers/archive/hp/ipkg.py
```diff
@@ -0,0 +1,103 @@
...
+    def extract(self, inpath: Path, outdir: Path):
+        with File.from_path(inpath) as file:
+            header = self._struct_parser.parse("ipkg_header_t", file, Endian.LITTLE)
+            file.seek(header.toc_offset, io.SEEK_SET)
+            for _ in range(header.toc_entries):
+                entry = self._struct_parser.parse(
+                    "ipkg_file_entry_t", file, Endian.LITTLE
+                )
+                next_header_offset = file.tell()
+                carve_chunk_to_file(
```
martonilles, line 67: need to check if entry.name does not cause any path-traversal issue, use `is_safe_path`

unblob/handlers/archive/hp/ipkg.py
```diff
@@ -0,0 +1,103 @@
...
+                    file=file,
+                    chunk=Chunk(
+                        start_offset=entry.offset, end_offset=entry.offset + entry.size
+                    ),
+                    carve_path=outdir.joinpath(Path(snull(entry.name).decode("utf-8"))),
+                )
+                file.seek(next_header_offset)
```
martonilles, line 74: same comment on seeking

martonilles (changes requested):

unblob/handlers/archive/hp/bdl.py
nyuware: I believe the 0x1 means file type, so it would stay 0x1 all the time, I removed these checks and set them in the pattern

unblob/handlers/archive/hp/ipkg.py
nyuware: I believe the 0x1 and 0x3 belong to some file type, its not really version, same comment as before, I removed them and set them in the pattern

unblob/handlers/archive/hp/bdl.py
nyuware: I was already doing that [here](https://github.com/onekey-sec/unblob/commit/66eda14d4dc60a41afd800dbc8faab7367e8c737) but Quentin said it was possible to do it in a single loop, which recommendation should I follow?

unblob/handlers/archive/hp/bdl.py
```diff
@@ -0,0 +1,104 @@
...
+        header = self.parse_header(file, endian=Endian.LITTLE)
+
+        if not is_valid_header(header):
+            raise InvalidInputFormat("Invalid BDL header.")
+
+        file.seek(start_offset + header.toc_offset, io.SEEK_SET)
+        end_offset = start_offset
+        for _ in range(header.toc_entries):
+            entry = self._struct_parser.parse("bdl_toc_entry_t", file, Endian.LITTLE)
+            end_offset = max(end_offset, entry.offset + entry.size)
```
qkaiser, line 102:
```suggestion
            end_offset = max(end_offset, sart_offset + entry.offset + entry.size)
```

unblob/handlers/archive/hp/bdl.py
```diff
@@ -0,0 +1,104 @@
...
+    EXTRACTOR = HPBDLExtractor()
+
+    def calculate_chunk(self, file: File, start_offset: int) -> Optional[ValidChunk]:
+        header = self.parse_header(file, endian=Endian.LITTLE)
+
+        if not is_valid_header(header):
+            raise InvalidInputFormat("Invalid BDL header.")
+
+        file.seek(start_offset + header.toc_offset, io.SEEK_SET)
+        end_offset = start_offset
```
qkaiser, line 99: It should be explicit that `end_offset` is not set to a good value:

```suggestion
        end_offset = -1
```

unblob/handlers/archive/hp/ipkg.py
```diff
@@ -0,0 +1,106 @@
...
+    def extract(self, inpath: Path, outdir: Path):
+        entries = []
+        with File.from_path(inpath) as file:
+            header = self._struct_parser.parse("ipkg_header_t", file, Endian.LITTLE)
+            file.seek(header.toc_offset, io.SEEK_SET)
+            for _ in range(header.toc_entries):
+                entry = self._struct_parser.parse(
+                    "ipkg_toc_entry_t", file, Endian.LITTLE
+                )
+                if not is_safe_path(outdir, Path(str(entry.name))):
```
qkaiser, line 62: Converting bytes to str with `str()` is not right. `b"hello"` will be converted into `'b"hello"'`.

```suggestion
                if not is_safe_path(outdir, Path(entry.name.decode('utf-8'))):
```

unblob/handlers/archive/hp/ipkg.py
```diff
@@ -0,0 +1,106 @@
...
+                    logger.warning(
+                        "Path traversal attempt, discarding.", output_path=entry.name
+                    )
+                    return
```
qkaiser, line 66: Just skip this entry, not the entire thing.

```suggestion
                    continue
```

unblob/handlers/archive/hp/ipkg.py
```diff
@@ -0,0 +1,106 @@
...
+        header = self.parse_header(file, endian=Endian.LITTLE)
+
+        if not is_valid_header(header):
+            raise InvalidInputFormat("Invalid IPKG header.")
+
+        file.seek(start_offset + header.toc_offset, io.SEEK_SET)
+        end_offset = start_offset
+        for _ in range(header.toc_entries):
+            entry = self._struct_parser.parse("ipkg_toc_entry_t", file, Endian.LITTLE)
+            end_offset = max(end_offset, entry.offset + entry.size)
```
qkaiser, line 104:
```suggestion
            end_offset = max(end_offset, start_offset + entry.offset + entry.size)
```

unblob/handlers/archive/hp/ipkg.py
```diff
@@ -0,0 +1,106 @@
...
+    EXTRACTOR = HPIPKGExtractor()
+
+    def calculate_chunk(self, file: File, start_offset: int) -> Optional[ValidChunk]:
+        header = self.parse_header(file, endian=Endian.LITTLE)
+
+        if not is_valid_header(header):
+            raise InvalidInputFormat("Invalid IPKG header.")
+
+        file.seek(start_offset + header.toc_offset, io.SEEK_SET)
+        end_offset = start_offset
```
qkaiser, line 101:
```suggestion
        end_offset = -1
```

qkaiser (approved):

unblob/handlers/archive/hp/ipkg.py
```diff
@@ -0,0 +1,108 @@
...
+    def extract(self, inpath: Path, outdir: Path):
+        entries = []
+        with File.from_path(inpath) as file:
+            header = self._struct_parser.parse("ipkg_header_t", file, Endian.LITTLE)
+            file.seek(header.toc_offset, io.SEEK_SET)
+            for _ in range(header.toc_entries):
+                entry = self._struct_parser.parse(
+                    "ipkg_toc_entry_t", file, Endian.LITTLE
+                )
+                if not is_safe_path(outdir, Path(snull(entry.name).decode("utf-8"))):
```
martonilles, line 62: I would put `Path(snull(entry.name).decode("utf-8"))` into a variable so no need to process it multiple times

unblob/handlers/archive/hp/ipkg.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def calculate_chunk(self, file: File, start_offset: int) -> Optional[ValidChunk]:
+        header = self.parse_header(file, endian=Endian.LITTLE)
+
+        if not is_valid_header(header):
+            raise InvalidInputFormat("Invalid IPKG header.")
+
+        file.seek(start_offset + header.toc_offset, io.SEEK_SET)
+        end_offset = -1
+        for _ in range(header.toc_entries):
```
martonilles, line 104: what if toc_entries == 0? we would have a strange end_offset?

unblob/handlers/archive/hp/ipkg.py
```diff
@@ -0,0 +1,108 @@
...
+                if not is_safe_path(outdir, Path(snull(entry.name).decode("utf-8"))):
+                    logger.warning(
+                        "Path traversal attempt, discarding.",
+                        outdir=outdir,
+                        outfile=Path(snull(entry.name).decode("utf-8")),
+                    )
+                    continue
+                entries.append(
+                    (
+                        outdir.joinpath(Path(snull(entry.name).decode("utf-8"))),
```
martonilles, line 71: if entry.name contains a `/` character we would fail as we do not create the parent directory. I do not know if it valid and we should handle it or we can just fail.

martonilles (dismissed):

unblob/handlers/archive/hp/ipkg.py
nyuware: I do this check `if header.toc_offset == 0 or header.toc_entries == 0:` on the `is_valid_header`  function in the BDL handler, I can do it too on the ipkg handler

unblob/handlers/__init__.py
qkaiser: There's no open specs. Some online content that talks about it:
- https://foxglovesecurity.com/2017/11/20/a-sheep-in-wolfs-clothing-finding-rce-in-hps-printer-fleet/ (hardly understood the whole format)
- https://labs.withsecure.com/content/dam/labs/docs/Printing-Shellz.pdf (first reversing)
- https://github.com/tylerwhall/hpbdl (extractor in Rust)

unblob/handlers/archive/hp/ipkg.py
```diff
@@ -0,0 +1,112 @@
...
+        with File.from_path(inpath) as file:
+            header = self._struct_parser.parse("ipkg_header_t", file, Endian.LITTLE)
+            file.seek(header.toc_offset, io.SEEK_SET)
+            for _ in range(header.toc_entries):
+                entry = self._struct_parser.parse(
+                    "ipkg_toc_entry_t", file, Endian.LITTLE
+                )
+                entry_path = Path(snull(entry.name).decode("utf-8"))
+                if "/" in entry_path.parts:
+                    raise InvalidInputFormat("File contains a forward slash")
```
qkaiser, line 66: This won't work if `entry.name = "tmp/test/lol`:

```
>>> Path("tmp/test/lol").parts
('tmp', 'test', 'lol')
```

```suggestion
                if entry_path.parent.name:
                    raise InvalidInputFormat("Entry name contains directories.")
```

qkaiser (approved):

PR #545, opened 2023-04-03
fix(reporting): Add StatReport for every Task

kukovecz:
Unblob should produce a StatReport for every Task which it needs to handle. Currently it does not do it in 2 cases:
    1. When the max depth limit is reached
    2. When the to-be-processed file's path is considered invalid

In both cases it would be nice to see the actual report of the not-processed file(s).

### DIFF
--- a/unblob/processing.py
+++ b/unblob/processing.py
@@ -235,6 +235,8 @@ class Processor:
         logger.exception("Unknown error happened", exc_info=exc)
 
     def _process_task(self, result: TaskResult, task: Task):
+        stat_report = StatReport.from_path(task.path)
+        result.add_report(stat_report)
         log = logger.bind(path=task.path)
 
         if task.depth >= self._config.max_depth:
@@ -245,9 +247,6 @@ class Processor:
         if not valid_path(task.path):
             log.warning("Path contains invalid characters, it won't be processed")
             return
-
-        stat_report = StatReport.from_path(task.path)
-        result.add_report(stat_report)
 
         if stat_report.is_dir:
             log.debug("Found directory")


qkaiser:
We had a discussion with @vlaci about the relevance of `is_valid_path` since this was introduced to handle crashes triggered by `yara-python` not handling those files well (see https://github.com/onekey-sec/unblob/commit/e903602c94e4ca550911e3e68e01f0b09e48012e).

I think we could try removing `is_valid_path` as part of this MR. Here's a sample CAB file with filenames holding non UTF-8 bytes (`0xFF 0xF2 0x01 0x02`): 
[fruits.cab.gz](https://github.com/onekey-sec/unblob/files/11137003/fruits.cab.gz)

kukovecz:
> We had a discussion with @vlaci about the relevance of `is_valid_path` since this was introduced to handle crashes triggered by `yara-python` not handling those files well (see [e903602](https://github.com/onekey-sec/unblob/commit/e903602c94e4ca550911e3e68e01f0b09e48012e)).
> 
> I think we could try removing `is_valid_path` as part of this MR. Here's a sample CAB file with filenames holding non UTF-8 bytes (`0xFF 0xF2 0x01 0x02`): [fruits.cab.gz](https://github.com/onekey-sec/unblob/files/11137003/fruits.cab.gz)

Yeah, makes sense!

I have did some testing and simply removing the `valid_path` in the `_process_task` is not enough, it can trigger another kind of unicode errors. 

My suggestion is to handle removing the unicode validation in a different Issue. I will create it and link this MR to it for more context.

vlaci (approved):

PR #548, opened 2023-04-04
feat(processing): Re-add non-POSIX file path handling

kukovecz:
The sole reason of the existence and usage of the valid_path function (to limit unblob not to handle files with non-POSIX path) was added because of yara-python limitations in commit
e903602c94e4ca550911e3e68e01f0b09e48012e.

Since we are not using yara-python anymore, these limitations can now be lifted!

The only consequence what I discovered during testing is that the logger's format needed to be extended. It was already implemented to encode Path-strings with surrogateescape, but paths can be logged in strings as well (for example when logging the extractor command, which contains the path of the file to be extracted).

### DIFF
--- a/tests/test_file_utils.py
+++ b/tests/test_file_utils.py
@@ -1,5 +1,4 @@
 import io
-from pathlib import Path
 from typing import List
 
 import pytest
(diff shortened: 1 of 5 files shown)


kukovecz:
> darwin

Makes sense to me. I have added a `pytest.skipif` into the new test to skip it on `Darwin`

qkaiser (approved):

e3krisztian (approved):

PR #552, opened 2023-04-07
upgrade sasquatch

qkaiser:
Upgrade sasquatch version to include our fix for https://github.com/onekey-sec/sasquatch/issues/14.

Added integration tests to cover the specific use of LZMA adaptive in squashfs.


vlaci (approved):
Not an issue but wanted to mention that you can do targeted updates of flake.lock using
`nix flake lock --update-input sasquatch --commit-lockfile`

Issue #16, opened 2021-11-22
Metadata file


kissgyorgy:
Store the metadata file in `extract_root` in one JSON file.

We don't want to pollute the extracted folder with lots of small files.  
It's nice if this is easy to read, so a JSON is easy to look at.

martonilles:
we have a root File and in case of processing a directory, then we have a list of root Files
FilesystemObject
- parent (Chunk, null for root)
- children (list of Chunk, could be zero)
- path
- type (File, Directory, Device, Symlink etc.)
- permission, ownership, timestamp, acl etc. (coming from the handler which extract metadata from the chunk, otherwise leave as null)
- magic/mime
- NB: we want to record metadata on "files" that are not written as part of the extraction (eg: char devices from squashfs)

Chunk
- parent (File)
- children (list of Files, could be zero)
- start/end offset
- length
- type (handler)
- tags: encryption, 
- metadata key/values (FIXME)

Questions:
- how can we get the metadata? (can we get it from the extractors, are they smart enough)?
- if metadata gathering is expensive, we should probably make those optional

qkaiser:
Almost all of the information described above is now part of the reporting feature of unblob.

The information that is missing right now:
1. meta-data about files that were not created because we run without elevated privileges (block devices, character devices) 
2. exact permission, ownership, and timestamp information on every file

I don't think item 1 has a lot of added value right now. Regarding item 2, we already have the structure in place to collect that information. What remains is making sure the extraction phase **preserve** that information so that we can simply `stat` the file for details.

I would take care of item 2 in two steps:
1. add permission, ownership, and timestamps to `StatReport`s
2. once it's there, spend time making sure we extract or use extractors in a way that they preserve that information whenever they can. We already collected that information at https://unblob.org/formats/, and it's something our intern can do :)

On top of that, I would like to add a specific feature to our meta-data collection effort: saving header information. The idea is to  have a `metadata` field as part of our `ChunkReport`s, which is simply a dict where the handler developer can put relevant information, such as parsed headers.

I submitted a PR to `dissect.cstruct` going into that direction (see https://github.com/fox-it/dissect.cstruct/pull/29).

The idea behind this is to expose metadata to further analysis steps through the unblob report (e.g. a binary analysis toolkit would read the load address and architecture from a uImage chunk to analyze the file extracted from that chunk with the right settings).

All of these changes are quite simple to implement since reporting is already there:

```diff
[...]

PR #557, opened 2023-04-14
feat(reporting): report meta-data information about chunks.

qkaiser:
Allow handlers to provide a dict value as part of a `ValidChunk` metadata attribute. That dictionary can contain any relevant metadata information from the perspective of the handler, but we advise handler writers to report parsed information such as header values.

This metadata dict is later reported as part of our `ChunkReports` and available in the JSON report file if the user requested one.

The idea is to expose metadata to further analysis steps through the unblob report. For example, a binary analysis toolkit would read the load address and architecture from a uImage chunk to analyze the file extracted from that chunk with the right settings.

**A note on the 'as_dict' implementation.**

The initial idea was to implement it in dissect.cstruct (see https://github.com/fox-it/dissect.cstruct/pull/29), but due to expected changes in the project's API I chose to implement it in unblob so we're not dependent on another project.

Related to #16 and initial discussion in https://github.com/onekey-sec/unblob/issues/16#issuecomment-1493368023

---

You can observe the changes like this:

```
poetry run unblob -vvv -e /tmp/out -f --report /tmp/report.json tests/integration/archive/sevenzip/__input__/cherry.7z

cat /tmp/report.json| jq '.[0].reports[-1].metadata'
{
  "magic": "7z��'\u001c",
  "version_maj": 0,
  "version_min": 4,
  "crc": 3845252377,
  "next_header_offset": 147,
  "next_header_size": 33,
  "next_header_crc": 2089891445
}
```

### DIFF
--- a/python/unblob/handlers/archive/sevenzip.py
+++ b/python/unblob/handlers/archive/sevenzip.py
@@ -24,6 +24,7 @@ from pathlib import Path
 from structlog import get_logger
 
 from unblob.extractors import Command
+from unblob.report import MetadataReport, register_report_type
 
 from ...extractors.command import MultiFileCommand
 from ...file_utils import Endian, InvalidInputFormat, StructParser
@@ -77,6 +78,14 @@ def calculate_sevenzip_size(header) -> int:
     return len(header) + header.next_header_offset + header.next_header_size
 
 
+@register_report_type
+class SevenZipMetadataReport(MetadataReport):
+    """Report for 7-Zip metadata."""
+
+    version_major: int
+    version_minor: int
+
+
 class SevenZipHandler(StructHandler):
     NAME = "sevenzip"
 
@@ -106,6 +115,14 @@ class SevenZipHandler(StructHandler):
         limitations=[],
     )
 
+    def get_metadata_reports(self, header) -> list[MetadataReport]:
+        return [
+            SevenZipMetadataReport(
+                version_major=int(header.version_maj),
+                version_minor=int(header.version_min),
+            )
+        ]
+
     def calculate_chunk(self, file: File, start_offset: int) -> ValidChunk | None:
         header = self.parse_header(file)
 
@@ -113,7 +130,13 @@ class SevenZipHandler(StructHandler):
 
         size = calculate_sevenzip_size(header)
 
-        return ValidChunk(start_offset=start_offset, end_offset=start_offset + size)
+        metadata_reports = self.get_metadata_reports(header)
+
+        return ValidChunk(
+            start_offset=start_offset,
+            end_offset=start_offset + size,
+            metadata_reports=metadata_reports,
+        )
 
 
 class MultiVolumeSevenZipHandler(DirectoryHandler):
--- a/python/unblob/models.py
+++ b/python/unblob/models.py
@@ -19,6 +19,7 @@ from .report import (
     CarveDirectoryReport,
     ChunkReport,
     ErrorReport,
+    MetadataReport,
     MultiFileReport,
     RandomnessReport,
     Report,
@@ -181,6 +182,7 @@ class ValidChunk(Chunk):
 
     handler: Handler = attrs.field(init=False, eq=False)
     is_encrypted: bool = attrs.field(default=False)
+    metadata_reports: list[MetadataReport] = attrs.field(factory=list, kw_only=True)
 
     def extract(self, inpath: Path, outdir: Path) -> ExtractResult | None:
         if self.is_encrypted:
@@ -201,6 +203,7 @@ class ValidChunk(Chunk):
             size=self.size,
             handler_name=self.handler.NAME,
             is_encrypted=self.is_encrypted,
+            metadata_reports=self.metadata_reports,
             extraction_reports=extraction_reports,
         )
 
@@ -246,6 +249,7 @@ class PaddingChunk(Chunk):
             is_encrypted=False,
             handler_name="padding",
             extraction_reports=[],
+            metadata_reports=[],
         )
 
 
@@ -266,6 +270,7 @@ class MultiFile(Blob):
             paths=self.paths,
             handler_name=self.handler.NAME,
             extraction_reports=extraction_reports,
+            metadata_reports=[],
         )
 
 
--- a/python/unblob/models.py
+++ b/python/unblob/models.py
@@ -18,6 +18,7 @@ from .parser import hexstring2regex
 from .report import (
     CarveDirectoryReport,
     ChunkReport,
+    EncryptionMetadataReport,
     ErrorReport,
     MetadataReport,
     MultiFileReport,
@@ -181,8 +182,15 @@ class ValidChunk(Chunk):
     """Known to be valid chunk of a File, can be extracted with an external program."""
 
     handler: Handler = attrs.field(init=False, eq=False)
-    is_encrypted: bool = attrs.field(default=False)
     metadata_reports: list[MetadataReport] = attrs.field(factory=list, kw_only=True)
+
+    @property
+    def is_encrypted(self) -> bool:
+        return any(
+            report.is_encrypted
+            for report in self.metadata_reports
+            if isinstance(report, EncryptionMetadataReport)
+        )
 
     def extract(self, inpath: Path, outdir: Path) -> ExtractResult | None:
         if self.is_encrypted:
@@ -202,7 +210,6 @@ class ValidChunk(Chunk):
             end_offset=self.end_offset,
             size=self.size,
             handler_name=self.handler.NAME,
-            is_encrypted=self.is_encrypted,
             metadata_reports=self.metadata_reports,
             extraction_reports=extraction_reports,
         )
@@ -246,7 +253,6 @@ class PaddingChunk(Chunk):
             start_offset=self.start_offset,
             end_offset=self.end_offset,
             size=self.size,
-            is_encrypted=False,
             handler_name="padding",
             extraction_reports=[],
             metadata_reports=[],
--- a/python/unblob/report.py
+++ b/python/unblob/report.py
@@ -6,7 +6,7 @@ import stat
 import traceback
 from enum import Enum
 from pathlib import Path
-from typing import TYPE_CHECKING, Annotated, Any
+from typing import TYPE_CHECKING, Annotated, Any, TypeVar
 
 if TYPE_CHECKING:
     from collections.abc import Iterable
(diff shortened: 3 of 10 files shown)


e3krisztian (dismissed):

unblob/file_utils.py
```diff
@@ -311,3 +311,26 @@ def read_until_past(file: File, pattern: bytes):
             return file.tell()
         if next_byte not in pattern:
             return file.tell() - 1
+
+
+def as_dict(obj) -> dict:
```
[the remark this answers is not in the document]
qkaiser: I moved it to models and used a `converter` as you suggested. Yes it's related to the fact that `Instance` objects from `dissect.cstruct` can't be pickled by multi-processing (they're not serializable). I initially submitted a PR to the maintainer of `dissect.cstruct` to make them serializable (see https://github.com/fox-it/dissect.cstruct/pull/29), but given the answer I thought it better to do it on our end.

unblob/models.py
```diff
@@ -190,7 +204,7 @@ def default(self, obj):
             try:
                 return obj.decode()
             except UnicodeDecodeError:
-                return str(obj)
+                return obj.decode("utf-8", errors="surrogateescape")
```
e3krisztian, line 207: `try`-`except` construct is no longer needed, a simple `obj.decode("utf-8", errors="surrogateescape")` will always succeed.

unblob/models.py
```diff
@@ -88,6 +100,7 @@ class ValidChunk(Chunk):
 
     handler: "Handler" = attr.ib(init=False, eq=False)
     is_encrypted: bool = attr.ib(default=False)
```
martonilles, line 102: just wondering that now we have metadata, is_encrypted is metadata or still a member of valid chunk?

unblob/report.py
```diff
@@ -181,6 +181,7 @@ class ChunkReport(Report):
     end_offset: int
     size: int
     is_encrypted: bool
+    metadata: dict = attr.ib(factory=dict)
```
martonilles, line 184: just wondering if we want to validate metadata dict, do we want to enforce that key is a string and value is of a certain type, or we are ok we anything, even nested meta-data?

unblob/handlers/archive/sevenzip.py
```diff
@@ -70,4 +70,6 @@ def calculate_chunk(self, file: File, start_offset: int) -> Optional[ValidChunk]
         # We read the signature header here to get the offset to the header database
         first_db_header = start_offset + len(header) + header.next_header_offset
         end_offset = first_db_header + header.next_header_size
-        return ValidChunk(start_offset=start_offset, end_offset=end_offset)
+        return ValidChunk(
+            start_offset=start_offset, end_offset=end_offset, metadata=header
```
martonilles, line 74: do we want to pass all attributes from the header as metadata?

unblob/report.py
martonilles: Maybe it would be great to somehow have a "namespace" or at least some convention on metadata variable naming?

What if we want to push data from multiple headers, or permissions, etc?

unblob/models.py
qkaiser: I would move it to metadata and adapt handlers accordingly.

unblob/handlers/archive/sevenzip.py
qkaiser: This point came up when discussing with @e3krisztian yesterday. I think it's better to only pass the most relevant header attributes rather than the whole instance.

unblob/report.py
qkaiser: I'm ok with enforcing a convention on metadata variable naming. Having a namespace would be too complicated since we can't foresee the metadata field names used by handlers.

I would enforce that metadata is a dict without nested data, keys must be strings and values must be base types.

I would convey information about files created (timestamps, permissions, owner) with something different since it involves way more complex structures.

unblob/models.py
qkaiser: After some code writing I think `is_encrypted` is better there as an attribute than within `metadata`.

unblob/report.py
qkaiser: Added a validator. See 99944e3

unblob/handlers/archive/sevenzip.py
qkaiser: See 75548d2

python/unblob/handlers/archive/sevenzip.py
orosam, line 1: I'd split the sevenzip-specific change off to a separate commit.

python/unblob/models.py
```diff
@@ -181,6 +181,9 @@ class ValidChunk(Chunk):
 
     handler: Handler = attrs.field(init=False, eq=False)
     is_encrypted: bool = attrs.field(default=False)
+    metadata: dict = attrs.field(
+        factory=dict, validator=attrs.validators.instance_of(dict)
+    )
```
orosam: I was thinking that we could use some simple class here instead of a plain dict. Apart from having something typed, it would be easier to introduce validations, without breaking its API.
Nothing fancy, just a simple pydantic class for instance that interested handlers can subclass and extend with their fields. And then need not worry about key typos.

Something like
```python
class ChunkMetadata(BaseModel):
    """Subclass this per handler to report chunk metadata.

    Add typed fields for whatever data the handler wants to report.
    The class has to be JSON-serializable.
    """

    model_config = ConfigDict(extra="forbid")
```

Then e.g. in sevenzip:
```python
class SevenZipMetadata(ChunkMetadata):
    version_major: int
    version_minor: int
```

And then, here:
```suggestion
    metadata: ChunkMetadata = attrs.field(
        factory=ChunkMetadata, validator=attrs.validators.instance_of(ChunkMetadata)
    )
```

This can then be converted to a dict in `as_report()`, see there.

What do you think?

python/unblob/models.py
```diff
@@ -201,6 +204,7 @@ def as_report(self, extraction_reports: list[Report]) -> ChunkReport:
             size=self.size,
             handler_name=self.handler.NAME,
             is_encrypted=self.is_encrypted,
+            metadata=self.metadata,
```
orosam: Then this would become:

```suggestion
            metadata=self.metadata.model_dump(mode="json"),
```

So ChunkReport would hold a simple dict.

python/unblob/models.py
martonilles: I am all for structure! :)

Wrote earlier that it would be great to have some field taxonomy, so that it is not just random metadata fields for every chunk type as that would make parsing and processing easier. It is ok to have chunk specific fields, but would be nice to standardize some fields, either through some pre-defined data class and class magic or some other means.

orosam: I explored some ideas, but couldn't settle on something I was content with...

* It either became a bit baroque (e.g. multi-level),
* Couldn't see (yet) what to standardize,
* Wasn't clear how to add (or change) standard fields later (and what that would mean for e.g. out-of-tree handlers).

But yes, pre-defined fields? types? would be nice. I'm very much open to ideas.

BTW, `ChunkMetadata` can also be nested. So we could even provide complete metadata blocks.

qkaiser: I would go with `ChunkMetadata` with every handler subclassing it. On top of that, I would attach the `ChunkMetadata` subclass to the handler (in a similar way to `EXTRACTOR`) so that a meta-data consumer can lookup the pydantic `BaseModel` it needs to validate it.

python/unblob/handlers/archive/sevenzip.py
qkaiser: done

qkaiser:
@orosam please check the new implementation, I think we're getting somewhere :)

python/unblob/models.py
```diff
@@ -201,6 +214,7 @@ def as_report(self, extraction_reports: list[Report]) -> ChunkReport:
             size=self.size,
             handler_name=self.handler.NAME,
             is_encrypted=self.is_encrypted,
+            metadata=self.metadata.model_dump(mode="json"),
```
qkaiser: `metadata` should be `metadata_reports`, a list of Report objects (probably `MetadataReport`).

python/unblob/models.py
```diff
@@ -201,6 +214,7 @@ def as_report(self, extraction_reports: list[Report]) -> ChunkReport:
             size=self.size,
             handler_name=self.handler.NAME,
             is_encrypted=self.is_encrypted,
```
qkaiser: this should probably be moved to a MetadataReport

python/unblob/report.py
```diff
@@ -230,12 +234,18 @@ class ChunkReport(Report):
...
     is_encrypted: bool
     extraction_reports: list[Report]
+    metadata_reports: list[MetadataReport] = Field(default_factory=list)
 
     @field_validator("extraction_reports", mode="before")
     @classmethod
     def validate_extraction_reports(cls, value: Any) -> list[Report]:
         return validate_report_list(value)
 
+    @field_validator("metadata_reports", mode="before")
```
orosam, line 245: It's the same as `validate_extraction_reports()`. 

Wouldn't renaming that to `validate_reports` and adding `"metadata_reports"` as another arg work? Like e.g. in `ExtractCommandFailedReport.decode_bytes()`?

Same for MultiFileReport (and an update in `vulture_whitelist.py`).

tests/test_models.py
```diff
@@ -4,6 +4,7 @@
 import pytest
 
 from unblob.file_utils import InvalidInputFormat
+from unblob.handlers.archive.sevenzip import SevenZipMetadataReport
```
orosam, line 7: This `SevenZipMetadataReport` doesn't yet exist at this point, plus it's a bit awkward to depend on a handler specific class in a model test like this.

We can just create a test metadata report class here, register it manually in the test, and use that.
Like in `test_custom_report_registration_and_deserialization()`.

Or if we want to test the decorator style, that should also work fine in the test context.

python/unblob/report.py
```diff
@@ -222,20 +223,32 @@
     chi_square: RandomnessMeasurements
 
 
-class ChunkReport(Report):
+class MetadataReport(Report): ...
```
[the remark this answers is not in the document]
qkaiser: Fixed

python/unblob/report.py
qkaiser: done

tests/test_models.py
qkaiser: done

orosam (approved):

PR #558, opened 2023-04-17
Report calculated entropies

e3krisztian:
Extended existing code, so that entropy can be reported for outside use.

Refactor:
- shuffled around code to do only one thing in entropy-calculation functions

Fixes:
- typing for alternative imports from `unblob.math`
- logging of entropy charts, which was unfortunately broken due to global state in the `plotext` library

### DIFF
--- a/tests/test_processing.py
+++ b/tests/test_processing.py
@@ -2,22 +2,30 @@ import platform
 import sys
 import zipfile
 from pathlib import Path
-from typing import Collection, List, Tuple
+from typing import Collection, List, Tuple, Type, TypeVar
 
 import attr
 import pytest
 
+from unblob import handlers
 from unblob.models import UnknownChunk, ValidChunk
 from unblob.processing import (
     ExtractionConfig,
     calculate_buffer_size,
     calculate_entropy,
     calculate_unknown_chunks,
-    draw_entropy_plot,
+    format_entropy_plot,
     process_file,
     remove_inner_chunks,
 )
-from unblob.report import ExtractDirectoryExistsReport, StatReport
+from unblob.report import (
+    EntropyReport,
+    ExtractDirectoryExistsReport,
+    StatReport,
+    UnknownChunkReport,
+)
+
+T = TypeVar("T")
 
 
 def assert_same_chunks(expected, actual, explanation=None):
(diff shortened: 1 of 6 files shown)


unblob/processing.py
```diff
@@ -345,15 +347,26 @@ def _process_chunks(
 
         for chunk in unknown_chunks:
             carved_unknown_path = carve_unknown_chunk(self.carve_dir, file, chunk)
-            self._calculate_entropy(carved_unknown_path)
-            self.result.add_report(chunk.as_report())
+            entropy = self._calculate_entropy(carved_unknown_path)
+            self.result.add_report(chunk.as_report(entropy=entropy))
```
qkaiser, line 351: I don't understand why you add the `EntropyReport` to the results in two different ways. One line 336-337 you check if entropy is set and then add it as is, on this line you don't check if it is set and add it using `chunk.as_report()`.

unblob/math.py
```diff
@@ -1,3 +1,7 @@
+from typing import Callable
+
+shannon_entropy: Callable[[bytes], float]
```
qkaiser, line 3: Out of curiosity, was the lack of typing reported by ruff or causing issues ? What got you to write this ?

tests/test_processing.py
```diff
@@ -129,32 +129,34 @@ def test_calculate_buffer_size(
     )
 
 
-def test_draw_entropy_plot_error():
+def test_format_entropy_plot_error():
     with pytest.raises(TypeError):
-        draw_entropy_plot([])
+        format_entropy_plot(percentages=[], buffer_size=1024)
```
qkaiser, line 134: I would also add a test case where we calculate the entropy of an empty file.

```python
p = Path("/tmp/empty")
p.touch()
report = calculate_entropy(p)
```

unblob/processing.py
e3krisztian: yes, because the previous instance is not chunk related, the entropy report is added to the result directly *if* it was calculated, while in the current case the entropy report is part of the chunk-report (or not), but the chunk report can be added unconditionally.

unblob/math.py
e3krisztian: This is older than ruff, it is pyright LSP (editor only problem), it has bothered me for a year.

This is how importing in `processing` looks to  me without it:
shannon_entropy-pyright

tests/test_processing.py
e3krisztian: I am not sure about testing exactly this bit (in fact I wanted to delete this test as well :)), as it looks a very corner-case, which is already handled by existing code at https://github.com/onekey-sec/unblob/blob/a2c396fdd765006336bade6648aa9e41a6ad7847/unblob/processing.py#L275 but some test will be added to check, that entropy is reported when processing a file.

qkaiser: Makes sense.

qkaiser (approved):

e3krisztian:
Since the approval a test has been added for reporting entropy for unknown chunks and for files, that are not extracted further - except for empty files.
And also rebased to main.

PR #566, opened 2023-04-25
fix(extractor): handle OSError in fix_extracted_directory.

qkaiser:
The following error was triggered when enumerating extraction directory using Path.rglob("*"):

```
OSError: [Errno 36] File name too long: 'image.gz_extract/sav_extract/long_link'
```

This is due to pathlib calling stat() on every file, recursively:

```
stat("image.gz_extract/sav_extract/long_link", 0x7ffe5dbf94b0) = -1 ENAMETOOLONG (File name too long)
```

It seems to be unique to `pathlib` so I switched over to `glob`.

I added an integration test file that was triggering the bug. It's taken from e2fsprogs (https://github.com/tytso/e2fsprogs/tree/master/tests).


unblob/testing.py
```diff
@@ -21,7 +22,9 @@ def configure_logging():  # noqa: PT004
 
 
 def gather_integration_tests(test_data_path: Path):
-    test_input_dirs = list(test_data_path.glob("**/__input__"))
+    test_input_dirs = [
+        Path(p) for p in glob.glob(test_data_path.as_posix() + "/**/**/__input__")
```
e3krisztian, line 26:
```py
        Path(p) for p in glob.iglob(f"{test_data_path}/**/__input__")
```

e3krisztian (changes requested):

unblob/extractor.py
```diff
@@ -86,7 +87,7 @@ def fix_symlink(path: Path, outdir: Path, task_result: TaskResult) -> Path:
 
 def fix_extracted_directory(outdir: Path, task_result: TaskResult):
     fix_permission(outdir)
-    for path in outdir.rglob("*"):
+    for path in [Path(p) for p in glob.glob(outdir.as_posix() + "/**", recursive=True)]:
```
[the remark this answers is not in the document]
qkaiser: thank you so much !! I was struggling a bit there.

qkaiser:
Of course Darwin is breaking on these weird files.

qkaiser:
There is a difference in how debugfs works:

### OSX (Darwin Kernel Version 20.6.0)

```
$ debugfs -R 'rdump / f_badsymlinks.img_extract' tests/integration/filesystem/extfs/__input__/f_badsymlinks.img
$ ls -al f_badsymlinks.img_extract/
total 0
drwxr-xr-x 12 runner wheel  384 May 17  2002 .
drwx------ 10 runner wheel  320 Apr 27 14:20 ..
lrwxr-xr-x  1 runner wheel    3 Apr 27 14:20 attr_fastlink -> foo
lrwxr-xr-x  1 runner wheel   62 Apr 27 14:20 attr_link -> aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
lrwxr-xr-x  1 runner wheel   62 Apr 27 14:20 bad_link_size -> aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
lrwxr-xr-x  1 runner wheel    0 Apr 27 14:20 empty_fastlink -> ''
lrwxr-xr-x  1 runner wheel    0 Apr 27 14:20 empty_link -> ''
lrwxr-xr-x  1 runner wheel    0 Apr 27 14:20 extra_link_block -> ''
lrwxr-xr-x  1 runner wheel   62 Apr 27 14:20 high_link -> aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
lrwxr-xr-x  1 runner wheel   63 Apr 27 14:20 long_fastlink -> 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaҾ'$'\b'
lrwxr-xr-x  1 runner wheel 1023 Apr 27 14:20 long_link -> aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
drwxr-xr-x  2 runner wheel   64 Jul 27  2001 lost+found
```

### Linux (Linux 5.19.0-40-generic)

```
$ debugfs -R 'rdump / f_badsymlinks.img_extract' tests/integration/filesystem/extfs/__input__/f_badsymlinks.img
$ ls -al f_badsymlinks.img_extract
total 100
drwxr-xr-x  3 quentin quentin  4096 Mai 17  2002 .
drwxrwxrwt 81 root    root    69632 Apr 27 16:22 ..
lrwxrwxrwx  1 quentin quentin     3 Apr 27 16:22 attr_fastlink -> foo
lrwxrwxrwx  1 quentin quentin    62 Apr 27 16:22 attr_link -> aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
lrwxrwxrwx  1 quentin quentin    62 Apr 27 16:22 bad_link_size -> aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
lrwxrwxrwx  1 quentin quentin    62 Apr 27 16:22 high_link -> aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
lrwxrwxrwx  1 quentin quentin    63 Apr 27 16:22 long_fastlink -> 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaҾ'$'\b'
lrwxrwxrwx  1 quentin quentin  1023 Apr 27 16:22 long_link -> aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
drwxr-xr-x  2 quentin quentin  4096 Jul 27  2001 lost+found
```

unblob/extractor.py
```diff
@@ -62,7 +62,13 @@ def fix_symlink(path: Path, outdir: Path, task_result: TaskResult) -> Path:
...
         return path
 
-    target = Path(os.readlink(path))
+    raw_target = os.readlink(path)
+    if not raw_target:
+        logger.error("Symlink with empty target, removing.")
+        path.unlink()
+        return path
+
+    target = Path(raw_target)
```
vlaci, line 71: Would worth a new case in `test_fix_symlink`

unblob/extractor.py
```diff
@@ -86,7 +93,7 @@ def fix_symlink(path: Path, outdir: Path, task_result: TaskResult) -> Path:
 
 def fix_extracted_directory(outdir: Path, task_result: TaskResult):
     fix_permission(outdir)
-    for path in outdir.rglob("*"):
+    for path in (Path(p) for p in glob.iglob(f"{outdir}/**", recursive=True)):
```
e3krisztian, line 96: There is a regression problem here compared to the previous version: `Path.glob()` returned hidden files (those, that start with `.`), while `glob.*glob` does not.

Looking at the `glob` module doc, there is a parameter `include_hidden`, but it was added only in 3.11!!!

I think the only remaining option besides going still lower-level is using `os.walk`.

unblob/extractor.py
qkaiser: It's such an edge case that I can't reproduce it through Python code.

call | OS | exception
----|-----|------------
`Path('test').symlink_to('')` | Linux | `FileNotFoundError`
`Path('test').symlink_to('')` | Darwin| `OSError`
`os.symlink('test', '')` | Linux| `FileNotFoundError`
`os.symlink('test', '')` | Darwin | `FileNotFoundError`

It's not worth implementing a test based on a file since it's already done by the `f_badsymlink.img` file in integration tests. If we have a regression, the extfs integration test will fail on Darwin.

unblob/extractor.py
qkaiser: Fix and detailed explanation in commit abbc974

qkaiser:
Oh ffs.

unblob/extractor.py
```diff
@@ -85,12 +91,24 @@ def fix_symlink(path: Path, outdir: Path, task_result: TaskResult) -> Path:
 
 
 def fix_extracted_directory(outdir: Path, task_result: TaskResult):
+    def _fix_extracted_directory(directory: Path, task_result: TaskResult):
```
e3krisztian, line 94:
```suggestion
    def _fix_extracted_directory(directory: Path):
```

`task_result` is accessible from outer scope.

unblob/extractor.py
```diff
@@ -85,12 +91,24 @@ def fix_symlink(path: Path, outdir: Path, task_result: TaskResult) -> Path:
...
+        if not directory.exists():
+            return
+        for path in (directory / p for p in os.listdir(directory)):
+            try:
+                fix_permission(path)
+                if path.is_symlink():
+                    fix_symlink(path, outdir, task_result)
+                    continue
+                if path.is_dir():
+                    _fix_extracted_directory(path, task_result)
```
e3krisztian, line 104:
```suggestion
                    _fix_extracted_directory(path)
```

unblob/extractor.py
```diff
@@ -85,12 +91,24 @@ def fix_symlink(path: Path, outdir: Path, task_result: TaskResult) -> Path:
 
 
 def fix_extracted_directory(outdir: Path, task_result: TaskResult):
+    def _fix_extracted_directory(directory: Path, task_result: TaskResult):
+        if not directory.exists():
+            return
+        for path in (directory / p for p in os.listdir(directory)):
```
e3krisztian, line 97: For the future: `os.scandir` (instead of `os.listdir`) would be a better match for the task, but it would require reworking `fix_permission` as well.

e3krisztian (approved):
It looks good! The comments are minor, and this PR could be merged without addressing those.

unblob/extractor.py
```diff
@@ -85,12 +91,24 @@ def fix_symlink(path: Path, outdir: Path, task_result: TaskResult) -> Path:
...
+            try:
+                fix_permission(path)
+                if path.is_symlink():
+                    fix_symlink(path, outdir, task_result)
+                    continue
+                if path.is_dir():
+                    _fix_extracted_directory(path, task_result)
+            except OSError as e:
+                if e.strerror == "File name too long":
```
e3krisztian, line 106: Use `errno` instead of error string:
```py
import errno

if e.errno == errno.ENAMETOOLONG:
```

unblob/extractor.py
qkaiser: Did not know about `scandir`, but yeah it makes sense to move to it in the future.

e3krisztian:
Thanks for the fixes!

PR #567, opened 2023-04-25
feat(cli): provide human readable reporting in console output

qkaiser:
Provide some information about identified chunks, extracted content, and encountered errors. All the information is obtained from the ProcessResult object.

Example:

```
poetry run unblob --report /tmp/report.json -f -e /tmp/out sample.img
Extracted files: 3616
Extracted directories: 609
Extracted links: 782
Extraction directory size: 298.19 MB.
Chunks identification ratio: 90.74%
Chunks distribution
	- EXTFS: 150.00 MB (54.17%)
	- ELF32: 73.98 MB (26.72%)
	- UNKNOWN: 25.64 MB (9.26%)
	- FAT: 16.00 MB (5.78%)
	- LZO: 9.28 MB (3.35%)
	- XZ: 1.03 MB (0.37%)
	- TAR: 860.00 KB (0.30%)
	- BZIP2: 93.56 KB (0.03%)
	- GZIP: 24.21 KB (0.01%)	
Encountered errors: 1
	- Severity.WARNING: MaliciousSymlinkRemoved
```

This code was sitting on my machine for the last month. I figured I would push it to start a discussion 😄


e3krisztian (approved):
This is cool, and it can/should be merged!

qkaiser:
I will use it as my default branch over the coming days to see if anything breaks. I think I've seen a division by zero earlier, probably a good thing to write some tests thinking about unhappy paths.

vlaci:
Instead of just restricting the default log level, it'd make sens to write a log file next to the extraction with a bit higher verbosity. I'd keep the console as clear as possible. My dream is to add some more intelligible progress reporting to the console via `rich` and-or `promp-poolkit` progress bars.

qkaiser:
> Instead of just restricting the default log level, it'd make sens to write a log file next to the extraction with a bit higher verbosity.

How would you do it ?

> My dream is to add some more intelligible progress reporting to the console via rich and-or promp-poolkit progress bars.

Me too !

vlaci:
> > Instead of just restricting the default log level, it'd make sens to write a log file next to the extraction with a bit higher verbosity.
> 
> How would you do it ?

Okay, maybe I've missed something, but it seems like it is really not obvious how to do with structlog...

qkaiser:
I introduced logging to a file and a progress bar using `rich` when the `-v` option is not used.

The progress bar runs in the main thread and updates the total and completed value based on submitted and completed tasks of the worker pool. Yes, the progress bar can go back but that's how it is. We can't predict what we will encounter down the extraction directory.

There are a few open UI/UX questions:
- how well does `rich` and logging play together ? It seems it breaks the console when I used the progress bar.
- how should we name the logging file ?
- should we always create the logging file ? should it be linked to an option switch ?

vlaci:
If we route `structlog` logs to standard logging handlers, we can add different levels to different output streams:
https://github.com/onekey-sec/unblob/pull/597/files#diff-9debb99c720bedc5736dc89eab3987d1e9b3c2ebce05997fb6e79f69b0781b79R122
here, an additional handler could be defined which writes to a file, and the console handler's level could be restricted.

unblob/logging.py
```diff
@@ -100,14 +106,40 @@ def configure_logger(verbosity_level: int, extract_root: Path):
...
+        foreign_pre_chain=shared_processors,
+        processors=[
+            structlog.stdlib.ProcessorFormatter.remove_processors_meta,
+            structlog.dev.ConsoleRenderer(colors=False),
+        ],
     )
 
+    handler = logging.StreamHandler(sys.stdout)
+    handler.setFormatter(formatter)
+    file_handler = logging.handlers.WatchedFileHandler("unblob.log")  # type: ignore
```
qkaiser, line 135: An open question is whether we should always use the same file or create a temp file and provide the name as part of console output.

unblob/logging.py
```diff
@@ -1,4 +1,5 @@
 import logging
+import logging.config
```
e3krisztian, line 2: unused?

unblob/models.py
```diff
@@ -29,7 +29,7 @@
 #
 
 
-@attr.define(frozen=True)
+@attr.define(frozen=False)
```
e3krisztian, line 32: ?

unblob/logging.py
qkaiser: weirdly enough, this import made `logging.handlers.WatchedFileHandler` work while I should have been importing with `from logging.handlers import WatchedFileHandler`.

Fixed the import, allowing me to remove the `# type: ignore` next to `WatchedFileHandler`.

unblob/logging.py
martonilles: maybe the user can specify the filename and we can default to a filename similar to the input filename? This way just running it would not overwrite existing file, but for scripting they can specify and do not need to guess it?

e3krisztian: Log files based on the input is a good idea, and it is tolerable beside the extraction directory, but I also feel it is nice to have for only some and only in a few cases and not in general. It does hurt a bit over `unblob.log`: it is somewhat harder to properly script against (thinking about shell escapes here).

Making it configurable is time and code without clear use in sight, so why?

I still prefer a single well known "unblob.log".

qkaiser: I just added a `--log` option so user can provide an explicit log file. The value defaults to `unblob.log` if it's not provided.

The log file default verbosity is 2 so that it's not bloated with escape sequences and colors coming from `dissect.cstruct` structures dumped in the logs. Verbosity level 3 is only used by contributors during development anyway.

 I had to set the `exception_formatter` explicitly to `structlog.dev.plain_traceback`. Otherwise `rich` takes over and formats exceptions with displays and colors which severely impact performance.

e3krisztian (approved):

PR #569, opened 2023-05-02
Entropy fixes

e3krisztian:


qkaiser:
Good catch !

qkaiser (approved):
Looks good to me.

PR #574, opened 2023-05-08
fix(extractor): do not set permission on symlink

vlaci:
fix_permission is called before fix_symlink, hence the symlink might be an absolute target and setting an permission would fail, so we should ignore setting permission on symlink.

Setting follow_symlinks=False on chmod() would be a solution, but not all platforms support it as of now.

Ignoring permission setting on the target is ok, as that would be taken care separately.


kukovecz (approved):

kukovecz:
> Setting follow_symlinks=False on chmod() would be a solution, but not all platforms support it as of now.

Shame :smiling_face_with_tear:

PR #576, opened 2023-05-09
yaffs handler improvements

qkaiser:
Submitted within the same MR since it was present in the same sample.

It also had a YAFFS false positive sample that gets parsed but triggers a `NodeIDAbsentError` exception when calling `self.file_entries.expand_tree(mode=Tree.DEPTH)` during extraction. This exception is raised by `treelib` if the tree is empty.    
We handle this edge case by checking if the tree is empty after the chunk identification, and raise an InvalidInputFormat if it is.

### DIFF
--- a/unblob/handlers/filesystem/yaffs.py
+++ b/unblob/handlers/filesystem/yaffs.py
@@ -299,6 +299,7 @@ class YAFFSParser:
 
     def parse(self, store: bool = False):  # noqa: C901,FBT001,FBT002
         self.init_tree()
+        entries = 0
         for offset, page, spare in iterate_over_file(self.file, self.config):
             try:
                 data_chunk = self.build_chunk(
@@ -325,8 +326,11 @@ class YAFFSParser:
 
                 if store:
                     self.insert_entry(self.build_entry(header, data_chunk))
+                entries += 1
             elif store:
                 self.data_chunks[data_chunk.object_id].append(data_chunk)
+        if not entries:
+            raise InvalidInputFormat("YAFFS filesystem with no entries.")
         self.end_offset = self.file.tell()
 
     def auto_detect(self) -> YAFFSConfig:


unblob/handlers/filesystem/squashfs.py
```diff
@@ -27,7 +28,12 @@ def calculate_chunk(self, file: File, start_offset: int) -> Optional[ValidChunk]
         header = self.parse_header(file, endian)
 
         end_of_data_offset = start_offset + header.bytes_used
-        file.seek(end_of_data_offset)
+        try:
+            file.seek(end_of_data_offset, io.SEEK_SET)
+        except ValueError:
+            raise InvalidInputFormat(
+                "Squashfs bytes_used value is larger than file size."
+            ) from ValueError
```
martonilles, line 36: I do not really understand this. The File seek already throws either a SeekError or a ValueError (in case we also read an EOFError).

Which is handled in finder.py:
```py
def _calculate_chunk(
    handler: Handler, file: File, real_offset, task_result: TaskResult
) -> Optional[ValidChunk]:
    file.seek(real_offset)
    try:
        return handler.calculate_chunk(file, real_offset)
    except InvalidInputFormat as exc:
        logger.debug(
            "File format is invalid",
            exc_info=exc,
            handler=handler.NAME,
            _verbosity=2,
        )
    except EOFError as exc:
        logger.debug(
            "File ends before header could be read",
            exc_info=exc,
            handler=handler.NAME,
            _verbosity=2,
        )
    except SeekError as exc:
        logger.debug(
            "Seek outside file during chunk calculation",
            exc_info=exc,
            handler=handler.NAME,
            _verbosity=2,
        )
```

Why is it better to have an InvalidFormat instead of any of the above?

qkaiser: Removed the squashfs commit and kept the YAFFS one.

PR #577, opened 2023-05-09
bumped version to 23.5.9

vlaci:


qkaiser:
Can we include https://github.com/onekey-sec/unblob/pull/576 in this version ?

qkaiser (approved):

PR #578, opened 2023-05-10
chore(deps): upgrade jefferson to 0.4.3

qkaiser:


qkaiser:
darn it. ubi-reader requires `python-lzo` but we did not catch it because it's defined within their `setup.py` like this:

```python
install_requires=['python-lzo'],
```

Poetry did not catch it either. The fact that ubi-reader worked so far is just a happy accident based on the fact they we used to rely on python-lzo for crc32 and adler32 calls.

vlaci:
> darn it. ubi-reader requires `python-lzo` but we did not catch it because it's defined within their `setup.py` like this:
> 
> ```python
> install_requires=['python-lzo'],
> ```
> 
> Poetry did not catch it either. The fact that ubi-reader worked so far is just a happy accident based on the fact they we used to rely on python-lzo for crc32 and adler32 calls.

Yeah, it is completely absent from the metadata:

```console
$ cat ubi_reader-0.8.5/lib/python3.10/site-packages/ubi_reader-0.8.5.dist-info/METADATA
Metadata-Version: 2.1
Name: ubi-reader
Version: 0.8.5
Summary: Extract files from UBI and UBIFS images.
Home-page: https://github.com/jrspruitt/ubi_reader
Author: Jason Pruitt
Author-email: jrspruitt@gmail.com
License: GNU GPL
Keywords: UBI UBIFS
Platform: Linux
Requires: lzo
License-File: LICENSE

Collection of Python scripts for reading information about and extracting data from UBI and UBIFS images.
```

I can add a shim to lzallright to be compatible on the api surface with the lzo package, and optionally provide that package name also :thinking:

qkaiser:
@vlaci let's wait on https://github.com/jrspruitt/ubi_reader/pull/73

qkaiser:
`python-lzo` is finally out with the upgrade to `ubi-reader` version 0.8.7

qkaiser:
@vlaci I got bored of waiting for Nixpkgs so I created the required files under `nix/`.

nix/lzallright/default.nix
```diff
@@ -0,0 +1,50 @@
...
+
+  meta = with lib; {
+    description = ''
+      A Python 3.8+ binding for lzokay library which is an MIT licensed
+      a minimal, C++14 implementation of the LZO compression format.
+    '';
+    homepage = "https://github.com/vlaci/lzallright";
+    license = licenses.mit;
+    maintainers = with maintainers; [ vlaci ];
+  };
```
vlaci, line 49: You can get rid of these `meta` sections, we don't really need them for anything here.

vlaci (approved):

PR #579, opened 2023-05-10
 Fix subclassing bug in Report subclasses + pyright upgrade

e3krisztian:
When a frozen attr class is further specialized, the new class also must be as frozen.

This issue was hidden from pyright by a file-level ignore.

Updated pyright to the current version, which revealed further linter problems. They are fixed in separate commits.


e3krisztian:
Updating pyright apparently uncovered some new issues elsewhere...

.github/workflows/run-tests.yml
```diff
@@ -44,7 +44,7 @@ jobs:
         uses: ./.github/actions/setup-dependencies
 
       - name: Check - pyright
-        run: PYRIGHT_PYTHON_FORCE_VERSION=1.1.188 poetry run pyright .
+        run: PYRIGHT_PYTHON_FORCE_VERSION=1.1.307 poetry run pyright .
```
vlaci, line 47: If we set a new version which tracks the upstream pyright version in dev dependencies, we no longer need this variables.

pyproject.toml
```diff
@@ -34,7 +34,7 @@ optional = true
 
 [tool.poetry.group.dev.dependencies]
 pytest = "^6.2.4"
-pyright = "^0.0.12"
+pyright = "^1.1.291"
```
vlaci, line 37: Why are we asking for an older minimum version than we want to run with? We are locked on the latest one and have to deal with the type errors from the latest one anyway.

e3krisztian: Updated to match the current version.

Or did you propose to not lock in the project file at all?

e3krisztian:
Included fixes for newly discovered linter problems in several handlers.

unblob/handlers/archive/xiaomi/hdr.py
```diff
@@ -106,7 +106,7 @@ def extract(self, inpath: Path, outdir: Path):
 
     def parse(self, file: File) -> Iterable[Tuple[Path, Chunk]]:
         header = self._struct_parser.parse(self.header_struct, file, Endian.LITTLE)
-        for offset in header.blob_offsets:
+        for offset in cast(Iterable, header.blob_offsets):
```
vlaci, line 109: I don't get these errors. For me pyright tells, it's `Any`. It is interesting, that you got `Never`. This means, that pyright thinks, that the method is diverting, i.e always raises an exception.

unblob/handlers/archive/netgear/trx.py
```diff
@@ -47,7 +47,11 @@ def extract(self, inpath: Path, outdir: Path):
             eof = file.tell()
             i = 0
             offsets = sorted(
-                [offset for offset in [*header.offsets, eof] if offset > 0]
+                [
+                    offset
+                    for offset in [*cast(Iterable, header.offsets), eof]
```
vlaci, line 52: Same as below

unblob/handlers/archive/xiaomi/hdr.py
e3krisztian: There were some changes for `StructParser.parse()`: previously its return type was unspecified, and was interpreted as `Any`, it has been changed to `dissect.struct.Instance`.

The issue with `Instance` is, that it is a rather generic class. All the properties on it is generated according to the spec `parse()` receives. So `blob_offset` here can be found only inside of a struct spec string, and not in any Python code, so pyright have no chance knowing about it.

e3krisztian: These error messages have been copied out from a failed github action output: https://github.com/onekey-sec/unblob/actions/runs/4940612918/jobs/8832518378

Issue #419, opened 2022-08-14
Add support for EWF / E01 disk image format


jgrover:
This is a common disk image format found in the digital forensics domain.  Would bring a whole new set of users for unblob if this were added.

QKaiser:
Standard reference documented here: https://github.com/libyal/libewf/blob/main/documentation/Expert%20Witness%20Compression%20Format%20(EWF).asciidoc

Format is made of segments containing sections. We can get to the end offset by browsing through sections.

I didn't find third party extractors so far. What we could do is use ewfexport to translate from ewf to raw data and get other handlers (e.g. extfs, ntfs) pick up chunks from raw.

PR #582, opened 2023-05-17
feat(handler): Add support for EWF format

nyuware:
Temporary solve for #419 
Expert Witness Format (EWF) works with multiple forensics tools such as Encase, FTK, Linen, they mostly share the same specification.
The format is divided into multiple section, each section contain the size of it's own section and the offset of the next section. 
Right now this handler works for the following formats : 
- Linen 5-6
- Encase 2-7
- EWFX
- FTK
This handler currently doesn't support the following formats : 
- Encase 1
- SMART
- EWF
It also doesn't support the zlib compression.

This is only a temporary solution, the proper solution would be to use the "table" section, which holds the information to each chunk offset as well as the compression level of the chunk itself. 
The table is built like this but can change depending on the version used
TABLE HEADER -> how many offset in table
ARRAY OF ENTRIES -> hold the relative offset to each chunk and the most significant bit indicates if compressed or not. 

Some generous people made the documentation of the format [here](https://github.com/libyal/libewf/blob/main/documentation/Expert%20Witness%20Compression%20Format%20(EWF).asciidoc)

### DIFF
--- a/unblob/handlers/archive/ewf/ewf.py
+++ b/unblob/handlers/archive/ewf/ewf.py
@@ -1,0 +1,202 @@
+import io
+import math
+import zlib
+from enum import Enum
+from pathlib import Path
+
+from dissect.cstruct import Instance
+from structlog import get_logger
+
+from unblob.file_utils import (
+    Endian,
+    File,
+    FileSystem,
+    InvalidInputFormat,
+    StructParser,
+    iterate_file,
+)
+from unblob.models import Extractor, HexString, StructHandler, ValidChunk
+
+logger = get_logger()
+
+C_DEFINITIONS = r"""
+    typedef struct ewf_header {
+        char magic[8];
+        uint8 field_start;
+        uint16 segment_number;
+        uint16 field_end;
+    } ewf_header_t;
+
+    typedef struct data_descriptor {
+        char definition[16];
+        uint64 next_offset;
+        uint64 section_size;
+        char padding[40];
+        uint32 adler_32;
+    } data_descriptor_t;
+
+    typedef struct volume_descriptor {
+        uint32 unknown;
+        uint32 chunk_count;
+        uint32 sectors_per_chunks;
+        uint32 bytes_per_sectors;
+        uint32 sectors_count;
+    } volume_descriptor_t;
+
+    typedef struct table_descriptor {
+        uint32 number_of_entries;
+        char padding[16];
+        uint32 adler_32;
+    } table_descriptor_t;
+
+    typedef struct table_entry {
+        char offset[3];
+        char compression_type[1];
+    } table_entry_t;
+    typedef struct hash_descriptor {
+        char md5_hash[16];
+        char unknown[16];
+        uint32 adler_32;
+    } hash_descriptor_t;
+"""
+
+EWF_HEADER_LEN = 13
+DESCRIPTOR_LEN = 76
+
+
+class Definition(Enum):
+    DONE = b"done".ljust(16, b"\x00")
+    TABLE = b"table".ljust(16, b"\x00")
+    SECTORS = b"sectors".ljust(16, b"\x00")
+    VOLUME = b"volume".ljust(16, b"\x00")
+
+
+class ZlibMagic(Enum):
+    LOW = b"\x78\x01"
+    DEFAULT = b"\x78\x9c"
+    BEST = b"\x78\xda"
+    COMPRESSION = b"\x78\x5e"
+
+
+def find_chunk_size(header: Instance) -> int:
+    size = header.sectors_per_chunks
+    log = math.log(size, 2)
+    power = math.pow(2, log + 9)
+    return int(power)
+
+
+def is_valid_header(header: Instance) -> bool:
+    if header.field_start != 0x01 or header.field_end != 0x0:
+        return False
+    return True
+
+
+class EWFExtractor(Extractor):
+    def __init__(self, header_struct: str):
+        self.header_struct = header_struct
+        self._struct_parser = StructParser(C_DEFINITIONS)
+
+    def table_descriptor(
+        self, file: File, position: int, outdir: Path, sectors_per_chunk: int
+    ) -> Instance:
+        fs = FileSystem(outdir)
+        entries = []
+        header = self._struct_parser.parse("table_descriptor_t", file, Endian.LITTLE)
+        entry_path = Path("ewf.decrypted")
+
+        for _ in range(header.number_of_entries):
+            entry = self._struct_parser.parse("table_entry_t", file, Endian.LITTLE)
+            entries.append(entry.offset)
+
+        with fs.open(entry_path) as output_file:
+            for offset in entries:
+                file.seek(
+                    position
+                    + int.from_bytes(offset, byteorder="little")
+                    - DESCRIPTOR_LEN,
+                    io.SEEK_SET,
+                )
+
+                magic_bytes = file.read(2)
+                compressed = any(magic_bytes == magic.value for magic in ZlibMagic)
+
+                for chunk in iterate_file(
+                    file,
+                    position
+                    + int.from_bytes(offset, byteorder="little")
+                    - DESCRIPTOR_LEN,
+                    sectors_per_chunk,
+                ):
+                    if compressed:
+                        compressed_chunk = zlib.decompress(chunk)
+                        output_file.write(compressed_chunk)
+                    output_file.write(chunk)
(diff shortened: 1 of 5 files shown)


unblob/handlers/archive/ewf/ewf.py
```diff
@@ -0,0 +1,157 @@
...
+    def __init__(self, header_struct: str):
+        self.header_struct = header_struct
+        self._struct_parser = StructParser(C_DEFINITIONS)
+
+    def extract(self, inpath: Path, outdir: Path):
+        sectors_per_chunk = 0
+        volume_descriptor = None
+        outpath = outdir.joinpath(f"{inpath.name}.decrypted")
+        with File.from_path(inpath) as file:
+            file.seek(13)  # we skip the initial header
```
qkaiser, line 83: This should be a constant.

unblob/handlers/archive/ewf/ewf.py
```diff
@@ -0,0 +1,157 @@
...
+                            remaining_size = data_descriptor.section_size - len(
+                                data_descriptor
+                            )
+                            while remaining_size > 0:
+                                chunk_size = min(sectors_per_chunk, remaining_size)
+                                for chunk in iterate_file(
+                                    file, file.tell(), chunk_size
+                                ):
+                                    outfile.write(chunk[:-4])
+                                remaining_size -= chunk_size
```
qkaiser, line 117: This is way too complex, I'm even amazed that flake did not report it. I would split the work in dedicated functions, depending on the definition type.

qkaiser (changes requested):

unblob/handlers/archive/ewf/ewf.py
nyuware: Divided the code into different sections

qkaiser:
@nyuware please squash your commits and rebase on main

nyuware:
The code pushed works for the zlib compressed fluxes.
The problems left are the following : 

1. adler-32 not implemented
2. when the file is not compressed by zlib, the `sector_per_chunks` can be bigger than the remaining size of the section, meaning that the function outfile.write(chunks) will collect more than it's supposed to.

In the past, I was using this method to calculate the remaining size of the section, but it seems that the section_size is incorrect.

`if volume_descriptor is not None:
                            remaining_size = data_descriptor.section_size - len(
                                data_descriptor
                            )
                            while remaining_size > 0:
                                chunk_size = min(sectors_per_chunk, remaining_size)
                                for chunk in iterate_file(
                                    file, file.tell(), chunk_size
                                ):
                                    outfile.write(chunk[:-4])
                                remaining_size -= chunk_size`

unblob/handlers/archive/ewf/ewf.py
```diff
@@ -0,0 +1,202 @@
...
+                "data_descriptor_t", file, Endian.LITTLE
+            )
+
+        return ValidChunk(start_offset=start_offset, end_offset=file.tell())
+
+
+class EWFEHandler(EFWHandlerBase):
+    NAME = "ewfe"
+
+    PATTERNS = [HexString("45 56 46 09 0d 0a ff 00")]
```
qkaiser, line 188: You can enforce a strict pattern and get rid of `is_valid_header`:

```suggestion
    PATTERNS = [HexString("45 56 46 09 0d 0a ff 00 01 [2] 00")]
```

unblob/handlers/archive/ewf/ewf.py
```diff
@@ -0,0 +1,202 @@
...
+
+    C_DEFINITIONS = C_DEFINITIONS
+    HEADER_STRUCT = "ewf_header_t"
+    EXTRACTOR = EWFExtractor("ewf_header_t")
+
+
+class EWFLHandler(EFWHandlerBase):
+    NAME = "ewfl"
+
+    PATTERNS = [HexString("4C 56 46 09 0d 0a ff 00")]
```
qkaiser, line 198: You can enforce a strict pattern and get rid of `is_valid_header`:

```suggestion
    PATTERNS = [HexString("4C 56 46 09 0d 0a ff 00 01 [2] 00")]
```

unblob/handlers/archive/ewf/ewf.py
```diff
@@ -0,0 +1,202 @@
...
+                        output_file.write(compressed_chunk)
+                    output_file.write(chunk)
+
+    def extract(self, inpath: Path, outdir: Path):
+        with File.from_path(inpath) as file:
+            file.seek(EWF_HEADER_LEN)  # we skip the initial header
+            data_descriptor = self._struct_parser.parse(
+                "data_descriptor_t", file, Endian.LITTLE
+            )
+            logger.debug("data_descriptor_t", header=data_descriptor, _verbosity=3)
```
qkaiser, line 141: You can remove this logger call if you place the next one at the start of the while loop.

unblob/handlers/archive/ewf/ewf.py
```diff
@@ -0,0 +1,202 @@
...
+
+            # the file is made of section, we loop over all the sections
+            while data_descriptor.definition != Definition.DONE.value:
+                file.seek(data_descriptor.next_offset, io.SEEK_SET)
+                data_descriptor = self._struct_parser.parse(
+                    "data_descriptor_t", file, Endian.LITTLE
+                )
+                logger.debug("data_descriptor_t", header=data_descriptor, _verbosity=3)
```
qkaiser, line 149: Move it up before the seek.

unblob/handlers/archive/ewf/ewf.py
```diff
@@ -0,0 +1,202 @@
...
+                    volume_descriptor = self._struct_parser.parse(
+                        "volume_descriptor_t", file, Endian.LITTLE
+                    )
+                    sectors_per_chunk = find_chunk_size(volume_descriptor)
+
+                if data_descriptor.definition == Definition.SECTORS.value:
+                    position = file.tell()
+
+                if data_descriptor.definition == Definition.TABLE.value:
+                    self.table_descriptor(file, position, outdir, sectors_per_chunk)
```
qkaiser, line 161: Can you explain how EWF is structured ? Can we observe VOLUME, SECTORS, and TABLE in random orders ? Do they follow a strict ordering ? Is it possible to have multiple VOLUMEs ? What about the others ?

unblob/handlers/archive/ewf/ewf.py
```diff
@@ -0,0 +1,202 @@
...
+        self.header_struct = header_struct
+        self._struct_parser = StructParser(C_DEFINITIONS)
+
+    def table_descriptor(
+        self, file: File, position: int, outdir: Path, sectors_per_chunk: int
+    ) -> Instance:
+        fs = FileSystem(outdir)
+        entries = []
+        header = self._struct_parser.parse("table_descriptor_t", file, Endian.LITTLE)
+        entry_path = Path("ewf.decrypted")
```
qkaiser, line 105: Is EWF encrypted ?

unblob/handlers/archive/ewf/ewf.py
```diff
@@ -0,0 +1,202 @@
...
+
+        for _ in range(header.number_of_entries):
+            entry = self._struct_parser.parse("table_entry_t", file, Endian.LITTLE)
+            entries.append(entry.offset)
```
qkaiser, line 109: I would do the conversion here to ease understanding later on:

 ```suggestion
            entries.append(int.from_bytes(entry.offset, byteorder="little"))
```

File: unblob/handlers/archive/ewf/ewf.py [hunk 10]
```diff
@@ -0,0 +1,202 @@
...
+                    + int.from_bytes(offset, byteorder="little")
+                    - DESCRIPTOR_LEN,
+                    sectors_per_chunk,
+                ):
+                    if compressed:
+                        compressed_chunk = zlib.decompress(chunk)
+                        output_file.write(compressed_chunk)
+                    output_file.write(chunk)
+
+    def extract(self, inpath: Path, outdir: Path):
```
qkaiser, line 135: There is a logic / naming issue here. The **extraction** is performed in `table_descriptor` while the **parsing** is performed in `extract`. Please adjust your function so they reflect what they're actually doing. Ideally you would parse the EWF structure and return data that can be used by `extract` in order to create files in `outdir`.

unblob/handlers/archive/ewf/ewf.py
nyuware: Named it `ewf.extracted`

PR #586, opened 2023-05-22
fix(handlers): handle tar archives with same file present more than once

qkaiser:
A `PermissionError: [Errno 13] Permission denied` is thrown when expanding a tarfile to which a file had been added more than once, *if* the file is not writeable.  (tarfile expands the file twice, but the second time finds a non-writeable file in the way.)

Decided to handle this gracefully by catching the PermissionError and logging it.

See https://bugs.python.org/issue30438 for a detailed discussion of this issue.

### DIFF
--- a/tests/integration/archive/tar/__input__/dual_entry.tar
+++ b/tests/integration/archive/tar/__input__/dual_entry.tar
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:8684178ad6cf02586da93ba255e90c9c345e31fa83b4beffe881cf8d0d6592fd
+size 9728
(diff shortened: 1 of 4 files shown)


qkaiser:
We're discussing a redesign of the tar extractor internally so this issue will probably be fixed differently.

Excerpt from @e3krisztian take:

>  Currently extraction works by creating a custom derived class `_SafeTarFile` (or something similar), and calling `extractall()` on it, while its internal is patched with our workarounds. This is a reactive design - we interpret and handle errors when they occur.

> The proposed new design is not using this derived class at all, but use `TarFile.getmembers()` and process the resulting tar member list before calling `extractall()` with the list. This is a proactive design - we do not allow any errors to occur inside `TarFile.extract()`, or at least not something we know about.

> This processing is what we do in `_SafeTarFile`: remove any recognized duplicates, and other problem makers, and report them.

> The plan is to call `extractall()` in the end, because it does some permission related work, that `.extract()` does not, and also create missing directories, but there is a catch that could force us to reimplement these extra bits in the end and use only `.extract()`: we want to be able to extract members to a different path, than `extractall()` would extract by default: tar entries with absolute paths. These absolute entries we currently skip/do not allow because it is a kind of path traversal, and to extract them, we need to extract them elsewhere as `TarFile` would do. Archives with all absolute paths are currently extract to an empty extraction.

e3krisztian:
The plan quoted above by @qkaiser did not work, as path traversal check is not really possible in advance.
In the end went with the second best: extract entries one by one.

#592 contains an alternative fix to the problem of this MR, with the test data, and partially the commit message taken from here.

PR #592, opened 2023-06-01
Tar handler improvements

e3krisztian:
Improvements:
- handle duplicated entries (with test data taken from #586)
- (path traversal) handle tar entries with absolute paths
- (path traversal) handle link targets with absolute paths

At the source level there was also a refactor to slightly better decouple the code from `tarfile.TarFile` implementation details.

### DIFF
--- a/unblob/handlers/archive/tar.py
+++ b/unblob/handlers/archive/tar.py
@@ -1,3 +1,4 @@
+import contextlib
 import os
 import tarfile
 from pathlib import Path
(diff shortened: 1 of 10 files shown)


unblob/handlers/archive/tar.py
```diff
@@ -85,13 +86,8 @@ def _find_end_of_padding(file, *, find_from: int) -> int:
...
     def extract(self, inpath: Path, outdir: Path):
-        tf = SafeTarFile.open(inpath.as_posix())
-        try:
-            tf.extractall(outdir.as_posix())
-        except FileExistsError as file_exists_error:
-            logger.warning(
-                "FileExistsError during tar archive extraction", error=file_exists_error
-            )
+        with contextlib.closing(SafeTarFile(inpath)) as tarfile:
+            tarfile.extractall(outdir)
```
qkaiser, line 90: Might as well implement `__enter__` and `__exit__` in `SafeTarFile`, no ?

qkaiser (commented):
I like the approach in 1d87e117a94db794a654e323be1b0268322788dd, we should probably apply it to other extractors at some point (I don't have specific examples in mind though).

unblob/handlers/archive/tar.py
e3krisztian: yes, they are the same, but going with `__enter__` and `__exit__` is more verbose, and I always forgot the order of the 3 argument names for `__exit__`. It is not fun looking them up and not using them at all.

`contextlib.closing` was the most concise solution, and is relatively well known, I think.

qkaiser (approved):

PR #597, opened 2023-06-06
Experimental Landlock based sandboxing

vlaci:
Implementation of #594  together with https://github.com/onekey-sec/unblob-native/pull/11

Having to first create the extraction directory complicates things _a lot_. 
I am unsure what approach we should take here but it can be seen, that the first directory needs somewhat special treatment.

Alternative integration approaches:
- set `LANDLOCK_ACCESS_FS_MAKE_DIR` on the parent of the extraction root as an escape hatch
-  introduce a workdir where extraction takes place one level deeper to achieve simpler code.


qkaiser:
I would be in favor of option 1 since it's how unblob works. The extraction path provided with `-e` is where we create the extraction directory so we do need `LANDLOCK_ACCESS_FS_MAKE_DIR`.

As long as we're clear about the fact that unblob limits itself to the path provided with `-e` and not the directory within, it's acceptable to me.

unblob/processing.py
```diff
@@ -193,7 +207,6 @@ def prepare_report_file(config: ExtractionConfig, report_file: Optional[Path]) -
     # check that the report directory can be written to
     try:
         report_file.write_text("")
-        report_file.unlink()
```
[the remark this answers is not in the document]
vlaci: At this point the rules are not applied. If I were to keep unlinking here `LANDLOCK_ACCESS_FS_MAKE_REG`

vlaci:
> I would be in favor of option 1 since it's how unblob works. The extraction path provided with `-e` is where we create the extraction directory so we do need `LANDLOCK_ACCESS_FS_MAKE_DIR`.
> 
> As long as we're clear about the fact that unblob limits itself to the path provided with `-e` and not the directory within, it's acceptable to me.

We need this for the parent of extraction directory:
> LANDLOCK_ACCESS_FS_MAKE_DIR: Create (or rename) a directory.

And this for the parent directory of report file:
>LANDLOCK_ACCESS_FS_MAKE_REG: Create (or rename or link) a regular file.

The latter seems a bit much to me

vlaci:
Hehe, this change is incompatible with code coverage measurement :D
```
 INTERNALERROR>   File "/home/runner/.cache/pypoetry/virtualenvs/unblob-PkeEArhf-py3.8/lib/python3.8/site-packages/coverage/sqldata.py", line 1064, in _connect
INTERNALERROR>     raise DataError(f"Couldn't use data file {self.filename!r}: {exc}") from exc
INTERNALERROR> coverage.exceptions.DataError: Couldn't use data file '/home/runner/work/unblob/unblob/.coverage.fv-az626-878.3677.365646': unable to open database file
```

qkaiser:
Two things to do before tagging as ready for review:

- [ ] solve the `ModuleNotFoundError: No module named 'unblob_native.sandbox'` error by properly exposing it
- [ ] add unit testing (unsupported platforms, different access checks)

qkaiser:
https://github.com/onekey-sec/unblob-native/pull/11

qkaiser:
Will rebase once version `0.1.2` of unblob-native is in.

unblob/processing.py
```diff
@@ -112,6 +118,30 @@ def get_extract_dir_for(self, path: Path) -> Path:
...
+            AccessFS.read_write(report_file),
+            AccessFS.make_reg(report_file.parent),
+        ]
+
+    if "pytest" in sys.modules:
+        restrictions += [
+            AccessFS.read_write("/tmp"),  # noqa: S108
+            AccessFS.read_write("/build"),
+            AccessFS.read_write(Path(__file__).parent.parent.resolve().as_posix()),
+        ]
```
vlaci, line 140: restrictions could be part of `ExtractionConfig` we pass for process_file, so test and report specific values could be set-up in fixture and command line processing.

qkaiser: I like your thinking.

unblob/processing.py
```diff
@@ -136,6 +166,13 @@ def process_file(
         )
         return ProcessResult()
 
+    try:
+        if not hasattr(process_file, "_sandboxed"):
+            sandbox(extract_dir, report_file)
+            process_file._sandboxed = True  # noqa: SLF001
+    except SandboxError:
+        logger.warning("Sandboxing FS access is unavailable on this system, skipping.")
+
```
vlaci, line 175: Is it a problem to call this multiple times? If so I'd separate concerns a bit clearer:

```python
import functools

def call_once(fn):
    fn.__called_once__ = False
    
    @functools.wraps(fn)
    def wrapper(*args, **kwargs):
        if fn.__called_once__:
            return
        fn(*args, **kwargs)
        fn.__called_once__ = True

@call_once
def sandbox(...):
    ...
```

This way, we can even test the once-call functionality independently.

unblob/processing.py
vlaci: Working on it

qkaiser: gave it a try yesterday but failed :)

vlaci: I found an interesting thing: the only test, that actually calls the cli and does a full extraction is the `tests/test_cli.py::test_skip_extension` case, everything else works with a mocked process_file :)

unblob/cli.py
```diff
@@ -277,6 +278,22 @@ def cli(
...
+        # Extracted contents
+        AccessFS.read_write(extract_root.as_posix()),
+        AccessFS.make_dir(extract_root.parent.as_posix()),
+    ]
+
+    if report_file:
+        sandbox_access_restrictions += [
+            AccessFS.read_write(report_file),
+            AccessFS.make_reg(report_file.parent),
+        ]
```
qkaiser, line 295: We need a similar condition check and restrictions for `log_path`.

unblob/processing.py
```diff
@@ -118,28 +119,28 @@ def get_extract_dir_for(self, path: Path) -> Path:
...
+@call_once
+def try_enter_sandbox(config: ExtractionConfig):
+    if not config.sandbox_access_restrictions:
+        return
+    try:
+        restrict_access(*config.sandbox_access_restrictions)
+    except SandboxError:
+        logger.warning("Sandboxing FS access is unavailable on this system, skipping.")
+        restrict_access(*config.sandbox_access_restrictions)
```
qkaiser, line 143: This will lead to an unhandled `SandboxError` on non supported platforms.

qkaiser:
rebased and solved conflicts

qkaiser:
Monthly reminder about landlock in unblob, this time brought to you by @l0kod talk at SSTIC (https://www.sstic.org/2024/presentation/landlock-design/).

Rebased and solved conflicts.

l0kod:
FYI, other archive tools are sandboxing themselves too: https://github.com/ouch-org/ouch/pull/723

qkaiser:
rebased and solved conflicts. IMO this is ready to be merged.

vlaci:
I had an idea to make sandboxing more composable: spawn a new thread, drop privileges, run `process_file` there, then return the results. This way, the caller's privileges are not affected. I've ran into a mess making the process pool cleanly terminate when receiving a SIGTERM. I'll try to make it work for some more time, if I can't, then I drop this idea.

vlaci:
It's interesting that it fails on aarch64-linux. Maybe because of qemu emulation?

qkaiser:
> It's interesting that it fails on aarch64-linux. Maybe because of qemu emulation?

created a linux host on aarch64 (qemu) here, will do some experiments tomorrow

qkaiser:
@vlaci must be related to Nix, cause the tests are passing here:

```
$ uname -avr
Linux debian 6.1.0-26-arm64 #1 SMP Debian 6.1.112-1 (2024-09-30) aarch64 GNU/Linux
$ poetry run pytest tests/ -k test_sandbox
==== test session starts ====
platform linux -- Python 3.11.2, pytest-8.3.3, pluggy-1.5.0
rootdir: /home/quentin/unblob
configfile: pyproject.toml
plugins: cov-5.0.0
collected 919 items / 916 deselected / 3 selected                                                                                                                                                                                                                                                                                                                                          

tests/test_cli.py .                                                                                                                                                                                                                                                                                                                                                                  [ 33%]
tests/test_sandbox.py ..
==== 3 passed, 916 deselected in 190.78s (0:03:10) ====
```

aarch64 Debian running on `qemu-system-aarch64`

qkaiser:
Even though it *seems* to be supported by QEMU on aarch64 since https://github.com/qemu/qemu/commit/3a2f19b7ee3abbba5fd6f3e02fcdd26dc392a990, landlock syscalls are actually not supported in user mode emulation.

For demo purposes, let's write this quick-and-dirty PoC:

```c
#include <stdio.h>
#include <unistd.h>
#include <sys/syscall.h>
#include <errno.h>

int main() {
    int result = syscall(SYS_landlock_create_ruleset, NULL, 0, 0);

    if (result == -1) {
        printf("System call failed: %d (errno: %d)\n", result, errno);
    } else {
        printf("System call succeeded.\n");
    }

    return 0;
}
```

Compile it, make sure it's aarch64, run it in user mode emulation:

```
$ aarch64-linux-gnu-gcc -static poc.c
$ file a.out 
a.out: ELF 64-bit LSB executable, ARM aarch64, version 1 (GNU/Linux), statically linked, BuildID[sha1]=c42a43dc682c2dc504699d7b816e99a2e32c6bf0, for GNU/Linux 3.7.0, not stripped
$ qemu-aarch64-static ./a.out        
System call failed: -1 (errno: 38)
```

errno `38` is `-ENOSYS` or "function not implemented".

So problem is on QEMU user mode emulation not doing the translation for landlock syscalls. However, it should be handled gracefully by unblob-native I think.

qkaiser:
I built the `sandboxer` example from rust-landlock and it works on qemu-aarch64-static by returning an error:

```
$ LL_FS_RO="/bin:/lib:/usr:/proc:/etc:/dev/urandom" \
LL_FS_RW="/dev/null:/dev/full:/dev/zero:/dev/pts:/tmp" \
LL_TCP_BIND="9418" LL_TCP_CONNECT="80:443" \
qemu-aarch64-static \
target/aarch64-unknown-linux-musl/debug/examples/sandboxer bash -i
Error: Landlock is not supported by the running kernel.
```

So the bug is probably in unblob-native or between unblob-native Rust and unblob-native Python. The exception is giving me the same feeling:

```
where False = isinstance(AttributeError("'int' object has no attribute 'results'"), PermissionError)
```

vlaci:
I'll add tests on the rust side as well then :)

vlaci:
> I built the `sandboxer` example from rust-landlock and it works on qemu-aarch64-static by returning an error:
> 
> ```
> $ LL_FS_RO="/bin:/lib:/usr:/proc:/etc:/dev/urandom" \
> LL_FS_RW="/dev/null:/dev/full:/dev/zero:/dev/pts:/tmp" \
> LL_TCP_BIND="9418" LL_TCP_CONNECT="80:443" \
> qemu-aarch64-static \
> target/aarch64-unknown-linux-musl/debug/examples/sandboxer bash -i
> Error: Landlock is not supported by the running kernel.
> ```
> 
> So the bug is probably in unblob-native or between unblob-native Rust and unblob-native Python. The exception is giving me the same feeling:
> 
> ```
> where False = isinstance(AttributeError("'int' object has no attribute 'results'"), PermissionError)
> ```

Ahh, we need this error check at the end...
https://github.com/landlock-lsm/rust-landlock/blob/94721d26b2fd1151e71bd7a3aa5a43c463a22347/examples/sandboxer.rs#L133-L135

l0kod:
> Ahh, we need this error check at the end... https://github.com/landlock-lsm/rust-landlock/blob/94721d26b2fd1151e71bd7a3aa5a43c463a22347/examples/sandboxer.rs#L133-L135

Please note that kernels not supporting Landlock should not be an error for programs sandboxing themselves, only for sandboxers that must create sandboxes or error out (like this example). It can be a warning though.

vlaci:
> > Ahh, we need this error check at the end... https://github.com/landlock-lsm/rust-landlock/blob/94721d26b2fd1151e71bd7a3aa5a43c463a22347/examples/sandboxer.rs#L133-L135
> 
> Please note that kernels not supporting Landlock should not be an error for programs sandboxing themselves, only for sandboxers that must create sandboxes or error out (like this example). It can be a warning though.

Yep, unblob itself already just logs the problem, but I like clear failures in tests.
https://github.com/onekey-sec/unblob/pull/597/files#diff-2d4e67217491bddba5546f44b16c52ae6be29d568b52949950866ec5e4c24b0fR92-R95

vlaci:
Resolved-by https://github.com/onekey-sec/unblob-native/pull/65

unblob/pool.py
```diff
@@ -157,3 +197,20 @@ def make_pool(process_num, handler, result_callback) -> Union[SinglePool, MultiP
...
+
+
+orig_signal_handlers = {}
+
+
+def _on_terminate(signum, frame):
+    with contextlib.suppress(StopIteration):
+        while True:
+            pool = next(iter(pools))
+            pool.close(immediate=True)
```
qkaiser, line 209: Maybe this way so `if callable...` is reachable:

```suggestion
        for pool in pools:
            pool.close(immediate=True)
```

unblob/sandbox.py
```diff
@@ -0,0 +1,118 @@
...
+        exception = None
+        result = None
+
+        def _run_in_thread(callback, *args, **kwargs):
+            nonlocal exception, result
+
+            self._try_enter_sandbox()
+            try:
+                result = callback(*args, **kwargs)
+            except BaseException as e:
```
qkaiser, line 71: Since generic `Exception`s are caught within `processing.py` and that the only exception class that inherits `BaseException` is `ShutdownRequired`, wouldn't it  work with this ?

```suggestion
            except ShutdownRequired as e:
```

unblob/pool.py
vlaci: not really, as close mutates the pools set

unblob/sandbox.py
vlaci: `ShutDownRequired` is no more. I wanted to catch everything, like `GeneratorExit` and `SystemExit` as well. It is unlikely that these would be thrown, but if they are left uncaught, the thread will crash without notice.

vlaci:
The one thing requires thorough manual testing is keeping exit on CTRL-C and SIGTERM working. Many shenanigans are added for that purpose.

unblob/sandbox.py
```diff
@@ -0,0 +1,118 @@
...
+    def __init__(
+        self,
+        config: ExtractionConfig,
+        log_path: Path,
+        report_file: Optional[Path],
+        extra_restrictions: Iterable[AccessFS] = (),
+    ):
+        self.restrictions = [
+            # Python, shared libraries, extractor binaries and so on
+            AccessFS.read("/"),
```
e3krisztian, line 44: Does it also imply execution of third party tools?

tests/test_cli.py
```diff
@@ -425,3 +426,29 @@ def test_clear_skip_magics(
...
+        side_effect=lambda *_args, **_kwargs: unrelated_file.write_text(
+            "sandbox escape"
+        )
+    )
+    with mock.patch.object(unblob.cli, "process_file", process_file_mock):
+        result = runner.invoke(unblob.cli.cli, params)
+
+    assert result.exit_code != 0
+    assert isinstance(result.exception, PermissionError)
+    process_file_mock.assert_called_once()
```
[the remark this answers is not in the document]
vlaci: You I right, I had a tunnel-vision for the exceptions.

unblob/processing.py
```diff
@@ -110,7 +110,6 @@ def get_extract_dir_for(self, path: Path) -> Path:
         return extract_dir.expanduser().resolve()
 
 
-@terminate_gracefully
```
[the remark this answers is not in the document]
vlaci: This line meant to be in the previous patch

unblob/sandbox.py
vlaci: Yes it does. We are using a high-level API, so we don't need to bother with that granularity.

https://docs.rs/landlock/0.4.1/src/landlock/fs.rs.html#106-115

In theory, we could filter execution, but it looks like a can of worms, as we need to be sure, what tools could be transitively executed

tests/test_sandbox.py
```diff
@@ -0,0 +1,57 @@
...
+
+    sandbox.run(extraction_config.extract_root.mkdir, parents=True)
+    sandbox.run(directory_in_extract_root.mkdir, parents=True)
+
+    sandbox.run(file_in_extract_root.touch)
+    sandbox.run(file_in_extract_root.write_text, "file content")
+
+    # log-file is already opened
+    log_path.touch()
+    sandbox.run(log_path.write_text, "log line")
```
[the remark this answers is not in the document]
vlaci: Yep

unblob/pool.py
qkaiser: How about this ?

```py
while pool := next(iter(pools)):
    pool.close(immediate=True)
```

unblob/sandbox.py
qkaiser: > In theory, we could filter execution, but it looks like a can of worms, as we need to be sure, what tools could be transitively executed

Let's not go there. yet :)

unblob/pool.py
```diff
@@ -129,14 +167,16 @@ def submit(self, args):
         self._input.put(args)
 
     def process_until_done(self):
-        while not self._input.is_empty():
-            result = self._output.get()
-            self._result_callback(self, result)
-            self._input.task_done()
+        with contextlib.suppress(EOFError):
```
e3krisztian, line 170: Where do this `EOFError` come from?

unblob/pool.py
e3krisztian, line 1: This went into the wrong commit...

unblob/sandbox.py
```diff
@@ -0,0 +1,118 @@
+import ctypes
+import sys
+import threading
+from pathlib import Path
+from typing import Callable, Iterable, Optional, Type, TypeVar
+
+from structlog import get_logger
+from unblob_native.sandbox import (
+    AccessFS,
```
e3krisztian, line 9: I know this comes from Landlock terminology, but this `AccessFS` name looks so alien to me (is it a file-system?).
I would use `FSAccess`, if possible.

unblob/sandbox.py
```diff
@@ -0,0 +1,118 @@
...
+            AccessFS.make_dir(config.extract_root.parent),
+            AccessFS.read_write(log_path),
+            *extra_restrictions,
+        ]
+
+        if report_file:
+            self.restrictions += [
+                AccessFS.read_write(report_file),
+                AccessFS.make_reg(report_file.parent),
+            ]
```
e3krisztian, line 58: Since we have a deny-all, then allow exceptions mechanism, maybe a more specific name could be used.
```suggestion
        extra_passthrough: Iterable[AccessFS] = (),
    ):
        self.passthrough = [
            # Python, shared libraries, extractor binaries and so on
            AccessFS.read("/"),
            # Multiprocessing
            AccessFS.read_write("/dev/shm"),  # noqa: S108
            # Extracted contents
            AccessFS.read_write(config.extract_root),
            AccessFS.make_dir(config.extract_root.parent),
            AccessFS.read_write(log_path),
            *extra_passthrough,
        ]

        if report_file:
            self.passthrough += [
                AccessFS.read_write(report_file),
                AccessFS.make_reg(report_file.parent),
            ]
```

unblob/pool.py
vlaci: It will never be false anyway as `next` will return something or raise an error.

unblob/pool.py
vlaci: from multiprocessing's queue, as it is using pipes

unblob/pool.py
```diff
@@ -1,18 +1,24 @@
...
+import contextlib
 import multiprocessing as mp
 import os
 import queue
+import signal
 import sys
 import threading
 from multiprocessing.queues import JoinableQueue
-from typing import Any, Callable, Union
+from typing import Any, Callable, Set, Union
```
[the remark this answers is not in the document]
qkaiser: I would merge https://github.com/onekey-sec/unblob/pull/1008 without this then.

unblob/pool.py
```diff
@@ -87,14 +107,32 @@ def __init__(
...
+
         self._wait_for_workers_to_quit()
+        super().close(immediate=immediate)
+
+    def _terminate_workers(self):
+        for proc in self._procs:
+            proc.terminate()
+
+        self._input.close()
+        if sys.version_info >= (3, 9):
```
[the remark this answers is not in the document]
qkaiser: I would merge https://github.com/onekey-sec/unblob/pull/1008 without this then.

unblob/sandbox.py
qkaiser: This would require a rewrite of `unblob-native` which itself follows the naming convention coming from the landlock Rust crate (see https://github.com/onekey-sec/unblob-native/blob/main/src/sandbox/linux.rs#L2). I would keep things as they are right now.

unblob/pool.py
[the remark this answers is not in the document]
vlaci: We discussed since then, that we'll deprecate 3.8 independently of this

qkaiser:
@vlaci we're one conflict away from a merge I think :)

qkaiser (approved):

qkaiser (approved):

PR #622, opened 2023-06-23
Add execution timeout to Command Extractor

### LINKED ISSUE #621: No timeout defined for Command Extractor
The framework calls external extractors without setting an explicit execution timeout. If an external extractors goes haywire (e.g. [debugfs getting into an endless loop](https://github.com/tytso/e2fsprogs/pull/149)), unblob may never finish.

After an internal discussion, we decided to set a timeout value that is high enough not to block long running execution such as extraction of large disk images, but small enough to make sure unblob finish its execution at some point.

We agreed to a 12 hours timeout :nerd_face:

qkaiser:
Fix #621

### DIFF
--- a/unblob/extractors/command.py
+++ b/unblob/extractors/command.py
@@ -6,10 +6,18 @@ from typing import TYPE_CHECKING, List, Optional, Union
 from structlog import get_logger
 
 from unblob.models import DirectoryExtractor, ExtractError, Extractor
-from unblob.report import ExtractCommandFailedReport, ExtractorDependencyNotFoundReport
+from unblob.report import (
+    ExtractCommandFailedReport,
+    ExtractorDependencyNotFoundReport,
+    ExtractorTimedOut,
+)
 
 if TYPE_CHECKING:
     import io
+
+# value that is high enough not to block long running execution such as extraction of large
+# disk images, but small enough to make sure unblob finish its execution at some point.
+COMMAND_TIMEOUT = 12 * 60 * 60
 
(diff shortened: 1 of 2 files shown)


e3krisztian (approved):
I think it solves a concrete problem, so I do not request bigger immediate changes, but I am slightly unhappy about it:
- 12 hours is too much for a single extractor, I would expect them to be fast, and would prefer a 10 minute limit instead, if that limit is not enough, we should definitely look at what is wrong -  do we know of  something slow, and non-problematic?
- the limit is not configurable
- it is in a need of a test (monkey-patching a sub-second timeout and `sleep 60`  as external command would work)
- it is only for external commands?

qkaiser:
> I think it solves a concrete problem, so I do not request bigger immediate changes, but I am slightly unhappy about it:
> 
> * 12 hours is too much for a single extractor, I would expect them to be fast, and would prefer a 10 minute limit instead, if that limit is not enough, we should definitely look at what is wrong -  do we know of  something slow, and non-problematic?

Extraction time is dependent on the input file size. We chose this value because the objective is to make sure unblob stops at some point, not to spot slow extractors.

> * the limit is not configurable

Yes, I don't think we want it to be at this point.

> * it is in a need of a test (monkey-patching a sub-second timeout and `sleep 60`  as external command would work)

I can add it to this branch, what do you say ?

> * it is only for external commands?

Yes, endless loops have been observed in external commands. When that happens in our own code we can fix it :)

PR #632, opened 2023-07-26
fix(handler): handle truncated FAT images

qkaiser:
Some FAT samples that we have are truncated to match exactly the size of reserved sectors + FAT sectors + active cluster sectors. It's like the initial image minus the free clusters.
    
Another approach to compute the size of a FAT chunk is to parse the File Allocation Table (FAT) and consider that the FAT chunk size is equal to the amount of active clusters multiplied by cluster size + size of reserved sectors + size of FAT sectors.
    
This approach works great with truncated images but not so much with valid images as it triggers 'Unexpected end of archive' in 7z, even if the aforementioned end of archive that got stripped is full of null bytes.
    
Therefore, we only apply it as a fallback mechanism when the chunk size overflows the file.


unblob/handlers/filesystem/fat.py
```diff
@@ -155,6 +155,11 @@ def calculate_chunk(self, file: File, start_offset: int) -> Optional[ValidChunk]
 
         size = header.common.BytesPerSec * sector_count
 
+        # handle corrupted FAT images by considering the whole file as the FAT image.
+        file_size = file.size()
+        if start_offset == 0 and size > file_size:
```
martonilles, line 160: do we know the overflowed chunk size? maybe we can consider that

qkaiser: Looking at the Android OTA sample, we have these FAT16 images:

```
bluetooth.img:         DOS/MBR boot sector, code offset 0x3c+2, OEM-ID "MSDOS5.0", Bytes/sector 4096, sectors/cluster 4, root entries 512, Media descriptor 0xf8, sectors/FAT 3, sectors/track 63, heads 255, sectors 16384 (volumes > 32 MB), serial number 0xbc614e, unlabeled, FAT (16 bit)
modem_apq.img:         DOS/MBR boot sector, code offset 0x3c+2, OEM-ID "MSDOS5.0", sectors/cluster 32, root entries 512, Media descriptor 0xf8, sectors/FAT 28, sectors/track 63, heads 255, sectors 225280 (volumes > 32 MB), serial number 0xbc614e, unlabeled, FAT (16 bit)
modem.img:             DOS/MBR boot sector, code offset 0x3c+2, OEM-ID "MSDOS5.0", sectors/cluster 32, root entries 512, Media descriptor 0xf8, sectors/FAT 28, sectors/track 63, heads 255, sectors 225280 (volumes > 32 MB), serial number 0xbc614e, unlabeled, FAT (16 bit)
```

Looking at them with `7z`:

```
Path = bluetooth.img
Type = FAT
ERRORS:
Unexpected end of archive
Physical Size = 67108864
File System = FAT16
Cluster Size = 16384
Free Space = 66338816
Headers Size = 61440
Sector Size = 4096
ID = 12345678


Path = modem_apq.img
Type = FAT
ERRORS:
Unexpected end of archive
Physical Size = 115343360
File System = FAT16
Cluster Size = 16384
Free Space = 73269248
Headers Size = 160256
Sector Size = 512
ID = 12345678


Path = modem.img
Type = FAT
ERRORS:
Unexpected end of archive
Physical Size = 115343360
File System = FAT16
Cluster Size = 16384
Free Space = 3981312
Headers Size = 5714432
Sector Size = 512
ID = 12345678
```

So in terms of difference between size in header and actual size:

```
67108864 - 765952 = 66342912
115343360 - 42074112 = 73269248
115343360 - 111362048 = 3981312
```

In terms of ratio of actual size vs size stored in header:

```
>>> (765952 / 67108864) * 100
1.141357421875
>>> (42074112 / 115343360) * 100
36.47727272727273
>>> (111362048 / 115343360) * 100
96.54829545454545
```

martonilles: hmm, it looks like that the actual size is around the "Physical size" - "Free Space", at least for the modem_apq.img and modem.img and for bluetooth.img it is almost the same except off by one sector which is 4k in that case. 

Maybe we should use this as a fall-back calculation?

qkaiser: We need to implement a file allocation table parser to calculate the free space. It's not trivial but we'll get an enhanced chunk size calculation out of it. I'll look into it.

qkaiser: Pushed a new solution that fallback to FAT parsing to calculate the end offset. Everything is explained in the updated description.

unblob/handlers/filesystem/fat.py
```diff
@@ -132,10 +149,72 @@ def valid_header(self, header: Instance) -> bool:
             return self.valid_fat32_header(header)
         return True
 
+    def get_fat_size(self, header: Instance) -> int:
```
martonilles, line 152: the name suggests it returns the size, but in reality it returns the sector count, either return real size, or change function name

unblob/handlers/filesystem/fat.py
```diff
@@ -155,6 +234,29 @@ def calculate_chunk(self, file: File, start_offset: int) -> Optional[ValidChunk]
...
 
+        file_size = file.size()
+        if start_offset + size > file_size:
+            logger.warning(
+                "Calculated chunk size overflows, falling back to active cluser based calculation."
+            )
+            file.seek(
+                start_offset + (header.common.BytesPerSec * header.common.RsvdSecCnt),
+                io.SEEK_SET,
+            )
```
martonilles, line 245: why seek here, if same seek is done in read_fat

unblob/handlers/filesystem/fat.py
```diff
@@ -132,10 +149,72 @@ def valid_header(self, header: Instance) -> bool:
...
+        if FsType(header.FileSysType) == FsType.FAT32:
+            return header.FATSz32
+        raise InvalidInputFormat("Invalid FAT type.")
+
+    def read_fat(self, header: Instance, start_offset: int, file: File) -> List[bytes]:
+        # we go past the reserved sectors (boot sector, fs information sector, custom)
+        file.seek(start_offset + (header.common.BytesPerSec * header.common.RsvdSecCnt))
+
+        fats = []
+        fat_size = self.get_fat_size(header) * header.common.BytesPerSec
```
martonilles, line 164: why calculate again fat_size if it is already calculated in parse_fat?

unblob/handlers/filesystem/fat.py
```diff
@@ -132,10 +149,72 @@ def valid_header(self, header: Instance) -> bool:
...
+
+        for _ in range(header.common.NumFATs):
+            content = file.read(fat_size)
+            if len(content) != fat_size:
+                raise InvalidInputFormat(
+                    "An error occured when reading the File Allocation Table."
+                )
+            fats += [content]
```
martonilles, line 172: we can use fats as set() as we just need one fat at the end

unblob/handlers/filesystem/fat.py
```diff
@@ -132,10 +149,72 @@ def valid_header(self, header: Instance) -> bool:
...
+        fat_size = self.get_fat_size(header) * header.common.BytesPerSec
+
+        fats = self.read_fat(header, start_offset, file)
+
+        entry_size = FAT_ENTRY_SIZES[FsType(header.FileSysType)]
+        total_entries = int(fat_size // entry_size)
+        fat = [0] * total_entries
+
+        main_file_allocation_table = fats[0]
+        # backup_file_allocation_table = fats[1]
```
martonilles, line 190: do we need it?

unblob/handlers/filesystem/fat.py
```diff
@@ -132,10 +149,72 @@ def valid_header(self, header: Instance) -> bool:
...
+
+        if FsType(header.FileSysType) == FsType.FAT32:
+            for i in range(total_entries):
+                entry = main_file_allocation_table[
+                    (i * entry_size) : (i * entry_size) + entry_size
+                ]
+                fat[i] = struct.unpack("<L", entry)[0] & 0x0F_FF_FF_FF
```
martonilles, line 197: we can user convert_int32

unblob/handlers/filesystem/fat.py
```diff
@@ -132,10 +149,72 @@ def valid_header(self, header: Instance) -> bool:
...
+        fats = self.read_fat(header, start_offset, file)
+
+        entry_size = FAT_ENTRY_SIZES[FsType(header.FileSysType)]
+        total_entries = int(fat_size // entry_size)
+        fat = [0] * total_entries
+
+        main_file_allocation_table = fats[0]
+        # backup_file_allocation_table = fats[1]
+
+        if FsType(header.FileSysType) == FsType.FAT32:
```
martonilles, line 192: match would be more elegant here ;)

martonilles (changes requested):
Would be nice also to add a test sample, where the FS is truncated like this

qkaiser:
Handling truncated filesystem images for every possible scenario would have required writing an entire FAT parsing stack. I started writing one but given the complexity I completely switched gears and moved to [pyfatfs](https://github.com/nathanhi/pyfatfs). It's an MIT licensed and maintained library.

It could become useful in the future if we want to use it as our extractor or to simply obtain file meta-data. In the meantime, we can rely on it to accurately identify the end offset of truncated filesystems.

qkaiser:
Had some issues with pyfatfs and Nix. Ended up forking the repository and making it use poetry at  https://github.com/qkaiser/pyfatfs/. Couldn't find a way to get rid of the `setuptools-scm` error, but maybe @vlaci has a way. Not sure forking is the right way to go.

vlaci:
Fixed build in https://github.com/onekey-sec/unblob/commit/171e2677e5277eedf3a715dd9fdf88683e8144e4

unblob/handlers/filesystem/fat.py
```diff
@@ -33,127 +51,25 @@ class FATHandler(StructHandler):
...
-            sector_count = header.TotSec32
-
-        if not self.valid_header(header):
-            raise InvalidInputFormat("Invalid FAT header.")
-
-        logger.debug("FAT header parsed", header=header, _verbosity=3)
-
-        if sector_count == 0x0:
-            sector_count = header.common.TotSectors
+            total_sectors = pyfat_fs.bpb_header["BPB_TotSec32"]
```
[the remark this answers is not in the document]
qkaiser: Fixed.

unblob/handlers/filesystem/fat.py
```diff
@@ -33,127 +51,25 @@ class FATHandler(StructHandler):
...
-        if header.BootSig not in VALID_NT_SIGNATURES:
-            return False
-        if header.FileSysType not in (b"FAT12   ", b"FAT16   "):
-            return self.valid_fat32_header(header)
-        return True
-
     def calculate_chunk(self, file: File, start_offset: int) -> Optional[ValidChunk]:
-        header = self.cparser_le.fat12_16_bootsec_t(file)
+        pyfat_fs = PyFat()
+        pyfat_fs.close = no_op
```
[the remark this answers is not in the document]
qkaiser: Fixed.

unblob/handlers/filesystem/fat.py
```diff
@@ -33,127 +51,25 @@ class FATHandler(StructHandler):
...
-            return False
-        if header.BootSig not in VALID_NT_SIGNATURES:
-            return False
-        if header.FileSysType not in (b"FAT12   ", b"FAT16   "):
-            return self.valid_fat32_header(header)
-        return True
-
     def calculate_chunk(self, file: File, start_offset: int) -> Optional[ValidChunk]:
-        header = self.cparser_le.fat12_16_bootsec_t(file)
+        pyfat_fs = PyFat()
```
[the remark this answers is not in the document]
qkaiser: Fixed.

unblob/handlers/filesystem/fat.py
```diff
@@ -1,21 +1,39 @@
...
-VALID_MEDIAS = [0xF0, 0xF8, 0xF9, 0xFA, 0xFB, 0xFC, 0xFD, 0xFE, 0xFF]
+def get_max_offset(
+    fs: PyFat, root_dir, max_cluster: int = 0, max_offset: int = 0
+) -> int:
+    dirs, files, _ = root_dir.get_entries()
 
+    for d in dirs:
+        if d.get_cluster() > max_cluster:
+            max_cluster = d.get_cluster()
+            max_offset = fs.get_data_cluster_address(max_cluster) + d.get_size()
```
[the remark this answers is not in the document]
qkaiser: I like the first approach very much, it's quite elegant :) I tested it out on a bunch of in-the-wild truncated FAT samples and it works flawlessly. Adapted the code accordingly.

e3krisztian (approved):
The commit needs a description update, as the algorithm has slightly changed and it no longer depends on the file size.

PR #633, opened 2023-07-28
Report extraction problems from handlers/extractors

e3krisztian:
Main changes are:
- `Extractor.extract` can optionally return an `ExtractionResult` which has a `reports` attribute.
- introduction of `Sandbox`,  which prevents out of extraction directory file system object creation/reference and records any such attempts
- support of problem reporting for `tar`, `romfs`, `yaffs`, `ipkg`, `hdr`, with most of them converted to use `Sandbox`

### DIFF
--- a/tests/test_file_utils.py
+++ b/tests/test_file_utils.py
@@ -1,4 +1,5 @@
 import io
+import os
 from pathlib import Path
 from typing import List
 
@@ -7,8 +8,10 @@ import pytest
 from unblob.file_utils import (
     Endian,
     File,
+    FileSystem,
     InvalidInputFormat,
     StructParser,
+    chop_root,
     convert_int8,
     convert_int16,
     convert_int32,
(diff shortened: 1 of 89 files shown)


qkaiser:
I like the approach of `SandboxFS` very much ! It makes writing secure unblob extractors easier for newcomers. Maybe use the name `SandboxFS` rather than `Sandbox` to convey the fact that it's about files rather than a generic sandbox with limits on execution / privileges / network.

Are there extractors other than tar, romfs, yaffs, ipkg, and hdr that could benefit from it ?

I'll do a deeper pass for review today.

e3krisztian:
Thanks, I hoped it will be useful later as well :)

This was the full evolution of the class name: `ExtractionFS` -> `SandboxFS` -> `FSSandbox` -> `DirectorySandbox` -> `Sandbox`.

I did not really like the `FS` suffix/prefix. I feel, that `DirectorySandbox` was the most descriptive of the bunch, but `Sandbox` was much shorter and won for me in the end as unique enough inside `unblob`.

I would prefer to keep the current `Sandbox` name, while the implementation is not something complex and generic, it is inside the helper module `unblob.file_utils` to show its scope, and if a more specific name is desired, it can be renamed on import.

Most of the `Handler`s use `Command` as extractor, either directly, or indirectly, or the extractor creates only a single output file with a deterministic name, thus not worth converting, so I think the list of converted extractors are exhaustive at the moment.

unblob/handlers/filesystem/romfs.py
```diff
@@ -250,92 +265,52 @@ def create_symlink(self, extract_root: Path, output_path: Path, inode: FileHeade
         else:
             target_path = extract_root.joinpath(target).resolve()
 
-        if not is_safe_path(extract_root, target_path):
-            logger.warning(
-                "Path traversal attempt through symlink.", target_path=target_path
-            )
+        if not self.fs.check_path(target_path, "symlink"):
```
qkaiser, line 268: Shouldn't this be handled by `self.fs.create_symlink(src=target_path, dst=output_path)` ?

unblob/handlers/filesystem/yaffs.py
```diff
@@ -474,80 +473,40 @@ def get_file_bytes(self, entry: YAFFSEntry) -> Iterable[bytes]:
...
-            out_path.mkdir(exist_ok=True)
+            fs.mknod(out_path, entry.st_mode, entry.st_rdev)
+        elif entry.object_type == YaffsObjectType.DIRECTORY:
+            fs.mkdir(out_path, exist_ok=True)
         elif entry.object_type == YaffsObjectType.FILE:
-            logger.debug("creating file", file_path=out_path, _verbosity=3)
-            with out_path.open("wb") as f:
-                for chunk in self.get_file_bytes(entry):
-                    f.write(chunk)
+            fs.write_chunks(out_path, self.get_file_bytes(entry))
```
qkaiser, line 502: I would rename `get_file_bytes` to `get_file_chunks` because it's confusing.

qkaiser (commented):
If we all agree on the API of `Sandbox`, can you update the documentation as part of this PR ?

I would put it under "Writing extractors" in docs/development.md

martonilles:
I like the Sandbox very much, not sure though about the name.

The other thing I was wondering about, what if we pass the Sandbox instance to the extractor instead of just the outdir? It would require api change, but might decrease boilerplate (not sure about CommandExtractor).

Also I was wondering if for the Sandbox apis do we want to pass absolute path or just a path within the sandbox? Maybe it would be easier if the extractor logic would not need to know about the outdir

unblob/handlers/filesystem/romfs.py
e3krisztian: The problem with refactors/rewrites over many file, is that small details are glossed over. I have made this error many times, so I wanted to stay safe here and keep the original logic.

The code continues as
```py
        if not self.fs.check_path(target_path, "symlink"):
            return
        # we create relative paths to make the output directory portable
        output_path.symlink_to(os.path.relpath(target_path, start=output_path.parent))
```
It was left unchanged intentionally, because there was a path calculation *after* the check.
I wanted to keep the first check, then continue with creating the symlink.

However you are right, with your intuition, that the symlink should be created via `self.fs.create_symlink()`, so it was changed.

unblob/handlers/filesystem/yaffs.py
e3krisztian: Agreed, and done.

qkaiser:
> The other thing I was wondering about, what if we pass the Sandbox instance to the extractor instead of just the outdir? It would require api change, but might decrease boilerplate (not sure about CommandExtractor).

I was not sure about this in the beginning since `Sandbox` would only be effective when the extractor does the extraction itself without relying on external tools or libraries. However, I like the idea of changing the API so that extractors receive a `Sandbox` instance, making it the only way for someone writing an extractor to interact with the extraction directory. Going into that direction, we need to define what is public/private within the Sandbox so that whenever someone does `sandbox.outdir.whatever` it's picked up and reported by pyright.

Of course `CommandExtractor` would be an exception to that since it would require access to `outdir`.

> Also I was wondering if for the Sandbox apis do we want to pass absolute path or just a path within the sandbox? Maybe it would be easier if the extractor logic would not need to know about the outdir

We could do things so that all paths provided to the Sandbox APIs are considered relative to the extraction directory. If an absolute path is provided, its root should be considered to be the outdir.

What do you think ?

unblob/file_utils.py
```diff
@@ -332,3 +353,147 @@ def read_until_past(file: File, pattern: bytes):
...
+        self.problems = []
+
+    def record_problem(self, problem: ExtractionProblem):
+        self.problems.append(problem)
+
+        msg = f"{problem.problem} {problem.resolution}"
+        if isinstance(problem, LinkExtractionProblem):
+            logger.warning(msg, path=problem.path, link_path=problem.link_path)
+        else:
+            logger.warning(msg, path=problem.path)
```
e3krisztian, line 384: `ExtractionProblem` should grow a `.log_me_with(logger)` method to avoid this `isinstance()` and support potential new detailed subclasses of `ExtractionProblem`.

e3krisztian:
Thanks for the comments @qkaiser @martonilles and @vlaci.
I agree with all of your comments, and they will be implemented in this MR:
- rename `Sandbox` to `FileSystem`
- accept only relative paths (in contrast to the current absolute paths), that will be interpreted as paths relative to `FileSystem.root`

These proposals/ideas are nice, and will be/should be implemented later:
- changing the `Extractor.extract(self, inpath: Path, outdir: Path)` signature
- sandboxing third party extractors with Landlock

unblob/handlers/filesystem/romfs.py
```diff
@@ -257,41 +253,41 @@ def walk_dir(self, addr: int, parent: Optional[FileHeader] = None):
...
+            target_path = Path(target)
 
         # we create relative paths to make the output directory portable
         target_path = Path(os.path.relpath(target_path, start=output_path.parent))
         self.fs.create_symlink(src=target_path, dst=output_path)
 
-    def create_hardlink(self, extract_root: Path, output_path: Path, inode: FileHeader):
+    def create_hardlink(self, output_path: Path, inode: FileHeader):
         if inode.spec_info in self.inodes:
             target = str(self.inodes[inode.spec_info].path).lstrip("/")
```
qkaiser, line 270: I guess this will be handled by `chop_root` later on so the `lstrip` is no longer required.

qkaiser:
@e3krisztian would be nice to follow atomic commit message structure using `feat`, `ci`, `doc`, `chore`, `fix`, `refactor` to indicate the type of operation and the affected module (`docs`, `handler`, `processing`, ...) between parenthesis. This greatly helps when looking for specific changes later on.

unblob/file_utils.py
```diff
@@ -386,60 +423,66 @@ def record_problem(self, problem: ExtractionProblem):
...
+        return None
 
     def write_bytes(self, path: Path, content: bytes):
         logger.debug("creating file", file_path=path, _verbosity=3)
-        if self.check_path(path, "write_bytes"):
-            path.write_bytes(content)
+        safe_path = self.get_checked_path(path, "write_bytes")
+
+        if safe_path:
+            safe_path.write_bytes(content)
```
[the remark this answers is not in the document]
e3krisztian: Yes, and it was like that, but changed for the less terse version, because I felt too many things was crammed into the `if` condition, while it should be simple.

Walrus is controversial, and in this case I do not think it is worth it:  more harm is done to readabillity, than it is gained through conciseness.

unblob/handlers/filesystem/romfs.py
```diff
@@ -257,41 +253,41 @@ def walk_dir(self, addr: int, parent: Optional[FileHeader] = None):
             self.inodes[addr] = file_header
         return file_header.next_filehdr
 
-    def create_symlink(self, extract_root: Path, output_path: Path, inode: FileHeader):
+    def create_symlink(self, output_path: Path, inode: FileHeader):
         target = inode.content.decode("utf-8").lstrip("/")
```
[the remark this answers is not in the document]
e3krisztian: Symlinks require special attention. Further down this function many path manipulations happen with `target` before it gets passed to `fs.create_symlink` and eventually `chop_root`, so I would not touch it.

unblob/handlers/filesystem/romfs.py
e3krisztian: Agreed, and removing this even conversion to `str` and back to `Path` can be avoided.

qkaiser (approved):

unblob/handlers/filesystem/romfs.py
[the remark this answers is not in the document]
e3krisztian: Thanks for being persistent! There was a bug in `create_symlink` that was covered up by these lines.

tests/test_file_utils.py
```diff
@@ -338,3 +363,198 @@ def test_get_endian_resets_the_file_pointer(self):
...
+        """Document a remaining path traversal scenario through a symlink chain.
+
+        unblob.extractor.fix_symlinks() exists to cover up cases like this.
+        """
+        (sandbox_parent / "outer-secret").write_text("private key")
+
+        # The path traversal is possible because at the creation of "secret" "future" does not exist
+        # so it is not yet possible to determine if it will be a symlink to be allowed or not.
+        # When the order of the below 2 lines are changed, the path traversal is recognized and prevented.
+        sandbox.create_symlink(Path("future/../outer-secret"), Path("secret"))
```
qkaiser, line 527: I initially felt that this should have been handled by the Sandbox, but since we still need to call `fix_extracted_directory` and `fix_symlink` for third party extractors, I'm okay with the way things are. Just noting it for further references.

unblob/file_utils.py
```diff
@@ -332,3 +357,204 @@ def read_until_past(file: File, pattern: bytes):
...
+
+    def write_chunks(self, path: Path, chunks: Iterable[bytes]):
+        logger.debug("creating file", file_path=path, _verbosity=3)
+        safe_path = self.get_checked_path(path, "write_chunks")
+
+        if safe_path:
+            with safe_path.open("wb") as f:
```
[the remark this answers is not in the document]
qkaiser: We discussed this offline and came to an agreement on staying with `wb` here.

Excerpt:

> I saw your xb comment and was thinking about it, but wanted to be on the safe side: `xb` raises some `OSError` derivative exception (`FileExistsException`?) when the file already exists, and we have formats that have different versions of the same file (e.g. tar and zip). We could still extract these formats by overwriting the existing file. I remember this for some extractors, but do not know where.

PR #639, opened 2023-08-08
Draft: feat(handler): add support for non-standard squashfs v2 signature.

qkaiser:
Will remove draft tag when https://github.com/onekey-sec/sasquatch/issues/19 is solved when sasquatch dependency is updated here.


qkaiser:
No longer in draft. I would wait for https://github.com/onekey-sec/unblob/pull/1352 to land first so we don't have too much conflicts.

PR #642, opened 2023-08-18
Enhance FileSystem class

e3krisztian:
The following enhancements are added:
- add `FileSystem.open` to create files and return a file object to write into
- automatically create missing parent directories
- `get_checked_path` is no longer public (and was renamed to remove the temptation to use for something else)

### DIFF
--- a/unblob/file_utils.py
+++ b/unblob/file_utils.py
@@ -429,7 +429,9 @@ class FileSystem:
     def _ensure_parent_dir(self, path: Path):
         path.parent.mkdir(parents=True, exist_ok=True)
 
-    def get_checked_path(self, path: Path, path_use_description: str) -> Optional[Path]:
+    def _get_extraction_path(
+        self, path: Path, path_use_description: str
+    ) -> Optional[Path]:
         fs_path = self._fs_path(path)
         if fs_path.is_safe:
             return fs_path.absolute_path
@@ -444,7 +446,7 @@ class FileSystem:
 
     def write_bytes(self, path: Path, content: bytes):
         logger.debug("creating file", file_path=path, _verbosity=3)
-        safe_path = self.get_checked_path(path, "write_bytes")
+        safe_path = self._get_extraction_path(path, "write_bytes")
 
         if safe_path:
             self._ensure_parent_dir(safe_path)
@@ -452,7 +454,7 @@ class FileSystem:
 
     def write_chunks(self, path: Path, chunks: Iterable[bytes]):
         logger.debug("creating file", file_path=path, _verbosity=3)
-        safe_path = self.get_checked_path(path, "write_chunks")
+        safe_path = self._get_extraction_path(path, "write_chunks")
 
         if safe_path:
             self._ensure_parent_dir(safe_path)
@@ -462,7 +464,7 @@ class FileSystem:
 
     def carve(self, path: Path, file: File, start_offset: int, size: int):
         logger.debug("carving file", path=path, _verbosity=3)
-        safe_path = self.get_checked_path(path, "carve")
+        safe_path = self._get_extraction_path(path, "carve")
 
         if safe_path:
             self._ensure_parent_dir(safe_path)
@@ -470,7 +472,7 @@ class FileSystem:
 
     def mkdir(self, path: Path, *, mode=0o777, parents=False, exist_ok=False):
         logger.debug("creating directory", dir_path=path, _verbosity=3)
-        safe_path = self.get_checked_path(path, "mkdir")
+        safe_path = self._get_extraction_path(path, "mkdir")
 
         if safe_path:
             self._ensure_parent_dir(safe_path)
@@ -478,7 +480,7 @@ class FileSystem:
 
     def mkfifo(self, path: Path, mode=0o666):
         logger.debug("creating fifo", path=path, _verbosity=3)
-        safe_path = self.get_checked_path(path, "mkfifo")
+        safe_path = self._get_extraction_path(path, "mkfifo")
 
         if safe_path:
             self._ensure_parent_dir(safe_path)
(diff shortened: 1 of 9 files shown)


unblob/file_utils.py
```diff
@@ -426,6 +427,9 @@ def has_root_permissions(self):
     def _fs_path(self, path: Path) -> _FSPath:
         return _FSPath(root=self.root, path=path)
 
+    def _ensure_parent_dir(self, path: Path):
```
martonilles, line 430: do we want to always do this, even if the extracted chunk is broken and actually miss that filesystem or just handle the "wrong order" case?

I am ok both, just want to make sure that it is a conscious decision

unblob/file_utils.py
```diff
@@ -426,6 +427,9 @@ def has_root_permissions(self):
...
+        path.parent.mkdir(parents=True, exist_ok=True)
+
     def get_checked_path(self, path: Path, path_use_description: str) -> Optional[Path]:
```
martonilles, line 433: I know it was in the previous PR, but this function for me is more like a get_absolute_path (which also checks for traversal). It is not just a checking, rather a conversion with a check

unblob/file_utils.py
```diff
@@ -558,3 +570,23 @@ def create_hardlink(self, src: Path, dst: Path):
...
+        logger.debug("create/open binary file for writing", file_path=path)
+        safe_path = self.get_checked_path(path, "open")
+
+        if safe_path:
+            self._ensure_parent_dir(safe_path)
+            mode = "r+b" if safe_path.exists() else "w+b"
+            return safe_path.open(mode)
+
+        # promised of not failing on bad paths:
+        return tempfile.TemporaryFile("w+b")
```
martonilles, line 592: not sure why we want to return a temp file instead of failing/raising execption?

Also do we want to allow overwriting existing file by default? I think it should be controllable by the caller

unblob/file_utils.py
e3krisztian: These were the reasons:
- for properly working handlers (ordering is right) it is safe to do
- for handlers that are dealing with formats that do not necessarily have directory entries, this makes the code much simpler

unblob/file_utils.py
e3krisztian: This is a function that does 2 common things, because it is removing a common boilerplate (smelly I agree). However the emphasis is definitely on the checking side, so `get_absolute_path` would be a very misleading name, as there is no explicit intention to return an *absolute* path, the goal is to return an *actionable* path, or `None` if it is not possible. Returning an absolute path is the easiest implementation of the actionable path, but being absolute is not an important property, and could be replaced with a working directory relative actionable path as well.

Possible improvements:
- make it protected/private, as it is not something to use externally
- change name to `_get_actionable_path`

unblob/file_utils.py
e3krisztian: The goal of `FileSystem` helper is to ease writing extractors, by removing the burden of dealing with the path traversal problem.

This means, that operations should succeed even if there is a path traversal, but the path traversal attempt is recorded for later reporting.

The problem here is that this operation differs from the rest, as it also returns an object to be used later in the extractor.
If we would raise an exception here or return value None, that would force a check (actually for path traversal) at the extractor side, as without that further operations would fail. So there was a need for a dummy return value, and due to its simplicity the stdlib `TemporaryFile` was used as a safe substitute.

Raising an exception would make this operation potentially fatal for the extractor, while this could be only one bad file of the many to be extracted, so that is not a good solution.

e3krisztian: I agree, the implicit smartness of opening the file is too much, we should have a control in place to open existing or require new file, maybe also fail on existing as well.

Something like
```py
def open_binary(self, path, non_existing=True, truncate_existing=True):
```
Where `non_existing` would result in `mode="xb+"` (raises error on existing), while `truncate_existing` would result in `mode="wb+"`, while both being False with an existing file would be `mode="rb+"`.

unblob/file_utils.py
```diff
@@ -429,80 +453,71 @@ def _fs_path(self, path: Path) -> _FSPath:
...
-            path=fs_path.format_path(),
-            problem=f"Potential path traversal through {path_use_description}",
-            resolution="Skipped.",
-        )
-        self.record_problem(report)
-        return None
+        if not fs_path.is_safe:
+            report = ExtractionProblem(
+                path=fs_path.format_path(),
+                problem=f"Potential path traversal through {path_use_description}",
```
martonilles, line 462: I would include in the report the path pointing to lost+found so this way we can get the original path easily

unblob/file_utils.py
```diff
@@ -368,12 +371,33 @@ def chop_root(path: Path):
...
+    dir_hash = hashlib.sha224(str(dir_path).encode(errors="ignore")).hexdigest()
+
+    # adapted from https://stackoverflow.com/questions/5574042/string-slugification-in-python
+    dir_slug = str(dir_path)
+    dir_slug = unicodedata.normalize("NFKD", dir_slug)
+    dir_slug = dir_slug.encode("ascii", "ignore").lower()
+    dir_slug = re.sub(rb"[^a-z0-9]+", b"-", dir_slug).strip(b"-")
+    dir_slug = re.sub(rb"[-]+", b"-", dir_slug).decode()
+
+    return Path(f".unblob-lost+found/{dir_slug}_{dir_hash}/{filename}")
```
martonilles, line 389: I would swap dir_slug and dir_hash, as that would be easier to read the slug/filename assuming most slug would be the original directory:
```
UNBLOB_LOST_FOUND = ".unblob-lost+found"

...


return Path(UNBLOB_LOST_FOUND) / f"{dir_hash}_{dir_slug}" / filename
```

martonilles (approved):
looks nice, two minor comments, but overall I like it very much! :)

unblob/file_utils.py
e3krisztian: done.

qkaiser (approved):

martonilles (approved):

Issue #640, opened 2023-08-11
`7z` fails to extract initramfs


NiklasGollenstede:
**Describe the bug**
I was browsing Nix projects when I found this one and thought "that's cool, should be useful to inspect `initrd`s, but unfortunately, that currently does not really work very much.

The below command tries to extract a zstd-compressed NixOS initramfs with prepended Intel CPU microcode, but:

* `7z` fails to extract the main CPIO archive, but `unblob` still exits with 0/success.
* The prepended microcode CPIO archive is reported to be found, but then missing in the output.
* There are `unknown` chunks that are really just zero-padding.

**To Reproduce**
```
 nix run github:onekey-sec/unblob/23.5.31 -- $( nix build --no-link --print-out-paths github:srid/nixos-config/1a6879bbd1c0f87f67533a7b91bc438e042b3bf6#nixosConfigurations.actual.config.system.build.initialRamdisk )/initrd
```

<details><summary>Command output</summary>

```
2023-08-10 23:58.54 [info     ] Start processing file          file=/nix/store/msmx1ylsyhxk6hx3p4nz39vqi2gkzn3j-initrd-linux-6.1.43/initrd pid=3806009
2023-08-10 23:58.54 [warning  ] Found unknown Chunks           chunks=[0x6f1200-0x6f1800] pid=3806014
2023-08-10 23:58.54 [info     ] Extracting unknown chunk       chunk=0x6f1200-0x6f1800 path=initrd_extract/7279104-7280640.unknown pid=3806014
2023-08-10 23:58.54 [info     ] Extracting valid chunk         chunk=0x6f1800-0x12284cb path=initrd_extract/7280640-19039435.zstd pid=3806014
2023-08-10 23:58.54 [info     ] Extracting valid chunk         chunk=0x0-0x6f1200 path=initrd_extract/0-7279104.cpio_portable_ascii pid=3806014
2023-08-10 23:58.54 [warning  ] Found unknown Chunks           chunks=[0x19b3000-0x19b4000] pid=3806016
2023-08-10 23:58.54 [info     ] Extracting unknown chunk       chunk=0x19b3000-0x19b4000 path=initrd_extract/7280640-19039435.zstd_extract/zstd.uncompressed_extract/26947584-26951680.unknown pid=3806016
[...]

PR #648, opened 2023-08-31
replace CPIO extractor (7z) by unblob based extractor.

qkaiser:
Multiple issues were identified in how 7z handles CPIO:
- processing of CPIO archives depends on the file name, and the rules are not explicit (name.cpio is OK, name.cpio.truncated is OK, but name.cpio.ext is not)
- special device files (e.g. block devices, character devices) are extracted as empty files rather than created using mknod

We therefore decided to implement a CPIO extractor based on unblob's `Extractor`. Since CPIO has so many different format, each format has its own extractor:

- binary CPIO (bin)- BinaryCPIOExtractor
- portable ASCII CPIO (odc) - PortableOldASCIIExtractor
- portable ASCII CPIO (newc) - PortableASCIIExtractor
- portable ASCII with CRC (crc) - PortableASCIIWithCRCExtractor

Each extractor use a different parser, adapted to the format.

Related to #640

### DIFF
--- a/unblob/handlers/archive/cpio.py
+++ b/unblob/handlers/archive/cpio.py
@@ -1,15 +1,35 @@
 import io
-from typing import Optional
+import os
+import stat
+from pathlib import Path
+from typing import Optional, Type
 
+import attr
+from dissect.cstruct import Instance
 from structlog import get_logger
 
-from ...extractors import Command
-from ...file_utils import OffsetFile, decode_int, round_up, snull
-from ...models import File, HexString, StructHandler, ValidChunk
+from ...file_utils import (
+    Endian,
+    FileSystem,
+    InvalidInputFormat,
+    StructParser,
+    decode_int,
+    iterate_file,
+    round_up,
+    snull,
+)
+from ...models import (
+    Extractor,
+    ExtractResult,
+    File,
+    Handler,
+    HexString,
+    ValidChunk,
+)
 
 logger = get_logger()
 
-CPIO_TRAILER_NAME = b"TRAILER!!!"
+CPIO_TRAILER_NAME = "TRAILER!!!"
 MAX_LINUX_PATH_LENGTH = 0x1000
 
 C_ISBLK = 0o60000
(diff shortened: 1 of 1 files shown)


unblob/handlers/archive/cpio.py
```diff
@@ -92,20 +179,56 @@ def calculate_chunk(  # noqa: C901
...
+            self.offset_file, current_offset - self.start_offset
         )
+        if self.start_offset == self.end_offset:
+            raise InvalidInputFormat("Invalid CPIO archive.")
+
+    def dump_entries(self, fs: FileSystem):
+        for entry in self.entries:
+            # skip entries with "." as filename
+            if not entry.path.name:
+                continue
```
e3krisztian, line 208: I was surprised by this fragment. I would have expected `.` as name for `Path(".")` as `..` is returned for `Path("..")`.
This behavior is not documented at https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.name, so it might be a bug (https://bugs.python.org/issue37130), that is fixed later.
I propose the below change, which will work with both the current bad behavior, and with the fixed `Path.name` behavior:

```suggestion
            # skip entries with "." as filename, currently Path(".").name is ""
            #  see https://bugs.python.org/issue37130
            if entry.path.name in ("", "."):
                continue
```

unblob/handlers/archive/cpio.py
```diff
@@ -253,19 +338,145 @@ def _calculate_name_size(header) -> int:
...
-    NAME = "cpio_portable_ascii_crc"
-    PATTERNS = [HexString("30 37 30 37 30 32 // 07 07 02")]
+    @staticmethod
+    def _calculate_rdev(header) -> int:
+        return os.makedev(
+            decode_int(header.c_rdev_maj, 16), decode_int(header.c_rdev_min, 16)
+        )
+
+
+class PortableASCIIWithCRCParser(CPIOParserBase):
```
e3krisztian, line 354: This is an exact duplicate of `PortableASCIIParser` above  (or I have missed the difference).

I think the original intention was that in this class there is an implementation of a CRC check against `c_chksum`, that is called from `CPIOParserBase.parse`, which has a no-op base implementation, but it was never there.
Even in the CRC check case, all the below implementation should go, and `PortableASCIIWithCRCParser` should derive from `PortableASCIIParser`, and implement only `check_crc`.

unblob/handlers/archive/cpio.py
qkaiser: Yeah it's the most counter-intuitive part of pathlib to me. I implemented your suggestion.

unblob/handlers/archive/cpio.py
qkaiser: Makes sense. Implemented your suggestion with `check_crc`.

unblob/handlers/archive/cpio.py
```diff
@@ -70,19 +151,28 @@ def calculate_chunk(  # noqa: C901
...
+
+            if filename == CPIO_TRAILER_NAME:
+                current_offset += self._pad_content(c_filesize)
+                break
 
-                filename = snull(tmp_filename)
-                if filename == CPIO_TRAILER_NAME:
-                    current_offset += self._pad_content(header, c_filesize, c_namesize)
-                    break
+            self.file.seek(current_offset, io.SEEK_SET)
```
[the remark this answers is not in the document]
qkaiser: Got rid of all the unnecessary `seek`s within the handler (there was quite a lot ...).

unblob/handlers/archive/cpio.py
```diff
@@ -136,31 +267,30 @@ def _calculate_name_size(header) -> int:
...
-            uint16 c_mtimes[2];
-            uint16 c_namesize;
-            uint16 c_filesize[2];
-        } old_cpio_header_t;
-    """
-    HEADER_STRUCT = "old_cpio_header_t"
+    def check_crc(self, header) -> bool:
+        return bool(header)
 
+    def calculate_crc(self, start_offset: int, size: int) -> int:
```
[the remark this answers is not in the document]
qkaiser: Fixed !

e3krisztian (changes requested):

unblob/handlers/archive/cpio.py
```diff
@@ -92,23 +179,59 @@ def calculate_chunk(  # noqa: C901
...
+                fs.mknod(entry.path, mode=entry.mode & 0o777, device=entry.rdev)
+            else:
+                logger.warning("unknown file type in CPIO archive")
 
     @classmethod
-    def _pad_file(cls, file: OffsetFile, end_offset: int) -> int:
+    def _pad_file(cls, file: File, end_offset: int) -> int:
         """CPIO archives can have a 512 bytes block padding at the end."""
         file.seek(end_offset, io.SEEK_SET)
         padded_end_offset = round_up(end_offset, cls._FILE_PAD_ALIGN)
```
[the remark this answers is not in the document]
qkaiser: Done and tested implementation with CPIO samples prefixed and suffixed with odd/even length, and the case you mentioned with start offset at 256.

unblob/handlers/archive/cpio.py
```diff
@@ -92,37 +179,78 @@ def calculate_chunk(  # noqa: C901
...
+            else:
+                logger.warning("unknown file type in CPIO archive")
+
+    def _pad_file(self, end_offset: int) -> int:
         """CPIO archives can have a 512 bytes block padding at the end."""
-        file.seek(end_offset, io.SEEK_SET)
-        padded_end_offset = round_up(end_offset, cls._FILE_PAD_ALIGN)
+        self.file.seek(end_offset, io.SEEK_SET)
+        padded_end_offset = self.start_offset + round_up(
+            end_offset, self._FILE_PAD_ALIGN
```
e3krisztian, line 237:
```suggestion
            size=end_offset - self.start_offset,
            alignment=self._FILE_PAD_ALIGN,
```

qkaiser: Applied your suggestion. Thanks for you help during this review !

unblob/handlers/archive/cpio.py
```diff
@@ -253,19 +344,128 @@ def _calculate_name_size(header) -> int:
...
+    """A common base for all CPIO formats.
+
+    The format should be parsed the same, there are small differences how to calculate
+    file and filename sizes padding and conversion from octal / hex.
+    """
+
+    EXTRACTOR = None
+
+    def calculate_chunk(self, file: File, start_offset: int) -> Optional[ValidChunk]:
+        raise NotImplementedError
```
e3krisztian, line 410: I think the current code is clean, there is only the redundant looking implementation of `calculate_chunk()` for every CPIO `Handler` below.

Maybe they could be eliminated by implementing a generic one here:

```suggestion
    EXTRACTOR: _CPIOExtractorBase

    def calculate_chunk(self, file: File, start_offset: int) -> Optional[ValidChunk]:
        parser = self.EXTRACTOR.PARSER(file, start_offset)
        parser.parse()
        return ValidChunk(
            start_offset=start_offset,
            end_offset=parser.end_offset,
        )
```

PR #652, opened 2023-09-08
update wiki

qkaiser:
Update wiki with the following:
- mention installation in Kali Linux since unblob is available through Kali
- update publications section with latest workshops
- update supported formats with all the vendor formats @nyuware added during their internship


e3krisztian (approved):
`CPIO` could now also be changed to `unblob extractor`

Issue #654, opened 2023-09-25
Lack of support for tar with UNIX-compatible format


qkaiser:
tar archives can be created using either the USTAR format or UNIX-compatible format. Right now unblob only supports the ones using USTAR by matching on the `ustar` magic.

We need to extend the tar handler in order to match on UNIX-compatible header. This is a bit more complex since there is no magic value within these headers.

Specifically, a UNIX-compatible header looks like this:

```c
typedef struct unix_header {
char name[100];
char mode[8];
char uid[8];
char gid[8];
char size[12];
char mtime[12];
char chksum[8];
char link[1];
char linkname[100];
} unix_header_t;
```

PR #655, opened 2023-09-25
fix(handlers): add support for unix-compatible (aka v7) tar files.

qkaiser:
v7 tar headers do not have the 'ustar' magic that we match on for modern tar files. In order to match on those v7 archive, we build a regular expression that matches on mode, uid, gid, mtime, size files given their properties:

- fixed size (e.g. 8 for mode)
- optionally prepended by whitespaces
- suffixed by null bytes (null terminated)
- ASCII encoded digits (0x30 to 0x39)

In order to build a pattern that can be handled by hyperscan without using a notation such as `\w{1,99}` for path name (see ¹ for detailed explanation), we rely on utility function to build a regular expression to match all possible combination using the or (`|`) operator.

Note: hyperscan will yield a `Pattern is too large` exception when trying to use `{1,99}` notation. Even though we found out that using `.*` works, it would have an important performance impact on pattern matching. That's why we decided to go with the OR operator approach with combination.

----

Although it may look simple, this change is significant and will require:

- [x] testing with real fws (performance and regression)
- [x] reviewing by someone else (@martonilles might be the best candidate, as @vlaci will be away)
- [x] ~potentially re-merging the tar handlers (somewhat optional, and needs to be tested) - see ²~

----

The current solution is the product of @qkaiser and @e3krisztian collaboration, so neither of us should be the final approver.

See #654 

¹ https://intel.github.io/hyperscan/dev-reference/compilation.html
² the v7 matching rule probably supersedes the one defined for ustar so we can probably leave it out


e3krisztian:
Made an attempt at using only the new version without the uStar handler, and after looking into `typeflag` related problems it looks working for the test inputs and also on a non-trivial sample, I think it could be possible to merge again the implementations.

e3krisztian:
Numbers are stored in octal, but there is a GNU extension, that does base-256 encoding to store bigger than 8G numbers, base-256 is harder to cover (there are prefix bytes), and also might not worth it: the uStar version already knows about this encoding, the now introduced does not need it for old archives.

Note, that this new format appears only for bigger tar files > 8GB.

So we might not need to merge the implementations in the end.

qkaiser:
Tested with samples with tar content that I have here (~3GB) and did not observe regressions. This patch has very limited performance impact. These are the stats from running executions with a 2.4GB real world tar sample:

performance results on main:

```
Performance counter stats for 'poetry run unblob -v -e /tmp/out -f -k -d 1 /home/quentin/Downloads/cvd-host_package.tar':

         13.130,53 msec task-clock                       #    1,002 CPUs utilized             
               238      context-switches                 #   18,126 /sec                      
                13      cpu-migrations                   #    0,990 /sec                      
            43.256      page-faults                      #    3,294 K/sec                     
    53.669.223.451      cycles                           #    4,087 GHz                       
   142.150.660.523      instructions                     #    2,65  insn per cycle            
     3.396.889.993      branches                         #  258,702 M/sec                     
        34.637.512      branch-misses                    #    1,02% of all branches           

      13,109785030 seconds time elapsed

      10,849108000 seconds user
       2,282617000 seconds sys
```

performance results on this branch:

```
 Performance counter stats for 'poetry run unblob -v -e /tmp/out -f -k -d 1 /home/quentin/Downloads/cvd-host_package.tar':

         13.170,81 msec task-clock                       #    1,000 CPUs utilized             
               628      context-switches                 #   47,681 /sec                      
                17      cpu-migrations                   #    1,291 /sec                      
            41.616      page-faults                      #    3,160 K/sec                     
    53.611.922.822      cycles                           #    4,071 GHz                       
   142.090.165.881      instructions                     #    2,65  insn per cycle            
     3.365.689.495      branches                         #  255,542 M/sec                     
        34.373.818      branch-misses                    #    1,02% of all branches           

      13,168817128 seconds time elapsed

      10,874326000 seconds user
       2,297380000 seconds sys
```

unblob/handlers/archive/tar.py
```diff
@@ -136,6 +136,36 @@ def calculate_chunk(self, file: File, start_offset: int) -> Optional[ValidChunk]
         header_size = snull(header.size)
         decode_int(header_size, 8)
 
+        def signed_sum(octets) -> int:
+            return sum(b if b < 128 else 256 - b for b in octets)
+
+        if header.chksum[6:8] not in (b"\x00 ", b" \x00"):
+            logger.error(
+                "Tar handler: invalid checksum format",
```
vlaci, line 144: Nitpick, sidetrack: we should actually bind the handler name in `_calculate_chunk`, so that all logs contain it from every handler.

Here in this PR I'd add `handler=self.NAME`, instead of this ad-hoc namespacing.

unblob/handlers/archive/tar.py
```diff
@@ -214,10 +214,10 @@ class TarUnixHandler(_TarHandler):
...
             #  (pattern would be too big)   char name[100]
             + _padded_field(r"[0-7]", 8)  # char mode[8]
-            + _padded_field(r"[0-9]", 8)  # char uid[8]
-            + _padded_field(r"[0-9]", 8)  # char gid[8]
-            + _padded_field(r"[0-9]", 12)  # char size[12]
-            + _padded_field(r"[0-9]", 12)  # char mtime[12]
+            + _padded_field(r"[0-7]", 8)  # char uid[8]
+            + _padded_field(r"[0-7]", 8)  # char gid[8]
+            + _padded_field(r"[0-7]", 12)  # char size[12]
+            + _padded_field(r"[0-7]", 12)  # char mtime[12]
```
vlaci, line 220: Nitpick: I don't think keeping this in a sepaarate commit adds any value.

unblob/handlers/archive/tar.py
```diff
@@ -146,3 +140,70 @@ def calculate_chunk(self, file: File, start_offset: int) -> Optional[ValidChunk]
...
+    # to get to the start of the file.
+    PATTERN_MATCH_OFFSET = -MAGIC_OFFSET
+
+
+def _re_protect(regexp: str):
+    return f"({regexp})"
+
+
+def _re_alternatives(regexps):
+    return _re_protect("|".join(_re_protect(regexp) for regexp in regexps))
```
vlaci, line 191: From the name `_re_protect` I have not associated to capture groups, but quoting or something similar. I'd rename this to `_re_group` instead.

Also, why do you need grouping the inner alternatives?

If I am not mistaken the output of this is:
`((re1)|(re2)|...|(reN))` Why doesn't `(re1|re2|...|reN)` suffice?

vlaci:
I have some questions and a general improvement idea, otherwise I am happy with the changes here.

vlaci:
> Tested with samples with tar content that I have here (~3GB) and did not observe regressions. This patch has very limited performance impact. These are the stats from running executions with a 2.4GB real world tar sample:
> 
> [...]

Another interesting sample could be some "garbage" data, filled with octal bytes which does not get picked up by other handlers, but triggers match on v7 all the time and then aborts on the checksum calculations.

unblob/handlers/archive/tar.py
e3krisztian: agreed, this commit was kept separate for review, it is meant to be either dropped or squashed.

qkaiser:
> > Tested with samples with tar content that I have here (~3GB) and did not observe regressions. This patch has very limited performance impact. These are the stats from running executions with a 2.4GB real world tar sample:
> > [...]
> 
> Another interesting sample could be some "garbage" data, filled with octal bytes which does not get picked up by other handlers, but triggers match on v7 all the time and then aborts on the checksum calculations.

Should be easy to build samples like this, will check when I have some time.

unblob/handlers/archive/tar.py
e3krisztian: Protecting the regular expression from appending/prefixing is the intention here, grouping is an implementation detail, so the name is right.

The internal regexps might not need to be grouped, but I am always unsure about precedences of regexp constructs (especially `|`), this is the reason. It should not hurt either, as hyperscan supports only non-capture groups.
So while it could be unneeded, I'd prefer it this way.

vlaci: This should be made obvious somehow. I have not figured out this intent. At least an explanatory comment would be welcome.

e3krisztian: Renamed the function and added a doc-string explaining this, thanks for the question and suggestion, this had become better because of those.

qkaiser:
I created two 75MB samples with a tar header followed by random bytes from `/dev/urandom`:
- first sample has an invalid checksum
- second sample has a valid checksum

#### Invalid checksum (654-tar-handler)

```
Performance counter stats for 'poetry run unblob -v -e /tmp/out -f -d 1 -k /tmp/tar_invalid.bin' (10 runs):

        139.327,15 msec task-clock:u                     #    9,269 CPUs utilized               ( +-  9,58% )
                 0      context-switches:u               #    0,000 /sec                      
                 0      cpu-migrations:u                 #    0,000 /sec                      
           470.571      page-faults:u                    #    6,207 K/sec                       ( +-  9,57% )
   361.014.537.121      cycles:u                         #    4,762 GHz                         ( +-  9,58% )
   414.359.307.124      instructions:u                   #    2,10  insn per cycle              ( +-  9,57% )
    73.635.561.486      branches:u                       #  971,237 M/sec                       ( +-  9,57% )
     1.556.295.394      branch-misses:u                  #    3,84% of all branches             ( +-  9,60% )

            15,031 +- 0,244 seconds time elapsed  ( +-  1,62% )
```

#### Invalid checksum (main)

```
Performance counter stats for 'poetry run unblob -v -e /tmp/out -f -d 1 -k /tmp/tar_invalid.bin' (10 runs):

         59.823,84 msec task-clock:u                     #   10,031 CPUs utilized               ( +-  9,55% )
                 0      context-switches:u               #    0,000 /sec                      
                 0      cpu-migrations:u                 #    0,000 /sec                      
           457.648      page-faults:u                    #   13,883 K/sec                       ( +-  9,57% )
   226.910.378.653      cycles:u                         #    6,884 GHz                         ( +-  9,57% )
   394.810.857.655      instructions:u                   #    3,16  insn per cycle              ( +-  9,58% )
    70.929.772.008      branches:u                       #    2,152 G/sec                       ( +-  9,58% )
     1.485.695.361      branch-misses:u                  #    3,81% of all branches             ( +-  9,58% )

             5,964 +- 0,110 seconds time elapsed  ( +-  1,84% )
```

#### valid checksum (654-tar-handler):

```
Performance counter stats for 'poetry run unblob -v -e /tmp/out -f -d 1 -k /tmp/tar_randombody.bin' (10 runs):

         64.096,36 msec task-clock:u                     #   10,034 CPUs utilized               ( +-  9,65% )
                 0      context-switches:u               #    0,000 /sec                      
                 0      cpu-migrations:u                 #    0,000 /sec                      
           470.814      page-faults:u                    #   13,464 K/sec                       ( +-  9,57% )
   236.536.039.198      cycles:u                         #    6,765 GHz                         ( +-  9,60% )
   413.304.132.300      instructions:u                   #    3,18  insn per cycle              ( +-  9,57% )
    73.431.199.316      branches:u                       #    2,100 G/sec                       ( +-  9,57% )
     1.513.254.304      branch-misses:u                  #    3,75% of all branches             ( +-  9,58% )

            6,3877 +- 0,0941 seconds time elapsed  ( +-  1,47% )
```

#### valid checksum (main):

```
Performance counter stats for 'poetry run unblob -v -e /tmp/out -f -d 1 -k /tmp/tar_randombody.bin' (10 runs):

         57.794,17 msec task-clock:u                     #   10,035 CPUs utilized               ( +-  9,51% )
                 0      context-switches:u               #    0,000 /sec                      
                 0      cpu-migrations:u                 #    0,000 /sec                      
           454.125      page-faults:u                    #   14,146 K/sec                       ( +-  9,57% )
   224.755.433.959      cycles:u                         #    7,001 GHz                         ( +-  9,56% )
   394.615.926.845      instructions:u                   #    3,18  insn per cycle              ( +-  9,57% )
    70.878.218.909      branches:u                       #    2,208 G/sec                       ( +-  9,57% )
     1.485.271.655      branch-misses:u                  #    3,81% of all branches             ( +-  9,57% )

            5,7590 +- 0,0461 seconds time elapsed  ( +-  0,80% )
```

### Summary

| branch | sample | time elapsed |
|------------|-------------|-------------------|
| 654-tar-handler | invalid checksum | 15,031 +- 0,244 seconds time elapsed  ( +-  1,62% ) |
| main | invalid checksum | 5,964 +- 0,110 seconds time elapsed  ( +-  1,84% ) |
| 654-tar-handler | valid checksum | 6,3877 +- 0,0941 seconds time elapsed  ( +-  1,47% ) |
| main | valid checksum | 5,7590 +- 0,0461 seconds time elapsed  ( +-  0,80% ) |

So a clear performance impact (152%) on samples with invalid checksum. Smaller performance impact (10%) for samples with valid checksum.

qkaiser:
To me it's okay to keep the code this way. Performance impact under specific conditions in exchange of support for a new format is a good trade off.

Asking approval from initial reviewer now.

vlaci (approved):

PR #657, opened 2023-09-26
fix(handlers): fix extfs handler for samples with spaces in their names.

qkaiser:
The Command extractor definition for our extfs handler was defined so that a sample with space in its name would make debugfs go haywire since the path was not properly quoted.

- `debugfs -R 'rdump / a_sample_with_no_space.ext4_extract' sample.ext4` -> works
- `debugfs -R 'rdump / a sample with space.ext4_extract' sample.ext4` -> does not work

Fixed by quoting the output dir :)

### DIFF
--- a/unblob/handlers/filesystem/extfs.py
+++ b/unblob/handlers/filesystem/extfs.py
@@ -64,7 +64,7 @@ class EXTHandler(StructHandler):
 
     PATTERN_MATCH_OFFSET = -MAGIC_OFFSET
 
-    EXTRACTOR = Command("debugfs", "-R", "rdump / {outdir}", "{inpath}")
+    EXTRACTOR = Command("debugfs", "-R", 'rdump / "{outdir}"', "{inpath}")
 
     def valid_header(self, header) -> bool:
         if header.s_state not in [0x1, 0x2]:


vlaci:
And what happens if there is a `'` in the name? :)
Maybe `shlex.quote` or something should be used here?

qkaiser:
> And what happens if there is a `'` in the name? :) Maybe `shlex.quote` or something should be used here?

We're calling [shlex.join](https://docs.python.org/3/library/shlex.html#shlex.join) on the extract command (see https://github.com/onekey-sec/unblob/blob/main/unblob/extractors/command.py#L38) so the presence of `'` is not problematic.

This is what it looks like when commands are executed with this fix on a sample with a single quote in its name:

```
sasquatch -le -no-exit-code -f -d '/tmp/out/SAMPLE Software '"'"'B V1.2.1.bin_extract/0-127647744.squashfs_v4_le_extract' '/tmp/out/SAMPLE Software '"'"'B V1.2.1.bin_extract/0-127647744.squashfs_v4_le'
```

```
debugfs -R 'rdump / "/tmp/out/SAMPLE Software '"'"'B V1.2.1.bin_extract/0-127647744.squashfs_v4_le_extract/root.ext4_extract"' '/tmp/out/SAMPLE Software '"'"'B V1.2.1.bin_extract/0-127647744.squashfs_v4_le_extract/root.ext4'
```

qkaiser:
Just to add some more info: the problem when calling `debugfs` is that `shlex.join` operates on and quotes `'rdump / {outdir}'` as a whole and not `outdir` itself since we're providing `debugfs` a "subcommand" . That's why we have to quote it ourselves.

The result stays secure since our command goes through `shlex.join`, which returns a value that is:

> [...] shell-escaped to protect against injection vulnerabilities

vlaci (approved):

PR #658, opened 2023-09-27
chore(deps): do not use SemVer dependency specification for CalVer dependencies

orosam:
The caret notation is meant for SemVer dependencies, it is not appropriate for [Calendar Versioning](https://calver.org/).

Specifying the lower bound only should be sufficient, as both [attrs](https://github.com/python-attrs/attrs/blob/main/.github/SECURITY.md) and [structlog](https://github.com/hynek/structlog/blob/main/.github/SECURITY.md) has strict backwards-compatibility policies.


e3krisztian (approved):
Nice one! These were too strict version pins.

qkaiser:
TIL

orosam:
Thanks for the quick review & merge.

PR #677, opened 2023-11-12
build and publish docker image for AARCH64 platforms

### LINKED ISSUE #417: Build and publish docker images for ARM platforms
I've seen a few reports of people wanting to give unblob a try on AARCH64 targets (mostly Apple M1 or M2) but couldn't because our Docker images are only built for x86 / x86-64.

We should investigate whether or not our current Dockerfile can be built into ARM images. If it works, we should edit the Github CI workflow to publish images for x86, ARMv6, ARMv7, and AARCH64.

Interesting resources on the subject:

- https://docs.docker.com/desktop/multi-arch/
- https://jitsu.com/blog/multi-platform-docker-builds
- https://itnext.io/building-multi-cpu-architecture-docker-images-for-arm-and-x86-3-building-in-github-action-ci-a382feab5af9

qkaiser:
Resolve #417

### DIFF
--- a/.github/workflows/build-publish-image.yml
+++ b/.github/workflows/build-publish-image.yml
@@ -9,14 +9,20 @@ on:
       - '**'
 
 env:
-  DOCKER_IMAGE: ghcr.io/onekey-sec/unblob:latest
+  DOCKER_IMAGE: ghcr.io/onekey-sec/unblob
 
 jobs:
-  build-and-push-image:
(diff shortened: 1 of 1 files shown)


qkaiser:
You can find more information about multi-platforms build at https://docs.docker.com/build/ci/github-actions/multi-platform/

vlaci (approved):
It is astonishing how complicated is to use these actions :)

PR #679, opened 2023-11-14
plugins: `is_dir` was always true

vlaci:


qkaiser:
How did you pick it up ? Ruff ?

qkaiser (approved):

vlaci:
> How did you pick it up ? Ruff ?

:smile: Experimented with `mypy`:

```
unblob/plugins.py:58: error: Function "is_dir" could always be true in boolean context  [truthy-function]
```

PR #681, opened 2023-11-17
add unlink function to FileSystem API.

qkaiser:


tests/test_file_utils.py
```diff
@@ -609,6 +609,32 @@ def test_open_outside_sandbox(self, sandbox: FileSystem):
...
+        assert sandbox.problems == []
+
+    def test_unlink_outside_sandbox(self, sandbox: FileSystem):
+        path = Path("../file")
+
+        with sandbox.open(path) as f:
+            f.write(b"content")
+
+        sandbox.unlink(path)
+        assert not (sandbox.root / path).exists()
```
e3krisztian, line 635: I think the logic in this test should be different: we should have an existing file outside the extraction directory, and we should check, that after calling `unlink` through a relative link, the file will not be removed.

unblob/file_utils.py
```diff
@@ -618,3 +618,10 @@ def open(  # noqa: A003
 
         self._ensure_parent_dir(safe_path)
         return safe_path.open(mode)
+
+    def unlink(self, path):
+        """Delete file within extraction path."""
+        logger.debug("unlink binary file", file_path=path)
```
e3krisztian, line 624:
```suggestion
        logger.debug("unlink file", file_path=path)
```

e3krisztian (approved):
Looks good! :)

PR #684, opened 2023-12-01
improve tar handler to support sparse archives.

### LINKED ISSUE #683: Missing support for sparse tar archives
Unblob does not support sparse tar archive because of the way the `tarfile` library handles those archives.

Our assumption is that the end of the archive corresponds to the last member offset + last member size, like so:

```python
try:
    tf = tarfile.TarFile(mode="r", fileobj=file)
except tarfile.TarError:
    return -1

last_member = None

try:
    for member in tf:
        last_member = member
except (tarfile.TarError, SeekError):
    # recover what's already been parsed
    pass

if last_member is None:
    return -1

last_file_size = round_up(last_member.size, BLOCK_SIZE)
end_of_last_tar_entry = last_member.offset_data + last_file_size
```

This assumption does not hold true with sparse archives since `tarfile` set the `TarInfo` size to the original file size within  `_proc_sparse`:

```python
def _proc_sparse(self, tarfile):
    """Process a GNU sparse header plus extra headers.
    """
    # We already collected some sparse structures in frombuf().
    structs, isextended, origsize = self._sparse_structs
    del self._sparse_structs

    # Collect sparse structures from extended header blocks.
    while isextended:
        buf = tarfile.fileobj.read(BLOCKSIZE)
        pos = 0
        for i in range(21):
            try:
                offset = nti(buf[pos:pos + 12])
                numbytes = nti(buf[pos + 12:pos + 24])
            except ValueError:
                break
            if offset and numbytes:
                structs.append((offset, numbytes))
            pos += 24
        isextended = bool(buf[504])
    self.sparse = structs

    self.offset_data = tarfile.fileobj.tell()
    tarfile.offset = self.offset_data + self._block(self.size)
    self.size = origsize # <---- here
    return self
```

We therefore need a way to know the sparse'd size if the archive is a sparse archive, not the original file size.

We can check if any member of the archive is sparse by using the `issparse()` function. I suppose supporting sparsed archive would mean browsing through every entry in the archive, not just the last member.

qkaiser:
Resolve #683 

TODO:
- [x] add sparse samples to integration tests

### DIFF
--- a/tests/integration/archive/tar/__input__/sparse.tar
+++ b/tests/integration/archive/tar/__input__/sparse.tar
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:217eb921a851ca38185b7585a554160f2f88f76c3b123c1930468a2cc4d80e46
+size 10240
--- a/tests/integration/archive/tar/__output__/sparse.tar_extract/dummy1
+++ b/tests/integration/archive/tar/__output__/sparse.tar_extract/dummy1
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:16fa66a7dc98d93f2a4c5d20baf5177f59c4c37fc62face65690c11c15fe6ff9
+size 51200
--- a/tests/integration/archive/tar/__output__/sparse.tar_extract/dummy2
+++ b/tests/integration/archive/tar/__output__/sparse.tar_extract/dummy2
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:16fa66a7dc98d93f2a4c5d20baf5177f59c4c37fc62face65690c11c15fe6ff9
+size 51200
--- a/tests/integration/archive/tar/__output__/sparse.tar_extract/file1
+++ b/tests/integration/archive/tar/__output__/sparse.tar_extract/file1
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:ed68b430af3630fde9b598ea74340a7be809408e493f3adff7aadbd6cfeef931
+size 45
--- a/tests/integration/archive/tar/__output__/sparse.tar_extract/file2
+++ b/tests/integration/archive/tar/__output__/sparse.tar_extract/file2
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:ed68b430af3630fde9b598ea74340a7be809408e493f3adff7aadbd6cfeef931
+size 45
--- a/unblob/handlers/archive/tar.py
+++ b/unblob/handlers/archive/tar.py
@@ -59,8 +59,7 @@ def _get_end_of_last_tar_entry(file) -> int:
     if last_member is None:
         return -1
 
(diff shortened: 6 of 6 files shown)


martonilles:
just wondering, but isn't it a bug in python tar implementation? is there an upstream bug for this?

I guess the difference comes from where tar takes the size:

```
    @classmethod
    def frombuf(cls, buf, encoding, errors):
        """Construct a TarInfo object from a 512 byte bytes object.
        """
        if len(buf) == 0:
            raise EmptyHeaderError("empty header")
        if len(buf) != BLOCKSIZE:
            raise TruncatedHeaderError("truncated header")
        if buf.count(NUL) == BLOCKSIZE:
            raise EOFHeaderError("end of file header")

        chksum = nti(buf[148:156])
        if chksum not in calc_chksums(buf):
            raise InvalidHeaderError("bad checksum")

        obj = cls()
        obj.name = nts(buf[0:100], encoding, errors)
        obj.mode = nti(buf[100:108])
        obj.uid = nti(buf[108:116])
        obj.gid = nti(buf[116:124])
        obj.size = nti(buf[124:136])
        obj.mtime = nti(buf[136:148])
        obj.chksum = chksum
        obj.type = buf[156:157]
        obj.linkname = nts(buf[157:257], encoding, errors)
        obj.uname = nts(buf[265:297], encoding, errors)
        obj.gname = nts(buf[297:329], encoding, errors)
        obj.devmajor = nti(buf[329:337])
        obj.devminor = nti(buf[337:345])
        prefix = nts(buf[345:500], encoding, errors)

        # Old V7 tar format represents a directory as a regular
        # file with a trailing slash.
        if obj.type == AREGTYPE and obj.name.endswith("/"):
            obj.type = DIRTYPE

        # The old GNU sparse format occupies some of the unused
        # space in the buffer for up to 4 sparse structures.
        # Save them for later processing in _proc_sparse().
        if obj.type == GNUTYPE_SPARSE:
            pos = 386
            structs = []
            for i in range(4):
                try:
                    offset = nti(buf[pos:pos + 12])
                    numbytes = nti(buf[pos + 12:pos + 24])
                except ValueError:
                    break
                structs.append((offset, numbytes))
                pos += 24
            isextended = bool(buf[482])
            origsize = nti(buf[483:495])
            obj._sparse_structs = (structs, isextended, origsize)

        # Remove redundant slashes from directories.
        if obj.isdir():
            obj.name = obj.name.rstrip("/")

        # Reconstruct a ustar longname.
        if prefix and obj.type not in GNU_TYPES:
            obj.name = prefix + "/" + obj.name
        return obj
```

qkaiser:
@martonilles to me it's not an upstream bug, it's just that the developers chose to consider that a `TarInfo` size attribute represents the original file size that was put in the archive, **not** the sparse'd version of it that's actually in the archive. Probably so they can display nice files listing of the content à la 7z, even with sparse archives.

They do this in `_proc_sparse`, which is called from `_proc_member`. In the function, once the sparse structs are parsed they simply do:

```python
self.size = origsize
```

That's why unblob fails on sparse tar archives since the handler finds an end offset past the end of the archive, given that the original file size is bigger than the sparse'd size.

qkaiser:
Found a better solution with @e3krisztian

e3krisztian (approved):

PR #689, opened 2023-12-16
feat(handler): add multi-part gzip handler.

qkaiser:
It's possible to create multi-part gzip with `split`, which will create multiple gzip compressed files with a 'aa', 'ab', 'ac', .. suffix.

We match on `.gz.aa` in a directory, get all the files with same name but different suffix, order them and feed them to 7z.

This is very close to what we were already doing with multi-part 7zip archives.


unblob/handlers/compression/gzip.py
```diff
@@ -124,3 +147,29 @@ def calculate_chunk(self, file: File, start_offset: int) -> Optional[ValidChunk]
             start_offset=start_offset,
             end_offset=file.tell(),
         )
+
+
+class MultiVolumeGzipHandler(DirectoryHandler):
+    NAME = "multi-gzip"
+    EXTRACTOR = MultiGZIPExtractor()
+
+    PATTERN = Glob("*.gz.aa")
```
e3krisztian, line 156: I was wondering if there is a reason restricting support only to `split`'s default numbering, which excludes both numeric and hex numbering (`*.gz.00`), as well as explicitly set, non-default suffix length (`*.gz.000`) , and additional suffixes (`*.gz.aa-suffix`) .

Could this pattern be generalized to
```suggestion
    PATTERN = Glob("*.gz.*")
```
?

qkaiser: Since the `paths` list would be sorted, we can definitely use that pattern and not restrict on a specific numbering.

unblob/handlers/compression/gzip.py
```diff
@@ -124,3 +147,33 @@ def calculate_chunk(self, file: File, start_offset: int) -> Optional[ValidChunk]
...
         )
+
+
+class MultiVolumeGzipHandler(DirectoryHandler):
+    NAME = "multi-gzip"
+    EXTRACTOR = MultiGZIPExtractor()
+
+    PATTERN = Glob("*.gz.*")
+
+    def calculate_multifile(self, file: Path) -> Optional[MultiFile]:
```
e3krisztian, line 158: With the generalization of `PATTERN` to cover all `split` output, `calculate_multifile()` would generate many potential `MultiFile` root files, including `file.gz.01`, `file.gz.ab`, ... resulting in potential conflicts between them (given some other conditions explained below).

Therefore this methods needs to be more picky, and a `MultiFile` should be recognized *only* if the input `file` `Path` is the first of the gzip parts - and has a gzip header.

For parts generated with `split`ting up a `gzip`ped file, it is usually not a problem, because we are filtering out all the other roots, and they are unlikely to have a valid gzip header.

However, https://www.gnu.org/software/gzip/manual/html_node/Advanced-usage.html mentions, that

> Multiple compressed files can be concatenated.

Which means, that parts can be generated with `split`ting the input first, then `gzip`ping the resulting parts (potentially in parallel for time-optimized builds), which would result in slightly different sized part files, but which all have valid gzip headers, these parts would all generate conflicting `MultiFile`-s with the current code.

----

We might also probably be prepared to exclude non-part files from `MultiFile.paths` like `file.gz.md5`, `file.gz.sha1`, `file.gz.txt`, `file.gz.README`, etc. here.

qkaiser: Agree. I don't think we need to make a special case for files with `.md5` or `.README` extensions since the input file provided to `7z` is always the first one in the sorted list and `7z` has its own logic to process split content.

I'll create integration test files to cover the different scenarios you mention here and work from there.

qkaiser: Done !

qkaiser:
Copying the exchange I had with @e3krisztian outside Github:

@e3krisztian:
> I saw the gzip changes, let's talk about it tomorrow.
> The problem is with these output files:
`tests/integration/compression/gzip/__output__/multi-volume-split-then-gzip.tar_extract/one.txt.gz.aa_extract/one.txt.aa`
`tests/integration/compression/gzip/__output__/multi-volume-split-then-gzip.tar_extract/one.txt.gz.aa_extract/one.txt.ab`
...
they should not be like this - it is not a multifile this way.
The expected output would be `tests/integration/compression/gzip/__output__/multi-volume-split-then-gzip.tar_extract/one.txt.gz_extract/one.txt`

@qkaiser:

> I looked at it and here's the problem:
> - we provide the first path to 7z through MultiFileCommand, this is because 7z is smart enough to detect multi-volume archives or compressed streams
> - if we wanted to decompress "split then compressed" multi-volumes, we can provide a wildcard to 7z by adapting MultiFileCommand so that it runs something like this: 7z x -p -y 'mv.7z*' -o/tmp/out
> - the problem with this wildcard approach is that it blocks 7z from working with legit multi-volume archives, because it will consider each matching file independently if we provide a wildcard

> So we can't have both. I think having split then compressed multi-volume is an edge case and should only be handled when we observe one. Until then it will still be handled, but each file decompressed independently, without causing issues.

e3krisztian (approved):

Issue #624, opened 2023-06-26
add option to only "scan" the binary like binwalk would do for example


dummys:
Hi,

I would like to skip using binwalk and use your tools rather, but I'm missing one of the binwalk feature, ability to simply scan the whole binary and print all know files inside.
I checked the command line of unblob and found one that should do the same, the "-s" option, but I get directory of extraction still created. Is is possible to avoid creating and thus clobbering the working directory with "*_extract" folder ?

qkaiser:
Hi @dummys ! That's something we talked about in the early days of unblob development but never got to implement.

I'll keep the issue open to track it, but please note that it's not on our priority list.

aserper:
+1

qkaiser:
@dummys @aserper please have a look at #692 and do not hesitate to comment. I'd like to know what are your expectations regarding the console output in "scan" mode.

> Is is possible to avoid creating and thus clobbering the working directory with "*_extract" folder ?

This is handled by 6aabbc7c692078b9e7c5b2a64b4df239b500a1c2 right now. This will probably land in main in 2024.

PR #692, opened 2023-12-24
Improve unblob "skip-extraction" mode of operation

qkaiser:
The following improvements have been made in order to let user simply "scan" a file:
1. do not create the root level extraction dir when executed with the `--skip-extraction` option.
2. display a summary of identified chunks when executed with the `--skip-extraction` option.

This is related to #624.

Right now the information displayed to the user is fairly limited:

image

But the introduction of metadata reporting in chunks (see https://github.com/onekey-sec/unblob/pull/557) would allow us to display detailed information in the "description" column, such as endianness, version, compression, block size, creation size, etc.

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -333,3 +333,37 @@ def test_skip_extension(
     result = runner.invoke(unblob.cli.cli, params)
     assert extracted_files_count == len(list(tmp_path.rglob("*")))
     assert result.exit_code == 0
+
+
+@pytest.mark.parametrize(
+    "args, skip_extraction, fail_message",
+    [
+        ([], False, "Should *NOT* have skipped extraction"),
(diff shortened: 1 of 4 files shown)


aserper:
Makes sense and looks good! 👍

e3krisztian (commented):
I thinks this is fine as is (only code reordering proposed), however there is no test for `skip_extraction` at all, so future changes could inadvertently break the newly added functionality.

qkaiser:
@e3krisztian I applied your recommendations, introduced tests for both the cli and processing parts of `skip_extraction` option and fixed a typo in the CLI (`skip_extraction` should have been `skip-extraction` from the beginning).

e3krisztian (approved):
Cool! :)

Issue #600, opened 2023-06-07
Do not extract Rust `.rlib` files


vlaci:
Similarly to `.a` files, we should keep `.rlib` files as is:
https://rustc-dev-guide.rust-lang.org/backend/libs-and-metadata.html#rlib

vlaci:
```console
❯ file target/debug/deps/libapprox-422485dfe0ddcbef.rlib
target/debug/deps/libapprox-422485dfe0ddcbef.rlib: current ar archive
❯ xdg-mime query filetype target/debug/deps/libapprox-422485dfe0ddcbef.rlib
application/x-archive```

Issue #647, opened 2023-08-30
Extend --skip-magic option for better matching


7homasSutter:
It would be nice to have a more fine grained option to select what file-types should be skipped. The current `--skip-magic` option allows to select specific skipped types but it does not allow to extend the default list of magics. It would be nice to have the possibility to extend the default list instead of overwriting it. Moreover, using the magic prefix is confusing in scenarios where the magic bytes of a file is the same (for example, zip and apk files).

**Is your feature request related to a problem? Please describe.**
Issue #262 has described the exact same problem and the solution was to integrate the `--skip-magic` option which is a nice way to solve the problem but I suggest extending the feature.

The problem that I'm facing is, that I want to extract an Android image but don't want to extract certain file types (e.g., .apk, .ttf). However, using the `--skip-magic` option isn't really user friendly because I would need to define a list of `--skip-magic` parameters for every filetype to exclude as well as for the default list of magic defined by unblob.

Consider the following example: Let's assume I have a .zip file that contains only three files (an .xlsx, an .apk., and a .jar file). We then want to extract this file with unblob but don't want to extract the .apk, .jar, and .xlsx file. As user, I would expect that it is sufficient to add `--skip-magic "APK" --skip-magic "JAR"` to skip these file extensions. However, adding these two parameters doesn't match apk and jar files as it seems. Moreover, when setting a `--skip-magic` parameter it overwrites the default list of skip-magic in unblob. Thus, unblob extracts all the files including the .xlsx, which is not what we want. 
```
docker run --platform=linux/amd64 --rm --pull always -v /Volumes/ExtremeSSD/test/output/:/data/output -v /Volumes/ExtremeSSD/test/input/:/data/input ghcr.io/onekey-sec/unblob:latest --skip-magic "APK" --skip-magic "JAR" "/data/input/Test.apk.zip"
```
To overcome this problem, we have to figure out the correct magic prefix for apk and jar files. So we figured out that adding the magic "Android" and "Java" would actually skip the apk and jar files. However, we would need to add for all defaults another `--skip-magic` parameter to prevent overwriting the default magic list and skip as well the .xlsx file. The list of defaults to skip is quiet long. Thus, we would need to add around 20 --skip-magic parameters to skip all the defaults.
```
[...]

PR #695, opened 2023-12-24
Add ability to skip files based on extension.

qkaiser:
As indicated in #647, the ability to skip files from being processed based on their extension seems to be the right approach.

### DIFF
--- a/unblob/processing.py
+++ b/unblob/processing.py
@@ -77,6 +77,7 @@ DEFAULT_SKIP_MAGIC = (
     "Windows Embedded CE binary image",
     "Intel serial flash for PCH ROM",
 )
+DEFAULT_SKIP_EXTENSION = (".rlib",)
 
 
 @attr.define(kw_only=True)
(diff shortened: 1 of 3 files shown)


qkaiser:
Should also resolve #600 without having a blanket filtering of `ar archives`.

unblob/processing.py
```diff
@@ -292,9 +297,16 @@ def _process_task(self, result: TaskResult, task: Task):
         should_skip_file = any(
             magic.startswith(pattern) for pattern in self._config.skip_magic
         )
+        should_skip_file |= any(
+            task.path.suffix == suffix for suffix in self._config.skip_extension
+        )
```
e3krisztian, line 302:
```suggestion
        should_skip_file |= task.path.suffix in self._config.skip_extension
```

unblob/processing.py
```diff
@@ -87,6 +88,10 @@ class ExtractionConfig:
     entropy_plot: bool = False
     max_depth: int = DEFAULT_DEPTH
     skip_magic: Iterable[str] = DEFAULT_SKIP_MAGIC
+    skip_extension: Iterable[str] = attr.field(
+        default=DEFAULT_SKIP_EXTENSION,
+        converter=lambda value: [extension for extension in value if extension],
+    )
```
e3krisztian, line 94: Why are empty extensions removed?
```suggestion
    skip_extension: Iterable[str] = DEFAULT_SKIP_EXTENSION
```

qkaiser: Providing an empty extension would mean skipping all files without an extension, does it make sense from a usage perspective ?

qkaiser: did some thinking, I'm not against it. Applying.

PR #698, opened 2024-01-04
chore(doc): fix documentation inconsistencies and missing dependencies

nyuware:
`sudo` was missing for both `dpkg` commands and I also added the option `--all-extras` in the development guide for `poetry install`


qkaiser:
It's explicit that users need to use sudo, they can copy paste the script and run it with elevated privileges. We don't use extras in unblob-core. Closing.

PR #706, opened 2024-01-19
fix(handler): fix UBI PEB size calculation.

qkaiser:
Computing the most frequent interval can only work if we return each and every interval value observed in a list. Returning it in a set hides the repeating interval patterns.

The `get_intervals` function now _actually_ acts like numpy.diff.

### DIFF
--- a/tests/test_iter_utils.py
+++ b/tests/test_iter_utils.py
@@ -6,11 +6,11 @@ from unblob.iter_utils import get_intervals
 @pytest.mark.parametrize(
     "values, expected",
     [
-        ([], set()),
-        ([0, 0], {0}),
-        ([0, 0, 0], {0}),
-        ([1, 2, 3], {1}),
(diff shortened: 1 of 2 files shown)


qkaiser:
To provide more context, here's how `get_intervals` is used by the UBI handler:

```python
def _guess_peb_size(self, file: File) -> int:
        # Since we don't know the PEB size, we need to guess it. At the moment we just find the
        # most common interval between every erase block header we find in the image. This _might_
        # cause an issue if we had a blob containing multiple UBI images, with different PEB sizes.
        all_ubi_eraseblock_offsets = list(iterate_patterns(file, self._UBI_EC_HEADER))

        offset_intervals = get_intervals(all_ubi_eraseblock_offsets)
        if not offset_intervals:
            raise InvalidInputFormat

        return statistics.mode(offset_intervals)
```

e3krisztian (approved):
Nice catch! :)

PR #722, opened 2024-01-25
add search_chunks fuzzing with atheris

qkaiser:
oss-fuzz recommends project maintainers to place their fuzzing harness into their project repository.

This is an optimized fuzzing harness to test search_chunks by creating an mmap'ed File from the bytes received by the fuzzer and submitting it to search_chunks.

Other fuzzing harness can be added to the fuzzing directory later on and they will be executed by oss-fuzz.

This is what I plan to submit to oss-fuzz: https://github.com/google/oss-fuzz/compare/master...qkaiser:oss-fuzz:unblob

### DIFF
--- a/fuzzing/search_chunks_fuzzer.py
+++ b/fuzzing/search_chunks_fuzzer.py
@@ -1,0 +1,64 @@
+#!/usr/bin/env python3
+import logging
+import sys
+from pathlib import Path
+
+import atheris
+import structlog
+
+
+def set_unblob_log_level(level=logging.CRITICAL):
+    logger = logging.getLogger("unblob")
+
+    def logger_factory():
+        return logger
+
+    structlog.configure(logger_factory=logger_factory)
+    logger.setLevel(level)
+
+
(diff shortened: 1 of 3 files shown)


fuzzing/search_chunks_fuzzer.py
```diff
@@ -0,0 +1,63 @@
...
+        task = Task(
+            path=Path("/dev/shm/nonexistent"), depth=0, blob_id=""  # noqa: S108
+        )
+        result = TaskResult(task)
+        search_chunks(file, len(data), config.handlers, result)
+
+
+set_unblob_log_level()
+atheris.Setup(sys.argv, test_search_chunks)
+atheris.Fuzz()
```
e3krisztian, line 63: In python scripts usually there is a guard to prevent execution on import like below, but in this case it might not be that useful:
```suggestion
if __name__ == "__main__":
    set_unblob_log_level()
    atheris.Setup(sys.argv, test_search_chunks)
    atheris.Fuzz()
```

PR #740, opened 2024-01-30
Fix docker build

qkaiser:
The digests files needs to have unique names that do not contain forward slashes. Both digests files needs to be pulled and merged.

Fixed according to https://docs.docker.com/build/ci/github-actions/multi-platform/


qkaiser:
@nyuware you can see it fully working (finally) at https://github.com/onekey-sec/unblob/actions/runs/7709608959/job/21011336732

nyuware (approved):

PR #755, opened 2024-02-11
fix(models): adapt chunk's contains() definition.

### LINKED ISSUE #753: Error `chunk has higher start_offset than end_offset` with end_offset=0
**Describe the bug**
Unblob reports an error `Chunk has higher start_offset than end_offset` with an end_offset value of 0 for at least 58 DLINK firmware images and fails to extract files.

**To Reproduce**
1. Download a sample firmware to trigger the bug with: `wget https://legacyfiles.us.dlink.com/DCS-5009L/REVA/FIRMWARE/DCS-5009L_REVA_FIRMWARE_1.00.B1.zip`
2. Launch unblob with command `unblob -v DCS-5009L_REVA_FIRMWARE_1.00.B1.zip`
3. See error:

```
2024-02-10 23:27.13 [error    ] Unknown error happened         pid=2767672
  File "/unblob/unblob/processing.py", line 246, in process_task
  File "/unblob/unblob/processing.py", line 310, in _process_task
    _FileTask(self._config, task, stat_report.size, result).process()
  File "/unblob/unblob/processing.py", line 522, in process
    unknown_chunks = calculate_unknown_chunks(outer_chunks, self.size)
  File "/unblob/unblob/processing.py", line 688, in calculate_unknown_chunks
    unknown_chunk = UnknownChunk(
  File "<attrs generated init unblob.models.UnknownChunk>", line 9, in __init__
    self.__attrs_post_init__()
  File "/unblob/unblob/models.py", line 69, in __attrs_post_init__
    raise InvalidInputFormat(
unblob.file_utils.InvalidInputFormat: Chunk has higher start_offset than end_offset: 0xac5f2-0x0
```

**Expected behavior**
A standard linux-based filesystem should be extracted. If `binwalk` is run on this image it finds a CPIO archive within LZMA compressed data that contains ~700 files.

**Environment information:**
 - OS: Ubuntu 22.04
 - Docker

```
Linux b4935d734f27 6.2.2 #3 SMP PREEMPT_DYNAMIC Wed Mar  8 12:03:22 EST 2023 x86_64 x86_64 x86_64 GNU/Linux

DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=22.04
DISTRIB_CODENAME=jammy
DISTRIB_DESCRIPTION="Ubuntu 22.04.3 LTS"

The following executables found installed, which are needed by unblob:
    7z                          ✓
    debugfs                     ✓
    jefferson                   ✓
    lz4                         ✓
    lziprecover                 ✓
    lzop                        ✓
    sasquatch                   ✓
    sasquatch-v4be              ✓
    simg2img                    ✓
    ubireader_extract_files     ✓
    ubireader_extract_images    ✓
    unar                        ✓
    zstd                        ✓
``` 

**Additional context**
I found this bug while doing some large-scale evaluations of filesystems produced by binwalk and unblob using [fw2tar](https://github.com/AndrewFasano/fw2tar).

qkaiser:
We were considering that a chunk A contains a chunk B if chunk B start after chunk A and ends before or on the same offset than chunk A.

This caused issues with compressed DMG images where two handlers would rightfully identify two chunks:
- a bzip chunk starting at offset 0 and ending before the DMG plist
- a DMG chunk starting at offset 0 and ending after the DMG footer

We therefore adapted our definition of contains() where a contained chunk can start at the same offset but ends before, or start after the containing chunk but can end on the same offset.

Resolve #753

### DIFF
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -18,30 +18,47 @@ class TestChunk:
     @pytest.mark.parametrize(
         "chunk1, chunk2, result",
         [
-            (
+            pytest.param(
                 Chunk(start_offset=0, end_offset=10),
                 Chunk(start_offset=1, end_offset=2),
                 True,
+                id="starts-after-ends-before",
             ),
-            (
+            pytest.param(
                 Chunk(start_offset=0, end_offset=10),
                 Chunk(start_offset=11, end_offset=12),
                 False,
+                id="starts-after-ends-after",
             ),
-            (
+            pytest.param(
                 Chunk(start_offset=0, end_offset=10),
                 Chunk(start_offset=15, end_offset=20),
                 False,
+                id="starts-after-ends-after",
             ),
-            (
+            pytest.param(
                 Chunk(start_offset=1, end_offset=2),
                 Chunk(start_offset=3, end_offset=5),
                 False,
+                id="starts-after-ends-after",
             ),
-            (
+            pytest.param(
                 Chunk(start_offset=0, end_offset=10),
                 Chunk(start_offset=1, end_offset=10),
                 True,
+                id="starts-after-ends-same",
+            ),
+            pytest.param(
+                Chunk(start_offset=0, end_offset=10),
+                Chunk(start_offset=0, end_offset=9),
+                True,
+                id="starts-same-ends-before",
+            ),
+            pytest.param(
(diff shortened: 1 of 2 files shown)


e3krisztian (approved):
Nice catch!

PR #756, opened 2024-02-11
fix(fs): duplicate entries handling in FileSystem API.

### LINKED ISSUE #754: FileExistsError during extraction with Netgear firmware
**Describe the bug**
During extraction of at least 29 NETGEAR firmware images, unblob may try creating the same output file twice triggering an exception. As a result, some files that should be extracted are not.

**To Reproduce**
1. Download a sample firmware to trigger the bug with: `wget https://www.downloads.netgear.com/files/GDC/M4100/M4100-V10.0.2.20.zip`
2. Launch unblob with command `unblob -v M4100-V10.0.2.20.zip`
3. See error:

```
2024-02-10 23:39.13 [error    ] Unknown error happened while extracting chunk pid=2295991
  File "/unblob/unblob/processing.py", line 607, in _extract_chunk
    if result := chunk.extract(inpath, extract_dir):
  File "/unblob/unblob/models.py", line 115, in extract
    return self.handler.extract(inpath, outdir)
  File "/unblob/unblob/models.py", line 452, in extract
    return self.EXTRACTOR.extract(inpath, outdir)
  File "/unblob/unblob/handlers/archive/cpio.py", line 384, in extract
    parser.dump_entries(fs)
  File "/unblob/unblob/handlers/archive/cpio.py", line 215, in dump_entries
    fs.carve(entry.path, self.file, entry.start_offset, entry.size, mode=entry.mode & 0o777)
  File "/unblob/unblob/file_utils.py", line 511, in carve
    carve(safe_path, file, start_offset, size, mode=mode)
  File "/unblob/unblob/file_utils.py", line 294, in carve
    with carve_path.open("xb") as f:
  File "/usr/lib/python3.10/pathlib.py", line 1119, in open
    return self._accessor.open(self, mode, buffering, encoding, errors,
FileExistsError: [Errno 17] File exists: '/tmp/tmp1151iav4/M4100_V10.0.2.20.zip_extract/m4100v10.0.2.20.stk_extract/1201148-2097967.lzma_extract/lzma.uncompressed_extract/lib/libthread_db-1.0.so'
```

**Expected behavior**
This error should not be raised, instead additional files should be extracted. I made a simpel change in file_utils.py's carve method (see below) to return early if the target file already exists and with this change an extra 75 files are created in `[extract_dir]/m4100v10.0.2.20.stk_extract/1201148-2097967.lzma_extract/lzma.uncompressed_extract`. I doubt this is the right fix, but it shows that this bug prevents some files from being extracted.

**Environment information:**
 - OS: Ubuntu 22.04
 - Docker

```
Linux b4935d734f27 6.2.2 #3 SMP PREEMPT_DYNAMIC Wed Mar  8 12:03:22 EST 2023 x86_64 x86_64 x86_64 GNU/Linux

DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=22.04
DISTRIB_CODENAME=jammy
DISTRIB_DESCRIPTION="Ubuntu 22.04.3 LTS"

The following executables found installed, which are needed by unblob:
    7z                          ✓
    debugfs                     ✓
    jefferson                   ✓
    lz4                         ✓
    lziprecover                 ✓
    lzop                        ✓
    sasquatch                   ✓
    sasquatch-v4be              ✓
    simg2img                    ✓
    ubireader_extract_files     ✓
    ubireader_extract_images    ✓
    unar                        ✓
    zstd                        ✓
``` 

**Additional context**
I found this bug while doing some large-scale evaluations of filesystems produced by binwalk and unblob using [fw2tar](https://github.com/AndrewFasano/fw2tar).

My (likely-incorrect) patch that results in additional files being created:
```
diff --git a/unblob/file_utils.py b/unblob/file_utils.py
index 21e887b..3db4b98 100644
--- a/unblob/file_utils.py
+++ b/unblob/file_utils.py
@@ -291,6 +291,9 @@ def carve(carve_path: Path, file: File, start_offset: int, size: int):
     """Extract part of a file."""
     carve_path.parent.mkdir(parents=True, exist_ok=True)

+    if carve_path.exists():
+        print(f"Warning not replacing {carve_path}")
+        return
     with carve_path.open("xb") as f:
         for data in iterate_file(file, start_offset, size):
             f.write(data)
```

After fixing this, I got another error along the same vein in file_utils which I patched with:
```
diff --git a/unblob/file_utils.py b/unblob/file_utils.py
index 21e887b..3db4b98 100644
--- a/unblob/file_utils.py
+++ b/unblob/file_utils.py

@@ -579,7 +582,8 @@ class FileSystem:
         if safe_link:
             dst = safe_link.dst.absolute_path
             self._ensure_parent_dir(dst)
-            dst.symlink_to(src)
+            if not dst.exists():
+                dst.symlink_to(src)

     def create_hardlink(self, src: Path, dst: Path):
         """Create a new hardlink dst to the existing file src."""
```

qkaiser:
Early draft to take care of archives and filesystems holding duplicate entries.

We need to decide on a strategy when that happens:
- overwrite the previous entry with the new one ? (overwrite)
- do not overwrite the previous entry with the new one ? (skip)
- write the file with a new suffix (e.g. `.1` à la wget) ? (rewrite)

Do we want to compare hashes before creating a renamed entry or not ?

What happens if we have triplicate or quadruplicate entries ?

Resolve #754

### DIFF
--- a/tests/integration/archive/cpio/cpio_binary/__input__/duplicate_entries.cpio
+++ b/tests/integration/archive/cpio/cpio_binary/__input__/duplicate_entries.cpio
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:fab950a545cff6286c14c8736f532a1204c217de2e1a32f66e83655c7196a7ab
+size 1024
--- a/tests/integration/archive/cpio/cpio_binary/__output__/duplicate_entries.cpio_extract/colors/blue
+++ b/tests/integration/archive/cpio/cpio_binary/__output__/duplicate_entries.cpio_extract/colors/blue
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:a0bee6616b5e5eae6799cb4525a884a82e7161614f11122bbdf4383b2ac05998
+size 5
--- a/tests/integration/archive/cpio/cpio_binary/__output__/duplicate_entries.cpio_extract/colors/blue.txt
+++ b/tests/integration/archive/cpio/cpio_binary/__output__/duplicate_entries.cpio_extract/colors/blue.txt
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:b946a10c9133c481c9c95234cb8d61a5c063388721f22021ff30363e1bfaa3b0
+size 13
--- a/tests/integration/archive/cpio/cpio_binary/__output__/duplicate_entries.cpio_extract/colors/orange.txt
+++ b/tests/integration/archive/cpio/cpio_binary/__output__/duplicate_entries.cpio_extract/colors/orange.txt
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:6afd32ccf0462f3660d9fe54596a8c519d9478619c9798bc56bec0ba3ac1a7a3
+size 15
--- a/tests/integration/archive/cpio/cpio_binary/__output__/duplicate_entries.cpio_extract/colors/purple
+++ b/tests/integration/archive/cpio/cpio_binary/__output__/duplicate_entries.cpio_extract/colors/purple
@@ -1,0 +1,1 @@
+purple.txt
--- a/tests/integration/archive/cpio/cpio_binary/__output__/duplicate_entries.cpio_extract/colors/purple.txt
+++ b/tests/integration/archive/cpio/cpio_binary/__output__/duplicate_entries.cpio_extract/colors/purple.txt
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:596157e5858a0c647af6a4f55945b9c09ed70dc3f5d0e3db248ea5d5136e8ad3
+size 7
--- a/unblob/file_utils.py
+++ b/unblob/file_utils.py
@@ -628,7 +628,7 @@ class FileSystem:
 
     def unlink(self, path):
         """Delete file within extraction path."""
-        logger.debug("unlink file", file_path=path)
+        logger.debug("unlink file", file_path=path, _verbosity=3)
         safe_path = self._get_extraction_path(path, "unlink")
 
         safe_path.unlink(missing_ok=True)
--- a/unblob/handlers/archive/cpio.py
+++ b/unblob/handlers/archive/cpio.py
@@ -210,6 +210,11 @@ class CPIOParserBase:
             if entry.path.name in ("", "."):
                 continue
 
+            # There are cases where CPIO archives have duplicated entries
+            # We then unlink the files to overwrite them and avoid an error.
+            if not stat.S_ISDIR(entry.mode):
+                fs.unlink(entry.path)
+
             if stat.S_ISREG(entry.mode):
                 fs.carve(entry.path, self.file, entry.start_offset, entry.size)
             elif stat.S_ISDIR(entry.mode):


AndrewFasano:
Thanks for drafting a fix for this issue. If the files are the same, skipping the syntax error and keeping everything else the same seems like a good solution.

When the files are different, it's probably useful for most users to be able to see all versions. I don't love the `.1` syntax though, just because that makes the output directory less like how the files would look like in a standard system that was using the filesystem.

I don't have a deep understanding of these file formats. When two files exist with the same name what would actually happen in a typical system that used the format? Would only one version of the file show up and one be hidden? If that's the case (and there's a way to tell which one would show up), perhaps selecting the normal file for the standard unblob output location would be good and then adding the second (or third, etc) into a secondary output location?

AndrewFasano:
I tried this out and found there's still an issue with some firmware images: if `fs_path.absolute_path.exists()` and `fs_path.absolute_path` is a directory, the unlink call fails:
```
2024-02-12 02:15.14 [error    ] Unknown error happened while extracting chunk pid=2445276
  File "/unblob/unblob/processing.py", line 607, in _extract_chunk
    if result := chunk.extract(inpath, extract_dir):
  File "/unblob/unblob/models.py", line 118, in extract
    return self.handler.extract(inpath, outdir)
  File "/unblob/unblob/models.py", line 455, in extract
    return self.EXTRACTOR.extract(inpath, outdir)
  File "/unblob/unblob/handlers/archive/cpio.py", line 384, in extract
    parser.dump_entries(fs)
  File "/unblob/unblob/handlers/archive/cpio.py", line 217, in dump_entries
    fs.mkdir(
  File "/unblob/unblob/file_utils.py", line 524, in mkdir
    safe_path = self._get_extraction_path(path, "mkdir")
  File "/unblob/unblob/file_utils.py", line 485, in _get_extraction_path
    fs_path.absolute_path.unlink()
  File "/usr/lib/python3.10/pathlib.py", line 1206, in unlink
    self._accessor.unlink(self)
IsADirectoryError: [Errno 21] Is a directory: '/tmp/tmpg86lywif/DIR-655_REVB_FIRMWARE_2.00.ZIP_extract/dir655_revB_FW_200NA/DIR655B1_FW200NAb33.bin_extract/64-6212776.gzip_extract/gzip.uncompressed_extract/2906880-7696567.gzip_extract/gzip.uncompressed_extract/dev'
```

qkaiser:
> I don't have a deep understanding of these file formats. When two files exist with the same name what would actually happen in a typical system that used the format? Would only one version of the file show up and one be hidden? 

Yes, that's usually what happens. Silently.

> If that's the case (and there's a way to tell which one would show up), perhaps selecting the normal file for the standard unblob output location would be good and then adding the second (or third, etc) into a secondary output location?

Let's start with an implementation that simply overwrites and see what comes up from the team during review.

AndrewFasano:
Sounds good to me.

In terms of this patch and the error I found above, I initially tried fixing it by making it support deleting the directory when it already existed. That turned out to be a bad idea as it would end up deleting a directory full of lots of extracted files only to replace it with an empty directory. Things seem better if it overwrites duplicate files and ignores existing directories:

```diff
diff --git a/unblob/file_utils.py b/unblob/file_utils.py
index c96da27..1109280 100644
--- a/unblob/file_utils.py
+++ b/unblob/file_utils.py
@@ -477,13 +477,21 @@ class FileSystem:
         fs_path = self._fs_path(path)

         if fs_path.absolute_path.exists():
-            report = ExtractionProblem(
-                path=str(fs_path.relative_path),
-                problem=f"Attempting to create a file that already exists through {path_use_description}",
-                resolution="Overwrite.",
-            )
-            fs_path.absolute_path.unlink()
-            self.record_problem(report)
+            if fs_path.absolute_path.is_file():
+                report = ExtractionProblem(
+                    path=str(fs_path.relative_path),
+                    problem=f"Attempting to create a file that already exists through {path_use_description}",
+                    resolution="Overwrite.",
+                )
+                fs_path.absolute_path.unlink()
+                self.record_problem(report)
+
+            elif fs_path.absolute_path.is_dir():
+                report = ExtractionProblem(
+                    path=str(fs_path.relative_path),
+                    problem=f"Attempting to create a directory that already exists through {path_use_description}",
+                    resolution="Ignore",
+                )

         if not fs_path.is_safe:
             report = PathTraversalProblem(
```

AndrewFasano:
And there's another issue in here too - in `_get_checked_link` the `src` is the link target, while the `dst` is the file that's being created ([discussion here](https://github.com/onekey-sec/unblob/pull/763#issuecomment-1942368890)). These names are very confusing. So the file already exists error should only be raised if the `dst` already exists.

```diff
diff --git a/unblob/file_utils.py b/unblob/file_utils.py
index 65d7ad6..4ee7f97 100644
--- a/unblob/file_utils.py
+++ b/unblob/file_utils.py
@@ -565,7 +565,7 @@ class FileSystem:
     def _get_checked_link(self, src: Path, dst: Path) -> Optional[_FSLink]:
         link = _FSLink(root=self.root, src=src, dst=dst)

-        if link.src.absolute_path.exists():
+        if link.dst.absolute_path.exists():
             self.record_problem(link.format_report("File already exists."))
             return None
         if not link.is_safe:
```

nyuware:
The code I pushed is the works of @qkaiser and @AndrewFasano. 
1. It overwrites files that already exists.
2. discard if a files we are trying to extract already exists. 

The issue is that we are tripping two tests: 
1. `test_open_no_path_traversal`, fails because of `assert sandbox.problems == []`, which is no longer true because we raise `Overwriting already existing file` on file overwrites
2. `test_open` fails because of 
`with sandbox.open(path, "rb+") as f: assert f.read() == bytes(100) + b"text"` because the file already exists.
1. I think is a non issue
2. do we accept that a file CAN BE overwritten, thus dropping that part of the test, or do we not overwrite ?

e3krisztian:
The `FileSystem` class (as its doc-string says) is intended to limit the output to a well known directory, and the class' responsibility should end there.

The original problem is a CPIO extractor problem.
I think the resolution of conflicting output files should be handled mostly in the CPIO extractor.

The `cpio` extractor is using the `FileSystem.carve()` method, which will die, if the output already exists.
Fortunately there is already a `FileSystem.unlink(path)` method, which will delete `path`, but will not complain if `path` does not exist.
So I would simply put 
```py
fs.unlink(entry.path)
```
before
```py
fs.carve(entry.path, self.file, entry.start_offset, entry.size)
```
in the `cpio` extractor.

I would also add a test file with duplicate files to cover the new overwriting behavior.

----

An alternative implementation could be to smart up the `carve()` function and method with an `overwrite_existing=False` default parameter, and change the `carve` call in `cpio` to have the extra `overwrite_existing=True`.

qkaiser:
@e3krisztian yes that's the solution we agreed on with @nyuware , tyvm for your input ! :)

nyuware:
According to  @e3krisztian's comment, we decided to `overwrite` the files if they already exists by doing `fs.unlink(_file_)` before it gets `carved` again and fails to do so.  

The 3 pushes were me failing to rebase the branch and forgetting the integration tests, apologies about that.

e3krisztian (approved):

Issue #761, opened 2024-02-12
Symlink Troubles


AndrewFasano:
I believe there are a handful of bugs in how unblob handles symlinks. I'm not too confident about my understanding of unblob internals so I wanted to talk through these instead of just proposing fixes.

I've encountered all these failures with this firmware image: https://legacyfiles.us.dlink.com/DCS-5009L/REVA/FIRMWARE/DCS-5009L_REVA_FIRMWARE_1.00.B1.zip. I'm comparing the unblob output to that generated by running binwalk.

Issue 1: CPIO extractor calls `create_symlink` with backwards arguments - the source of a symlink is the filename (entry.path) while the destination is where the link points to (I think). Without this fix all the binaries in this firmware that symlink to `/bin/busybox` are missing in the output archive (i.e., `/sbin/arp`) as `_get_checked_link` in file_utils.py sees that `/bin/busybox` already exists and skips (checking the destination, not the source, since the arguments are swapped).

```diff
diff --git a/unblob/handlers/archive/cpio.py b/unblob/handlers/archive/cpio.py
index eacb30a..fecb5dc 100644
--- a/unblob/handlers/archive/cpio.py
+++ b/unblob/handlers/archive/cpio.py
@@ -223,7 +223,7 @@ class CPIOParserBase:
                         self.file[entry.start_offset : entry.start_offset + entry.size]
                     ).decode("utf-8")
                 )
-                fs.create_symlink(src=link_path, dst=entry.path)
+                fs.create_symlink(src=entry.path, dst=link_path)
             elif (
                 stat.S_ISCHR(entry.mode)
                 or stat.S_ISBLK(entry.mode)
```

Issue 2: `create_symlink` creates symlinks backwards, they currently point from `dst` to `src`. The call to `_get_checked_link` has it right and correctly checks if the `src` argument already exists. But, the code to actually create the link would previously create a link from `dst -> src` instead of from `src -> dst`. This would raise a `FileExistsError` if Issue 1 is fixed with no other changes.

```diff
diff --git a/unblob/file_utils.py b/unblob/file_utils.py
index 167357d..c6b8677 100644
--- a/unblob/file_utils.py
+++ b/unblob/file_utils.py
@@ -593,12 +593,13 @@ class FileSystem:
             # but they are relocatable
             src = self._path_to_root(dst.parent) / chop_root(src)

-        safe_link = self._get_checked_link(src=dst.parent / src, dst=dst)
[...]

PR #763, opened 2024-02-12
Improve support for safe symlink extraction

AndrewFasano:
This PR aims to fix a few issues around symlink extraction discussed in #761. I don't think this PR is perfect, but I hope it's an improvement over the current state of things.

Now in #768 ~~954c1cd5a06bcdb52048fc23da44661c73c94f31 rewrites the logic to sanitize symlinks to be relative and kept within the extraction directory. This is done using the `os` module instead of `Pathlib` as Pathlib.resolve would fail if a symlink target was missing (which doesn't prevent us from safely converting it to a relative link). With this change I no longer see false positives around MaliciousSymlinks, instead symlinks are created safely within the extraction directory. If a relative symlink originally tried accessing a directory above its own root (i.e., `./bin/sh -> ../../../../../bin/bash`), we update the link so it remains within the extraction directory.~~

This may have just been an artifact of swapping src/dst? ~~bbe18c61eb4ee18c7c2acd0364dd84df446d5bf4 and 76c29fe745176a7fc4cb43e1ad5a267ac534a1a3 change how the `.dst` field of a symlink is calculated in file_utils and in _safe_tarfile - previously it was made by combining the extraction root with the symlink destination. This would lose critical information about the path of the symlink source. For example a symlink at `./sbin/shell -> ../bin/sh` is safely within the extraction directory while a symlink at `/shell -> ../bin/sh` is trying to go up too high.~~

Now in #770 ~~fc60755c1608822c26934c2e28df5055eebed5e7 fixes a bug where tarfile absolute symlinks would be improperly dropped which I observed with a system that had`/var/log -> /tmp`. 0e6e0265269392d2c27b2d954e18652db064a8cd fixes a bug with relative symlinks that I observed on a system that had `/var/tmp -> ../../tmp`~~

~~56617a332fb3da19a1092d934e95e3d1ffd43ee5 and 2ce66d6ed32e9448172beb82a20f9b406bcfd72b are trying to fix a mix up between symlink source and destination when calling `create_symlink`. I fixed the CPIO extractor but it looks like things may be [backwards in other extractors](https://github.com/onekey-sec/unblob/issues/761#issuecomment-1939201482) as well.~~

To test these changes, I created a CPIO archive with the following script:
```sh
#!/bin/bash

# Set up a temporary directory for our archive contents
WORKDIR=$(mktemp -d)
echo "Working directory: $WORKDIR"

# Create necessary directories and dummy file for symlink targets
mkdir -p "$WORKDIR/bin" "$WORKDIR/sbin" "$WORKDIR/usr/bin"
echo "This is BusyBox" > "$WORKDIR/bin/busybox"
echo "Dummy content" > "$WORKDIR/usr/bin/dummy"

# Descriptive names for symlink targets
# 1) Symlink in the same directory
ln -s busybox "$WORKDIR/bin/symlink_same_dir"

# 2) Symlink that points to a valid parent directory + file
ln -s ../bin/busybox "$WORKDIR/sbin/symlink_up_to_busybox"

# 3) Symlink with extra parent directories that would still be valid
ln -s ../../../bin/busybox "$WORKDIR/sbin/symlink_extra_up_to_busybox"

# 4) Absolute symlink to a file
ln -s /bin/busybox "$WORKDIR/bin/symlink_absolute"

# 5) Broken symlink (target does not exist)
ln -s non_existent_file "$WORKDIR/bin/symlink_broken"

# 6) Symlink pointing to another symlink (chained symlinks)
ln -s symlink_same_dir "$WORKDIR/bin/symlink_to_symlink"

# 7) Circular symlink (A -> B, B -> A)
ln -s symlink_circular_b "$WORKDIR/bin/symlink_circular_a"
ln -s symlink_circular_a "$WORKDIR/bin/symlink_circular_b"

# 8) Second symlink to busybox
ln -s busybox "$WORKDIR/bin/symlink_same_dir2"

# Navigate to the work directory
cd "$WORKDIR"

# Create the CPIO archive
find . | cpio -ov --format=newc > test_archive.cpio

# Move the archive to the current directory (assuming it's where the script is run)
mv test_archive.cpio "$OLDPWD"
cd "$OLDPWD"

# Clean up the working directory
rm -rf "$WORKDIR"
echo "Archive created: test_archive.cpio"
```

If I extract this with the head of unblob (d0f30869747957ccea6537a2fe6959f542e89a84) and run
`find ../test/test_archive.cpio_extract/ -type f,l -exec ls -al {} \;` I get:

```
-rw-r--r-- 1 root root 0 Feb 12 17:09 ../test/test_archive.cpio_extract/test_archive.cpio
lrwxrwxrwx 1 root root 17 Feb 12 17:09 ../test/test_archive.cpio_extract/bin/symlink_broken -> non_existent_file
-rw-r--r-- 1 root root 16 Feb 12 17:09 ../test/test_archive.cpio_extract/bin/busybox
lrwxrwxrwx 1 root root 7 Feb 12 17:09 ../test/test_archive.cpio_extract/bin/symlink_to_symlink -> busybox
lrwxrwxrwx 1 root root 7 Feb 12 17:09 ../test/test_archive.cpio_extract/bin/symlink_same_dir2 -> busybox
lrwxrwxrwx 1 root root 18 Feb 12 17:09 ../test/test_archive.cpio_extract/bin/symlink_circular_a -> symlink_circular_b
lrwxrwxrwx 1 root root 7 Feb 12 17:09 ../test/test_archive.cpio_extract/bin/symlink_same_dir -> busybox
lrwxrwxrwx 1 root root 7 Feb 12 17:09 ../test/test_archive.cpio_extract/bin/symlink_absolute -> busybox
-rw-r--r-- 1 root root 14 Feb 12 17:09 ../test/test_archive.cpio_extract/usr/bin/dummy
lrwxrwxrwx 1 root root 14 Feb 12 17:09 ../test/test_archive.cpio_extract/sbin/symlink_up_to_busybox -> ../bin/busybox
```

After applying the changes in this PR I get the following result with two additional (and expected) files extracted: `symlink_circular_b` and `symlink_extra_up_to_busybox`. All the files are still contained within the extraction directory.
```
-rw-r--r-- 1 root root 0 Feb 12 17:10 ../test/test_archive.cpio_extract/test_archive.cpio
lrwxrwxrwx 1 root root 18 Feb 12 17:10 ../test/test_archive.cpio_extract/bin/symlink_circular_b -> symlink_circular_a
lrwxrwxrwx 1 root root 17 Feb 12 17:10 ../test/test_archive.cpio_extract/bin/symlink_broken -> non_existent_file
-rw-r--r-- 1 root root 16 Feb 12 17:10 ../test/test_archive.cpio_extract/bin/busybox
lrwxrwxrwx 1 root root 16 Feb 12 17:10 ../test/test_archive.cpio_extract/bin/symlink_to_symlink -> symlink_same_dir
lrwxrwxrwx 1 root root 7 Feb 12 17:10 ../test/test_archive.cpio_extract/bin/symlink_same_dir2 -> busybox
lrwxrwxrwx 1 root root 18 Feb 12 17:10 ../test/test_archive.cpio_extract/bin/symlink_circular_a -> symlink_circular_b
lrwxrwxrwx 1 root root 7 Feb 12 17:10 ../test/test_archive.cpio_extract/bin/symlink_same_dir -> busybox
lrwxrwxrwx 1 root root 7 Feb 12 17:10 ../test/test_archive.cpio_extract/bin/symlink_absolute -> busybox
-rw-r--r-- 1 root root 14 Feb 12 17:10 ../test/test_archive.cpio_extract/usr/bin/dummy
lrwxrwxrwx 1 root root 14 Feb 12 17:10 ../test/test_archive.cpio_extract/sbin/symlink_extra_up_to_busybox -> ../bin/busybox
lrwxrwxrwx 1 root root 14 Feb 12 17:10 ../test/test_archive.cpio_extract/sbin/symlink_up_to_busybox -> ../bin/busybox
```

### DIFF
--- a/unblob/extractor.py
+++ b/unblob/extractor.py
@@ -47,68 +47,73 @@ def is_recursive_link(path: Path) -> bool:
     return False
 
 
-def fix_symlink(path: Path, outdir: Path, task_result: TaskResult) -> Path:
-    """Rewrites absolute symlinks to point within the extraction directory (outdir).
+def sanitize_symlink_target(base_dir, current_dir, target):
+    # Normalize all paths to their absolute forms
+    base_dir_abs = os.path.abspath(base_dir)
+    current_dir_abs = os.path.abspath(current_dir)
+    target_abs = os.path.abspath(os.path.join(current_dir, target)) \
+                    if not os.path.isabs(target) else os.path.abspath(target)
 
-    If it's not a relative symlink it is either removed it it attempts
-    to traverse outside of the extraction directory or rewritten to be
-    fully portable (no mention of the extraction directory in the link
-    value).
-    """
-    if is_recursive_link(path):
-        logger.error("Symlink loop identified, removing", path=path)
-        error_report = MaliciousSymlinkRemoved(
-            link=path.as_posix(), target=os.readlink(path)
-        )
-        task_result.add_report(error_report)
-        path.unlink()
-        return path
-
-    raw_target = os.readlink(path)
-    if not raw_target:
-        logger.error("Symlink with empty target, removing.")
-        path.unlink()
-        return path
-
-    target = Path(raw_target)
-
-    if target.is_absolute():
-        target = Path(target.as_posix().lstrip("/"))
+    # Check if the target is absolute and within the base_dir
+    if os.path.isabs(target):
+        if target_abs.startswith(base_dir_abs):
+            return os.path.relpath(target_abs, current_dir_abs)
+        else:
+            # Target is absolute but outside base_dir - we'll pretend base_dir is our root
+            # and adjust the target to be within base_dir
+            abs = base_dir + "/" + os.path.relpath(target_abs, os.path.commonpath([target_abs, base_dir_abs]))
+            # We want to return the relative path from current_dir to the adjusted target
+            return os.path.relpath(abs, current_dir_abs)
     else:
-        target = path.resolve()
+        # Target is relative
+        if target_abs.startswith(base_dir_abs):
+            # Target is relative and does not escape base_dir
+            return os.path.relpath(target_abs, current_dir_abs)
+        else:
+            # Target is relative and escapes base_dir
+            # Say we have foo/passwd -> ../../../etc/passwd with root at /host/test_archive
+            # from /host/test_archive/foo/passwd, we want to return ../etc/passwd which is the
+            # relative path from /host/test_archive/foo to /host/test_archive/etc/passwd
+            # without escaping /host/test_archive
 
-    safe = is_safe_path(outdir, target)
+            for drop_count in range(0, len(target.split('/'))):
+                # We drop '..'s from the target by prepending placeholder directories until we get something valid
+                abs = current_dir + "/" + "/".join(["foo"] * drop_count) + target
+                resolved = os.path.abspath(abs)
+                if resolved.startswith(base_dir_abs):
+                    break
+            else:
+                raise ValueError(f"Could not resolve symlink target {target} within base_dir {base_dir}")
 
-    if not safe:
-        logger.error("Path traversal attempt through symlink, removing", target=target)
-        error_report = MaliciousSymlinkRemoved(
-            link=path.as_posix(), target=target.as_posix()
-        )
-        task_result.add_report(error_report)
-        path.unlink()
-    else:
-        relative_target = os.path.relpath(outdir.joinpath(target), start=path.parent)
-        path.unlink()
-        path.symlink_to(relative_target)
-    return path
-
+            # We need to add the /placeholder to the relative path because we need
+            # to act like a file within base_dir is our root (as opposed to base_dir itself)
+            return os.path.relpath(resolved, base_dir_abs + '/placeholder')
 
 def fix_extracted_directory(outdir: Path, task_result: TaskResult):
     def _fix_extracted_directory(directory: Path):
         if not directory.exists():
             return
-        for path in (directory / p for p in os.listdir(directory)):
-            try:
-                fix_permission(path)
-                if path.is_symlink():
-                    fix_symlink(path, outdir, task_result)
-                    continue
-                if path.is_dir():
-                    _fix_extracted_directory(path)
-            except OSError as e:
-                if e.errno == errno.ENAMETOOLONG:
-                    continue
-                raise e from None
+
+        base_dir = os.path.abspath(outdir)
+        for root, dirs, files in os.walk(base_dir, topdown=True):
+            fix_permission(Path(root))
+            for name in dirs+files:
+                try:
+                    full_path = os.path.join(root, name)
+                    if os.path.islink(full_path):
+                        # Make symlinks relative and constrain them to the base_dir
+                        target = os.readlink(full_path)
+                        new_target = sanitize_symlink_target(base_dir, root, target)
+                        if new_target != target:
+                            os.remove(full_path)
+                            os.symlink(new_target, full_path)
+                            logger.info("Updated symlink", path=full_path, target=new_target)
+                        else:
+                            logger.debug("Symlink is already sanitized", path=full_path, target=new_target)
+                except OSError as e:
+                    if e.errno == errno.ENAMETOOLONG:
+                        continue
+                    raise e from None
 
     fix_permission(outdir)
     _fix_extracted_directory(outdir)
--- a/unblob/file_utils.py
+++ b/unblob/file_utils.py
@@ -568,11 +568,21 @@ class FileSystem:
         """Create a symlink dst with the link/content/target src."""
         logger.debug("creating symlink", file_path=dst, link_target=src, _verbosity=3)
 
-        if src.is_absolute():
-            # convert absolute paths to dst relative paths
-            # these would point to the same path if self.root would be the real root "/"
-            # but they are relocatable
-            src = self._path_to_root(dst.parent) / chop_root(src)
+        if dst.is_absolute():
+            # If the symlink destination is absolute, we need to make it relative to the root
+            # so it can be safely created in the extraction directory.
+            # If the resulting path points to outside of the extraction directory, we skip it.
+            dst = self.root / chop_root(dst)
+            if not is_safe_path(self.root, dst):
+                self.record_problem(
+                    LinkExtractionProblem(
+                        problem="Potential path traversal through symlink",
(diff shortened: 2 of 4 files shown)


qkaiser:
@AndrewFasano from within the unblob directory, you can do `pre-commit install` followed by `pre-commit run --all`. This will perform multiple tasks like type checking, code linting, import sorting, etc.

[pre-commit](https://pre-commit.com/) can be installed with `pip`.

It will modify the code and you can create fixups for existing commits depending on the part it touches.

AndrewFasano:
Ruff seems to be quite unhappy with my use of `os.path` and wants me to use `Pathlib` instead, but there are some critical differences between how those two modules behave around symlinks. Seems like it might take some time to go through all these linter errors and test what can actually be switched to Pathlib versus what needs to stay as `os.path`.

e3krisztian:
I think we should ste
I think there is some confusion on how `FileSystem.create_symlink()` should work.

e3krisztian:
```py
FileSystem.create_symlink(src, dst)
```

is supposed to be following the unix command line order and naming

```
cp src dst
mv src dst
ln src dst
ln -s src dst
```

Except these are named differently in the `man` page for `ln`, as `src` and `dst` is arguably confusing for links:

```
SYNOPSIS
       ln [OPTION]... [-T] TARGET LINK_NAME
```

So `src=TARGET` and `dst=LINK_NAME` for `create_symlink`, we definitely should have used the `man ln` names for arguments.

AndrewFasano:
Thanks for the info @e3krisztian. These names are definitely confusing and now that I cleaned up my changes, I can see that  you're right - the order was fine before. Sorry about that! There might still be value in revising the change from https://github.com/onekey-sec/unblob/commit/bbe18c61eb4ee18c7c2acd0364dd84df446d5bf4 to add the `root` to the base of the target path before checking if it's safe, but maybe even that's unnecessary? At one point I thought these changes were required for some symlinks to properly be created, but that might've been a mistake since it seems like they're largely a no-op other than switching the arg order.

I think the other changes are still relevant though!

e3krisztian:
@AndrewFasano this is a valued input and contribution for `unblob` and these changes/problems are taken seriously!

You might also be correct with some hidden argument swapping somewhere still.

The below symlinks are intentionally marked as problematic in the current code (`3)` being the case of path traversal), maybe `unblob` is too strict by throwing these out, and we should sanitize the link targets as you propose.

```sh
# 3) Symlink with extra parent directories that would still be valid
ln -s ../../../bin/busybox "$WORKDIR/sbin/symlink_extra_up_to_busybox"

# 7) Circular symlink (A -> B, B -> A)
ln -s symlink_circular_b "$WORKDIR/bin/symlink_circular_a"
ln -s symlink_circular_a "$WORKDIR/bin/symlink_circular_b"
```

e3krisztian:
Added a review comment for 954c1cd5a06bcdb52048fc23da44661c73c94f31 - `sanitize_symlink_target`, and extracted that commit into a separate branch, and PR: #768.

e3krisztian:
> https://github.com/onekey-sec/unblob/commit/bbe18c61eb4ee18c7c2acd0364dd84df446d5bf4 and https://github.com/onekey-sec/unblob/commit/76c29fe745176a7fc4cb43e1ad5a267ac534a1a3 change how the .dst field of a symlink is calculated in `file_utils` and in `_safe_tarfile` - previously it was made by combining the extraction root with the symlink destination. This would lose critical information about the path of the symlink source. For example a symlink at `./sbin/shell` -> `../bin/sh` is safely within the extraction directory while a symlink at `/shell` -> `../bin/sh` is trying to go up too high.

Cherry-picking these commit over `main` and running the tests 6 tests fail. These failures are due to `x / y / z` becoming absolute path, if `y` is absolute path. Maybe they were needed due to some other changes in this PR?

I have tried to reproduce the problems locally based on the description given with the tests below applied to `main` branch. These tests pass, so I do not understand what is the problem supposed to be solved by these commits.

```diff
diff --git a/tests/test_file_utils.py b/tests/test_file_utils.py
index 9a20da74..1fc94138 100644
--- a/tests/test_file_utils.py
+++ b/tests/test_file_utils.py
@@ -25,7 +25,7 @@ from unblob.file_utils import (
     round_down,
     round_up,
 )
-from unblob.report import PathTraversalProblem
+from unblob.report import LinkExtractionProblem, PathTraversalProblem


 @pytest.mark.parametrize(
@@ -503,6 +503,30 @@ class TestFileSystem:
         assert os.readlink(output_path) == "target file"
         assert sandbox.problems == []

+    def test_create_symlink_target_inside_sandbox(self, sandbox: FileSystem):
+        # ./sbin/shell -> ../bin/sh
+        sandbox.mkdir(Path("bin"))
+        sandbox.write_bytes(Path("bin/sh"), b"posix shell")
+        sandbox.mkdir(Path("sbin"))
+        sandbox.create_symlink(Path("../bin/sh"), Path("sbin/shell"))
+
+        output_path = sandbox.root / "sbin/shell"
+        assert output_path.read_bytes() == b"posix shell"
+        assert output_path.exists()
+        assert os.readlink(output_path) == "../bin/sh"
+        assert sandbox.problems == []
+
+    def test_create_symlink_target_outside_sandbox(self, sandbox: FileSystem):
+        # /shell -> ../bin/sh
+        sandbox.mkdir(Path("bin"))
+        sandbox.write_bytes(Path("bin/sh"), b"posix shell")
+        sandbox.create_symlink(Path("../bin/sh"), Path("/shell"))
+
+        assert any(p for p in sandbox.problems if isinstance(p, LinkExtractionProblem))
+        output_path = sandbox.root / "shell"
+        assert not output_path.exists()
+        assert not output_path.is_symlink()
+
     def test_create_symlink_absolute_paths(self, sandbox: FileSystem):
         sandbox.write_bytes(Path("target file"), b"test content")
         sandbox.create_symlink(Path("/target file"), Path("/symlink"))
```

AndrewFasano:
@e3krisztian I can try updating those commits to work on main and try digging up a filesystem where I was seeing the issue that was supposed to address. At a minimum bbe18c61eb4ee18c7c2acd0364dd84df446d5bf4 would need to be updated to swap src/dst since this PR has them swapped.

AndrewFasano:
I think this PR was a bit too big and skewed by my mix up with the src/dst variables. I updated the original comment to point to the new PRs for the relevant components. The changes worth discussing are in #768 and #770.

Issue #326, opened 2022-04-05
distinguish between _carve and _extract directory


e3krisztian:
just wondering, but do not we want to distinguish between _carve and _extract directory?

_Originally posted by @martonilles in https://github.com/IoT-Inspector/unblob/pull/323#discussion_r841237372_

qkaiser:
This has been implemented with https://github.com/onekey-sec/unblob/pull/1017, leaving it up to the user to configure custom suffix for `_carve` and `_extract` if they want.

PR #766, opened 2024-02-13
fix(processing): delete successfully processed files.

qkaiser:
Delete the source file after extraction if the extraction was successful and the chunk being extracted covers the whole file.

Technical notes: A call to `extract()` does not return anything if everything went well (no unhandled exception, no extraction errors). Under those conditions, we could delete the source file if the `--keep-extracted-chunks` option is not set.

Relying on `keep-extracted-chunks` to make this decision is not ok since we're not _technically_ operating on a chunk here. A chunk is something that was **carved** out of a file, here we're operating on a file that was **extracted** from an archive/filesystem.

Closely related to https://github.com/onekey-sec/unblob/issues/326, triggered by https://github.com/onekey-sec/unblob/discussions/687


unblob/processing.py
```diff
@@ -606,6 +606,9 @@ def _extract_chunk(self, file, chunk: ValidChunk):  # noqa: C901
         try:
             if result := chunk.extract(inpath, extract_dir):
                 extraction_reports.extend(result.reports)
+            elif chunk.is_whole_file and not self.config.keep_extracted_chunks:
+                logger.debug("Removing successfully processed file.", path=inpath)
+                inpath.unlink()
```
e3krisztian, line 611: It is pretty bad and misleading, that `chunk.extract` returns `ExtractResult | None`, it should always return a result, or raise an exception. I think `None` was kept for backward compatibility, when `ExtractResult` was introduced, but some/most `Handler`s are transformed to always return `ExtractResult`.

I think the above history led to this code, which translates to me "if there is a result, report problems, otherwise delete the file", so this code implies "result=something bad happened".

I think the intention here is if there is a report during extraction or an exception is raised, it is not a clean extraction, and should be kept, as it is potentially worth further looking into:

```suggestion
            result = chunk.extract(inpath, extract_dir)
            chunk_reports = result.reports if result else []
            extraction_reports.extend(chunk_reports)
            successfully_extracted = not chunk_reports

            if successfully_extracted and chunk.is_whole_file and not self.config.keep_extracted_chunks:
                logger.debug("Removing successfully processed file.", path=inpath)
                inpath.unlink()
```

martonilles:
I would default to not deleting source files, a we can easily loose information and structure while doing so (there might be a reference to a file , which is further extracted and hence deleted).

Even though not sure we would benefit much by deleting files compared to the complexity

qkaiser:
https://github.com/onekey-sec/unblob/discussions/687#discussioncomment-8494360

PR #768, opened 2024-02-14
Sanitize symlink target

e3krisztian:
Split off of #763 . There are still problems to solve here, see https://github.com/onekey-sec/unblob/commit/954c1cd5a06bcdb52048fc23da44661c73c94f31#commitcomment-138623089 but tests should run with the exception of 2 failures.


unblob/extractor.py
```diff
@@ -1,4 +1,6 @@
 """File extraction related functions."""
+# ruff: noqa
+
```
qkaiser, line 3: No. We should adapt the code to use `pathlib` instead of straight up ignoring ruff. I can take care of it if you want.

unblob/extractor.py
```diff
@@ -15,6 +17,14 @@
...
 DIR_PERMISSION_MASK = 0o775
 
+# re-exported, or unused symbols
+__all__ = [
+    "is_safe_path",
+    "MaliciousSymlinkRemoved",
+    "fix_symlink",
+    "is_recursive_link",
+]
+
```
qkaiser, line 27: Why is this required ?

unblob/extractor.py
```diff
@@ -47,6 +47,12 @@ def is_recursive_link(path: Path) -> bool:
     return False
 
 
+def fix_symlink(path: Path, outdir: Path, task_result: TaskResult) -> Path:
+    # This is a temporary function for existing unit tests in tests/test_extractor.py
+    fix_extracted_directory(outdir, task_result)
+    return path
+
+
```
qkaiser: I suppose it's a temporary fix and the commit message should say so.

qkaiser (commented):
I'll mark the PR as draft until ruff is back and we have a cleaner view on what are the logic changes being applied.

AndrewFasano:
I believe there is still a bug in b11fe469a8716e7c00e70f4daa99e291a0277422: when extracting in a directory within `/tmp/`, a symlink of `/var` pointing to `/tmp/var` will be sanitize to point to `var` creating a symlink loop. This seems to be caused by the use of `os.path.commonpath` finding a similarity between the extraction directory and the absolute symlink target.

e3krisztian:
@qkaiser I wanted to make this PR just to have a place to discuss. It was extracted from a larger PR, and wanted to see CI test results, which our code checks (ruff) prevented, to push through the commit I have made some hacky fixes.

This definitely needs a rewrite, so should have made it draft initially.

I do not plan working on this soon, especially as the `tar` fix covering these problems are merged.

unblob/extractor.py
e3krisztian: answered as main comment.
these changes are to be dropped.

unblob/extractor.py
e3krisztian: answered as main comment.
these changes are to be dropped.

unblob/extractor.py
e3krisztian: answered as main comment.
these changes are to be dropped.

Issue #769, opened 2024-02-14
Safe Tarfile incorrectly blocks Symlink Traversal Attempt


AndrewFasano:
(I'm attempting to break #763 and #761 into smaller issues with concrete bugs and examples)

Filesystem: https://files.dlink.com.au/Products/DCS-6517/REV_B/Firmware/Firmware_2.00.03/DCS-6517B1_FW_v2.00.03.zip

Binwalk extraction produces 173 symlinks that unblob drops. For example `sbin/init -> ../bin/busybox`. Unblob does not produce these due to incorrect symlink handling in `_safe_tarfile.py`

```
2024-02-14 18:19.09 [warning  ] Traversal attempt through link path. Skipped. path=sbin/init pid=58
```

Tested with head of main and #768, both produce the same behavior (since this bug is specific to the logic in safe_tarfile)

e3krisztian:
With #775 merged, all but one symlinks are extracted:

```sh
DCS-6517B1_FW_v2.00.03.zip_extract/DCS-6517B1_FW_v2.00.03.pkg_extract/44-17853067.gzip_extract/gzip.uncompressed_extract$ find -type l -ls | wc -l
347
```

PR #770, opened 2024-02-14
SafeTarfile: Update relative symlinks instead of dropping

AndrewFasano:
This is a partial fix for #769. With this change, 172 of the 173 missing symlinks are correctly extracted

If I rebase this on top of #768 then the final symlink is extracted correctly (`/mnt/mtd/ubifs_app -> ubifs_app`).

While this fixes the bug, I'm not positive my path handling is correct/secure for all possible symlinks.

### DIFF
--- a/unblob/handlers/archive/_safe_tarfile.py
+++ b/unblob/handlers/archive/_safe_tarfile.py
@@ -76,23 +76,93 @@ class SafeTarFile:
 
         # prevent traversal attempts through links
         if tarinfo.islnk() or tarinfo.issym():
-            if Path(tarinfo.linkname).is_absolute():
+            rel_target = Path(tarinfo.linkname)
+            if rel_target.is_absolute():
+                # If target is absolute, we'll rewrite to be relative to the symlink
+
+                # Strip leading '/' to make the path relative to extract directory
+                rel_target = (extract_root / rel_target.relative_to("/")).relative_to(
+                    extract_root
+                )
+
+                # Now we need to find the path from the symlink to the target
+                # If the symlink is in a directory like /foo/bar and the target is at /target
+                # we need to make rel_target ../target (relative to /foo) instead of just being /target (relative to /)
+
+                # Let's calculate depth of the symlink itself and the depth of the target
+                symlink_depth = len(tarinfo.name.split("/"))
+                target_depth = len(rel_target.parts)
+
+                # If the symlink is deeper than the target, we need to go up by the difference in depth
+                if symlink_depth > target_depth:
+                    # We need to go up by the difference in depth
+                    rel_target = (
+                        Path(
+                            "/".join(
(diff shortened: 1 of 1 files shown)


AndrewFasano:
Updated to hopefully fix the pyright check. I can't for the life of me figure out how to get pytest to work locally though.

unblob/handlers/archive/_safe_tarfile.py
```diff
@@ -76,23 +76,93 @@ def extract(self, tarinfo: tarfile.TarInfo, extract_root: Path):  # noqa: C901
...
+            tarinfo.linkname = rel_target.as_posix()
+            logger.info(
+                "Link target is relative", linkname=tarinfo.linkname, name=tarinfo.name
+            )
+
+            # Resolve the link target to an absolute path
+            resolved_path = (extract_root / tarinfo.name).parent / rel_target
+
+            # If the resolved path points outside of extract_root, we need to fix it!
+            if not is_safe_path(extract_root, resolved_path):
```
e3krisztian, line 124: This is the real fix, that saves the busybox symlinks.

I am not sure the fix attempt of violating symlinks is needed below.

e3krisztian (commented):
Have looked at and tried to understand what is going on here (as this is quite a big change) - I have made some comments.

However in the end I have ended up with an alternative PR #775 based on your work here (added you as co-author to the fix-commits if you do not mind), but with smaller changes.

@AndrewFasano may I ask you to look at #775 and validate the code/run some tests with fw-s to check if that indeed fixes the symlink extraction problems?

e3krisztian:
> If I rebase this on top of https://github.com/onekey-sec/unblob/pull/768 then the final symlink is extracted correctly (`/mnt/mtd/ubifs_app` -> `ubifs_app`).

Looking at the file
```sh
$ tar tvf DCS-6517B1_FW_v2.00.03.zip_extract/DCS-6517B1_FW_v2.00.03.pkg_extract/44-17853067.gzip_extract/gzip.uncompressed | rg ubifs_app
lrwxrwxrwx richard/richard       0 2017-05-24 10:52 mnt/ubifs_app -> mtd
lrwxrwxrwx richard/richard       0 2017-05-24 10:52 mnt/mtd/ubifs_app -> ubifs_app
lrwxrwxrwx richard/richard       0 2017-07-11 14:17 www/classes/panoviewer_lib.exe -> /mnt/ubifs_app/panoviewer_lib.exe
lrwxrwxrwx richard/richard       0 2017-07-11 14:17 www/classes/panoviewer.dmg -> /mnt/ubifs_app/panoviewer.dmg
```
It looks bad at the archive level.
I do not know the reason behind removing symlink loops, but it does look like a symlink loop.
Do you have any idea how such a symlink could be useful?

e3krisztian:
#775 got merged, which was inspired by and has the same goal as this PR.

@AndrewFasano thank you!

PR #775, opened 2024-02-15
Fix SafeTarFile symlinks extraction

e3krisztian:
Based on the work of @AndrewFasano in PR #770 this should fix most of the tar symlink extraction problems reported.

Related issue: #769

### DIFF
--- a/unblob/handlers/archive/_safe_tarfile.py
+++ b/unblob/handlers/archive/_safe_tarfile.py
@@ -83,10 +83,9 @@ class SafeTarFile:
                     "Converted to extraction relative path.",
                 )
                 tarinfo.linkname = f"./{tarinfo.linkname}"
-            if not is_safe_path(
-                basedir=extract_root,
-                path=extract_root / tarinfo.linkname,
-            ):
+
+            resolved_path = (extract_root / tarinfo.name).parent / tarinfo.linkname
+            if not is_safe_path(basedir=extract_root, path=resolved_path):
                 self.record_problem(
                     tarinfo,
                     "Traversal attempt through link path.",
--- a/unblob/handlers/archive/_safe_tarfile.py
+++ b/unblob/handlers/archive/_safe_tarfile.py
@@ -63,7 +63,7 @@ class SafeTarFile:
                 "Absolute path.",
                 "Converted to extraction relative path.",
             )
-            tarinfo.name = f"./{tarinfo.name}"
+            tarinfo.name = str(Path(tarinfo.name).relative_to("/"))
 
         # prevent traversal attempts through file name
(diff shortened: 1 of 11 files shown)


unblob/handlers/archive/_safe_tarfile.py
```diff
@@ -82,7 +82,18 @@ def extract(self, tarinfo: tarfile.TarInfo, extract_root: Path):  # noqa: C901
...
                     "Converted to extraction relative path.",
                 )
-                tarinfo.linkname = f"./{tarinfo.linkname}"
+
+                def calculate_linkname():
+                    root = extract_root.resolve()
+                    path = (extract_root / tarinfo.name).resolve()
+                    common_path = Path(os.path.commonpath([root, path]))
+                    # normally root == common_path
+                    # if it is not, the output will be bad
```
qkaiser, line 91: We could add an assert here to make sure it is and be noisy if it's not, which is indicative of something wrong with `is_safe_path`.

```python
assert root == common_path
```

qkaiser (commented):
Small comment otherwise LGTM. Nice work on tests.

unblob/handlers/archive/_safe_tarfile.py
e3krisztian: Yes, this made me thinking of adding an assertion as well.

However an assert would blow up extraction, instead it is explicitly reported now.

A fixup commit is added for review, it needs to be squashed before merge (if accepted).
Converting to draft to prevent accidental merge, but it is supposed to be complete and waiting for review.

e3krisztian:
Rebased for linear history.

unblob/handlers/archive/_safe_tarfile.py
```diff
@@ -77,23 +77,33 @@ def extract(self, tarinfo: tarfile.TarInfo, extract_root: Path):  # noqa: C901
...
                 def calculate_linkname():
                     root = extract_root.resolve()
                     path = (extract_root / tarinfo.name).resolve()
-                    common_path = Path(os.path.commonpath([root, path]))
-                    # normally root == common_path
-                    # if it is not, the output will be bad
-                    depth = max(0, len(path.parts) - len(common_path.parts) - 1)
+
+                    if path.parts[: len(root.parts)] != root.parts:
+                        return None
```
qkaiser, line 86: This would be easier to understand I think:

```python
if common_path != root:
    return None
```

But maybe I'm missing something.

qkaiser (approved):
As agreed, squash before merge.

unblob/handlers/archive/_safe_tarfile.py
e3krisztian: yeah, except `common_path` is no more with the last change.

PR #787, opened 2024-02-25
fix(gzip): empty files are invalid and should not be parsed

### LINKED ISSUE #786: mmap error during multi file calculation for multi-gzip
When extracting this file http://ixp.mikrotik.co.id/download/archive/6.18/all_packages-mipsbe-6.18.zip I get the following error:

```
2024-02-25 03:31.49 [warning  ] Unhandled Exception during multi file calculation handler=multi-gzip path=all_packages-mipsbe-6.18.zip_extract/system-6.18-mipsbe.npk_extract/4096-5779456.squashfs_v4_le_extract/home/web/webfig/index.html.gz.nocache pid=2582003 severity=<Severity.ERROR: 'ERROR'>
  File "/unblob/unblob/processing.py", line 377, in _calculate_multifile
    return dir_handler.calculate_multifile(path)
  File "/unblob/unblob/handlers/compression/gzip.py", line 179, in calculate_multifile
    if self.is_valid_gzip(file):
  File "/unblob/unblob/handlers/compression/gzip.py", line 159, in is_valid_gzip
    with File.from_path(path) as f:
  File "/unblob/unblob/file_utils.py", line 58, in from_path
    m = cls(base_file.fileno(), 0, access=access)
ValueError: cannot mmap an empty file
```

AndrewFasano:
Passing an empty file into `File.from_path` leads to an error that an empty file cannot be mmaped.

This could also be implemented in the from_path method, but I thought the `is_valid` method made since because an empty file isn't a valid gzip.

Fixes #786

### DIFF
--- a/unblob/file_utils.py
+++ b/unblob/file_utils.py
@@ -40,11 +40,17 @@ class SeekError(ValueError):
     """Specific ValueError for File.seek."""
 
 
+class InvalidInputFormat(Exception):
+    pass
+
+
 class File(mmap.mmap):
     access: int
 
     @classmethod
     def from_bytes(cls, content: bytes):
+        if not content:
+            raise InvalidInputFormat("Can't create File from empty bytes.")
         m = cls(-1, len(content))
         m.write(content)
         m.seek(0)
@@ -55,7 +61,10 @@ class File(mmap.mmap):
     def from_path(cls, path: Path, access=mmap.ACCESS_READ):
         mode = "r+b" if access == mmap.ACCESS_WRITE else "rb"
         with path.open(mode) as base_file:
-            m = cls(base_file.fileno(), 0, access=access)
+            try:
+                m = cls(base_file.fileno(), 0, access=access)
+            except ValueError as exc:
+                raise InvalidInputFormat from exc
             m.access = access
             return m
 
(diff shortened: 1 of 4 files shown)


qkaiser:
Thanks @AndrewFasano ! I took the liberty to work on your branch to also cover the sevenzip multi-volume handler and add some checks in the `File` class + testing.

This is something that came up when I worked on fuzzing (see https://github.com/onekey-sec/unblob/blob/main/fuzzing/search_chunks_fuzzer.py#L50) but I left it on the side.

Will request a review from the team.

unblob/file_utils.py
```diff
@@ -53,6 +59,8 @@ def from_bytes(cls, content: bytes):
 
     @classmethod
     def from_path(cls, path: Path, access=mmap.ACCESS_READ):
+        if path.stat().st_size == 0:
+            raise InvalidInputFormat("Can't create File from empty file.")
         mode = "r+b" if access == mmap.ACCESS_WRITE else "rb"
         with path.open(mode) as base_file:
             m = cls(base_file.fileno(), 0, access=access)
```
orosam, line 66: I think we can just catch the `ValueError("cannot mmap an empty file")` from mmap here:

```suggestion
        mode = "r+b" if access == mmap.ACCESS_WRITE else "rb"
        with path.open(mode) as base_file:
            try:
                m = cls(base_file.fileno(), 0, access=access)
            except ValueError as exc:
                raise InvalidInputFormat from exc
```

orosam (commented):
I have one small suggestion, otherwise it looks good to me.

orosam (approved):

PR #793, opened 2024-02-29
Fix `is_valid_gzip` on empty file

e3krisztian:
#787 was fixing the issue #786 not just at the location of the problem, but this reached the `File` class as well, modifying it to raise an unblob specific error, effectively aborting processing in a "well known manner".
I think this was wrong, and `File` worked as expected and was a rather generic class, and should remain so.
The fix also had changes in gzip and 7z handlers, ignoring empty files from the collected files, but that change was neither supported by tests nor explained - I think that was a redundant fix for the problem.

The root of the issue is not in the `File` class, but rather that the raised `ValueError` is not handled by the handler.

This MR reverts some changes and fixes only the failing function, covered with a test.

### DIFF
--- a/unblob/handlers/compression/gzip.py
+++ b/unblob/handlers/compression/gzip.py
@@ -156,7 +156,12 @@ class MultiVolumeGzipHandler(DirectoryHandler):
     PATTERN = Glob("*.gz.*")
 
     def is_valid_gzip(self, path: Path) -> bool:
-        with File.from_path(path) as f:
+        try:
+            file = File.from_path(path)
+        except ValueError:
+            return False
+
+        with file as f:
             try:
                 fp = SingleMemberGzipReader(f)
                 if not fp.read_header():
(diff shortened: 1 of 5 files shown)


unblob/handlers/compression/gzip.py
```diff
@@ -167,11 +172,7 @@ def is_valid_gzip(self, path: Path) -> bool:
 
     def calculate_multifile(self, file: Path) -> Optional[MultiFile]:
         paths = sorted(
-            [
-                p
-                for p in file.parent.glob(f"{file.stem}.*")
-                if p.resolve().exists() and p.stat().st_size > 0
-            ]
```
e3krisztian, line 174: Please note, that if this were the only change (and a test with an empty file), this would have been a good fix.

As it is also possible to get it right at a lower level function, this filtering is no longer necessary.

qkaiser (approved):

Issue #792, opened 2024-02-26
unblob:search_chunks_fuzzer: Out-of-memory in search_chunks_fuzzer


qkaiser:
Project: unblob
Fuzzing Engine: libFuzzer
Fuzz Target: search_chunks_fuzzer
Job Type: libfuzzer_asan_unblob
Platform Id: linux

Crash Type: Out-of-memory (exceeds 2560 MB)
Crash Address: 
Crash State:
  search_chunks_fuzzer
  
Sanitizer: address (ASAN)

Affected revision: [3c2db4a6a8cfe00b90514419687d9ec271404d75](https://github.com/onekey-sec/unblob/commit/3c2db4a6a8cfe00b90514419687d9ec271404d75)

Reproducer test case is below:

```
00000000  20 20 20 20 20 20 20 20  20 20 20 20 20 20 ff 20  |              . |
00000010  00 00 ff ff 00 69 fd ff  ff ff ff ff ff 00 00 00  |.....i..........|
00000020  00 00 00 00 00 00 ff ff  ff ff ff ff ff ff ff ff  |................|
00000030  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
*
00000060  ff ff ff ff ff 00 00 00  00 00 20 00 00 00 04 00  |.......... .....|
00000070  00 83 4a 6a 5d 83 4a 6a  5d 01 00 ff 00 00 00 00  |..Jj].Jj].......|
00000080  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
*
000000c0  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff 00 00  |................|
000000d0  00 00 00 00 00 00 00 00  00 00 00 00 00 00 00 00  |................|
000000e0  00 00 00 00 00 00 00 02  00 00 00 00 00 00 00 00  |................|
000000f0  00 00 00 00 00 00 00 00  00 ff ff ff ff ff ff ff  |................|
00000100  ff ff ff ff ff ff ff ff  ff ff ff ff ff ff ff ff  |................|
*
00000170  ff ff ff ff ff ff 00 00  00 00 00 20 00 00 00 04  |........... ....|
00000180  00 00 83 4a 6a 5d 83 4a  6a 5d 01 00 ff ff 53 ef  |...Jj].Jj]....S.|
[...]

PR #803, opened 2024-03-04
fix(handler): introduce upper size limit for s_log_block_size

nyuware:
The maximum block size limit on a 64bit system is `65536`.
When creating a fake filesystem with `mkfs.ext4` and setting the block size with `-b 65536`, the maximum value that `s_log_block_size` will reach is `6`.

Fixes : https://github.com/onekey-sec/unblob/issues/792

### DIFF
--- a/unblob/handlers/filesystem/extfs.py
+++ b/unblob/handlers/filesystem/extfs.py
@@ -84,6 +84,12 @@ class EXTHandler(StructHandler):
                 "ExtFS header major version too high", rev_level=header.s_rev_level
             )
             return False
+        if header.s_log_block_size > 6:
+            logger.debug(
+                "ExtFS header s_log_block_size is too large",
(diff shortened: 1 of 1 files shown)


qkaiser:
Makes perfect sense since `1024 << 6 = 65536`

unblob/handlers/filesystem/extfs.py
```diff
@@ -84,6 +84,8 @@ def valid_header(self, header) -> bool:
                 "ExtFS header major version too high", rev_level=header.s_rev_level
             )
             return False
+        if header.s_log_block_size > 6:
+            return False
```
qkaiser, line 88:
```suggestion
        if header.s_log_block_size > 6:
            logger.debug("ExtFS header s_log_block_size is too large", s_log_block_size=header.s_log_block_size)
            return False
```

qkaiser (approved):

PR #820, opened 2024-04-02
fix(handler): add support for zip64 values in Central Directory Headers

### LINKED ISSUE #819: Failed to extract zip64 due to mising values in `is_zip64_`
**Describe the bug**
We found in our corpus some zip files that fail to extract with out current zip handler. 

After some investigation, the file seems to be a `zip64` with `FFFF` values in the `Central directory header`. 
sanity check:  returning `True` in `is_zip64_eocd` correctly carves the zip file. 

In conclusion, we should add coverage for zip files that are `zip64` "informations" in their `central directory headers`

nyuware:
`ZIP64` is the standard that replaces zip, its designated to "fix" the 4gb size limit of zip.
In the format, it means that some values can be changed to FFFF, for example :
`offset_of_cd` or `central_directory_size` can be `0xFFFFFFFF`, indicating it's a zip64. 

Some time last week, some zip files were not being extracted, after analysis it seemed that their `central directory` also had FFFF values.

So in this PR, I added a new definition, `is_zip64_cd`, to check if the corresponding values indicate `zip64`

New integration tests are also added, we simply use `zip -fz` to create those zip64 files. 

Fixes #819

### DIFF
--- a/unblob/handlers/archive/zip.py
+++ b/unblob/handlers/archive/zip.py
@@ -36,9 +36,9 @@ class ZIPHandler(StructHandler):
             uint16 internal_file_attr;
             uint32 external_file_attr;
             uint32 relative_offset_local_header;
-            char file_name[file_name_length];
-            char extra_field[extra_field_length];
-        } cd_file_header_t;
+            // char file_name[file_name_length];
+            // char extra_field[extra_field_length];
+        } partial_cd_file_header_t;
 
         typedef struct end_of_central_directory
         {
@@ -94,8 +94,12 @@ class ZIPHandler(StructHandler):
     ) -> bool:
         file.seek(start_offset + end_of_central_directory.offset_of_cd, io.SEEK_SET)
         for _ in range(end_of_central_directory.total_entries):
-            cd_header = self.cparser_le.cd_file_header_t(file)
-            if cd_header.flags & self.ENCRYPTED_FLAG:
+            file_header = self.cparser_le.partial_cd_file_header_t(file)
+            file.seek(
+                file_header.file_name_length + file_header.extra_field_length,
+                io.SEEK_CUR,
+            )
+            if file_header.flags & self.ENCRYPTED_FLAG:
                 return True
         return False
 
@@ -109,6 +113,14 @@ class ZIPHandler(StructHandler):
             or end_of_central_directory.total_entries == 0xFFFF
             or end_of_central_directory.central_directory_size == 0xFFFFFFFF
             or end_of_central_directory.offset_of_cd == 0xFFFFFFFF
+        )
+
+    @staticmethod
+    def is_zip64_cd_file(file_header: Instance):
+        # see https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT section 4.3.9.2
+        return (
+            file_header.file_size == 0xFFFFFFFF
+            or file_header.compress_size == 0xFFFFFFFF
         )
 
     def _parse_zip64(self, file: File, start_offset: int, offset: int) -> Instance:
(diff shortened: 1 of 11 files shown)


unblob/handlers/archive/zip.py
```diff
@@ -147,7 +155,21 @@ def calculate_chunk(self, file: File, start_offset: int) -> Optional[ValidChunk]
...
-            if self.is_zip64_eocd(end_of_central_directory):
+            # some values in the CD can be FFFF, indicating its a zip64
+            # if the offset of the CD is 0xFFFFFFFF, its definitely one
+            # otherwise we check every other header indicating zip64
+            if end_of_central_directory.offset_of_cd == 0xFFFFFFFF:
+                file.seek(offset, io.SEEK_SET)
+                end_of_central_directory = self._parse_zip64(file, start_offset, offset)
+                break
+
+            file.seek(end_of_central_directory.offset_of_cd, io.SEEK_SET)
```
qkaiser, line 166: If the ZIP file is prefixed by garbage data, you'll seek to the wrong place.

```suggestion
            file.seek(start_offset + end_of_central_directory.offset_of_cd, io.SEEK_SET)
```

qkaiser: I'd add a sample with some prefixed data to test that.

qkaiser:
@nyuware Your changes are also making some unit tests fail, specifically the ones running on corrupt ZIP files. Please investigate why.

unblob/handlers/archive/zip.py
nyuware: Modified the code with the `start_offset`, I will add the integration files

nyuware: Added integration tests using dd with /dev/random to have 1024 bytes of random data in front of the file.

unblob/handlers/archive/zip.py
```diff
@@ -147,7 +155,30 @@ def calculate_chunk(self, file: File, start_offset: int) -> Optional[ValidChunk]
             file.seek(offset, io.SEEK_SET)
             end_of_central_directory = self.parse_header(file)
 
-            if self.is_zip64_eocd(end_of_central_directory):
+            if end_of_central_directory.offset_of_cd < offset:
```
qkaiser, line 158:
```suggestion
            if start_offset + end_of_central_directory.offset_of_cd < offset:
```

No ?

unblob/handlers/archive/zip.py
```diff
@@ -147,7 +155,30 @@ def calculate_chunk(self, file: File, start_offset: int) -> Optional[ValidChunk]
...
+                    )
+                    central_directory = self.cparser_le.cd_file_header_t(file)
+                    if self.is_zip64_cd(central_directory):
+                        file.seek(offset, io.SEEK_SET)
+                        end_of_central_directory = self._parse_zip64(
+                            file, start_offset, offset
+                        )
+                        break
+                except Exception:
+                    logger.info("Malformed zip container, discarding")
```
qkaiser, line 172: `Exception` is too generic and the try/catch block is too broad. If it's related to seeking out of range, the `if` condition you introduced should be sufficient. If you want to guard it even more you can check that the central directory offset is positive and less than the current EOCD offset.

Since the central directory offset is used twice, you could renamed it this way:

```python
absolute_offset_of_cd = start_offset + end_of_central_directory.offset_of_cd
```

And I would change the condition to:

```
if 0 > absolute_offset_of_cd < offset:
```

nyuware: Nice, changed it according to your suggestion to get ride of `try expect`

qkaiser (approved):

unblob/handlers/archive/zip.py
```diff
@@ -137,6 +145,23 @@ def _parse_zip64(self, file: File, start_offset: int, offset: int) -> Instance:
...
 
+    def is_zip64(self, file, start_offset, offset, end_of_central_directory):
+        absolute_offset_of_cd = start_offset + end_of_central_directory.offset_of_cd
+
+        if 0 < absolute_offset_of_cd < offset:
+            file.seek(absolute_offset_of_cd, io.SEEK_SET)
+            try:
+                central_directory = self.cparser_le.cd_file_header_t(file)
+            except EOFError:
+                return False
```
e3krisztian, line 156: The source of this exception is an invalid header, that could occur for damaged zip files or for zip file in zip file.
The definition of `cd_file_header_t` has 2 variable length fields, that depend on size values embedded in the same record, these sizes could have damaged value leading to read past EOF.

As these fields are not used at all, and they do cost time to read, and could also fail, I propose to modify the definition and remove (comment out) the last 2 fields (`file_name`, `extra_field`).
Unfortunately removing these fields would lead to incomplete record reads, which cause problems when the zip cd is read through in a loop (`is_encrypted`), so that needs an extra seek adjustment to skip these 2 fields as well.
As an extra safety we could also check the parsed magic values for records and fail on mismatch instead of continuing with bad content.

----

I think we should also look at all the cd file headers to decide if we have zip64 format, e.g. we could have an archive with 2 files, the first one being a short one, while the second overflowing one of the 32bit size fields, making the archive zip64.
The current code would cut off the zip64 records at the end, leading to a potential extraction failure on the truncated chunk.

e3krisztian (changes requested):

e3krisztian (approved):

PR #829, opened 2024-04-11
fix(zlib): Limit zlib pattern at the start of the file

nyuware:
The zlib handler has a weak pattern, it can hog a lot of CPU time, we thus want to improve it's pattern by forcing it to match the start of the file only.

As a result, extraction can be much faster but might results in missed chunks.


qkaiser (approved):
LGTM

PR #831, opened 2024-04-11
fix(permission): Ignore symlinks in fix_permission

### LINKED ISSUE #830: PermissionError raised when fixing permissions on symlinks pointing to non-accessible file/directory
If a filesystem holds a symlink pointing to a file located within a directory the process user has no access to, an unhandled `PermissionError` exception will be raised and stop the extraction.

In the attached sample, the `badlink` file points to `/etc/ssl/private/ssl-cert-snakeoil.key`. This file can't be read due to permissions on `/etc/ssl/private`:

```
ls -al /etc/ssl 
total 52
drwxr-xr-x   4 root root      4096 feb 28 09:57 .
drwxr-xr-x 204 root root     12288 apr 10 14:11 ..
drwxr-xr-x   3 root root     16384 jun  8  2023 certs
-rw-r--r--   1 root root     12419 feb  6  2023 openssl.cnf
drwx--x---   2 root ssl-cert  4096 jun  8  2023 private
```

When calling `.exists()`, pathlib follows the symlink and calls `stat()` on the destination. Since the destination is in a directory that can't be read, the `stat` call fails.

One solution would be to use the [follow_symlinks](https://docs.python.org/3/library/pathlib.html#pathlib.Path.exists) attribute of `exists()`, but it's only available since Python 3.12 so it's a no go.

Sample: [sample.zip](https://github.com/onekey-sec/unblob/files/14946236/sample.zip)

Stack trace:

```
File "/home/quentin/.local/lib/python3.10/site-packages/unblob/extractor.py", line 102, in _fix_extracted_directory
    fix_permission(path)
  File "/home/quentin/.local/lib/python3.10/site-packages/unblob/extractor.py", line 26, in fix_permission
    if not path.exists():
  File "/usr/lib/python3.10/pathlib.py", line 1290, in exists
    self.stat()
```

nyuware:
Fixes a regression introduced in https://github.com/onekey-sec/unblob/commit/486bb91531d6d0f677f68e724675983f04ae1474 where symlink were not being ignored when they should have been 

Fixes #830

### DIFF
--- a/tests/integration/filesystem/squashfs/squashfs_v4_le/__input__/bad_symlink.squashfs
+++ b/tests/integration/filesystem/squashfs/squashfs_v4_le/__input__/bad_symlink.squashfs
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:0c55f810260e48d4db63d02b92a0a07e0e8b04c46070a2f3ef1f2314eaab0d37
+size 4096
--- a/tests/integration/filesystem/squashfs/squashfs_v4_le/__output__/bad_symlink.squashfs_extract/badlink
+++ b/tests/integration/filesystem/squashfs/squashfs_v4_le/__output__/bad_symlink.squashfs_extract/badlink
@@ -1,0 +1,1 @@
+etc/ssl/private/ssl-cert-snakeoil.key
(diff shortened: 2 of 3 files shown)


qkaiser:
Can you add the test sample provided in #830 to the integration tests ? This way we demonstrate the fix.

qkaiser (approved):

PR #852, opened 2024-05-15
fix(ubi): add exception handler for SeekError

mucoze:
This check is needed to exit the loop when the end of the file is reached.

### DIFF
--- a/unblob/handlers/filesystem/ubi.py
+++ b/unblob/handlers/filesystem/ubi.py
@@ -81,7 +81,7 @@ class UBIFSHandler(StructHandler):
     """
     HEADER_STRUCT = "ubifs_sb_node_t"
 
-    EXTRACTOR = Command("ubireader_extract_files", "{inpath}", "-o", "{outdir}")
+    EXTRACTOR = Command("ubireader_extract_files", "{inpath}", "-w", "-o", "{outdir}")
 
     def calculate_chunk(self, file: File, start_offset: int) -> Optional[ValidChunk]:
         endian = get_endian(file, self._BIG_ENDIAN_MAGIC)
--- a/unblob/handlers/filesystem/ubi.py
+++ b/unblob/handlers/filesystem/ubi.py
@@ -7,7 +7,7 @@ from structlog import get_logger
 
 from unblob.extractors import Command
 
-from ...file_utils import InvalidInputFormat, get_endian, iterate_patterns
+from ...file_utils import InvalidInputFormat, SeekError, get_endian, iterate_patterns
 from ...iter_utils import get_intervals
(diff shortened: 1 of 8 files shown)


qkaiser:
My understanding is that if a `SeekError` is raised, then the guessed PEB size is wrong or the UBI image is truncated. If the guessed PEB size is wrong, we should raise an `InvalidInputFormat` rather than silently returning a wrong offset.

What did you observe ? Was it a truncated image or the PEB is wrong ?

mucoze:
PEB size is correct. 
I don't think the UBI images are truncated, given that they are valid samples and can also be properly extracted by `ubireader_extract_images` manually.

I'm observing `SeekError` on 3 different samples.

unblob/handlers/filesystem/ubi.py
```diff
@@ -137,8 +137,14 @@ def _walk_ubi(self, file: File, peb_size: int) -> int:
...
             if first_bytes == b"" or first_bytes != self._UBI_EC_HEADER:
                 break
-            file.seek(offset + peb_size)
-
+            try:
+                file.seek(offset + peb_size)
+            except SeekError:
+                last_chunk_size = file.size() - file.tell() - len(first_bytes)
+                # if aligned, we accept to go to the end even if not full PEB
+                if peb_size % last_chunk_size == 0:
```
martonilles, line 145: I do not understand this alignment check, what needs to be aligned here?

we only check that the peb_size must be the multiple of the last_chunk_size, but does it make the last chunk size valid?

eg: peb size is 256k, then even a 8 or 16byte last chunk would make it valid, what are we really checking here?

qkaiser: I got lost trying to make it work with a very specific sample so as not to have an unknown chunk at the end. Please forget about this :)

The only thing we need is to break on `SeekError`, which is what we do now.

e3krisztian:
I think the changes are fine, and could be merged.

I somewhat miss some means of test, that shows that it is working as intended.

Another potential problem I was thinking on is, when the `_walk_ubi` exits early due to the change resulting in a too small to be useful ubifs chunk. I think it is not a big issue, as it would fail at extraction time. 

Would a minimal size check in `_walk_ubi`/`calculate_chunk` be valid and useful here?

qkaiser:
@e3krisztian integration tests files demonstrating the behavior have been added by @mucoze 

> Would a minimal size check in _walk_ubi/calculate_chunk be valid and useful here?

Minimal size checks are usually arbitrary and not super useful IMO. As you said,  failing case is not a big issue as it would fail at extraction time.

qkaiser (approved):

Issue #875, opened 2024-06-11
Unable to install unblob with pip in Ubuntu 24.04


jrmoore:
**Describe the bug**
Installation via `pip` fails when it reaches `lief`. I believe the problem is there, because failure occurs during its build process, but I am filing this report because unblob installation doesn't seem to rely in `lief` being installed through `pip` and tries to build it unconditionally (`pip install lief` works, installs 0.14.1 at the time of writing).

**To Reproduce**
1. Create a new venv and enter its environment: `python3 -m venv test && source test/bin/activate`
2. [Install `lief` manually: `pip install lief`]
3. Attempt to install unblob with pip: `pip install unblob`

<details>
<summary>Error log</summary>

```
$ pip install unblob

DEPRECATION: Loading egg at /home/james/BinAnalysis/lib/python3.12/site-packages/binwalk-2.4.1+3acfb0c-py3.12.egg is deprecated. pip 24.3 will enforce this behaviour change. A possible replacement is to use pip for package installation.. Discussion can be found at https://github.com/pypa/pip/issues/12330
Collecting unblob
  Using cached unblob-24.6.10-py3-none-any.whl.metadata (1.3 kB)
Collecting arpy<3.0.0,>=2.3.0 (from unblob)
  Using cached arpy-2.3.0-py3-none-any.whl
Collecting attrs>=23.1.0 (from unblob)
  Using cached attrs-23.2.0-py3-none-any.whl.metadata (9.5 kB)
Collecting click<9.0.0,>=8.1.7 (from unblob)
  Using cached click-8.1.7-py3-none-any.whl.metadata (3.0 kB)
Collecting cryptography<43.0,>=41.0 (from unblob)
  Using cached cryptography-42.0.8-cp39-abi3-manylinux_2_28_x86_64.whl.metadata (5.3 kB)
Collecting dissect.cstruct<3.0,>=2.0 (from unblob)
  Using cached dissect.cstruct-2.0-py3-none-any.whl.metadata (6.0 kB)
Collecting jefferson<0.5.0,>=0.4.5 (from unblob)
  Using cached jefferson-0.4.5-py3-none-any.whl.metadata (1.6 kB)
Collecting lark<2.0.0,>=1.1.8 (from unblob)
  Using cached lark-1.1.9-py3-none-any.whl.metadata (1.9 kB)
Collecting lief<0.13.0,>=0.12.3 (from unblob)
  Using cached lief-0.12.3.zip (16.2 MB)
[...]

PR #877, opened 2024-06-13
chore(deps): upgrade lief to 0.14.1

qkaiser:
We had to adapt our code since LIEF no longer returns "???" on Enum lookup with unknown values, but rather raise a RuntimeError.

Related to https://github.com/onekey-sec/unblob/issues/875


qkaiser:
Need to update LIEF in Nix packages too - https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/libraries/lief/default.nix

vlaci:
https://github.com/NixOS/nixpkgs/pull/326322

vlaci:
LIEF update is in nixpkgs in #896

vlaci (approved):

Issue #878, opened 2024-06-14
API assistance?


AndrewFasano:
I'm trying to use the unblob API for what I think should be a fairly straightforward task but I'm having some difficulties and hoping to get some help. I haven't found many examples of API usage so I'm hoping this issue might also help other users get started with the API from the code I have and learn from my mistakes.

My goal here is to use unblob to do a recursive extraction of a blob but to produce a clean copy of each extraction without any sub-extractions (i.e., have no `_extract` files within any of my output directories). Instead I want each of the extractions to be stored one directory deep within an output directory (e.g., output/extraction1, output/extraction2). I've previously implemented something like this by just running unblob then parsing the generated outputs looking for files named `*_extract` but I think it should be much cleaner to do this with the API.

I've written the code below which successfully logs a lot of information about the extraction process and almost gets me what I want, but I find that extraction files sometimes still end up in my output so I suspect I'm doing something wrong or missing something obvious here.

I have 3 specific questions, but any advice or guidance would be much appreciated! Thanks

1. Am I missing a much easier way to do this with the API?
2. Are `blob_id` values supposed to be unique per blob? It seems like the same blob_id will show up with distinct paths for example if a blob is carved into 2 files, both the base blob and the 2 generated files will have the same blob_id. Am I just misunderstanding this interface?
3. Are there examples of API usage somewhere?

Example usage after installing unblob dependencies, unblob itself, and saving the below script as extracator.py
```sh
wget https://dlcdnets.asus.com/pub/ASUS/wireless/RT-N66U_C1/FW_RT_N66U_C1_300438510000.zip
python3 extractor.py FW_RT_N66U_C1_300438510000.zip output
```

In the generated output directory I see one of the extracted directories contains two `_extract` directories

```
$ ls output/extracted/*
output/extracted/00dd3c22e77b9382941bf19afb8370b1897535fb.unblob:
Firmware_Release

output/extracted/31c56af333e9f4652626f6e0e10418e27dd1af33.unblob:
part0  part0_extract  part1  part1_extract

output/extracted/a0f5794c735f16dd4a7b12042ef23e1d95dcb134.unblob:
bin  cifs1  cifs2  dev  etc  home  jffs  lib  media  mmc  mnt  opt  proc  rom  root  sbin  sys  sysroot  tmp  usr  var  www

output/extracted/e008abdad83a71b718083b949b91dafaf071b3c5.unblob:
lzma.uncompressed
```

[...]

PR #891, opened 2024-07-02
Report carve

### LINKED ISSUE #326: distinguish between _carve and _extract directory

### LINKED ISSUE #554: Some directories are not reported
When extracting chunks, there is a logic for handling the whole chunks differently, [here](https://github.com/onekey-sec/unblob/blob/f58f352f04c7ac62c2669a1ab548d64acc3830be/unblob/processing.py#L359-L367). This results that in some cases some directories are not reported. 

Reproduce this with this test file: [test.zip](https://github.com/onekey-sec/unblob/files/11178758/test.zip). This is actually from the integration test suit, but I had to zip it for github to allow me attach it.

If I run this file with unblob and check the report, I get the following item:

<details>
<summary>A part of the generated report json</summary>

```json
{
  "task": {
    "path": "/tmp/fruits.lvl1.lzh",
    "depth": 0,
    "chunk_id": "",
    "__typename__": "Task"
  },
  "reports": [
    {
      "path": "/tmp/fruits.lvl1.lzh",
      "size": 146,
      "is_dir": false,
      "is_file": true,
      "is_link": false,
      "link_target": null,
      "__typename__": "StatReport"
    },
    {
      "magic": "  LHarc 1.x/ARX archive data  [lh0], 0x0 OS, with \"apple.txt\"\\012- data",
      "mime_type": "application/x-lzh-compressed",
      "__typename__": "FileMagicReport"
    },
    {
      "md5": "cf71709694cd2f3e98fcf87524194beb",
      "sha1": "701248bfd7dd7a7360ce237754a82425d1d13346",
      "sha256": "e016f42094b088058e7fa5d9c3f98bafaeac87899205192d95b8001f72058a0f",
      "__typename__": "HashReport"
    },
    {
      "chunk_id": "47941:3",
      "handler_name": "lzh",
      "start_offset": 96,
      "end_offset": 146,
      "size": 50,
      "is_encrypted": false,
      "extraction_reports": [],
      "__typename__": "ChunkReport"
    },
    {
      "chunk_id": "47941:2",
      "handler_name": "lzh",
      "start_offset": 47,
      "end_offset": 96,
      "size": 49,
      "is_encrypted": false,
      "extraction_reports": [],
      "__typename__": "ChunkReport"
    },
    {
      "chunk_id": "47941:1",
      "handler_name": "lzh",
      "start_offset": 0,
      "end_offset": 47,
      "size": 47,
      "is_encrypted": false,
      "extraction_reports": [],
      "__typename__": "ChunkReport"
    }
  ],
  "subtasks": [
    {
      "path": "/tmp/unblob/fruits.lvl1.lzh_extract/96-146.lzh_extract",
      "depth": 1,
      "chunk_id": "47941:3",
      "__typename__": "Task"
    },
    {
      "path": "/tmp/unblob/fruits.lvl1.lzh_extract/47-96.lzh_extract",
      "depth": 1,
      "chunk_id": "47941:2",
      "__typename__": "Task"
    },
    {
      "path": "/tmp/unblob/fruits.lvl1.lzh_extract/0-47.lzh_extract",
      "depth": 1,
      "chunk_id": "47941:1",
      "__typename__": "Task"
    }
  ],
  "__typename__": "TaskResult"
}
```
</details>

This means, when unblob handles `/tmp/fruits.lvl1.lzh`, it will create 3 subtasks:
* `/tmp/unblob/fruits.lvl1.lzh_extract/96-146.lzh_extract`
* `/tmp/unblob/fruits.lvl1.lzh_extract/47-96.lzh_extract`
* `/tmp/unblob/fruits.lvl1.lzh_extract/0-47.lzh_extract`

And will continue to run for those (sub)tasks. However a task for the `/tmp/unblob/fruits.lvl1.lzh_extract` directory is never created, so that directory is just there in the file system without actually being in the generated report.

AndrewFasano:
* Adds a new report type of CarveReport and reports when files are carved to fix #554 
* Changes the directory suffix used by carve to be `_carve` by default to fix #326 

I really wanted to change carving to generate a subtask where there's a task for the original file being processed that has subtasks for each carved file. Then each carved file is another task that generates subtasks for the extraction. But I couldn't find a sane way to do this - would love any feedback or help with it if y'all think that's a better way to approach this.

### DIFF
--- a/unblob/report.py
+++ b/unblob/report.py
@@ -236,6 +236,29 @@ class UnknownChunkReport(Report):
     entropy: Optional[EntropyReport]
 
 
+@attr.define(kw_only=True, frozen=True)
+class CarveReport(Report):
+    """Describes a successful carving operation."""
+
+    carved_from: Path
+    carved_to: Path
+    start_offset: int
+    end_offset: int
+    handler_name: str
+
+    @classmethod
+    def from_chunk(
+        cls, carved_from: Path, carved_to: Path, chunk: "ValidChunk"  # noqa: F821
+    ):
+        return cls(
+            carved_from=carved_from,
+            carved_to=carved_to,
+            start_offset=chunk.start_offset,
+            end_offset=chunk.end_offset,
+            handler_name=chunk.handler.NAME,
+        )
+
+
 @final
 @attr.define(kw_only=True, frozen=True)
 class MultiFileReport(Report):
--- a/unblob/processing.py
+++ b/unblob/processing.py
@@ -95,6 +95,7 @@ class ExtractionConfig:
     process_num: int = DEFAULT_PROCESS_NUM
     keep_extracted_chunks: bool = False
     extract_suffix: str = "_extract"
+    carve_suffix: str = "_extract"
     handlers: Handlers = BUILTIN_HANDLERS
     dir_handlers: DirectoryHandlers = BUILTIN_DIR_HANDLERS
     verbose: int = 1
@@ -584,7 +585,7 @@ class _FileTask:
                 CarveReport.from_chunk(self.task.path, inpath, chunk)
             )
 
-            extract_dir = self.carve_dir / (inpath.name + self.config.extract_suffix)
+            extract_dir = self.carve_dir / (inpath.name + self.config.carve_suffix)
             carved_path = inpath
 
         if extract_dir.exists():
--- a/unblob/processing.py
+++ b/unblob/processing.py
@@ -34,6 +34,7 @@ from .models import (
 from .pool import make_pool
 from .report import (
     CalculateMultiFileExceptionReport,
+    CarveReport,
     EntropyReport,
     ExtractDirectoryExistsReport,
     FileMagicReport,
@@ -579,6 +580,10 @@ class _FileTask:
             carved_path = None
         else:
             inpath = carve_valid_chunk(self.carve_dir, file, chunk)
+            self.result.add_report(
+                CarveReport.from_chunk(self.task.path, inpath, chunk)
+            )
+
             extract_dir = self.carve_dir / (inpath.name + self.config.extract_suffix)
             carved_path = inpath
 


e3krisztian:
The addition of `CarveReport` looks clean enough (will have a second look to verify).

However the change from `_extract` to `_carve` directory suffix is backward incompatible, and while generally I like the idea (I have opened the linked issue it solves), I think it would be better for the default for `ExtractionConfig.carve_suffix` to be `_extract` (=no change to current output), and make it a command line option to override it.

----

I had about the same idea of simplifying the double step extraction (carve then extract chunks). I thought, if a file is not fully recognized by any handler (=has multiple chunks), it should be categorized as "unknown" (or "composite") and handled by a "default" handler, which would recognize and extract (carve) chunks, and could also assign handlers to them. This would be exactly what you wanted - one task for each file. I went so far as to [make an experimental refactor](https://github.com/onekey-sec/unblob/pull/464) to work like this 2 years ago, but it become too big of a change with untidy commits to review, and also probably with some bad decisions and thus was abandoned without much consideration. One of the problems to solve is how to pass the handler between processes to avoid the duplication of the expensive handler selection. With the current solution it is not needed: handler selection and extraction happens in the same process.

I do think understanding and reasoning about a flattened extraction process would be easier, but it would be a big work to rewrite the code now. 

Could you explain why you would like to handle chunk-files by separate (sub-)tasks? Maybe we can come up with a solution for that problem.

AndrewFasano:
Thanks for the feedback. I updated the PR to set the default carve_suffix to be `_extract`. I also fixed a type issue pyright caught, hopefully it will pass the CI checks now.

Thanks for pointing me at #464 - I can now see how complex that change would be and will avoid going down that path! What I'm trying to accomplish here is described in #878, but the short version is that I want to collect a clean version of each extraction produced by unblob. Specifically I'm trying to recover multiple partitions from firmware and package them up into archives for subsequent analysis. I _don't_ want to have any `*_extract` directories or the `####-####.<type>` files created by unblob in the packaged archives.

I've managed to do this using a terrible incantation of `find` to identify `*_extract` directories and then run `tar` on each with various `--exclude` flags to filter out unblob artifacts within the directory, but I figure there must be a better way either using the unblob API or by parsing the extraction report. If you have any tips, please let me know.

qkaiser:
I'm back ! Did anything happen with this @e3krisztian ?

e3krisztian:
> I'm back ! Did anything happen with this @e3krisztian ?

Welcome back @qkaiser !
There was no progress on it, unfortunately.

I think this carve suffix needs to be configurable from the CLI, and did not got around to it.

e3krisztian:
There are 2 features added in this PR (which if I understand properly could each solve the problem of some directories not being reported):
- report carving
- make carve suffix configurable

I think it would be cleaner as 2 independent PR-s, or choosing only one of them.

----

I have made an attempt at adding the command line, rebase, do some cleanups, then to look at it again (see https://github.com/onekey-sec/unblob/tree/report_carve), and do some local experiments.

Based on the experiment results, I would go with reporting the carves (as the title says), as it is almost there, only 2 tests need to be modified to know about the new report type.

Unfortunately the suffix change has bigger problems:
  -  extract root is by default the current directory, and it is the carve/extract directory (which was the same) that is checked if exists (do we need `--force` to delete it?). This somewhat worked so far, and we could find a `firmware.bin_extract` directory at the current directory when extracting `firmware.bin` even in a non-empty current directory. Now this would change, as we would either find `firmware.bin_extract` or `firmware.bin_carves` (`_carves` used for suffix) depending on unblob's recognition of the file format, and existence of extra padding. This I think is confusing. It would be cleaner to require a non-existing root directory, but that is not backward compatible.
  -  currently extraction and carving shares the same `_extract` suffix, which somewhat makes things easy, but when these carve and extract directories are derived differently, the internal variable naming becomes a total mess (one might argue, that the mess exists now) E.g. this line needs some work to get right: `self.carve_dir = config.get_extract_dir_for(self.task.path)`

(the above is not strictly what I saw, but it is probably because of some further mixing up of carve and extraction suffix)

qkaiser:
> There are 2 features added in this PR (which if I understand properly could each solve the problem of some directories not being reported):
> 
> * report carving
> * make carve suffix configurable
> 
> I think it would be cleaner as 2 independent PR-s, or choosing only one of them.

Agree. Let's make 2 independent merge requests.

> I have made an attempt at adding the command line, rebase, do some cleanups, then to look at it again (see https://github.com/onekey-sec/unblob/tree/report_carve), and do some local experiments.
> 
> Based on the experiment results, I would go with reporting the carves (as the title says), as it is almost there, only 2 tests need to be modified to know about the new report type.
> 
> Unfortunately the suffix change has bigger problems:
> 
> * extract root is by default the current directory, and it is the carve/extract directory (which was the same) that is checked if exists (do we need `--force` to delete it?). This somewhat worked so far, and we could find a `firmware.bin_extract` directory at the current directory when extracting `firmware.bin` even in a non-empty current directory. Now this would change, as we would either find `firmware.bin_extract` or `firmware.bin_carves` (`_carves` used for suffix) depending on unblob's recognition of the file format, and existence of extra padding. This I think is confusing. It would be cleaner to require a non-existing root directory, but that is not backward compatible.

It's confusing if we don't explain anything to the end user. I would be okay with this state of things if the non-verbose output would display an "output directory" line with the absolute path to where we put the extracted/carved content. We can also provide context in the interactive help and the online documentation.

> * currently extraction and carving shares the same `_extract` suffix, which somewhat makes things easy, but when these carve and extract directories are derived differently, the internal variable naming becomes a total mess (one might argue, that the mess exists now) E.g. this line needs some work to get right: `self.carve_dir = config.get_extract_dir_for(self.task.path)`

I would definitely argue that the mess exists now. We could move to a more generic naming like `get_output_dir_for` if it helps.

unblob/report.py
```diff
@@ -236,6 +236,29 @@ class UnknownChunkReport(Report):
...
+    @classmethod
+    def from_chunk(
+        cls, carved_from: Path, carved_to: Path, chunk: "ValidChunk"  # noqa: F821
+    ):
+        return cls(
+            carved_from=carved_from,
+            carved_to=carved_to,
+            start_offset=chunk.start_offset,
+            end_offset=chunk.end_offset,
+            handler_name=chunk.handler.NAME,
```
e3krisztian, line 258: I had another look at this implementation and its report-output, and I do not see how duplicating the chunk attributes is useful.

This is also storing the path to the carved file, while what was missing is some record for the parent directory, which is created to hold the carved files and their extractions.

Maybe I miss some use-case for the stored attributes, but I am inclined to drop this `CarveReport` and introduce instead a `CarveDirectoryReport`, with the missing carve directory path.

e3krisztian:
@AndrewFasano thank you for the initial implementation.
Closing in favor of #1017, which still has one of your commits unchanged (8024aede679f0d1b13d253826edf8635a20fd48e).
I hope that implementation would also work for you, please make comments there if not.

Issue #888, opened 2024-06-25
Python package is incompatible with dissect-squashfs


evan-goode:
**Describe the bug**

Hi, I'm trying to use the Python packages `unblob` and `dissect-squashfs` in the same project, but I can't since `unblob` [depends](https://github.com/onekey-sec/unblob/blob/357001eff45f35d3854245dfb9111f4a143223a1/pyproject.toml#L14) on a rather old (2020) version of `dissect.cstruct` and `dissect-squashfs` depends on `dissect.cstruct == 4.*`. Poetry gives me the following error:

```
Because no versions of dissect-squashfs match >1.5,<1.6.dev1 || >1.6.dev1,<1.6.dev2 || >1.6.dev2,<2.0
 and dissect-squashfs (1.6.dev1) depends on dissect.cstruct (==4.*), dissect-squashfs (>1.5,<1.6.dev2 || >1.6.dev2,<2.0) requires dissect.cstruct (==4.*).
And because dissect-squashfs (1.6.dev2) depends on dissect.cstruct (==4.*)
 and dissect-squashfs (1.5) depends on dissect.cstruct (==3.*), dissect-squashfs (>=1.5,<2.0) requires dissect.cstruct (>=3.0.dev,<5).
Because no versions of unblob match >24.6.10,<25.0.0
 and unblob (24.6.10) depends on dissect.cstruct (>=2.0,<3.0), unblob (>=24.6.10,<25.0.0) requires dissect.cstruct (>=2.0,<3.0).
Thus, unblob (>=24.6.10,<25.0.0) is incompatible with dissect-squashfs (>=1.5,<2.0).
So, because sample-project depends on both dissect-squashfs (^1.5) and unblob (^24.6.10), version solving failed.
```

Would it be possible to update unblob to use the latest version of dissect.cstruct?

**To Reproduce**

Use this pyproject.toml:

```
[tool.poetry]
name = "sample-project"
version = "0.1.0"
description = ""
authors = ["Author Name <author@example.com>"]
# readme = "README.md"
# license = "BSD"
packages = [
  { include = "sample_package" }
]

[tool.poetry.dependencies]
python = "^3.10"
dissect-squashfs = "^1.5"

[build-system]
requires = ["poetry-core"]
build-backend = "poetry.core.masonry.api"
```

and run `poetry add unblob`.

**Expected behavior**
I would like to be able to use both dissect-squashfs and unblob in the same codebase.

**Environment information (please complete the following information):**
[...]

PR #895, opened 2024-07-10
chore(dissect.cstruct): Get rid of `Instance` type annotation

vlaci:
- It was already interpreted essentially the same way as `typing.Any`, because of its `__getattr__` method
- Newer versions no longer have the `Instance` proxy object
- `cstruct().<x>` calls are annotated with `Any` in new versions
- The actual runtime type will be `Structure` in newer versions that needs to be changed in `logging` (not changed for now)

Relates-to: #888

### DIFF
--- a/tests/handlers/filesystem/test_jffs2.py
+++ b/tests/handlers/filesystem/test_jffs2.py
@@ -1,7 +1,6 @@
 import binascii
 
 import pytest
-from dissect.cstruct import Instance
 from helpers import unhex
 
 from unblob.file_utils import Endian, File
@@ -59,7 +58,7 @@ def get_valid_jffs2_old_be_header():
     )
 
 
-def calculate_crc(header: Instance):
+def calculate_crc(header):
     return (binascii.crc32(header.dumps()[:-4], -1) ^ -1) & 0xFFFFFFFF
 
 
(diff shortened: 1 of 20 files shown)


e3krisztian:
I did not like the `Instance` type annotations and celebrate its demise, as they were too generic to be meaningful, and what else do we have as values as type instances?

However instead of going explicitly `Any` (which is not really useful as a type annotation either for automated check or for human readers), I would prefer introducing a specific type alias for `Any`, by the name `Structure` or `CStruct`.  Although it would not be much help as a machine checkable type annotation, but it would at least signal a type for the human reader.
After upgrading `dissect.cstruct`, the new `Structure` could be aliased/used instead.

vlaci:
Unfortunately this `Structure` is sprinkled with so heavy metaclass magic[^1], that even its `__mro__` is unavailable as they are overwriting the instance dict directly... Fortunately, the `isinstance(..., Structure)` check will work for logging, but we won't be able to use it for type hints, as the generated fields are absent on the class definition. 

Type-hinting with a custom `Any` type doesn't really add anything to it besides requiring extra imports for type annotations to work. I can add it though if you'd think it'd add any value

[^1]: https://github.com/fox-it/dissect.cstruct/blob/7f6a584f77d48e8cce0ca11fd129b6f2a27e82a1/dissect/cstruct/types/structure.py#L367

e3krisztian (approved):

PR #901, opened 2024-07-19
fix(handler): improve zip64 detection

### LINKED ISSUE #900: zip file being missidentified as zip64
We have identified a possibility for the zip handler to miss label a zip file (zip, zip64).

Its possible for a zip file to contain FFFF values (indicating it's a zip64) but it does not hold a zip64 End of Central Directory. 

If that happens, the zip handler will just raise an InvalidInputFormat and skip the valid zip file

nyuware:
We improve the zip64 detection by trying to parse the zip64 early. 
If the zip file contains a zip64 header, we are sure its a zip64. 
If it doesn't hold a zip64, we are sure its not a zip64 and we fall back to zip32. 

fixes #900

### DIFF
--- a/unblob/handlers/archive/zip.py
+++ b/unblob/handlers/archive/zip.py
@@ -114,9 +114,9 @@ class ZIPHandler(StructHandler):
             or end_of_central_directory.offset_of_cd == 0xFFFFFFFF
         )
 
-    @staticmethod
-    def is_zip64_cd_file(file_header):
+    def has_zip64_tag(self, file):
         # see https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT section 4.3.9.2
+        file_header = self.cparser_le.partial_cd_file_header_t(file)
         return (
             file_header.file_size == 0xFFFFFFFF
             or file_header.compress_size == 0xFFFFFFFF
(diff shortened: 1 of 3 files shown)


unblob/handlers/archive/zip.py
```diff
@@ -115,13 +116,18 @@ def is_zip64_eocd(end_of_central_directory: Instance):
...
         # see https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT section 4.3.9.2
-        return (
+        file_header = self.cparser_le.partial_cd_file_header_t(file)
+        has_zip64_tag = (
             file_header.file_size == 0xFFFFFFFF
             or file_header.compress_size == 0xFFFFFFFF
         )
+        if has_zip64_tag:
+            with contextlib.suppress(InvalidInputFormat):
+                self._parse_zip64(file, start_offset, offset)
```
e3krisztian, line 128: We need to parse the zip64 structure to decide if it is a zip64, but if it is, we will parse again later:
https://github.com/onekey-sec/unblob/pull/901/files#diff-350302c56e7fd3d4ae27392f6c700612f4ed86cd04cea098f32c5cc67894d30fR182-R184

This work should be possible to spare by some reorganization.

e3krisztian (approved):

PR #909, opened 2024-07-23
fix(extract): ensure that reports are pickleable

vlaci:
Unfortunately `dissect.cstruct` 4.0 types are not pickleable

### DIFF
--- a/unblob/report.py
+++ b/unblob/report.py
@@ -13,6 +13,12 @@ import attr
 class Report:
     """A common base class for different reports."""
 
+    def __attrs_post_init__(self):
+        for field in attr.fields(type(self)):
+            value = getattr(self, field.name)
+            if isinstance(value, int):
+                object.__setattr__(self, field.name, int(value))
+
     def asdict(self) -> dict:
         return attr.asdict(self)
 


e3krisztian (commented):
I do not fully understand the original problem it solves.

We are passing the process boundary when finishing tasks and returning reports. So we have some reports that contain references to `dissect.cstruct.Instance`/`Structure` instead of raw Python atoms?
We probably should not, and in that case a more specific fix would be more appropriate?

An example/stack trace for what it fixes exactly may be enlightening ;)

vlaci:
Spectacularly, even primitive types are runtime-generated:

```
2024-07-24 09:30.07 [debug    ] creating special file          pid=1766981 special_path=nuts/hazelnut
2024-07-24 09:30.07 [warning  ] Root privileges are required to create block and char devices. Skipped. device=0x1f00 mode=0x21a4 path=nuts/hazelnut pid=1766981
Process Process-1:
  File "/nix/store/hyppxdhgqx7nbg394j97yhmi6vizl074-python3-3.11.9/lib/python3.11/multiprocessing/process.py", line 314, in _bootstrap
    self.run()
  File "/nix/store/hyppxdhgqx7nbg394j97yhmi6vizl074-python3-3.11.9/lib/python3.11/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/home/vlaci/devel/git/github.com/onekey-sec/unblob/unblob/pool.py", line 62, in _worker_process
    output.put(result)
  File "/nix/store/hyppxdhgqx7nbg394j97yhmi6vizl074-python3-3.11.9/lib/python3.11/multiprocessing/queues.py", line 371, in put
    obj = _ForkingPickler.dumps(obj)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/nix/store/hyppxdhgqx7nbg394j97yhmi6vizl074-python3-3.11.9/lib/python3.11/multiprocessing/reduction.py", line 51, in dumps
    cls(buf, protocol).dump(obj)
_pickle.PicklingError: Can't pickle <class 'types.uint32'>: attribute lookup uint32 on types failed
```

We may need a less brutal approach, as this currently fails when `mock.ANY` is used in tests :(

unblob/report.py
```diff
@@ -13,6 +13,12 @@
 class Report:
     """A common base class for different reports."""
 
+    def __attrs_post_init__(self):
+        for field in attr.fields(type(self)):
+            value = getattr(self, field.name)
+            if isinstance(value, int):
+                object.__setattr__(self, field.name, int(value))
+
```
e3krisztian, line 21: I have looked at the `report` module classes, and I think the only place it could be a problem is the `ChunkReport` class, which is generated from `Chunk` derivatives.

So addressing this problem in the `Chunk` class (`start_offset`, `end_offset`) could be closer to the root cause, and would be less magical.

I would also use https://www.attrs.org/en/stable/init.html#converters on the concrete fields instead of `__attrs_post_init__` (even though that already exists there to do some cross-field validation)

vlaci: The above issue is actually with the error reports we get from the failing `mknod` call, but in theory it could bleed into the chunks themselves

kissgyorgy: The problem can be anywhere where `dissect.cstruct` fills an `int` field, not just one specific report.
We could explicitly put a converter for every such report, but this is far convenient to do, more error-proof and more future-proof, because we don't have to check every single Pull Request for these converter.

kissgyorgy (approved):

PR #910, opened 2024-07-26
feat(handler): add support for squashfs v1

jstucke:
I was wondering why unblob did not find the squashFS file system in a firmware image. Turns out it is squashFS v1.0 and there is no handler for that. It seems that only pattern was missing because once I added it, the file was unpacked without problems.
The header seems to be identical to v2.

The firmware image: PK5000 firmware from https://www.centurylink.com/home/help/internet/modems-and-routers/retired-centurylink-modems.html (the squashFS is at offset 0x90014)

### DIFF
--- a/tests/handlers/filesystem/test_squashfs.py
+++ b/tests/handlers/filesystem/test_squashfs.py
@@ -6,6 +6,7 @@ from helpers import unhex
 
 from unblob.file_utils import round_up
 from unblob.handlers.filesystem.squashfs import (
+    SquashFSv1Handler,
     SquashFSv3Handler,
     SquashFSv4BEHandler,
     SquashFSv4LEHandler,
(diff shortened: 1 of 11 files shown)


kissgyorgy:
Very nice patch, thank you for your contribution! We are looking into this and merge as soon as we can.

jstucke:
I'm struggling a bit with creating a test file for the integration tests. I found the source code of squashfs-tools v1.3 [here](https://sourceforge.net/projects/squashfs/files/squashfs/) and got it to compile after a bit of tinkering but sasquatch will not unpack the created files:

```
SquashFS version [1.0] / inode count [4] suggests a SquashFS image of the same endianess
Read on filesystem failed because EOF
read_block: failed to read block @0x3f01b2
get_metadata: failed to read block
FATAL ERROR: read_inode: failed to read inode 4129202:0
```

I probably did something wrong when building mksquashfs because the files also won't mount. It's a version from 2004 and successfully building it seems to be problematic nowadays :neutral_face:

jstucke:
It seems I finally got it to work correctly after including a zlib version from 2004 and compiling as i386 with `-std=gnu99` :see_no_evil:

kissgyorgy:
Great job! :D It is indeed shows up as squashfs v1. The test fails, because somehow the lfs meta-files got into the fs, not the actual content of the files. Can you repackage it? If you try outside the repository, git lfs shouldn't be trouble.

jstucke:
> The test fails, because somehow the lfs meta-files got into the fs, not the actual content of the files.

Oops, I'm not really used to working with git lfs and somehow copied files that were not checked out properly. It should be fixed now :sweat_smile:

kissgyorgy (approved):
Great job, thank you!

PR #927, opened 2024-08-21
fix(tar): fix verbosity in checksum checker in tar handler

nyuware:
The logger debug was being shown at every level of verbosity. The underscore fixes this.


kissgyorgy (approved):
It was a lot of spam.

PR #946, opened 2024-09-11
Move sasquatch dependency to nixpkgs

vlaci:
Draft until https://github.com/NixOS/nixpkgs/pull/219169 is merged and available in channels.


qkaiser (commented):
nitpick: edit the commit message so as not to confuse the reader now that it no longer requires overriding nixpkgs

qkaiser (approved):

PR #970, opened 2024-10-01
Use nixpkgs as a base to build Unblob from

vlaci:
As https://github.com/NixOS/nixpkgs/pull/345554 is around the corner, we can get rid of most of the old cruft around our nix packaging.


vlaci:
I've heavily rewritten the Nix section of the docs.

flake.nix
```diff
@@ -7,14 +7,6 @@
     url = "github:onekey-sec/unblob-native";
     inputs.nixpkgs.follows = "nixpkgs";
   };
```
qkaiser, line 9: do we still need to declare unblob-native here ?

qkaiser: since unblob-native is part of https://github.com/NixOS/nixpkgs/pull/345554 and pulled by unblob, I would think we can drop it in a similar fashion than `lzallright` and `pyperscan`

vlaci: Not really, but I kept in as these two components are tied together very closely. Will work on #1034 next anyway.

qkaiser: ok

qkaiser (approved):

PR #972, opened 2024-10-01
fix(docker): system-wide install of unblob within container.

### LINKED ISSUE #971: uid for user inside container
**Is your feature request related to a problem? Please describe.**
When using unblob in a container, I need to rebuild the image because of uid missmatches.

**Describe the solution you'd like**
A way to specify uid when starting the container. Example `--user=1001`. This currently doesnt work since the unblob binary is inside the home-directory of unblob with strict permissions.

Something like this makes a new image that works.

```
FROM ghcr.io/onekey-sec/unblob:24.9.30
USER root
RUN usermod --uid 1001 unblob
USER unblob
```

There are multiple ways of solving this.

* Replacing the entrypoint that runs the usermod command above based on the current uid. Then execute "unblob"
* Package the binary in a location that is available for all users in the container. Example /usr/local/bin

**Describe alternatives you've considered**
Continue rebuilding the image, even tho thats not a good solution.

**Additional context**
The files I am scanning is part of a bigger pipeline where files are changed by multiple other images. The uid of the files needs to match up.

qkaiser:
Some users have issues with bind mounts and UID/GID mismatch between their host users and the container user that defaults to 1000.

We fix it by installing unblob system-wide within the container, making it available to any user within the container.

Users that want a one-to-one mapping of their host user to the container user can do so with the `-u` docker run parameter:

```
docker run -u $UID:$GID [...] ghcr.io/onekey-sec/unblob:latest
```

Resolve #971

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,12 +9,10 @@ WORKDIR /data/output
 COPY unblob/install-deps.sh /
 RUN sh -xeu /install-deps.sh
 
-USER unblob
-ENV PATH="/home/unblob/.local/bin:${PATH}"
-
 # You MUST do a poetry build before to have the wheel to copy & install here (CI action will do this when building)
 COPY dist/*.whl /tmp/
 RUN pip --disable-pip-version-check install --upgrade pip
-RUN pip install /tmp/unblob*.whl
+RUN pip install /tmp/unblob*.whl --prefix /usr/local
 
+USER unblob
 ENTRYPOINT ["unblob"]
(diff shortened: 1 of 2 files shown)


kissgyorgy (changes requested):
The change seems good to me, I would write one sentence to the documentation how to use it this way at the end of Docker image section. Maybe even mention that the bind-mounted output dir must have the same UID as passed in.

qkaiser:
@kissgyorgy done

kissgyorgy (approved):

PR #979, opened 2024-10-10
Use nixfmt instead of nixpkgs-fmt

vlaci:
- It won't complain if there is no nix on your machine
  (but then why would you touch nix files?)
- It'll run only for changed files

The code style is changed to match with RFC-166[^1]

[^1]: https://github.com/NixOS/rfcs/pull/166


qkaiser (approved):
I very much like this RFC. Code is definitely easier to read.

PR #985, opened 2024-10-14
chore(nix): make sure ruff is runnable on NixOS

vlaci:
Before `devenv`'s been introduced, `autoPatchelfHook` made sure that external binaries are executable. `Devenv` has its own magic sauce that makes Python dependencies "just work", however `ruff` is not a python binary, so it needs special care.


qkaiser (approved):
Looks like it's fixing my issues :)

PR #995, opened 2024-10-27
Compute and expose χ² probability in EntropyReport

qkaiser:
Introduce another entropy measure based on Chi Square probability by using unblob-native's chi_square_probability function. This function returns the Chi Square distribution probability.
    
Chi-square tests are effective for distinguishing compressed from encrypted data because they evaluate the uniformity of byte distributions more rigorously than Shannon entropy.
    
In compressed files, bytes often cluster around certain values due to patterns that still exist (albeit less detectable), resulting in a non-uniform distribution. Encrypted data, by contrast, exhibits nearly perfect uniformity, as each byte value from 0–255 is expected to appear with almost equal frequency, making it harder to detect any discernible patterns.
    
The chi-square distribution is calculated for the stream of bytes in the chunk and expressed as an absolute number and a percentage which indicates how frequently a truly random sequence would exceed the value calculated. The percentage is the only value that is of interest from unblob's perspective, so that's why we only return it.

According to ent doc⁰:
    
> We [can] interpret the percentage as the degree to which the sequence tested is suspected of being non-random. If the percentage is greater than 99% or less than 1%, the sequence is almost certainly not random. If the percentage is between 99% and 95% or between 1% and 5%, the sequence is suspect. Percentages between 90% and 95% and 5%  and 10% indicate the sequence is “almost suspect”.
    
[0] - https://www.fourmilab.ch/random/
    
This entropy measure is introduced by modifying the `EntropyReport` class so that it contains two `EntropyMeasures`:
- *shannon*: for Shannon entropy, which was already there 
- *chi_square*: for Chi Square entropy, which we introduce
    
The `format_entropy_plot` has been adjusted to display two lines within the entropy graph. One for Shannon, the other for Chi Square.


qkaiser:
Entropy levels for a plaintext lorem ipsum:

image

Entropy levels of that file, now XOR'ed:

image

Entropy levels of that plaintext file, now gzip'ed:

image

Entropy levels of that plaintext file, now AES encrypted:

image

The χ² probability convey more precise information about the nature of analyzed data. We can spot weak encryption like XOR even when Shannon entropy reports constant high levels of entropy. We can differentiate between compressed and encrypted given the χ² probability distribution, while they share similar Shannon entropy levels.

unblob/report.py
```diff
@@ -205,6 +205,12 @@ def lowest(self):
         return min(self.percentages)
 
 
+@attr.define(kw_only=True, frozen=True)
+class EntropyReport(Report):
+    shannon: EntropyMeasurements
+    chi_square: EntropyMeasurements
```
e3krisztian, line 211: This makes sense (maybe using a different name), but it is a breaking change for `Report` users.
I mean, it is probably fine to break it, but we can also be backward compatible if an extra report is returned instead of wrapping it in the changed old one.

```py
class _RandomnessReport(Report): ...  # renamed from EntropyReport, now an implementation detail
class EntropyReport(_RandomnessReport): pass
class ChiSquareReport(_RandomnessReport): pass
```

e3krisztian (changes requested):
I would prefer a wording change both in the commit message and in sources to use *randomness* or *randomness measure* instead of *entropy*.

----

In the commit message of 448c6850cfd4a23a22e1b77cd2049ce15925d8b7 chi^2 probability is categorized as a kind of entropy, probably because as we already have entropy as a randomness measure.

unblob/report.py
qkaiser: Pushed a fixup (93aec8d) where I use this approach and also rename all occurences of "entropy" to "randomness" to be accurate while keeping backward compatibility.

The `ExtractionConfig` is still initialized with `entropy_depth` so we don't break the API, but internally it's an alias to `randomness_depth`.

The CLI accepts both `--randomness-depth` and `--entropy-depth` for the same field, so we don't break CLI for existing users.

Idea is to keep backward compatibility while limiting confusion when devs work around `_RandomnessReport`s.

e3krisztian: I am really sorry to lead you into a rabbit hole.

I see, that it was not possible to just add the new chi-2 report, as `UnknowChunk` was already expected to contain an `EntropyReport`.

In the end, we are also back with wrapping the 2 reports into `RandomnessReport` with "measurements" inside (as it was originally, as can be seen in type annotation like `randomness: Optional[Tuple[EntropyReport, ChiSquareReport]]`), only without the original names.

I suggest using something like the original solution (one report with named measurements inside), or at least using a `RandomnessReport` class with `entropy` and `chi_square` fields in place of `Tuple[EntropyReport, ChiSquareReport]`.
As backward compatibility can not be maintained, so I would also make the further changes as well (`ExtractionConfig.randomness_depth`, and removing `--entropy-depth`).

qkaiser: It's done. The commit message has been adjusted too.

unblob/processing.py
```diff
@@ -725,28 +731,46 @@ def calculate_entropy(path: Path) -> EntropyReport:
...
+
+            chi_square_probability = mt.chi_square_probability(chunk)
+            chisquare_probability_percentage = round(chi_square_probability * 100, 2)
+            chi_square_percentages.append(chisquare_probability_percentage)
+
+    report = RandomnessReport(
+        shannon=RandomnessMeasurements(
+            percentages=shannon_percentages,
+            block_size=block_size,
+            mean=mean(shannon_percentages),
```
e3krisztian, line 748: I think the original calculation resulted in different values for means, as the naive mean ignoring file size can be highly biased:
if the file size is 1025, it is processed as 2 chunks:
- 1024 bytes
- 1 bytes

The 1 byte entropy percentage should be 100%, which pushes the overall `mean` higher than it really is for the whole file.  We can somewhat counteract this bias by weighting the values with the size of the chunk.

The above is valid for `chi_square` as well.

unblob/report.py
e3krisztian: Cool! :)

----

As the commit message was mentioned, I looked at it, and I think this part could be updated with the *randomness* word in place of *entropy*:

> This entropy measure is introduced by modifying the EntropyReport class
> so that it contains two RandomnessMeasurements:
> - shannon: for Shannon entropy, which was already there
> - chi_square: for Chi Square entropy, which we introduce

qkaiser: Fixed.

unblob/processing.py
qkaiser: Fixed.

qkaiser:
@e3krisztian now using unblob-native version [0.1.5](https://github.com/onekey-sec/unblob-native/releases/tag/v0.1.5) that was released today. Ready to be merged I think.

e3krisztian (approved):

PR #987, opened 2024-10-18
fix(test): when run under old systems in devenv

vlaci:
The issue is twofold:
- `sh` can come in as a system binary, and it can be linked to on older libc
- Python when running under devenv has its `manylinux` support runtime binaries injected using `LD_PRELOAD`
- This forces system binaries to use the devenv shipped libc
- This causes an error, if old system binaries are loading new libc

To prevent the problem, use `Python` as a test binary, which is for sure linked to a good libc, making the tests more hermetic

Sample error:

    DEBUG    unblob.extractors.command:command.py:39 {'command': "sh -c '> /run/user/1000/pytest-of-kr/pytest-9/test_command_execution0/created'", 'event': 'Running extract command', 'level': 'debug', 'timestamp': '2024-10-18 13:30.03', 'pid': 31484}
    ERROR    unblob.extractors.command:command.py:67 {'severity': <Severity.WARNING: 'WARNING'>, 'command': "sh -c '> /run/user/1000/pytest-of-kr/pytest-9/test_command_execution0/created'", 'stdout': '', 'stderr': 'sh: symbol lookup error: /nix/store/s0dg13igw8mgzpvmxzd2njk1xwdv21a2-manylinux2014/lib/libc.so.6: undefined symbol: __tunable_is_initialized, version GLIBC_PRIVATE\n', 'exit_code': '0x7f', 'event': 'Extract command failed', 'level': 'error', 'timestamp': '2024-10-18 13:30.03', 'pid': 31484}

cc @e3krisztian

PR #999, opened 2024-10-30
devenv: remove `LD_LIBRARY_PATH` hack from running python environment

vlaci:
To completely elliminate issues like those which came up in #987 

Flake lock file updates:

• Updated input 'devenv':
    'github:vlaci/devenv/5186fecbe2835ac45018ea4940388f8523bf1624' (2024-10-04)
  → 'github:vlaci/devenv/953a526754c5ff5e64ab6925c4388b26ae2c45c6' (2024-10-30)

From the related PR[^1] in `devenv`:

Injecting `LD_LIBRARY_PATH` to the Python runtime environment is great to bypass the need of having to patch non-nix binaries loaded into that environment, however it breaks down, when Python executes any other program not compiled for the given Nix system, e.g. a shell script via `subprocess`.

To work this around, `devenv` will inject a `pth`[^2] file to the virtual environment it creates, which mangles the `LD_LIBRARY_PATH` variable, undoing any changes to it made by `devenv` but preserving changes from other sources.

[^1]: https://github.com/cachix/devenv/pull/1562
[^2]: https://docs.python.org/3/library/site.html

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -182,11 +182,11 @@
         "pre-commit-hooks": "pre-commit-hooks_2"
       },
       "locked": {
-        "lastModified": 1728033307,
-        "narHash": "sha256-7+qCx/he1iiDbF7gA9AAAmmKvEvzRoSl8HCtNHkaTgw=",
+        "lastModified": 1730287473,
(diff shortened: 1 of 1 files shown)


e3krisztian (approved):
`subprocess.run('git')` works with this change (also required by `pre-commit` tools)

e3krisztian:
Thanks!

PR #1008, opened 2024-11-15
Drop python 3.8 support

qkaiser:
Python 3.8 reached EOL on 2024-10-07. Our dependencies will likely drop support for that version as well, so it will be a burden to keep-up support for it sooner or later.


qkaiser:
Draft until we're ready to flip the switch.

qkaiser:
Once reviewed it can be merged. The support drop will reach end users only when we do a tagged release :)

e3krisztian (approved):

PR #1013, opened 2024-11-22
feat(ci): use macos-14 in build-nix.

qkaiser:
macos-12 runner are getting phased out, time to update


vlaci:
hmm, how macos-14 on M1 works with x86_64? Does it go through rosetta?

qkaiser:
> hmm, how macos-14 on M1 works with x86_64? Does it go through rosetta?

I think they run macos-14 on Intel based runners (maybe Mac Mini or similar) when you set the arch to x86_64.

qkaiser:
@vlaci anything else ?

vlaci (approved):

PR #1017, opened 2024-11-26
Report carve dir

e3krisztian:
Reworked #891 to report carve dirs instead of the carved files, as well as support different suffixes for carve and extraction directories.

### DIFF
--- a/tests/test_processing_suffixes.py
+++ b/tests/test_processing_suffixes.py
@@ -1,0 +1,100 @@
+from pathlib import Path
+
+import pytest
+
+from unblob.processing import ExtractionConfig, process_file
+from unblob.report import OutputDirectoryExistsReport
+from unblob.testing import check_output_is_the_same
+
+TEST_DATA_PATH = Path(__file__).parent / "files/suffixes"
+
+
+def _patch(extraction_config: ExtractionConfig, carve_suffix: str, extract_suffix: str):
+    extraction_config.keep_extracted_chunks = False
+    extraction_config.carve_suffix = carve_suffix
+    extraction_config.extract_suffix = extract_suffix
+
+
+@pytest.mark.parametrize(
+    "carve_suffix,extract_suffix,output_root_dir_name",
+    [
+        ("_extract", "_extract", "defaults"),
+        ("_c", "_e", "_c_e"),
+        ("_carve", "_extract", "_carve_extract"),
+    ],
+)
+def test_top_level_carve(
+    carve_suffix: str,
+    extract_suffix: str,
+    output_root_dir_name: str,
+    extraction_config: ExtractionConfig,
+):
+    _patch(extraction_config, carve_suffix, extract_suffix)
+    input_file = TEST_DATA_PATH / "__input__/chunks"
+    carve_dir_name = input_file.name + extraction_config.carve_suffix
+    extract_dir_name = input_file.name + extraction_config.extract_suffix
+    expected_output_dir = TEST_DATA_PATH / "__outputs__/chunks" / output_root_dir_name
+
+    reports = process_file(extraction_config, input_file)
+
+    assert reports.errors == []
+
+    assert (
+        carve_dir_name == extract_dir_name
+        or not (extraction_config.extract_root / extract_dir_name).exists()
+    )
+    check_output_is_the_same(expected_output_dir, extraction_config.extract_root)
+
+
+EXPECTED_COLLISION_PATHS: "dict[tuple[str, str], set]" = {
+    ("_extract", "_extract"): {
+        "collisions.zip_extract/chunks_carve/0-160.gzip_extract",
+    },
+    ("_carve", "_extract"): {
+        "collisions.zip_extract/chunks_carve",
+        "collisions.zip_extract/chunks_carve/0-160.gzip_extract",
+    },
+}
+
+
+@pytest.mark.parametrize(
+    "carve_suffix,extract_suffix,output_root_dir_name",
+    [
+        ("_extract", "_extract", "defaults"),
+        ("_c", "_e", "_c_e"),
+        ("_carve", "_extract", "_carve_extract"),
+    ],
+)
+def test_top_level_extract_and_collisions(
+    carve_suffix: str,
+    extract_suffix: str,
+    output_root_dir_name: str,
+    extraction_config: ExtractionConfig,
+):
+    _patch(extraction_config, carve_suffix, extract_suffix)
+    input_file = TEST_DATA_PATH / "__input__/collisions.zip"
+    carve_dir_name = input_file.name + extraction_config.carve_suffix
+    extract_dir_name = input_file.name + extraction_config.extract_suffix
+    expected_output_dir = (
+        TEST_DATA_PATH / "__outputs__/collisions.zip" / output_root_dir_name
+    )
+
+    reports = process_file(extraction_config, input_file)
+
+    # check collision problems - the input was prepared to have collisions
+    # during both the carving and extracting phases
+    problem_paths = {
+        e.path.relative_to(extraction_config.extract_root).as_posix()
+        for e in reports.errors
+        if isinstance(e, OutputDirectoryExistsReport)
+    }
+    key = (carve_suffix, extract_suffix)
+    assert problem_paths == EXPECTED_COLLISION_PATHS.get(key, set())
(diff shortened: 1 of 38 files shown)


unblob/processing.py
```diff
@@ -101,16 +101,21 @@ class ExtractionConfig:
     verbose: int = 1
     progress_reporter: Type[ProgressReporter] = NullProgressReporter
 
-    def get_extract_dir_for(self, path: Path) -> Path:
-        """Return extraction dir under root with the name of path."""
+    def intern(self, path: Path) -> Path:
```
qkaiser, line 105: function name does not say a lot about its purpose. How about `get_path_under_root` ?

unblob/processing.py
```diff
@@ -101,16 +101,21 @@ class ExtractionConfig:
...
-        extract_dir = self.extract_root / relative_path.with_name(extract_name)
-        return extract_dir.expanduser().resolve()
+        return (self.extract_root / relative_path).expanduser().resolve()
+
+    def get_extract_dir_for(self, carved_path: Path) -> Path:
+        path = carved_path
+        return self.intern(path.with_name(path.name + self.extract_suffix))
+
+    def get_carve_dir_for(self, path: Path) -> Path:
+        return self.intern(path.with_name(path.name + self.carve_suffix))
```
qkaiser, line 119: Should be something like:

```py
    def get_extract_dir_for(self, path: Path) -> Path:
        return self.intern(path.with_name(path.name + self.extract_suffix))

    def get_carve_dir_for(self, path: Path) -> Path:
        return self.intern(path.with_name(path.name + self.carve_suffix))
```

unblob/cli.py
```diff
@@ -265,19 +265,20 @@ def __init__(
     expose_value=False,
 )
 def cli(
+    *,
```
qkaiser, line 268: What does this means ?

unblob/processing.py
```diff
@@ -560,13 +560,42 @@ def _process_chunks(
...
+        if is_whole_file_chunk:
+            # skip carving, extract directly the whole file (chunk)
+            carved_path = self.task.path
+            for chunk in outer_chunks:
+                self._extract_chunk(
+                    carved_path,
+                    chunk,
+                    self.config.get_extract_dir_for(carved_path),
+                    # since we do not carve, we want to keep the input around
+                    remove_extracted_input=False,
```
qkaiser, line 555: That is true only if the user set the `-k` option or the `--keep-extracted-chunks`. No ?

unblob/processing.py
```diff
@@ -582,17 +611,14 @@ def _calculate_randomness(self, path: Path) -> Optional[RandomnessReport]:
...
-            extract_dir = self.carve_dir / (inpath.name + self.config.carve_suffix)
-            carved_path = inpath
-
+    def _extract_chunk(
+        self,
+        carved_path: Path,
+        chunk: ValidChunk,
+        extract_dir: Path,
+        *,
+        remove_extracted_input,
```
qkaiser, line 618: missing type hint

unblob/processing.py
qkaiser: It is actually confusing because in `_extract_chunk` you do:

```py
if remove_extracted_input and not self.config.keep_extracted_chunks:
```

unblob/cli.py
```diff
@@ -453,35 +464,36 @@ def print_report(reports: ProcessResult):
...
-        )
+        for handler, size in sorted(
+            chunks_distribution.items(), key=lambda item: item[1], reverse=True
+        ):
+            chunks_table.add_row(
+                handler.upper(), human_size(size), f"{(size/total_size) * 100:0.2f}%"
+            )
 
-    console.print(chunks_table)
+        console.print(chunks_table)
```
qkaiser, line 496: This whole logic change is unrelated to the display of output path. Should be placed in a separate commit.

unblob/cli.py
```diff
@@ -453,35 +464,36 @@ def print_report(reports: ProcessResult):
...
 
-    if total_size == 0:
-        return
+    console = Console()
 
-    summary = Panel(
-        f"""Extracted files: [#00FFC8]{total_files}[/#00FFC8]
+    if total_size:
+        summary = Panel(
+            f"""Extracted files: [#00FFC8]{total_files}[/#00FFC8]
```
qkaiser, line 471: I would put the output path in the summary panel, first line.

qkaiser (commented):
Nice work. I would maybe add some tests where we set the `ExtractionConfig` to use non-standard `carve_suffix` and `extract_suffix` ? See if we can spot abnormal behavior that would appear out of assumption made on the default suffix values.

qkaiser:
Commits f2cb06ef36501c447b8ee7161cb7f1b47852dc05 and 8024aede679f0d1b13d253826edf8635a20fd48e should be next to each other in the history I think. Easier to follow.

unblob/cli.py
e3krisztian: It means, that the rest of the arguments are keyword only arguments: must be supplied with a name.

The problem with parameters of generic types is, that at the call site it can be confusing, and not easily understandable without looking at the function signature.  Boolean and optional parameters are the most confusing.

`f(True, False, None, True)` would hard to understand without knowing the parameter names, which hopefully are self-describing.

unblob/processing.py
e3krisztian: added.

unblob/processing.py
e3krisztian: > That is true only if the user set the `-k` option or the `--keep-extracted-chunks`. No ?

No, we keep the carved chunks in this case, as here the file is from the extraction (*whole file chunk*), which we do not carve, thus we want to keep input file even if extracted (we want to remove only carved files).

> It is actually confusing because in `_extract_chunk` you do:
> 
> ```python
> if remove_extracted_input and not self.config.keep_extracted_chunks:
> ```

Thanks, this is good catch, moved the condition to the other call site in `_carve_then_extract_chunk`, where it belongs.

unblob/processing.py
e3krisztian: yeah, it was like that, however I wanted to
- say more about the input of `get_extract_dir_for`
- preserve the visual similarity of the logic by not using different variable name

I was not happy with this, so converted as suggested

unblob/processing.py
e3krisztian: Changed to `_get_output_path`.

unblob/cli.py
e3krisztian: Separated. The logic change is a fix commit now, as in certain failure cases there was no user feedback due to an early return in the function.

Separated as fix-commit, moved `Chunks identification ratio` from `Summary` to after chunk distribution table, as chunk stat was causing the silent failure, this change lets both `Summary` and `Encountered errors` to be printed, while keeping chunk stat in one block.

unblob/cli.py
e3krisztian: Cool! Done.

qkaiser:
@e3krisztian cool ! I'll wait for the tests to land before I do a final review round.

tests/test_processing_suffixes.py
```diff
@@ -0,0 +1,109 @@
...
+    expected_output_dir_name: str,
+    extraction_config: ExtractionConfig,
+):
+    _patch(extraction_config, carve_suffix, extract_suffix)
+    input_file = TEST_DATA_PATH / "chunks"
+    carve_dir_name = input_file.name + extraction_config.carve_suffix
+    extract_dir_name = input_file.name + extraction_config.extract_suffix
+    expected_output_dir = (
+        TEST_DATA_PATH / "extractions" / expected_output_dir_name / carve_dir_name
+    )
```
qkaiser, line 41: It feels something's hidden here. How is unblob made to extract into `expected_output_dir_name` ? Shouldn't unblob extract into `chunks_c` or `chunks_carve` as top-level ? What am I missing ?

qkaiser:
@e3krisztian two small comments. I think we can rebase and merge when it's cleared.

tests/test_processing_suffixes.py
e3krisztian: Yep, I was not happy with this either. Changed the input-output layout and renamed the variable to avoid confusion.

e3krisztian:
Rebased and resolved raised problems in tests.

qkaiser (approved):

PR #1068, opened 2025-01-08
ci: Run linear history check only on pull requests

vlaci:
`GITHUB_BASE_REF` is not available for merge groups


qkaiser (approved):

vlaci:
Well, it looks like it's a tad bit more complicated :)

Issue #1034, opened 2024-12-20
RFC: Merge back unblob-native into this repo


vlaci:
The Unblob rust extension started its life as an opt-in piece in this repo. We extracted it into its own package due to (1) limitations in Poetry packaging native extensions, and (2) the higher barrier of entry people needed to build Unblob. 

1. I think it is easy to overcome this by now, we just need to use the `maturin` builder in pyproject.toml to build wheels
2. More and more Python applications/libraries require Rust, so I wouldn't consider this a niche now. Also, direnv integration helps greatly provisioning dev environments.

+1: Refactors, like dropping support for Python 3.8, wouldn't be missing the `unblob-native` part :o)

It looks like, we'd need to switch Poetry for `maturin` or `uv` for installing the project environment as well, if we want to avoid having a separate build output.

qkaiser:
I'm all for it. The most sensitive part will be porting unblob-native CI here and converting the current CI/CD pipeline definitions to use `maturin` without losing anything in the process.

Personally I'm all for limiting duplicate work. Right now the flake updates, the dependency upgrades from dependabot, the project settings, everything is done twice and it's annoying.

qkaiser:
Not sure what's your take but I'd be happy to try `uv`. My trust in tooling coming out of @astral-sh is rather strong.

kissgyorgy:
I'm a fan of monorepos, they are a productivity boost. I would have not split the repo first place, but I wasn't here to prevent it 😀

vlaci:
If we are to do this, we need to get off of Poetry for sure. Tried various ways, but there is still no way to build abi3 wheels with it, and there is no way to use a different builder.
Swithing to uv would be the sensible thing to do, but then we also need to integrate renovate for package updates, as dependabot doesn't handle uv locks.

Renovate would be nice, as we could integrate flake lock updates into it as well.

kissgyorgy:
We don't have to immediately integrate everything, the first step can be to just merge the two repos, later we can move to uv or whatever needed for a more seamless workflow.

vlaci:
> We don't have to immediately integrate everything, the first step can be to just merge the two repos, later we can move to uv or whatever needed for a more seamless workflow.

IMO, it would be easier the other way around: migrate to `uv`, then merge. That way, we can get rid of one round of CI test/publisher job merge hell.

kissgyorgy:
https://github.com/stvnksslr/uv-migrator

vlaci:
[...]

PR #1079, opened 2025-01-13
Poetry -> uv

vlaci:
It was surprisingly easy.

See #1034 for reasoning.


pyproject.toml
```diff
@@ -1,64 +1,57 @@
...
+  "python-magic>=0.4.27",
+  "rarfile>=4.1",
+  "rich>=13.3.5",
+  "structlog>=24.1.0",
+  "treelib>=1.7.0",
+  "ubi-reader>=0.8.9",
+  "unblob-native>=0.1.5",
+]
 description = "Extract files from any kind of container formats"
-license = "MIT"
```
kukovecz, line 6: Was this removed on purpose? The schema linked at the top of the file (https://json.schemastore.org/pyproject.json) allows `license` field here.

.github/actions/setup-dependencies/action.yml
```diff
@@ -13,35 +13,13 @@ runs:
       run: sudo unblob/install-deps.sh
       shell: bash
 
-    - name: Setup Python
-      uses: actions/setup-python@v5
+    - name: Install uv & Python
```
kukovecz, line 16: Did a quick search if any `poetry` stuff remained, the things I found:
* `flake.lock`: `poetry2nix`  - don't we need to sync the lock file?
* `bug-report-🐞.md` 
* `build-publish-image.yml` - actually does some `poetry build` stuff

pyproject.toml
vlaci: This a draft for a reason, although the license key is for custom license
https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license

.github/actions/setup-dependencies/action.yml
vlaci: We don't use poetry2nix, so it is some garbage.
Somewhere I lost the other changes, hmm.

e3krisztian (dismissed):

e3krisztian:
Custom `direnv` code using `poetry` would break, but that is expected.

https://github.com/direnv/direnv/wiki/Python#uv

https://github.com/direnv/direnv/issues/1250

pyproject.toml
qkaiser: > although the license key is for custom license

the doc you linked mentions SPDX explicitly

qkaiser (approved):
I agree with the move to `uv`, but would like to see @kukovecz comments addressed.

pyproject.toml
vlaci: I have zero idea, where I've read it or just dreamt about it, but it is cool then

.github/actions/setup-dependencies/action.yml
vlaci: poetry2nix is a transitive dependency of devenv

qkaiser (approved):

devenv.nix
```diff
@@ -7,33 +12,49 @@
...
     ]
     ++ unblob.runtimeDeps;
 
   tasks = {
+    "venv:link" = {
+      exec = ''
+        VENV_DIR="${config.devenv.root}/.venv"
+
+        if [[ -d "$VENV_DIR" && ! -L "$VENV_DIR" ]]; then
+          echo "${config.devenv.root}/.venv is a leftover directory. Please remove it."
```
e3krisztian, line 43: If we are sure it is a `leftover` without value, we could just `rm -rf` it. If we are not sure, the message could be softer:
```
echo "${config.devenv.root}/.venv is a directory. Please remove it if you want to use devenv."
```

.github/workflows/release.yml
```diff
@@ -15,6 +15,7 @@ jobs:
       - name: Setup dependencies
         uses: ./.github/actions/setup-dependencies
 
-      - run: poetry publish --build
+      - run: uv build
+      - run: uv publish
         env:
-          POETRY_PYPI_TOKEN_PYPI: ${{secrets.POETRY_PYPI_TOKEN_PYPI}}
+          UV_PUBLISH_TOKEN: ${{secrets.POETRY_PYPI_TOKEN_PYPI}}
```
e3krisztian, line 21: Should also update the secret...

e3krisztian (approved):

devenv.nix
vlaci: I just don't want to remove anything from a user's computer I don't know about.
I can add the latter part, but the user needn't know about what devenv is and even that we use it under the hood

.github/workflows/release.yml
vlaci: It would be nice to the secret having a descriptive name, e.g. `PYPI_UPLOAD_TOKEN` or something, but I have no authority over our pypi mirror.

@martonilles should be able to set-up a new token

devenv.nix
e3krisztian: I am fine with it as it is, it is probably in the nitpicking category.

I just found the word `leftover` and the instruction `Please remove it.` a bit too opinionated together.

kukovecz (approved):

devenv.nix
vlaci: Made it less opinionated

PR #1084, opened 2025-01-21
Pyright enable standard type checking mode

vlaci:
See individual commits for explanations

### DIFF
--- a/unblob/handlers/compression/_gzip_reader.py
+++ b/unblob/handlers/compression/_gzip_reader.py
@@ -15,7 +15,7 @@ class SingleMemberGzipReader(gzip._GzipReader):  # noqa: SLF001
         self._crc = zlib.crc32(data, self._crc)
         self._stream_size = self._stream_size + len(data)
 
-    def read(self):
+    def read_all(self):
         uncompress = b""
 
         while True:
@@ -39,7 +39,7 @@ class SingleMemberGzipReader(gzip._GzipReader):  # noqa: SLF001
 
     def read_until_eof(self):
         while not self._decompressor.eof:
-            self.read()
+            self.read_all()
 
     @property
     def unused_data(self):
--- a/unblob/models.py
+++ b/unblob/models.py
@@ -4,7 +4,7 @@ import json
 from collections.abc import Iterable
 from enum import Enum
 from pathlib import Path
-from typing import Optional, TypeVar
+from typing import Generic, Optional, TypeVar, Union
 
 import attrs
 from structlog import get_logger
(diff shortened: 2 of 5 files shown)


unblob/models.py
```diff
@@ -255,30 +255,30 @@ def get_output_dir(self) -> Optional[Path]:
 
 
 class _JSONEncoder(json.JSONEncoder):
-    def default(self, obj):
-        if attrs.has(type(obj)):
```
e3krisztian, line 259: It is OK, but I would just silence the problem here, as I think reading those `o`s has made the code worse.

What about renaming the parameter inside, like

```suggestion
    def default(self, o):
        obj = o   # rename inherited parameter `o`
        if attrs.has(type(obj)):
```

unblob/handlers/compression/_gzip_reader.py
```diff
@@ -15,7 +15,7 @@ def _add_read_data(self, data):
         self._crc = zlib.crc32(data, self._crc)
         self._stream_size = self._stream_size + len(data)
 
-    def read(self):
+    def read_all(self):
```
e3krisztian, line 18: There are more serious potential problems here than a linter problem: this gzip reader is an override of a private third-party class, whose implementation might change any moment.

So I think it is unjustified and somewhat dangerous to change the name of a known to work function (due to a linter problem).
I would just add a type ignore comment here for the original code.

vlaci: > gzip reader is an override of a private third-party class, whose implementation might change any moment.

That's why I thought it is more dangerous to override it than to just extend it with another function.

If in the future any part of this base class that we use would call `read()`, it would either behave incorrectly or explode in case if it would pass in an argument.

unblob/models.py
vlaci: I was thinking about this as well. I am okay with renaming it inside the function.

unblob/handlers/compression/_gzip_reader.py
e3krisztian: Yes, I agree with your logic. Yet I do not like changing code which works, because a linter says so.
Also, I think we do not run pyright in a matrix like the tests, so I think we would not catch problems as easily, as described, and thus we could also hide a newly popping up function accidentally.

vlaci: I did say the exact opposite of that: it would be hard to catch this kind of bug. My assessment is that renaming this function reduces that risk.
The change is not because of a linter warning, but because breaking the parent class's behavior is not a good idea in the first place.

unblob/models.py
qkaiser: worth renaming it in the function to keep linter happy while lowering mental load on our end when reading

unblob/handlers/compression/_gzip_reader.py
qkaiser: I like the fix as it is.

qkaiser:
Put my takes in threads started by @e3krisztian. Generally agree with the changes.

@vlaci not to be pedantic but you should fix the typo in your commit messages (`This check is currently turned of` -> `This check is currently turned off`).

qkaiser (approved):

PR #1085, opened 2025-01-22
fix(sandbox): add remove_dir permissions to sandbox

qkaiser:
When an extraction directory is empty, unblob will try to delete it.

This can lead to PermissionError due to insufficient permissions within the sandbox.

this needs to land in unblob-native first: https://github.com/onekey-sec/unblob-native/pull/82


vlaci:
IMO we dropped the ball on this one.

PR #1091, opened 2025-01-27
chore(renovate): do-not autoapprove lz4 updates

vlaci:
They currently don't provide aarhc64 wheels

https://github.com/python-lz4/python-lz4/pull/298

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -7,6 +7,11 @@
       "matchUpdateTypes": ["minor", "patch"],
       "autoApprove": true,
       "automerge": true
+    },
+    {
+      "matchPackageNames": ["lz4"],
+      "prBodyNotes": ":warning: Check aarch64 compatibility :warning:",
+      "autoApprove": false
     }
   ],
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -7,6 +7,11 @@
       "matchUpdateTypes": ["minor", "patch"],
       "autoApprove": true,
       "automerge": true
+    },
+    {
+      "matchPackageNames": ["lz4"],
+      "prBodyNotes": ":warning: Check aarch64 compatibility :warning:",
+      "autoApprove": false
```
kissgyorgy, line 14: This can happen any time with any other package. What if we would just auto approve minor versions and major versions should be manually accepted?

kissgyorgy (changes requested):

renovate.json
vlaci: this was a minor version update

vlaci:
Closed in favor of #1094

PR #1095, opened 2025-01-28
chore(deps): exclude lz4 4.4.3 which has build issues on aarch64

vlaci:
It has build issues on Ubuntu Noble for aarch64, so no wheels are published upstream:

         gcc -pthread -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -fPIC -Ilz4libs -I/opt/_internal/cpython-3.9.20/include/python3.9 -c lz4libs/lz4.c -o build/temp.linux-aarch64-cpython-39/lz4libs/lz4.o -O3 -Wall -Wundef
        gcc: internal compiler error: Segmentation fault signal terminated program cc1

We will see if the issue is fixed in future updates.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,6 +2,8 @@
 
 [project]
 authors = [{ name = "ONEKEY", email = "support@onekey.com" }]
+
+# Operators supported: https://packaging.python.org/en/latest/specifications/version-specifiers/#id5
 dependencies = [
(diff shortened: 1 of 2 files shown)


qkaiser (approved):
TIL you can explicitly exclude versions in pyproject files :)

PR #1096, opened 2025-02-01
Merge unblob-native

### LINKED ISSUE #1034: RFC: Merge back unblob-native into this repo

vlaci:
Resolves #1034


docs/development.md
```diff
@@ -41,9 +41,9 @@ where the exciting stuff is.
 - **Git LFS**: We have big integration test files, and we are using Git LFS to track them.
   [Install `git-lfs`](https://git-lfs.github.com/) from the website.
 
-- **Rust** (_for unblob-native_): unblob has a [Rust extension](https://github.com/onekey-sec/unblob-native) for
-  performance intensive processing. Building it is entirely optional and requires [`rustup`](https://rustup.rs/) to be
-  installed on the host system. Follow the instructions on the [rustup website](https://rustup.rs/) to install it.
+- **Rust** (_optional_): unblob has a Rust extension for performance intensive processing. Building it
```
qkaiser, line 44: "performance intensive" is weird, no ?

vlaci: yea...

rust/Cargo.toml
```diff
@@ -31,3 +31,7 @@ rand = "0.8.4"
 [[bench]]
 harness = false
 name = "benches_main"
+
+[lints.rust]
+# Required for Rust >= 1.84 && pyo3 < 0.23
+unexpected_cfgs = { level = "allow", check-cfg = ['cfg(addr_of)'] }
```
qkaiser, line 37: I would open a ticket about this so that we can keep track of it and actually work towards upgrading to 0.23

tests/rust/test_sandbox.py
```diff
@@ -2,7 +2,7 @@
 import errno
 import os
 import platform
-from ctypes import c_uint32
+import threading
```
qkaiser, line 5: Can you explain a bit more in the commit about what's happening here ? Are landlock limits applied to the thread only and not the main thread ?

docs/development.md
```diff
@@ -41,9 +41,9 @@ where the exciting stuff is.
 - **Git LFS**: We have big integration test files, and we are using Git LFS to track them.
   [Install `git-lfs`](https://git-lfs.github.com/) from the website.
 
-- **Rust** (_for unblob-native_): unblob has a [Rust extension](https://github.com/onekey-sec/unblob-native) for
-  performance intensive processing. Building it is entirely optional and requires [`rustup`](https://rustup.rs/) to be
-  installed on the host system. Follow the instructions on the [rustup website](https://rustup.rs/) to install it.
+- **Rust** some functionality of unblob is implemented in Rust. Building it is requires a Rust
```
qkaiser, line 44:
```diff
- - **Rust** some functionality of unblob is implemented in Rust. Building it is requires a Rust
+ - **Rust** some functionality of unblob is implemented in Rust. Building it requires a Rust
```

.github/workflows/CI.yml
```diff
@@ -0,0 +1,425 @@
...
+        uses: ./.github/actions/setup-git-lfs
+
+      - name: Run pytest
+        run: uv run pytest -vvv
+
+  build_linux_wheels:
+    name: Build wheels (linux)
+    if: github.event_name == 'push' || github.event_name == 'workflow_dispatch' || contains(github.event.*.labels.*.name, 'dependencies')
+    needs: [check_pre_commit]
+    runs-on: ubuntu-24.04
```
qkaiser, line 111: We use `ubuntu-latest` elsewhere but we use `ubuntu-24.04` here. We should standardize and use the same version everywhere.

qkaiser (commented):
Some nitpicks but otherwise looks good to me. Did a first pass yesterday and a second one this morning.

rust/Cargo.toml
vlaci: Renovate will remind us as well :)

tests/rust/test_sandbox.py
vlaci: added comment to the test

qkaiser (approved):

PR #1100, opened 2025-02-05
Export rust types

vlaci:
Given them a nice canonical place to import them from. Both internally and for external consumers of unblob API.

### DIFF
--- a/python/unblob/sandbox.py
+++ b/python/unblob/sandbox.py
@@ -7,17 +7,20 @@ from typing import Callable, Optional, TypeVar
 
 from structlog import get_logger
 
-from unblob._rust.sandbox import (
-    AccessFS,
-    SandboxError,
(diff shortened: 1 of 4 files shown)


python/unblob/sandbox.py
```diff
@@ -8,7 +8,9 @@
 from structlog import get_logger
 
 from unblob._rust.sandbox import (
-    AccessFS,
+    AccessFS as AccessFS,
+)
+from unblob._rust.sandbox import (
```
qkaiser, line 13: I don't understand why you can do `from ._rust` in `__init__.py` and you need to do `from unblob._rust` in this one.

vlaci: True, relative import would work in both cases.

qkaiser (approved):

PR #1128, opened 2025-02-11
chore(deps): upgrade ubi-reader to 0.8.10

qkaiser:
Integration tests had to be adapted since ubi-reader got smarter when extracting UBI images (padding is no longer written out to files).

See https://github.com/onekey-sec/ubi_reader/commit/40f6f598ff1aa86a53874d11c064cebd74f19378

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -36,18 +36,36 @@
         "aarch64-darwin"
       ];
 
+      # Temporary patches for nixpkgs required for current unblob
+      nixpkgsPatches = [ ];
+
       # Helper function to generate an attrset '{ x86_64-linux = f "x86_64-linux"; ... }'.
       forAllSystems = nixpkgs.lib.genAttrs supportedSystems;
 
       # Nixpkgs instantiated for supported system types.
       nixpkgsFor = forAllSystems (
         system:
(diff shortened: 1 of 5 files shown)


qkaiser:
needs https://github.com/NixOS/nixpkgs/pull/381211

vlaci:
I had an idea (pushed on top). Let me know what do you think of this approach

qkaiser:
> I had an idea (pushed on top). Let me know what do you think of this approach

What happens when the ubi-reader update lands in nixpkgs ? will we need to revert fc87bc352a5108ae53f7f8ec80da214aa7bb1421 ?

vlaci:
> > I had an idea (pushed on top). Let me know what do you think of this approach
> 
> What happens when the ubi-reader update lands in nixpkgs ? will we need to revert [fc87bc3](https://github.com/onekey-sec/unblob/commit/fc87bc352a5108ae53f7f8ec80da214aa7bb1421) ?

It will conflict, and yes, we need to revert adding the patch.

qkaiser:
@vlaci works for me. I can't approve my own PR so it's up to you now.

vlaci (approved):

PR #1129, opened 2025-02-11
fix(tests): run handlers integration tests in landlock sandbox

qkaiser:
Handlers tests were not previously run within the landlock sandbox, and even if it was they would have full access to the filesystem since our `is_sandbox_available()` was checking for landlock's API availability by requesting read-write access to '/'.

It's now fixed by:
- running `is_sandbox_available()` within its own thread, not to pollute the main thread
- run integration tests within the sandbox, one sandboxed thread per handler

The objective is to identify early any issues we may get into when running extraction within the sandbox.

### DIFF
--- a/python/unblob/testing.py
+++ b/python/unblob/testing.py
@@ -1,7 +1,6 @@
 import binascii
 import glob
 import io
-import platform
 import shlex
 import subprocess
 from pathlib import Path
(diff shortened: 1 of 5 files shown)


python/unblob/testing.py
```diff
@@ -222,14 +223,44 @@ def start(self, s):
...
+                "Sandboxing should work at least on Linux-x86_64"
+            )
+
+    t = threading.Thread(target=_run_in_thread)
+    t.start()
+    t.join()
+    __tracebackhide__ = True
+    if exception:
+        raise exception
     return is_sandbox_available
```
vlaci, line 266: I think this mode of checking would be cleaner:
https://github.com/onekey-sec/unblob/blob/main/tests/rust/test_sandbox.py#L38-L69

qkaiser: See https://github.com/onekey-sec/unblob/pull/1129/commits/ced1d96e505ecaa89efcbf141e457d6d15ca1c70

vlaci (approved):

PR #1134, opened 2025-02-20
Arpy based ar extractor

e3krisztian:


python/unblob/file_utils.py
```diff
@@ -269,8 +269,13 @@ def iterate_patterns(
         file.seek(initial_position)
 
 
+class Seekable(Protocol):
```
qkaiser, line 272: I think `read` and `seek` should be marked as `@abc.abstractmethod`. This will be consistent with how we define classes in unblob and should make CodeQL happy.

python/unblob/handlers/archive/ar.py
```diff
@@ -15,6 +17,56 @@
...
+
+        with arpy.Archive(inpath.as_posix()) as archive:
+            archive.read_all_headers()
+
+            for name in sorted(archive.archived_files):
+                archived_file = archive.archived_files[name]
+
+                try:
+                    path = Path(name.decode())
+                except UnicodeDecodeError:
```
qkaiser, line 48: I would copy [integration test files from arpy](https://github.com/viraptor/arpy/tree/master/test) into `tests/integration/archive/ar/__input__` to make sure we're covering these exceptions (and maybe others). They're just files, not code, so I don't see any licensing issue.

qkaiser (commented):
Do you think it would be possible to integrate this patch https://github.com/viraptor/arpy/pull/21/commits/23faf88a88488c41fc4348ea2b70996803f84f40 in this MR ? Maybe in a similar way than what we did when fixing traversals in `tarfile` ?

qkaiser:
@vlaci I'm confused, why are we pulling gdb as a dependency here ? https://github.com/onekey-sec/unblob/actions/runs/13444205574/job/37565632238

vlaci:
> @vlaci I'm confused, why are we pulling gdb as a dependency here ? https://github.com/onekey-sec/unblob/actions/runs/13444205574/job/37565632238

It comes in through devenv's `languages.rust` feature. I am considering going back to the good old days of our own nix-shell definition, as there is a bit much of a churn around devenv stuff.

python/unblob/file_utils.py
e3krisztian: `Protocol` and `ABC` are two different beasts.

This is a `Protocol` class, and they are defining an interface (variables and method signatures), and will match any types, that have this interface, they need not be related in the type hierarchy (that's the point). New, similar implementations differing only in the interface can be adapted.

`ABC`-s on the other hand are to be implemented by descendants (Base Class), and they define/require direct relation in the type hierarchy, which means, the implementation must know in advance what is to be implemented.

----

For reference: there is one other `Protocol` class in `unblob` sources: https://github.com/onekey-sec/unblob/blob/530eacf091d6e57984d28714e6bbcadbab265de2/python/unblob/ui.py#L9

The implementations in that file could all be subclasses of `ProgressReporter`, but I guess there is one implementation, that is defined by a 3rd party in some other lib.

----

I have silenced `CodeQL` by pressing the `Dismiss alert` button. These were false positives, as the code is supposed to have no effect: they are interface declarations to be checked by the type checkers.

e3krisztian:
Renamed `Seekable` to `RandomReader`, as I did not like the name.

qkaiser:
@e3krisztian left a message on the inital PR to arpy, let's wait for devenv to fix their stuff (or use less of it) so the Nix CI passes

python/unblob/file_utils.py
```diff
@@ -269,8 +269,15 @@ def iterate_patterns(
         file.seek(initial_position)
 
 
+class RandomReader(Protocol):
+    # File implements this interface
+
+    def read(self, n: Optional[int] = None) -> bytes: ...
+    def seek(self, pos: int, whence: int = io.SEEK_SET) -> int: ...
```
e3krisztian, line 276: I was thinking about getting `File` (and `OffsetFile` in sync with `io.*` instead, as they use different parameter names.
```py
read(self, size=-1, /)
seek(self, offset, whence=0, /)
```
but it turns out, `io.seek()` has inconsistent names for the first parameter: `offset`, `pos`, target`, and `cookie` anyway, so just gave up on it.

kukovecz (approved):
Added two really minor questions, but consider it approved.

qkaiser:
just so you know, I sent an email to arpy's maintainer to see if we can take over the maintenance or join the maintainers list. wait and see.

python/unblob/handlers/archive/ar.py
```diff
@@ -15,6 +17,56 @@
...
+    """
+
+    def __init__(self, arpy_file: arpy.ArchiveFileData):
+        self._arpy_file = arpy_file
+
+    def read(self, n: Optional[int] = None) -> bytes:
+        return self._arpy_file.read(n)
+
+    def seek(self, pos: int, whence: int = io.SEEK_SET) -> int:
+        return self._arpy_file.seek(pos, whence)
```
vlaci, line 33: You can get rid of this adapter by using positional only arguments combined with `@overload`

E.g. this change type checks:

```diff
diff --git i/python/unblob/file_utils.py w/python/unblob/file_utils.py
index 53868cb1..31ad6c09 100644
--- i/python/unblob/file_utils.py
+++ w/python/unblob/file_utils.py
@@ -12,7 +12,7 @@ import sys
 import unicodedata
 from collections.abc import Iterable, Iterator
 from pathlib import Path
-from typing import Literal, Optional, Protocol, Union
+from typing import Literal, Optional, Protocol, Union, overload
 
 from dissect.cstruct import cstruct
 from structlog import get_logger
@@ -272,8 +272,12 @@ def iterate_patterns(
 class RandomReader(Protocol):
     # File implements this interface
 
+    @overload
+    def read(self) -> bytes: ...
+    @overload
+    def read(self, n: int, /) -> bytes: ...
     def read(self, n: Optional[int] = None) -> bytes: ...
-    def seek(self, pos: int, whence: int = io.SEEK_SET) -> int: ...
+    def seek(self, pos: int, /, whence: int = io.SEEK_SET) -> int: ...
 
 
 def iterate_file(
diff --git i/python/unblob/handlers/archive/ar.py w/python/unblob/handlers/archive/ar.py
index 5c54cf5f..ed4218c9 100644
--- i/python/unblob/handlers/archive/ar.py
+++ w/python/unblob/handlers/archive/ar.py
@@ -1,4 +1,3 @@
-import io
 import os
 from pathlib import Path
 from typing import Optional
@@ -17,22 +16,6 @@ HEADER_LENGTH = 0x44
 SIGNATURE_LENGTH = 0x8
 
 
-class RandomReader:
-    """Adapter for file_utils.RandomReader.
-
-    Changes the parameter names, as they are different for arpy and unblob.File.
-    """
-
-    def __init__(self, arpy_file: arpy.ArchiveFileData):
-        self._arpy_file = arpy_file
-
-    def read(self, n: Optional[int] = None) -> bytes:
-        return self._arpy_file.read(n)
-
-    def seek(self, pos: int, whence: int = io.SEEK_SET) -> int:
-        return self._arpy_file.seek(pos, whence)
-
-
 class ArExtractor(Extractor):
     def extract(self, inpath: Path, outdir: Path) -> Optional[ExtractResult]:
         fs = FileSystem(outdir)
@@ -58,7 +41,7 @@ class ArExtractor(Extractor):
                 fs.write_chunks(
                     path,
                     chunks=iterate_file(
-                        RandomReader(archived_file),
+                        archived_file,
                         0,
                         archived_file.header.size,
                     ),
```

e3krisztian: This is way better, thanks!

python/unblob/handlers/archive/ar.py
```diff
@@ -15,6 +17,56 @@
 SIGNATURE_LENGTH = 0x8
 
 
+class RandomReader:
+    """Adapter for file_utils.RandomReader.
+
+    Changes the parameter names, as they are different for arpy and unblob.File.
```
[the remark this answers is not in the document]
e3krisztian: Yes, it was valid, the parameter names for `read` and `seek` differed for `arpy.ArchiveFileData` and `unblob.File`.

Deleted whole adapter class based on https://github.com/onekey-sec/unblob/pull/1134#discussion_r1973030842

python/unblob/handlers/archive/ar.py
```diff
@@ -15,6 +17,56 @@
 SIGNATURE_LENGTH = 0x8
 
 
+class RandomReader:
```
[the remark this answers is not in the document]
e3krisztian: It should not, but I do not understand in what way it could hurt.

Anyway, it went away applying https://github.com/onekey-sec/unblob/pull/1134#discussion_r1973030842

qkaiser (approved):

PR #1143, opened 2025-03-03
feat(handler): add geom_uzip handler

rxpha3l:
geom_uzip is a FreeBSD feature for creating compressed disk images (usually containing UFS). The compression is done in blocks, and the resulting .uzip file can be mounted via the GEOM framework on FreeBSD.

The mkuzip header includes a table with block counts and sizes. The header declares the block size (size of decompressed blocks) and total number of blocks. Block size must be a multiple of 512 and defaults to 16384 in mkuzip.
It has the following structure:
> Magic, which is a shebang that is stored on 10 bytes.
> Version, which can change and is stored on 13 bytes.
> Command, which can change and is stored on 105 bytes.
> Block size, stored on 4 bytes.
> Block count, stored on 4 bytes.
> Table of content (TOC), which depends on the file lentgh.
The TOC is a list of uint64_t offsets into the file for each block. To determine the length of a given block, read the next TOC entry and subtract the current offset from the next offset (this is why there is an extra TOC entry at the end). Each block is compressed using zlib. A standard zlib decompressor will decode them to a block of size block_size.

Unblob parses the TOC to determine end & start offset of the uzip file. It will find the compressed blocks, decompress them using zlib and parses them together to recover the decompressed file. Empty chunks are ignored, which is why the decompressed file with unlbob can be a little bit lighter than the original one.

[Sources]
https://github.com/freebsd/freebsd-src/blob/master/sys/geom/uzip/g_uzip.c


python/unblob/handlers/compression/geom_uzip.py
```diff
@@ -0,0 +1,79 @@
...
+    C_DEFINITIONS = C_DEFINITIONS
+    EXTRACTOR = UZIPExtractor()
+
+    def calculate_chunk(self, file: File, start_offset: int) -> Optional[ValidChunk]:
+        header = self.parse_header(file, Endian.BIG)
+        # take the last TOC block offset, end of file is that block offset + block size,
+        # starting from the start offset
+        end_offset = start_offset + header.toc[-1]
+        file.seek(end_offset, SEEK_SET)
+        end_offset = read_until_past(file, b"\x00")
```
martonilles, line 75: can we just leave it there and let padding detection pick it up?

or are we sure that we can not read into something?

tests/integration/compression/uzip/__input__/myfs.img.uzip
qkaiser, line 1: this needs to move under `tests/integration/compression/geom_uzip/uzip/__input__` see https://github.com/onekey-sec/unblob/actions/runs/13635740043/job/38114244254?pr=1143#step:5:1069

tests/integration/compression/uzip/__output__/myfs.img.uzip_extract/myfs.img
qkaiser, line 1: this needs to move under `tests/integration/compression/geom_uzip/uzip/__output__` see https://github.com/onekey-sec/unblob/actions/runs/13635740043/job/38114244254?pr=1143#step:5:1069

python/unblob/handlers/compression/geom_uzip.py
```diff
@@ -0,0 +1,79 @@
...
+10 bytes shebang, with newline suffix: #!/bin/sh\n
+13 bytes version, with newline suffix: #V2.0 Format\n or #L3.0 Format\n
+105 bytes command, with null bytes suffix: (kldstat -qm g_uzip||kldload geom_uzip)>&-&&mount_cd9660 /dev/`mdconfig -af $0`.uzip $1\nexit $?\n\x00\x00\x00\x00\x00\x00\x00\x00\x00
+"""
+
+C_DEFINITIONS = r"""
+    typedef struct uzip_header{
+        char magic[10];             /* '10 bytes '*/
+        char version[13];           /* 13 bytes */
+        char format[105];           /* 105 bytes */
```
martonilles, line 29: are we absolutely sure about these sizes? could these be shorter/longer? It is a "strange" way to parse a text

python/unblob/handlers/compression/geom_uzip.py
```diff
@@ -0,0 +1,79 @@
...
+        )
+    ]
+
+    HEADER_STRUCT = HEADER_STRUCT
+    C_DEFINITIONS = C_DEFINITIONS
+    EXTRACTOR = UZIPExtractor()
+
+    def calculate_chunk(self, file: File, start_offset: int) -> Optional[ValidChunk]:
+        header = self.parse_header(file, Endian.BIG)
+        # take the last TOC block offset, end of file is that block offset + block size,
```
qkaiser, line 71: This line is no longer true.

```suggestion
        # take the last TOC block offset, end of file is that block offset,
```

python/unblob/handlers/compression/geom_uzip.py
```diff
@@ -0,0 +1,79 @@
...
+        # starting from the start offset
+        end_offset = start_offset + header.toc[-1]
+        file.seek(end_offset, SEEK_SET)
```
qkaiser, line 74: We use `io.SEEK_SET` everywhere else. Stick to it.

```suggestion
        file.seek(end_offset, io.SEEK_SET)
```

python/unblob/handlers/compression/geom_uzip.py
```diff
@@ -0,0 +1,79 @@
...
+                infile.seek(current_offset, SEEK_SET)
+                outfile.write(decompress(infile.read(compressed_len)))
+        return ExtractResult(reports=fs.problems)
+
+
+class UZIPHandler(StructHandler):
+    NAME = "uzip"
+    PATTERNS = [
+        HexString(
+            "23 21 2F 62 69 6E 2F 73 68 0A 23 (56 32 | 4c 33) 2e 30 20 46 6f 72 6d 61 74 0A"
```
qkaiser, line 61: Would be nice to have a second sample to validate that it also works for samples with `#L3.0 Format` rather than `#V2.0 Format`.

python/unblob/handlers/compression/geom_uzip.py
```diff
@@ -0,0 +1,79 @@
...
+        header = parser.parse(HEADER_STRUCT, infile, Endian.BIG)
+        fs = FileSystem(outdir)
+        outpath = Path(inpath.stem)
+        with fs.open(outpath, "wb+") as outfile:
+            for current_offset, next_offset in zip(header.toc[:-1], header.toc[1:]):
+                compressed_len = next_offset - current_offset
+                if compressed_len == 0:
+                    continue
+                infile.seek(current_offset, SEEK_SET)
+                outfile.write(decompress(infile.read(compressed_len)))
```
qkaiser, line 53: This makes it more obvious that we're decompressing zlib streams here:

```suggestion
                outfile.write(zlib.decompress(infile.read(compressed_len)))
```

python/unblob/handlers/compression/geom_uzip.py
```diff
@@ -0,0 +1,79 @@
...
+        char version[13];           /* 13 bytes */
+        char format[105];           /* 105 bytes */
+        uint32_t block_size;        /* 4 bytes */
+        uint32_t block_count;       /* 4 bytes - Number of blocks */
+        uint64_t toc[block_count];  /* table of content */
+    } uzip_header_t;
+"""
+HEADER_STRUCT = "uzip_header_t"
+
+logger = get_logger()
```
qkaiser, line 37: Unused. You can remove it.

```suggestion
```

python/unblob/handlers/compression/geom_uzip.py
qkaiser, line 29: Since we validate the magic using the `PATTERNS` and do not look futher into these fields, I would simplify it to:

```suggestion
        char magic[128];
```

python/unblob/handlers/compression/geom_uzip.py
```diff
@@ -0,0 +1,79 @@
...
+
+    HEADER_STRUCT = HEADER_STRUCT
+    C_DEFINITIONS = C_DEFINITIONS
+    EXTRACTOR = UZIPExtractor()
+
+    def calculate_chunk(self, file: File, start_offset: int) -> Optional[ValidChunk]:
+        header = self.parse_header(file, Endian.BIG)
+        # take the last TOC block offset, end of file is that block offset + block size,
+        # starting from the start offset
+        end_offset = start_offset + header.toc[-1]
```
martonilles, line 73: what if there is 0 block_count, can we check a sane block_size and block_count limit?

python/unblob/handlers/compression/geom_uzip.py
```diff
@@ -0,0 +1,79 @@
...
+105 bytes command, with null bytes suffix: (kldstat -qm g_uzip||kldload geom_uzip)>&-&&mount_cd9660 /dev/`mdconfig -af $0`.uzip $1\nexit $?\n\x00\x00\x00\x00\x00\x00\x00\x00\x00
+"""
+
+C_DEFINITIONS = r"""
+    typedef struct uzip_header{
+        char magic[10];             /* '10 bytes '*/
+        char version[13];           /* 13 bytes */
+        char format[105];           /* 105 bytes */
+        uint32_t block_size;        /* 4 bytes */
+        uint32_t block_count;       /* 4 bytes - Number of blocks */
```
qkaiser, line 31: I would remove the comments. They do not bring more information.

```suggestion
        uint32_t block_size; 
        uint32_t block_count;
```

python/unblob/handlers/compression/geom_uzip.py
```diff
@@ -0,0 +1,79 @@
...
+    Extractor,
+    ExtractResult,
+    File,
+    HexString,
+    StructHandler,
+    ValidChunk,
+)
+
+""",
+The geom_uzip header follows the following structure:
```
martonilles, line 19: would be nice to link the format documentation site, description

python/unblob/handlers/compression/geom_uzip.py
```diff
@@ -0,0 +1,79 @@
+from io import SEEK_SET
```
martonilles, line 1: would be nice to add test samples with some extra padding before/after content

martonilles (changes requested):

python/unblob/handlers/compression/geom_uzip.py
```diff
@@ -0,0 +1,79 @@
...
+    StructHandler,
+    ValidChunk,
+)
+
+""",
+The geom_uzip header follows the following structure:
+10 bytes shebang, with newline suffix: #!/bin/sh\n
+13 bytes version, with newline suffix: #V2.0 Format\n or #L3.0 Format\n
+105 bytes command, with null bytes suffix: (kldstat -qm g_uzip||kldload geom_uzip)>&-&&mount_cd9660 /dev/`mdconfig -af $0`.uzip $1\nexit $?\n\x00\x00\x00\x00\x00\x00\x00\x00\x00
+"""
```
qkaiser, line 23: This should be removed since it's already documented in the commit message.


```suggestion
```

python/unblob/handlers/compression/geom_uzip.py
qkaiser: for real world samples that we got here, it's usually standalone (no chunks that follows). This helps since extraction is done on the source file itself rather than carving then extraction.

e3krisztian: If it is standalone, why do we have potential padding?

python/unblob/handlers/compression/geom_uzip.py
rxpha3l: I tried it with different files and it was always like this. The header has a total length of 128 bytes. In there you will find the values stored like above.

python/unblob/handlers/compression/geom_uzip.py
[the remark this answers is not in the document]
rxpha3l: This was a mistake. It is actually block offset + null byte padding (`end_offset = read_until_past(file, b"\x00")`) .

python/unblob/handlers/compression/geom_uzip.py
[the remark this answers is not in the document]
rxpha3l: > @rxpha3l the initial implementation we worked on used `iterate_file` no ?

as far as I know, we didn't made changes this far. should I change it ? 




> how big compressed_len could be? it is uint32 offsets, so this could be gigabytes of data compressed and potentially much bigger chunk uncompressed stored in memory, we should use the streaming decompression API

Is there a built in one in unblob or should I import one ?

python/unblob/handlers/compression/geom_uzip.py
rxpha3l: i changed it.

python/unblob/handlers/compression/geom_uzip.py
rxpha3l: ok, I just didn't wanted to import the entire lib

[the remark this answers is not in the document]
qkaiser: > Is there a built in one in unblob or should I import one ?

yes. It's called `iterate_file`.

```suggestion
                decompressor = zlib.decompressobj()
                for chunk in iterate_file(in_file, current_offset, compressed_len):
                    outfile.write(decompressor.decompress(chunk))
```

See https://docs.python.org/3/library/zlib.html#zlib.decompressobj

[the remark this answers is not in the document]
rxpha3l: i put it all in one line. Is that a good practice or should i still declare a new variable ? 
`outfile.write(zlib.decompressobj().decompress(chunk))`

python/unblob/handlers/compression/geom_uzip.py
qkaiser: wrong choice of words on my end. in simpler terms: uzip format pads the compressed file with null byte so that they're aligned to a specific boundary. Padding is an inherent part of uzip. It's cleaner if we keep it rather than carve uzip and null padding separately.

python/unblob/handlers/compression/geom_uzip.py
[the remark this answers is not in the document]
qkaiser: No. As stated in the documentation, the purpose of a decompressor is:

> Returns a decompression object, to be used for decompressing data streams that won’t fit into memory at once.

You have to instantiate the decompressor object once and re-use it for every chunk. It's a stateful object.

Instantiating a decompressor object for every chunk is costly in memory, CPU cycles, and is actually counter-productive to decompression.

python/unblob/handlers/compression/geom_uzip.py
qkaiser: Oh so there are different compression formats. You need to implement that in your extractor. Depending on the version, a different decompressor must be instantiated.

python/unblob/handlers/compression/geom_uzip.py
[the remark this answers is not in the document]
rxpha3l: kk, thanks for the clarification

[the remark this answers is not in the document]
qkaiser: you can do something like this:


```py
if header.version == VERSION_3:
    decompressor = lzma.LZMADecompressor()
elif header.version == VERSION_2:
    decompressor = zlib.decompressobj()
```

python/unblob/handlers/compression/geom_uzip.py
rxpha3l: what do you mean with linking ? Like putting the link to the compressor or the source code ?

python/unblob/handlers/compression/geom_uzip.py
rxpha3l: If you mean with content the compressed blocks in the file, there is always some padding (very very very rare not).  The file size (in bytes) must be a multiple of a certain value. I think it is 512 but I am not very sure. I tested it with this value and i always got a multiple of it.

And if you mean with `padding before the content`, where the content are  the compressed blocks, nothing will change. The TOC will get the right offsets. In the worst case we will have compressed blocks that are empty but 
`if compressed_len == 0:
          continue`
will ignore them.

python/unblob/handlers/compression/geom_uzip.py
rxpha3l: we need the margic value, since there are 3 file types I need to differ

python/unblob/handlers/compression/geom_uzip.py
qkaiser: We mean creating samples with padding before and after the uzip chunk. This way we can test the logic is sound for the end offset calculation.

You can create one this way:

```
dd if=/dev/random bs=1 count=64 of=/tmp/padding.bin
cat /tmp/padding.bin tests/integration/compression/uzip/__input__/myfs.img.uzip /tmp/padding.bin > tests/integration/compression/uzip/__input__/myfs.img.padded.uzip
```

rxpha3l: the test fails since the header (first 64 bytes) are not right. do I need to code a condition to detect such behavior ? Should I  carve out the padding at the beginning and at the end of the file ?

python/unblob/handlers/compression/geom_uzip.py
qkaiser: @rxpha3l can you try something like this ?

```
end_offset = round_up(start_offset + header.toc[-1], header.block_size)
```

and remove the lines 74 and 75.

python/unblob/handlers/compression/geom_uzip.py
qkaiser: this is not the responsibility of your handler to handle this, the test is probably failing because you did not generate the output files first.

```
find tests/integration/compression/uzip/__input__ -type f -exec  uv run unblob -vvv -f -k -e tests/integration/compression/uzip/__output__ {} \;
```

unblob will match on the uzip chunk, carve the prefix, carve the uzip chunk, carve the suffix, and then extract the uzip chunk.

python/unblob/handlers/compression/geom_uzip.py
qkaiser: if you change the code you must re-generate your integration tests files otherwise they do not reflect the state of your code, and therefore break.

qkaiser: after careful consideration and checking the different derivatives of this format, we decided to not try to include the padding suffix since it's not always there (not present with zstd).

python/unblob/handlers/compression/geom_uzip.py
```diff
@@ -0,0 +1,79 @@
...
+
+C_DEFINITIONS = r"""
+    typedef struct uzip_header{
+        char magic[10];             /* '10 bytes '*/
+        char version[13];           /* 13 bytes */
+        char format[105];           /* 105 bytes */
+        uint32_t block_size;        /* 4 bytes */
```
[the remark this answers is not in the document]
qkaiser: we have a smarter approach that does not require `block_size`, `block_count * block_size` to get the end offset is too naive as an approach.

python/unblob/handlers/compression/geom_uzip.py
qkaiser: @rxpha3l my understanding is that `block_size` must be a power of 2 and block_count must be non-null

python/unblob/handlers/compression/geom_uzip.py
qkaiser: Add a reference in the code to https://github.com/freebsd/freebsd-src/tree/master/sys/geom/uzip, that's all

qkaiser:
@vlaci what would be the easiest way to add pyzstd to unblob dependencies in Nix here ? It's not yet in upstream at https://github.com/NixOS/nixpkgs/blob/0fa90d642277de2c67e93204cc5870aba8af5878/pkgs/by-name/un/unblob/package.nix#L59 so we need a way to define it in this branch in the meantime.

qkaiser:
> @vlaci what would be the easiest way to add pyzstd to unblob dependencies in Nix here ? It's not yet in upstream at https://github.com/NixOS/nixpkgs/blob/0fa90d642277de2c67e93204cc5870aba8af5878/pkgs/by-name/un/unblob/package.nix#L59 so we need a way to define it in this branch in the meantime.

@rxpha3l I'm using this fix locally, but not sure if it's idiomatic Nix

```diff
diff --git a/overlay.nix b/overlay.nix
index 9c5051e..265cd79 100644
--- a/overlay.nix
+++ b/overlay.nix
@@ -29,6 +29,8 @@ final: prev:
         ];
       };
 
+      dependencies = (super.dependencies or []) ++ [ prev.python3.pkgs.pyzstd ];
+
       # remove this when packaging changes are upstreamed
       cargoDeps = final.rustPlatform.importCargoLock {
         lockFile = ./Cargo.lock;
```

python/unblob/handlers/compression/geom_uzip.py
rxpha3l: is now applied

python/unblob/handlers/compression/uzip.py
```diff
@@ -77,8 +78,17 @@ class UZIPHandler(StructHandler):
     C_DEFINITIONS = C_DEFINITIONS
     EXTRACTOR = UZIPExtractor()
 
+    def is_valid_header(self, block_size: int, block_count: int) -> bool:
```
qkaiser, line 81:
```suggestion
    def is_valid_header(self, header) -> bool:
```

rxpha3l: is the return condition good for you for the ìs_valid_header() function ?
`
        return (
            header.block_count > 0 and header.block_size > 0 and (header.block_size & (header.block_size - 1)) == 0
        )
`

qkaiser: yes, please adjust in the code

overlay.nix
```diff
@@ -29,6 +29,8 @@ final: prev:
         ];
       };
 
+      dependencies = (super.dependencies or []) ++ [ prev.python3.pkgs.pyzstd ];
```
vlaci, line 32:
```suggestion
      dependencies = (super.dependencies or []) ++ [ final.python3.pkgs.pyzstd ];
```

qkaiser (approved):

python/unblob/handlers/compression/uzip.py
```diff
@@ -0,0 +1,100 @@
...
+        outpath = Path(inpath.stem)
+
+        if header.magic == ZLIB_COMPRESSION:
+            decompressor_cls = zlib.decompressobj
+        elif header.magic == LZMA_COMPRESSION:
+            decompressor_cls = lzma.LZMADecompressor
+        elif header.magic == ZSTD_COMPRESSION:
+            decompressor_cls = pyzstd.ZstdDecompressor
+        else:
+            raise InvalidInputFormat("unsupported compression format")
```
e3krisztian, line 57: can be simplified if we have the map mentioned above to
```py
try:
   decompressor_cls = VERSION_TO_DECOMPRESSOR[header.magic.rstrip()]
except LookupError:
   raise InvalidInputFormat("unsupported compression format")
```

python/unblob/handlers/compression/uzip.py
```diff
@@ -0,0 +1,100 @@
...
+                    outfile.write(decompressor.decompress(chunk))
+        return ExtractResult(reports=fs.problems)
+
+
+class UZIPHandler(StructHandler):
+    NAME = "uzip"
+    PATTERNS = [
+        Regex(r"#!/bin/sh\x0A#V2.0"),
+        Regex(r"#!/bin/sh\x0A#L3.0"),
+        Regex(r"#!/bin/sh\x0A#Z4.0"),
```
e3krisztian, line 75: I would use the constants in a previous comment above

```
PATTERNS = [
    Regex(re.escape(UZIP_V2)),
    Regex(re.escape(UZIP_V3)),
    Regex(re.escape(UZIP_V3)),
]
```

----

Wrapping in `re.escape()` is not technically necessary in this case, but constant definitions could have regexp magic chars, causing unintended failure.

Best way would be defining a new `FixedString(Pattern)`, besides `Regex`, that would do the escaping internally.

python/unblob/handlers/compression/uzip.py
```diff
@@ -0,0 +1,100 @@
...
+        uint32_t block_count;
+        uint64_t toc[block_count];
+    } uzip_header_t;
+"""
+
+HEADER_STRUCT = "uzip_header_t"
+
+ZLIB_COMPRESSION = b"#!/bin/sh\x0a#V2.0\x20"
+LZMA_COMPRESSION = b"#!/bin/sh\x0a#L3.0\x0a"
+ZSTD_COMPRESSION = b"#!/bin/sh\x0a#Z4.0\x20"
```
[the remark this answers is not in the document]
e3krisztian: `UZIP_L3`, `UZIP_Z4`, I missed that they are not all starting `V`.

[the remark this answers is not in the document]
rxpha3l: isn't it better to name them by their compression method ? because in no documentation they differentiate by the version but by the compression method. could be confusing, no ?

python/unblob/handlers/compression/uzip.py
```diff
@@ -0,0 +1,100 @@
...
+    ]
+    HEADER_STRUCT = HEADER_STRUCT
+    C_DEFINITIONS = C_DEFINITIONS
+    EXTRACTOR = UZIPExtractor()
+
+    def is_valid_header(self, header) -> bool:
+        return (
+            header.block_count > 0
+            and header.block_size > 0
+            and (header.block_size & (header.block_size - 1)) == 0
```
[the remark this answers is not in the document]
rxpha3l: it is the bitwise method which is usually faster

python/unblob/handlers/compression/uzip.py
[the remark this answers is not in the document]
rxpha3l: I use the escape character since the header magic uses the entire 16 bytes. with this i am sure that the magic is correct. is it better to drop it ?

python/unblob/handlers/compression/uzip.py
rxpha3l: is it really  better to put the decompression in `re.escape()` ? they are stored as bytes, which is mandatory for previous operation. `re.escape()` expects a string, so in that case i have to use the `.decode()` function. Is it not more efficient to directly put the raw regex string in `Regex() ? or is there a better way to do it ? because in the other handlers i often see `Regex()` or `HexString()` used as  matching element.

python/unblob/handlers/compression/uzip.py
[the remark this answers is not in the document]
e3krisztian: Discussed, that it is not faster in Python, and mod 512 is probably enough.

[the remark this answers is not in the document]
e3krisztian: Will also be removed.

python/unblob/handlers/compression/uzip.py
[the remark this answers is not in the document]
e3krisztian: discussed, that use of constants is the important thing here, and not the name, so the names could remain.

python/unblob/handlers/compression/uzip.py
```diff
@@ -0,0 +1,100 @@
...
+
+        with fs.open(outpath, "wb+") as outfile:
+            for current_offset, next_offset in zip(header.toc[:-1], header.toc[1:]):
+                compressed_len = next_offset - current_offset
+                if compressed_len == 0:
+                    continue
+                decompressor = decompressor_cls()
+                for chunk in iterate_file(infile, current_offset, compressed_len):
+                    outfile.write(decompressor.decompress(chunk))
+        return ExtractResult(reports=fs.problems)
```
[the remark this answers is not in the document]
e3krisztian: The decompressor could hold on some uncompressed bytes, if given a buffer limit, but also in some other cases:
- `lzma.LZMADecompressor`: `.eof`, `.needs_data`, `.unused_data`, but only used if we gave a non-negative buffer size
- `pyzstd.ZstdDecompressor` has the same fields, and further says in the `.decompress()` docstring: 
> Decompress data, return a chunk of decompressed data if possible, or b'' otherwise.
>
> It stops after a frame is decompressed.
- `zlib.decompressorobj()` also has `.eof`, `unused_data`, but `.flush()` and `.unconsumed_tail` as well

So at least `pyzstd` says it will never return more than a frame's worth of decompressed data at once.

Maybe the others also have some surprising logic, and we need to do some more work to decompress a stream properly (looking at the `lzma.decompress` function source it is also silently ignoring after `eof` input - we should probably check for that as well - it is potentially a problem to have extra bytes).

python/unblob/handlers/compression/uzip.py
e3krisztian: The important thing is to use the same magic constant as pattern if possible.

python/unblob/handlers/compression/uzip.py
```diff
@@ -1,10 +1,10 @@
-import lzma
 import re
 import zlib
+from lzma import LZMADecompressor
```
qkaiser, line 3: why this change ?

qkaiser (commented):
check the comments and rebase so there is no fixup commit

python/unblob/handlers/compression/uzip.py
rxpha3l: i talked with Krisztian, and maybe it is better just to import de decompression object instead of the entire library. should i change it ?

e3krisztian (approved):

python/unblob/handlers/compression/uzip.py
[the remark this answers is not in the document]
e3krisztian: Support for `flush`-ing the decompressor state was added. It has non-dummy implementation only for zlib (the others simply return an empty bytes object).

martonilles (changes requested):
small leak, otherwise looks ok!

PR #1147, opened 2025-03-10
Add documentation on developing for unblob's plugin system

kernelmethod:
Add a new section to the Development documentation with information on how to use a new handler via unblob's plugin system, using the `MyformatHandler` example from the docs.

My understanding of the unblob plugin system comes from browsing the pluggy documentation and skimming the source of `unblob`, if there's any missing or incorrect information please let me know.


docs/development.md
```diff
@@ -399,6 +399,46 @@ We've implemented integration tests this way:
...
+from unblob.plugins import hookimpl
+
+class MyformatHandler(StructHandler):
+    # Format-handling code would go here
+    ...
+
+@hookimpl(specname="unblob_register_handlers")
+def add_handlers():
+    return [MyformatHandler]
+```
```
qkaiser, line 429: We do it this way:

```suggestion
@hookimpl
def unblob_register_handlers() -> list[type[Handler]]:
    return [MyformatHandler]
```

And you would need to import `Handler`:

```py
from unblob.models import Handler
```

qkaiser (commented):
Thanks for taking the time. Highly appreciated. Just a small comment regarding the hook implementation.

docs/development.md
kernelmethod: Ah, got it, thanks for the clarification. Just pushed up that fix.

qkaiser (approved):

PR #1155, opened 2025-03-26
feat(handler): add partclone handler

rxpha3l:
Partclone is a program similar to the well-known backup utility "Partition Image" a.k.a partimage. Partclone provides utilities to save and restore used blocks on a partition and is designed for higher compatibility of the file system by using existing libraries, e.g. e2fslibs is used to read and write the ext2 partition.
Partclone now supports ext2, ext3, ext4, hfs+, reiserfs, reiser4, btrfs, vmfs3, vmfs5, xfs, jfs, ufs, ntfs, fat(12/16/32), exfat, f2fs, nilfs.
Partclone has two version. This handler supports the second one and automatically checks the endian value.

Create partclone image:
`partclone.[file_system] -F -c -s [source_file] -o [output.img] -L /dev/stdout[default=/var/log/partclone.log]`

Restore partclone image:
`partclone.restore -W -r -s [source_fiile.img] -o [restored.img] -L /dev/stdout[default=/var/log/partclone.log]`

**Notes**
- It is worth to mention that a logfile `-L` should be used since it will lead in the most cases to errors
- Partclone has no standardized file extension but the official website uses `.img`
- The official documentation got the endian values wrong. Swapping them will solve this issue
 
The header has the following structure : 
> 16 bytes : Magic
> 14 bytes : Partclone version
>   4 bytes : Image version in str
>   2 bytes : Endian
> 16 bytes : File system type
>   8 bytes : File system size
>   8 bytes : File system total block count
>   8 bytes : Used blocks based on superblock
>   8 bytes : Used blocks based on bitmap
>   4 bytes : Block size
>   4 bytes : Feature size
>   2 bytes : Image version in bin
>   2 bytes : Number of bits for CPU
>   2 bytes : Checksum mode
>   2 bytes : Checksum size
>   4 bytes : Blocks per checksum
>   1 byte  :  Reseed checksum
>   1 byte  :  Bitmap mode
>   4 bytes : CRC32 of previous 106 bytes

Unblob parses the header to determine the start and end offset by adding:
- Header
- Bitmap size + Checksum
- Data size
- Checksum per N bytes. This is calculated by multiplying `Block size`*`Blocks per checksum`. Each N-Bytes a 4 byte checksum will occur (min: 1)
Finally the `CRC32` value in the header does not the final XOR with `0xFFFFFFFF`. This can lead to deviation if checked manually. 

[Sources]
https://github.com/Thomas-Tsai/partclone
https://partclone.org/
https://wiki.osdev.org/Ext2
https://piazza.com/class_profile/get_resource/il71xfllx3l16f/inz4wsb2m0w2oz
https://www.science.smith.edu/~nhowe/262/oldlabs/ext2.html
https://www.science.smith.edu/~nhowe/262/oldlabs/kernel/ext2_fs.h

### DIFF
--- a/docs/formats.md
+++ b/docs/formats.md
@@ -8,7 +8,7 @@ hide:
 unblob supports more than 30 formats. You can see their code in
 [`unblob/handlers/`](https://github.com/onekey-sec/unblob/blob/main/unblob/handlers/__init__.py).
 
-✅: Some or all metadata is preserved for the format.  
+✅: Some or all metadata is preserved for the format.
 ❌: Metadata is not preserved (limitation of the format).
 
 ## Archives
@@ -22,6 +22,7 @@ unblob supports more than 30 formats. You can see their code in
 | CAB     | ❌         | ❌                 | ❌        | [archive/cab.py][cab-handler]         | [`7z`][cab-extractor]       |
 | CPIO    | ✅         | ✅                 | ✅        | [archive/cpio.py][cpio-handler]       | unblob extractor            |
 | DMG     | ❌         | ❌                 | ❌        | [archive/dmg.py][dmg-handler]         | [`7z`][dmg-extractor]       |
+| PARTCLONE | ✅       | ❌                 | ❌        | [archive/partclone.py][partclone-handler] | [`partclone`][partclone-extractor] |
 | RAR     | ❌         | ❌                 | ❌        | [archive/rar.py][rar-handler]         | [`unar`][rar-extractor]     |
 | 7ZIP    | ❌         | ❌                 | ❌        | [archive/sevenzip.py][7zip-handler]   | [`7z`][7zip-extractor]      |
 | StuffIt | ❌         | ❌                 | ❌        | [archive/stuffit.py][stuffit-handler] | [`unar`][stuffit-extractor] |
@@ -39,6 +40,8 @@ unblob supports more than 30 formats. You can see their code in
 [cpio-handler]: https://github.com/onekey-sec/unblob/blob/main/unblob/handlers/archive/cpio.py
 [dmg-handler]: https://github.com/onekey-sec/unblob/blob/main/unblob/handlers/archive/dmg.py
 [dmg-extractor]: https://github.com/onekey-sec/unblob/blob/3008039881a0434deb75962e7999b7e35aca8271/unblob/handlers/archive/dmg.py#L67-L69
+[partclone-handler]: https://github.com/onekey-sec/unblob/blob/main/unblob/handlers/archive/partclone.py
+[partclone-extractor]: https://github.com/onekey-sec/unblob/blob/b21b6dc291583af6b7ec9b7c3d63ee8302328841/python/unblob/handlers/archive/partclone.py#L44
 [rar-handler]: https://github.com/onekey-sec/unblob/blob/main/unblob/handlers/archive/rar.py
 [rar-extractor]: https://github.com/onekey-sec/unblob/blob/3008039881a0434deb75962e7999b7e35aca8271/unblob/handlers/archive/rar.py#L32
 [7zip-handler]: https://github.com/onekey-sec/unblob/blob/main/unblob/handlers/archive/sevenzip.py
@@ -97,7 +100,7 @@ For compression formats, metadata cannot be preserved, as this information in mo
 | ---------------------- | ---------------------------------- | ----------------------------------------------- | ----------------------------------------------- |
 | Android sparse image   | ❌                                 | [filesystem/android/sparse.py][android-handler] | [`simg2img`][android-extractor]                 |
 | CRAMFS                 | ✅                                 | [filesystem/cramfs.py][cramfs-handler]          | [`7z`][cramfs-extractor]                        |
-| EROFS                  | ✅                                 | [filesystem/android/erofs.py][erofs-handler]    | [`fsck.erfos`][erofs-extractor]                 | 
+| EROFS                  | ✅                                 | [filesystem/android/erofs.py][erofs-handler]    | [`fsck.erfos`][erofs-extractor]                 |
 | ExtFS                  | ✅                                 | [filesystem/extfs.py][extfs-handler]            | [`debugfs`][extfs-extractor]                    |
 | FAT                    | ✅                                 | [filesystem/fat.py][fat-handler]                | [`7z`][fat-extractor]                           |
 | ISO9660                | ✅                                 | [filesystem/iso9660.py][iso9660-handler]        | [`7z`][iso9660-extractor]                       |
--- a/package.nix
+++ b/package.nix
@@ -1,5 +1,6 @@
 {
   lib,
+  stdenv,
   python3,
   makeWrapper,
   e2fsprogs-nofortify,
@@ -9,6 +10,7 @@
   lziprecover,
   lzop,
   p7zip16,
+  partclone,
   nix-filter,
   sasquatch,
   sasquatch-v4be,
@@ -22,7 +24,7 @@
 
 let
   # These dependencies are only added to PATH
-  runtimeDeps = [
+  runtimeDeps = lib.optional stdenv.isLinux partclone ++ [
     e2fsprogs-nofortify
     erofs-utils
     jefferson
--- a/python/unblob/file_utils.py
+++ b/python/unblob/file_utils.py
@@ -358,14 +358,28 @@ class StructParser:
 def get_endian(file: File, big_endian_magic: int) -> Endian:
     """Read a four bytes magic and derive endianness from it.
 
-    It compares the read data with the big endian magic.  It reads
-    four bytes and seeks back after that.
+    It compares the read data with the big endian magic and then seeks back
+    the amount of read bytes.
     """
     if big_endian_magic > 0xFF_FF_FF_FF:
         raise ValueError("big_endian_magic is larger than a 32 bit integer.")
     magic_bytes = file.read(4)
     file.seek(-len(magic_bytes), io.SEEK_CUR)
     magic = convert_int32(magic_bytes, Endian.BIG)
+    return Endian.BIG if magic == big_endian_magic else Endian.LITTLE
+
+
+def get_endian_short(file: File, big_endian_magic: int) -> Endian:
+    """Read a two bytes magic and derive endianness from it.
+
+    It compares the read data with the big endian magic and then seeks back
+    the amount of read bytes.
+    """
+    if big_endian_magic > 0xFF_FF:
+        raise ValueError("big_endian_magic is larger than a 16 bit integer.")
+    magic_bytes = file.read(2)
+    file.seek(-len(magic_bytes), io.SEEK_CUR)
+    magic = convert_int16(magic_bytes, Endian.BIG)
     return Endian.BIG if magic == big_endian_magic else Endian.LITTLE
 
 
--- a/python/unblob/handlers/archive/partclone.py
+++ b/python/unblob/handlers/archive/partclone.py
@@ -1,0 +1,81 @@
+import binascii
+import io
+from math import ceil
+from typing import Optional
+
+from unblob.extractors import Command
+from unblob.file_utils import File, InvalidInputFormat, get_endian_short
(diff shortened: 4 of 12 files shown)


python/unblob/handlers/archive/partclone.py
```diff
@@ -0,0 +1,82 @@
...
+            len(header) + (ceil(header.fs_total_block_count / 8)) + header.checksum_size
+        )  # Header + Bitmap + Checksum
+        end_offset += used_blocks  # Data size
+        checksum_count = (
+            round(used_blocks / (header.fs_block_size * header.blocks_per_checksum))
+            * header.checksum_size
+        )
+        end_offset += (
+            checksum_count if checksum_count > 0 else header.checksum_size
+        )  # At least 1 checksum
```
qkaiser, line 81: I think this whole thing can be simplified. There are 4 main sections:

1. Header (partclone_header_t) – Fixed size.
2. Bitmap – Describes used blocks.
3. Checksums (optional) – For block verification.
4. Block Data – Actual used blocks of the filesystem.

So something like this:

```python
# header
end_offset = start_offset + len(header)

# feature area, since it's V2
end_offset += header.feature_size

# bitmap, as bytes
end_offset += ceil(header.fs_total_block_count / 8)

if header.checksum_mode != 0:
    checksum_blocks = ceil(header.fs_used_block_count_bitmap / header.blocks_per_checksum)
    end_offset += checksum_blocks * header.checksum_size

# Data
end_offset += header.fs_used_block_count_bitmap * header.fs_block_size
```

It may need some improvements but it's easier to follow.

package.nix
```diff
@@ -0,0 +1,131 @@
...
+  runtimeDeps = [
+    e2fsprogs-nofortify
+    jefferson
+    lziprecover
+    lzop
+    p7zip
+    sasquatch
+    sasquatch-v4be
+    ubi_reader
+    partclone
```
qkaiser, line 35: Ideally the whole file should be committed as part of the nix commit, but then the partclone addition at this line should be part of commit b21b6dc291583af6b7ec9b7c3d63ee8302328841. This way we separate the nix rework from the addition of partclone to it.

qkaiser (commented):
Good job :)

python/unblob/handlers/archive/partclone.py
```diff
@@ -0,0 +1,82 @@
...
+        uint16 checksum_size;
+        uint32 blocks_per_checksum;
+        uint8 reseed_checksum;
+        uint8 bitmap_mode;
+        uint32 crc32;
+    } partclone_header_t;
+"""
+
+HEADER_STRUCT = "partclone_header_t"
+BIG_ENDIAN = 0xC0DE
```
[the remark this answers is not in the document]
rxpha3l: I named it BIG_ENDIAN since get_endian() wants a big endian value. should I still change it ?

python/unblob/handlers/archive/partclone.py
rxpha3l: `header.feature_size` has a value (for the floppy test) of `18`. this will not end-up at the end. run the test and it has to be `header.checksum_size`. 

for `header.checksum_mode` should I not do `None` instead of `0`? The doc says so

python/unblob/handlers/archive/partclone.py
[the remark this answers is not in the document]
qkaiser: name it `BIG_ENDIAN_MAGIC`

python/unblob/handlers/archive/partclone.py
qkaiser: `None` and `NULL` do not exist in raw binary representation, they're just `0`. You can see it's a `uint16` from the header definition, so if checksum is disabled, `checksum_mode` is equal to `0`.

rxpha3l: ok, thanks for the clarification

qkaiser:
partclone is not available on OSX in nixpkg - https://github.com/onekey-sec/unblob/actions/runs/14124724723/job/39571274718?pr=1155

Would need to provide a patch for https://github.com/NixOS/nixpkgs/blob/e6057380b59bdf9a3693ca3ef45d67e6a95a0d66/pkgs/by-name/pa/partclone/package.nix.

Will look into it. It will take some time since I need to spin up some VMs.

rxpha3l:
I found a python build in lib that can return information about the system:
`import platform`
`platform.platform()` # Prints system info

Could we not match on that and set the test on `skip`  for OSX ? Maybe creating a sort of blacklist or adding an optional flag in the handler. If true, it will skip the test ?

tests/test_handlers.py
```diff
@@ -35,6 +35,13 @@ def test_all_handlers(
     extraction_config: ExtractionConfig,
     request: pytest.FixtureRequest,
 ):
+    handlers_by_name = {handler.NAME: handler for handler in extraction_config.handlers}
+    if (
+        input_dir.parent.name in handlers_by_name
+        and not handlers_by_name[input_dir.parent.name].is_supported_here()
+    ):
+        pytest.skip(f"Handler '{input_dir.parent.name}' not supported on this platform")
+
```
e3krisztian, line 44: We have a single testing exception, so for now I would add only the below condition, and skip all the other changes in this commit, until we know we will need some more general solution.

```py
    handler_name = input_dir.parent.name
    if (sys.platform, handler_name) == ('darwin', 'partclone'):
        pytest.skip(f"Handler '{handler_name}' not supported on platform '{sys.platform}'")
```

e3krisztian: This commit = 8a57c5b3c8a00f6be52a46049f0ab3af8ae8a4a6

package.nix
```diff
@@ -4,6 +4,7 @@
   makeWrapper,
   e2fsprogs-nofortify,
   erofs-utils,
+  fetchFromGitHub,
```
e3krisztian: I do not understand this commit b91550e8e1e8d230915740af945b0cb12d8bd7a4 - the message and the content are not in sync.

Maybe the main change was already committed earlier, and only this line change left?

package.nix
```diff
@@ -1,15 +1,16 @@
 {
   lib,
+  stdenv,
   python3,
   makeWrapper,
   e2fsprogs-nofortify,
   erofs-utils,
-  fetchFromGitHub,
```
e3krisztian: it was just added a commit before!?

I think the previous commit b91550e8e1e8d230915740af945b0cb12d8bd7a4 could be squashed into this one.

python/unblob/file_utils.py
```diff
@@ -355,17 +355,23 @@ def parse(
         return struct_parser(file)
 
 
-def get_endian(file: File, big_endian_magic: int) -> Endian:
+def get_endian(file: File, big_endian_magic: int, endian_len: int = 4) -> Endian:
```
e3krisztian, line 358: I would probably introduce a new `get_endian2` for the 2 byte magic case instead of complicating the existing one.

You can keep it, but please change the signature, as the code will not work with arbitrary `endian_len` - the only supported options are: 2 and 4:

```py
def get_endian(file: File, big_endian_magic: int, endian_len: Literal[2] | Literal[4] = 4) -> Endian:
```

python/unblob/file_utils.py
```diff
@@ -355,17 +355,23 @@ def parse(
...
     """Read a four bytes magic and derive endianness from it.
 
     It compares the read data with the big endian magic.  It reads
     four bytes and seeks back after that.
     """
```
e3krisztian, line 363: This is no longer true.

python/unblob/file_utils.py
qkaiser: @rxpha3l please take care of this

package.nix
qkaiser: rebase gone wrong, will fix once @rxpha3l takes care of the other points

package.nix
qkaiser: Fixed by dropping the commit. It had nothing to do here.

python/unblob/file_utils.py
qkaiser: Fixed. Also added test cases to `TestGetEndian` to cover both `endian_len` cases.

python/unblob/file_utils.py
qkaiser: Fixed.

python/unblob/file_utils.py
qkaiser: typing hints are wrong

```
/home/runner/work/unblob/unblob/tests/test_file_utils.py:364:62 - error: Argument of type "int" cannot be assigned to parameter "endian_len" of type "Literal[2, 4]" in function "get_endian"
    Type "int" is not assignable to type "Literal[2, 4]"
      "int" is not assignable to type "Literal[2]"
      "int" is not assignable to type "Literal[4]" (reportArgumentType)
  /home/runner/work/unblob/unblob/tests/test_file_utils.py:390:48 - error: Argument of type "int" cannot be assigned to parameter "endian_len" of type "Literal[2, 4]" in function "get_endian"
    Type "int" is not assignable to type "Literal[2, 4]"
      "int" is not assignable to type "Literal[2]"
      "int" is not assignable to type "Literal[4]" (reportArgumentType)
```

python/unblob/file_utils.py
```diff
@@ -355,17 +355,25 @@ def parse(
         return struct_parser(file)
 
 
-def get_endian(file: File, big_endian_magic: int) -> Endian:
-    """Read a four bytes magic and derive endianness from it.
+def get_endian(
+    file: File, big_endian_magic: int, endian_len: Literal[2] | Literal[4] = 4
```
e3krisztian, line 359: Failing the type check for earlier python versions, `Literal[2, 4]` should work, and is even shorter.

tests/test_file_utils.py
```diff
@@ -345,29 +345,49 @@ class TestGetEndian:
...
+            pytest.param(
+                b"\xff\x00", 0x00FF, Endian.LITTLE, id="valid_little_endian_2"
+            ),
+            pytest.param(b"\x10\x00", 0x1000, Endian.BIG, id="valid_big_endian_2"),
         ],
     )
     def test_get_endian(self, content: bytes, big_endian_magic: int, expected: Endian):
         file = File.from_bytes(content)
-        assert get_endian(file, big_endian_magic) == expected
+        assert get_endian(file, big_endian_magic, endian_len=len(content)) == expected
```
e3krisztian, line 364: This is failing the type check, as it is not known to the type checker if `len(content)` is surely one of `2` and `4`.

This test needs to be split into 2, to test, one of magic length 2, and for magic length 4.

(Which also shows that `get_endian()` should have a new version instead of being extended, keeping the original and defining a new one for 2 byte magic)

docs/formats.md
```diff
@@ -22,6 +22,7 @@ unblob supports more than 30 formats. You can see their code in
 | CAB     | ❌         | ❌                 | ❌        | [archive/cab.py][cab-handler]         | [`7z`][cab-extractor]       |
 | CPIO    | ✅         | ✅                 | ✅        | [archive/cpio.py][cpio-handler]       | unblob extractor            |
 | DMG     | ❌         | ❌                 | ❌        | [archive/dmg.py][dmg-handler]         | [`7z`][dmg-extractor]       |
+| PARTCLONE | ✅       | ❌                 | ❌        | [archive/partclone.py][partclone-hanlder] | [`partclone`][partclone-extractor] |
```
e3krisztian, line 25: typo in `partclone-hanlder` -> `partclone-handler` (as a result the reference was also broken).

e3krisztian (changes requested):

tests/test_file_utils.py
e3krisztian: What about leaving the original `get_endian()` as is, and introducing a new `get_endian_short()`?

I prefer 2 smaller function, than one bigger, which is also harder to use/understand.

qkaiser: implemented `get_endian_short` and adapted code and tests, should be easier to navigate now. Commit message has been adjusted too.

e3krisztian (approved):
Thanks for the work and revisions! :)

PR #1166, opened 2025-04-09
feat(handler): add par2 directory handler

rxpha3l:
Parchive or PAR2, is a format for creating redundant data that helps detect and repair corrupted files. These archives typically accompany split-file sets (like multi-volume RAR or ZIP archives). Each PAR2 file is composed of multiple “packets.”  The header starts with `PAR2\x00PKT`

Create PAR2 file:
> `par2create backup.par2 file_to_protect`
While creating PAR2 backup of a file, parchive creates in the **same** directory multiple .par2 file. Every file that ends with .volXX+YY.par2 specifies form which block (X), how many more blocks it contains (Y). 
E.g. `backup.vol01+02.par2` would contain the first block + 2 more blocks etc.

Verify PAR2 file:
> `par2verify backup.par2`

Header :
>  8 bytes : Magic
>  8 bytes : Packet length
> 16 bytes : MD5 hash
> 16 bytes : Recovery set ID
> 16 bytes : Recovery Type

Unblob identifies each PAR2 file in a given directory by validating the magic signature, computing the MD5 hash and validating it with the MD5 hash in the header. Since parchive is a repair tool, no extractor is specified.

[Sources]
https://docs.fileformat.com/compression/par/
https://parchive.github.io/
https://github.com/Parchive/par2cmdline/tree/master/src
https://parchive.github.io/doc/Parity%20Volume%20Set%20Specification%20v2.0.html

### DIFF
--- a/python/unblob/handlers/__init__.py
+++ b/python/unblob/handlers/__init__.py
@@ -6,6 +6,7 @@ from .archive import (
     cab,
     cpio,
     dmg,
+    par2,
     partclone,
     rar,
     sevenzip,
@@ -126,4 +127,5 @@ BUILTIN_HANDLERS: Handlers = (
 BUILTIN_DIR_HANDLERS: DirectoryHandlers = (
     sevenzip.MultiVolumeSevenZipHandler,
     gzip.MultiVolumeGzipHandler,
+    par2.MultiVolumePAR2Handler,
 )
--- a/docs/handlers.md
+++ b/docs/handlers.md
@@ -45,6 +45,7 @@
     | [`NETGEAR TRX V1`](#netgear-trx-v1) | ARCHIVE | :octicons-check-16: |
     | [`NETGEAR TRX V2`](#netgear-trx-v2) | ARCHIVE | :octicons-check-16: |
     | [`NTFS`](#ntfs) | FILESYSTEM | :octicons-check-16: |
+    | [`PAR2 (MULTI-VOLUME)`](#par2-multi-volume) | ARCHIVE | :octicons-check-16: |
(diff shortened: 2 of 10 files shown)


rxpha3l:
I have no idea why there is a `temp` commit.

qkaiser (commented):
Please get rid of temporary commits (can be done with `fixup`) and delete the erofs leftovers (can be done by rebasing on upstream/main).

rxpha3l:
> Please get rid of temporary commits (can be done with `fixup`) and delete the erofs leftovers (can be done by rebasing on upstream/main).

everything is now as it should be

python/unblob/handlers/__init__.py
```diff
@@ -54,6 +55,7 @@
 BUILTIN_HANDLERS: Handlers = (
     cramfs.CramFSHandler,
     extfs.EXTHandler,
+    erofs.EROFSHandler,
```
qkaiser, line 58: leftover

qkaiser:
Commit a14e76a4c74626b09714cc05f42eda1ec20505e1 broke some typing, see:

```
/home/runner/work/unblob/unblob/python/unblob/models.py:427:34 - error: "get_dependencies" is not a known attribute of "None" (reportOptionalMemberAccess)
  /home/runner/work/unblob/unblob/python/unblob/models.py:442:39 - error: Argument of type "list[Path]" cannot be assigned to parameter "inpath" of type "Path" in function "extract"
    "list[Path]" is not assignable to "Path" (reportArgumentType)
```

We can have a look at it sometime this week.

PR #1167, opened 2025-04-11
chore(deps): Downgrade 7zip to v16.02 in nix packaging

kukovecz:


qkaiser (commented):
LGTM. However, I would like the commit to describe what those problems are.

> Found some problems with 7z based extractions using the nix environment
which does not happen in debian.

Is the extraction different ? Did options changed ?

qkaiser (approved):

qkaiser:
> LGTM. However, I would like the commit to describe what those problems are.
> 
> > Found some problems with 7z based extractions using the nix environment
> > which does not happen in debian.
> 
> Is the extraction different ? Did options changed ?

Investigated and edited the commit message:

    7zip maintainers changed the behavior or MBR extraction between version
    16.02 and 17.05:
    
    - on 16.02, 7z creates a file per MBR partition in the extraction
      directory, regardless of the amount of partitions. So we get these
      files named ("0.img", "1.img") after their partition entry ID.
    - on 17.05, 7z will automagically extract the partition itself if there
      is a single partition within the MBR and that partition is a supported
      filesystem format (e.g. extfs). So instead of having a "0.img" file,
      you immediately get the content of that partition written to disk.
    
    Since one of the main objective of unblob is to identify known/unknown
    chunks and provide detailed meta-data about them, we consider important
    to have that intermediate step provided by 16.02. This way we know the
    MRB contains a partition of a specific type, which itself was handled by
    a dedicated unblob handler.
    
    This behavior is kind of invisible on unblob-core since the MBR handler
    is only available on the ONEKEY platform right now.

    7z 17.05 also has issues when extracting ZIP files, where it does not
    accept full path for the "-o" option. For example, "-o/tmp/lol" won't
    work, but "cd /tmp/lol; 7z x -p -y sample.zip -o." will. On top of that,
    7z seems to set arbitrary permissions on the output directory.

vlaci:
Also worth mentioning that Debian and RHEL are both sticking to 16.02 as of now.

PR #1189, opened 2025-05-08
feat(handler): add upx decompression to elf executable

rxpha3l:
UPX is a free, secure, portable, extendable, high-performance
executable packer for several executable formats. 

The UPX header is written after the elf header. The offset depends on how big the ELF is. Luckily UPX writes in the last 4 bytes of the compressed file the offet to the header. 

Create UPX file :
> `upx --best foo.elf -ofoo.upx`

Decompress UPX file: 
> `upx -d foo.upx -ofoo.decompressed`

Key notes:
- UPX uses a lot of alignments, which is why computing the UPX header after the ELF header can cause discrepancies
- The UPX header is called loader and is 12 bytes long, with `UPX!` as magic
- File format is little endian
- Supports 32 bit & 64 bit platforms
- UPX uses `stat` to get information about the compressed file, which is why the total compressed file size is no where to be found
- Checksum is calculated over the stubloader and uses adler32
- Loader size is **not** the actual size, since it is aligned. The alignment must be subtracted. It will be used to determine the data length used in the checksum function

Header format :
> 4 bytes : Checksum
> 4 bytes : Magic
> 2 bytes : Loader size
> 1 byte : Version
> 1 byte : Format

Using python's `lief` library,  it possible to get the offset of the last segment. Subtracting the loader size of it, returns the stub loader offset **after** alignment. Subtracting the alignment return the **correct** loader offset used for the checksum. Following the official UPX github repository, in `src/p_lx_elf.cpp` line `385` to `447` is where the alignment is done. Before calculating the total alignment, `xct_off` checks for shared android library in the section names of the elf header. If there is one present, this will change the total alignment afterwards.  To summarize, UPX aligns the loader size to be a multiple of 4, with additional steps in between. The alignment is the difference between the loader size in the header and the value after the alignment calculation. The checksum size the loader size minus the alignment.

[Sources]
https://bbs.kanxue.com/thread-248779.htm
https://github.com/upx/upx/blob/devel/src/stub/src/include/linux.h
https://github.com/upx/upx/blob/devel/src/p_lx_elf.cpp
https://www.programmersought.com/article/58884806733/

### DIFF
--- a/python/unblob/handlers/executable/elf.py
+++ b/python/unblob/handlers/executable/elf.py
@@ -1,5 +1,6 @@
 import io
 import shutil
+import zlib
 from pathlib import Path
 from typing import Optional
 
@@ -8,13 +9,18 @@ import lief
 from structlog import get_logger
 
 from unblob.extractor import carve_chunk_to_file
+from unblob.extractors import Command
 from unblob.file_utils import (
     Endian,
     File,
+    InvalidInputFormat,
+    StructParser,
     convert_int8,
     convert_int32,
     convert_int64,
+    iterate_file,
     iterate_patterns,
+    read_until_past,
     round_up,
 )
 from unblob.models import (
@@ -36,8 +42,77 @@ KERNEL_MODULE_SIGNATURE_FOOTER = b"~Module signature appended~\n"
 KERNEL_INIT_DATA_SECTION = ".init.data"
 
 
+# [Ref] https://github.com/upx/upx/blob/devel/src/stub/src/include/linux.h
+UPX_C_DEFINITIONS = r"""
+typedef struct packhead{
+    char magic[4];
+    uint8_t version;
+    uint8_t format;
+    uint8_t level;
+    uint8_t method;
+    uint64_t unknown1;
+    uint32_t u_filesize1;
+    uint32_t c_filesize;
+    uint32_t u_filesize2;
+    uint32_t unknown2;
+    uint32_t l_info_offset;
+} packhead_t;
+
+typedef struct l_info{
+    uint32_t l_checksum;
+    char l_magic[4];
+    uint16_t l_lsize;
+    uint8_t l_version;
+    uint8_t l_format;
+} l_info_t;
+"""
+upx_parser = StructParser(UPX_C_DEFINITIONS)
+
+
+def parse_upx_packhead(file: File):
+    return upx_parser.parse("packhead_t", file, Endian.LITTLE)
+
+
+def parse_upx_l_info(file: File):
+    return upx_parser.parse("l_info_t", file, Endian.LITTLE)
+
+
 @attrs.define(repr=False)
 class ElfChunk(ValidChunk):
+    @staticmethod
+    def upx_checksum_validates(file: File, l_info, elf) -> bool:
+        size_pack2 = elf.last_offset_segment - l_info.l_lsize
+        size_aligment = round_up(size_pack2, 4)  # Forces to be mod 4
+        xct_off = any(section.name == "init" for section in elf.sections)
+        size_aligment += (4 & size_aligment) ^ (int(bool(xct_off)) << 2)  # 4 or 0
+        size_aligment += 8  # Added 2 times 4 byte (size of disp)
+        if xct_off:
+            size_aligment += 12
+        alignment = size_aligment - size_pack2
+        checksum_offset = elf.last_offset_segment - (l_info.l_lsize - alignment)
+        file.seek(checksum_offset, io.SEEK_SET)
+        adler32_checksum = 1
+        for chunk in iterate_file(file, checksum_offset, l_info.l_lsize - alignment):
+            adler32_checksum = zlib.adler32(chunk, adler32_checksum)
+        return adler32_checksum == l_info.l_checksum
+
+    def is_valid_upx(self, inpath: Path, elf) -> bool:
+        file = File.from_path(inpath)
+        file.seek(-4, io.SEEK_END)  # last 4 bytes indicates where linfo ends
+        l_info_start_offset = abs(
+            convert_int32(file.read(4), Endian.LITTLE)
+            - upx_parser.cparser_le.l_info_t.size
+        )
+        if l_info_start_offset > file.size():
+            return False
+        file.seek(l_info_start_offset, io.SEEK_SET)
+        upx_header = parse_upx_l_info(file)
+        if upx_header.l_magic != b"UPX!":  # Magic
+            return False
+        if not self.upx_checksum_validates(file, upx_header, elf):
+            raise InvalidInputFormat("Invalid UPX checksum")
+        return True
+
     def extract(self, inpath: Path, outdir: Path):
         # ELF file extraction is special in that in the general case no new files are extracted, thus
         # when we want to clean up all carves to save place, carved ELF files would be deleted as well,
@@ -57,6 +132,10 @@ class ElfChunk(ValidChunk):
         if is_kernel:
             with File.from_path(inpath) as file:
                 extract_initramfs(elf, file, outdir)
+
+        elif self.is_valid_upx(inpath=inpath, elf=elf):
+            extract_upx(inpath, outdir)
+
         elif not self.is_whole_file:
             # make a copy, and let the carved chunk be deleted
             outdir.mkdir(parents=True, exist_ok=False)
@@ -67,6 +146,12 @@ class ElfChunk(ValidChunk):
             # Even though the second chunk search one is short-circuited,
             # because the ELF handler will recognize it as a whole file
             # other handlers might burn some cycles on the file as well.
+
+
+def extract_upx(inpath: Path, outdir: Path):
+    extractor = Command("upx", "-d", "{inpath}", "-o{outdir}/{inpath.stem}.elf")
+    outdir.mkdir(parents=True, exist_ok=False)
+    extractor.extract(inpath, outdir)
 
 
 def extract_initramfs(elf, file: File, outdir):
@@ -257,6 +342,35 @@ class _ELFBase(StructHandler):
 
         return end_offset
 
+    def is_upx(self, file: File, start_offset: int, end_offset: int) -> bool:
+        """Check if UPX magic is present after ELF header."""
+        return b"UPX!" in file[start_offset : min(end_offset, start_offset + 4096)]
+
+    def get_upx_end_offset(self, file: File, start_offset: int, end_offset: int) -> int:
+        """Locate UPX footer in ELF file and returns UPX end offset or original end offset."""
+        upx_footer = b"\xff\x00\x00\x00\x00UPX!\x00\x00\x00\x00"
+        for packhead_offset in iterate_patterns(file=file, pattern=upx_footer):
+            file.seek(
+                packhead_offset + len(upx_footer), io.SEEK_SET
+            )  # seek to end of footer
+            file.seek(
+                read_until_past(file=file, pattern=b"\x00")
+            )  # sometimes more NULL bytes are added
+            packheader = parse_upx_packhead(file)
+            file_size_compressed = packheader.c_filesize + packheader.size
+            packhead_is_valid = (
+                (
+                    packheader.magic == b"UPX!"
+                    and packheader.u_filesize1 == packheader.u_filesize2
+                )
+                and (file_size_compressed == file.tell() - start_offset)
+                and (file_size_compressed % 4 == 0)
+            )
+            if packhead_is_valid:
+                return start_offset + file_size_compressed
+        # no matching UPX footer found
+        return end_offset
+
     def calculate_chunk(self, file: File, start_offset: int) -> Optional[ElfChunk]:
         endian = self.get_endianness(file, start_offset)
         file.seek(start_offset, io.SEEK_SET)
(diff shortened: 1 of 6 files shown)


python/unblob/handlers/executable/elf.py
```diff
@@ -62,6 +113,11 @@ def extract(self, inpath: Path, outdir: Path):
             # other handlers might burn some cycles on the file as well.
 
 
+def extract_upx(inpath: Path, outdir: Path):
+    extractor = Command("upx", "-d", "{inpath}", "-o{outdir}/upx.uncompressed")
```
qkaiser, line 117: I don't see any `upx.uncompressed` in the integration test output so something must be off.

python/unblob/handlers/executable/elf.py
```diff
@@ -31,6 +36,48 @@
 
 @attrs.define(repr=False)
 class ElfChunk(ValidChunk):
+    def upx_get_loader(self, file: File) -> int:
+        file.seek(-4, io.SEEK_END)  # last 4 bytes indicates where linfo ends
+        return convert_int32(file.read(4), Endian.LITTLE) - 12
```
qkaiser, line 41: No need for a dedicated function. Can be kept in `is_valid_upx`.

python/unblob/handlers/executable/elf.py
```diff
@@ -31,6 +36,48 @@
...
+        alignment = size_aligment - size_pack2
+        checksum_offset = elf.last_offset_segment - (l_lsize - alignment)
+        file.seek(checksum_offset, io.SEEK_SET)
+        return zlib.adler32(bytearray(file.read(l_lsize - alignment)), 1) == l_checksum
+
+    def is_valid_upx(self, inpath: Path, elf) -> bool:
+        file = File.from_path(inpath)
+        try:
+            upx_loader = self.upx_get_loader(file)
+            file.seek(upx_loader + 4, io.SEEK_SET)  # first 4 bytes is checksum
```
qkaiser, line 72: Instead of getting the beginning of `l_info` by subtracting 12 from the end of `l_info`, then adding 4 to get to the magic value, I'd recommend you get the `l_info_start_offset`, seek there and parse the whole `l_info` using `StructParser`. This will make everything easier to understand.

For example:

```python
file.seek(-4, io.SEEK_END)  # last 4 bytes indicates where linfo ends
l_info_end_offset = convert_int32(file.read(4), Endian.LITTLE)
l_info_start_offset = l_info_end_offset - 12

parser = StructParser(C_DEFINITIONS)
l_info = parser.parse("l_info_t", file, endian=Endian.LITTLE)
# here you can do your work by using the l_info object rather than seeking at arbitrary locations
```

python/unblob/handlers/executable/elf.py
```diff
@@ -31,6 +36,48 @@
...
 class ElfChunk(ValidChunk):
+    def upx_get_loader(self, file: File) -> int:
+        file.seek(-4, io.SEEK_END)  # last 4 bytes indicates where linfo ends
+        return convert_int32(file.read(4), Endian.LITTLE) - 12
+
+    def upx_calc_checksum(self, file: File, l_info: int, elf) -> bool:
+        file.seek(l_info, io.SEEK_SET)
+        l_checksum = convert_int32(file.read(4), Endian.LITTLE)
+        file.seek(l_info + 8, io.SEEK_SET)
+        l_lsize = convert_int16(file.read(2), Endian.LITTLE)
```
qkaiser, line 47: All of these values should come from the `l_info` header parsed with a `StructParser`.

python/unblob/handlers/executable/elf.py
```diff
@@ -31,6 +36,48 @@
...
+        size_pack2 = elf.last_offset_segment - l_lsize
+        size_aligment = size_pack2  # Calc size before alignment
+        while size_aligment % 4 != 0:  # Forces to be mod 4 because of alignment
+            size_aligment += 1
+        xct_off = False  # For shared android library -> .so
+        for section in elf.sections:
+            if section.name == ".init":
+                xct_off = True
+                break
```
qkaiser, line 56: I don't understand the comment "For shared android library -> .so", what does that mean ?

Can be simplified to:

```suggestion
        xct_off = any(section.name == "init" for section in elf.sections)
```

python/unblob/handlers/executable/elf.py
```diff
@@ -31,6 +36,48 @@
...
+        size_aligment += (
+            0 if xct_off else (4 & size_aligment) ^ (int(bool(xct_off)) << 2)
+        )  # 0 or 4
+        size_aligment += 8  # Added 2 times 4 byte (size of disp)
+        if xct_off:
+            size_aligment += 12
+        alignment = size_aligment - size_pack2
```
qkaiser, line 63: What is the purpose of this code ? What does it do ?

python/unblob/handlers/executable/elf.py
```diff
@@ -31,6 +36,48 @@
...
+        checksum_offset = elf.last_offset_segment - (l_lsize - alignment)
+        file.seek(checksum_offset, io.SEEK_SET)
+        return zlib.adler32(bytearray(file.read(l_lsize - alignment)), 1) == l_checksum
```
qkaiser, line 66: You don't need to convert to `bytearray`, the function accepts bytes. Also if `l_size` is large, I would iterate over the file in chunks. `adler32` supports rolling checksums by setting the initial value (first time it's 1 but then you feed it the output of your first calculation).

python/unblob/handlers/executable/elf.py
```diff
@@ -31,6 +36,48 @@
...
+
+    def upx_calc_checksum(self, file: File, l_info: int, elf) -> bool:
+        file.seek(l_info, io.SEEK_SET)
+        l_checksum = convert_int32(file.read(4), Endian.LITTLE)
+        file.seek(l_info + 8, io.SEEK_SET)
+        l_lsize = convert_int16(file.read(2), Endian.LITTLE)
+        size_pack2 = elf.last_offset_segment - l_lsize
+        size_aligment = size_pack2  # Calc size before alignment
+        while size_aligment % 4 != 0:  # Forces to be mod 4 because of alignment
+            size_aligment += 1
```
qkaiser, line 51: You calculate alignment, or what I suppose is padding at two separate locations. Here and starting on line 57, these should be merged and simplified. If you need a 4 byte alignment there is the `round_up` function in unblob.

qkaiser (changes requested):

python/unblob/handlers/executable/elf.py
rxpha3l: I directly subtracted the 12 bytes since I will never use the variable `l_info_end_offset` 

`l_info_start_offset = convert_int32(file.read(4), Endian.LITTLE) - 12`

python/unblob/handlers/executable/elf.py
rxpha3l: Now simplified :

`
1) xct_off = any(section.name == "init" for section in elf.sections)
2) size_aligment += (4 & size_aligment) ^ (int(bool(xct_off)) << 2) # 4 or 0
3) size_aligment += 8 # Added 2 times 4 byte (size of disp)
4) if xct_off:
5)     size_aligment += 12
`

1) checks for a shared library in the header
2) If **no** shared library is found, add 4 bytes to the aligned size, else 0
3) 8 bytes are added because in the source code a 4 bytes variable called `disp`  is added 2 times
4) If a shared library is present, 3 times the size of `disp` is added (3*4 bytes)

Here is the source code, line 390 -> 430
https://github.com/upx/upx/blob/devel/src/p_lx_elf.cpp

python/unblob/handlers/executable/elf.py
rxpha3l: This one serves as base for the alignment:
`size_aligment = round_up(size_pack2, 4)# Forces to be mod 4`
After this the size of a variable called `disp` is added multiple times on top of it.

python/unblob/handlers/executable/elf.py
rxpha3l: unblob carves not the entire upx file. a custom end offset calculation for upx files must be made since it is a special case. @e3krisztian helped me with this finding.

qkaiser: This does not make sense. The way the integration file `__output__` directory looks like right now means the `extract_upx` path is not taken but rather the path from lines 55 and 56.

python/unblob/handlers/executable/elf.py
qkaiser: Marked as resolved but still there.

python/unblob/handlers/executable/elf.py
qkaiser: Marked as resolved but still there.

python/unblob/handlers/executable/elf.py
qkaiser: > After this the size of a variable called disp is added multiple times on top of it.

Why ?

python/unblob/handlers/executable/elf.py
qkaiser: Marked as resolved but still there.

python/unblob/handlers/executable/elf.py
```diff
@@ -62,6 +139,12 @@ def extract(self, inpath: Path, outdir: Path):
             # other handlers might burn some cycles on the file as well.
 
 
+def extract_upx(inpath: Path, outdir: Path):
+    extractor = Command("upx", "-d", "{inpath}", "-o{outdir}/upx-decompressed.elf")
```
qkaiser, line 143: I think this would be nicer:

```suggestion
    extractor = Command("upx", "-d", "{inpath}", "-o{outdir}/{inpath.stem}.elf")
```

python/unblob/handlers/executable/elf.py
```diff
@@ -29,8 +36,74 @@
...
+        adler32_checksum = 1
+        for chunk in iterate_file(file, checksum_offset, l_info.l_lsize - alignment):
+            adler32_checksum = zlib.adler32(chunk, adler32_checksum)
+        return adler32_checksum == l_info.l_checksum
+
+    def is_valid_upx(self, inpath: Path, elf) -> bool:
+        file = File.from_path(inpath)
+        try:
+            file.seek(-4, io.SEEK_END)  # last 4 bytes indicates where linfo ends
+            l_info_start_offset = convert_int32(file.read(4), Endian.LITTLE) - 12
```
qkaiser, line 96: You can get the size of the struct this way, so we know where this is coming from:

```suggestion
            l_info_start_offset = convert_int32(file.read(4), Endian.LITTLE) - upx_parser.cparser_le.l_info_t.size
```

python/unblob/handlers/executable/elf.py
```diff
@@ -29,8 +36,74 @@
...
+            file.seek(l_info_start_offset, io.SEEK_SET)
+            upx_header = parse_upx_l_info(file)
+            if upx_header.l_magic != b"UPX!":  # Magic
+                return False
+            if not self.upx_checksum_validates(file, upx_header, elf):
+                raise InvalidInputFormat("Invalid UPX checksum")
+            return True  # noqa: TRY300
+        except SeekError:
+            return False
```
qkaiser, line 105: You can check the offsets rather than catching the `SeekError`:

```suggestion
        file.seek(-4, io.SEEK_END)  # last 4 bytes indicates where linfo ends
        l_info_start_offset = convert_int32(file.read(4), Endian.LITTLE) - 12
        if l_info_start_offset > file.size():
            return False
        file.seek(l_info_start_offset, io.SEEK_SET)
        upx_header = parse_upx_l_info(file)
        if upx_header.l_magic != b"UPX!":  # Magic
            return False
        if not self.upx_checksum_validates(file, upx_header, elf):
            raise InvalidInputFormat("Invalid UPX checksum")
        return True
```

python/unblob/handlers/executable/elf.py
```diff
@@ -262,6 +375,9 @@ def calculate_chunk(self, file: File, start_offset: int) -> Optional[ElfChunk]:
         if header.e_type == lief.ELF.Header.FILE_TYPE.REL.value:
             end_offset = self.get_signed_kernel_module_end_offset(file, end_offset)
 
+        if self.is_upx(file, start_offset, end_offset):
```
qkaiser, line 378: You have two functions that checks if it's a UPX file:
- `_ELFBase.is_upx`
- `ElfChunk.is_valid_upx`

I think it would make sense to only use one of them. I would keep `is_valid_upx` and move it outside the class. Use it in both handler and extractor. This way if the UPX header is wrong, we catch it as early as possible, rather than carving an invalid UPX and bailing out at extraction time

qkaiser (approved):

qkaiser:
@rxpha3l you forgot to re-generate the integration test files, they're still on the old naming convention:

```
Only in /tmp/pytest-of-runner/pytest-0/test_all_handlers_executable_e0/hello.upx_extract: hello.elf
Only in /home/runner/work/unblob/unblob/tests/integration/executable/elf/elf64/__output__/hello.upx_extract: upx-decompressed.elf
Only in /tmp/pytest-of-runner/pytest-0/test_all_handlers_executable_e0/hello_notes.upx_extract: hello_notes.elf
Only in /home/runner/work/unblob/unblob/tests/integration/executable/elf/elf64/__output__/hello_notes.upx_extract: upx-decompressed.elf
```

rxpha3l:
Integration tests are now up to date

qkaiser (approved):

PR #1192, opened 2025-05-13
Improve handlers documentation

qkaiser:
Further improvements to the handlers documentation.


python/unblob/handlers/archive/arc.py
```diff
@@ -51,8 +51,8 @@ class ARCHandler(StructHandler):
     EXTRACTOR = Command("unar", "-no-directory", "-o", "{outdir}", "{inpath}")
 
     DOC = HandlerDoc(
-        name=NAME,
-        description="Handles ARC archive files, which are legacy archive formats used to store multiple files with metadata such as file size, creation date, and CRC.",
+        name="ARC",
+        description="ARC is legacy archive format used to store multiple files with metadata such as file size, creation date, and CRC.",
```
rasakozi, line 55: missing 'a'

rasakozi:
Some further improvement ideas for the extraction:

1. Some links don't work in the table because Markdown has some special rules with brackets when it comes to linking the headings.

Basically all we need to do is leave the brackets out, so for example the link for `CPIO (BINARY)` should be `[`CPIO (BINARY)`](#cpio-binary) ` as opposed to [`CPIO (BINARY)`](#cpio-(binary)). If you open the `handlers.md` file, you can actually check the correct formatting by simply deleting everything in the second part of the link - `#cpio-binary` in the case of the previously mentioned example - and retyping it starting from the `#`. VS code will come up with an auto suggestion for the link.

Another similar problem is headers with hyphens like `SQUASHFS (V3 - BROADCOM)`, because they also don't gel with the links. To not make our job harder than necessary, I'd simply remove the spaces: `SQUASHFS (V3-BROADCOM)`.

2. There are some empty lines between the admonitions, sometimes 3, sometimes 4, sometimes only 1, I don't think these can cause any trouble but it's a thing that caught my eye :D.

3. Since the table takes up a lot of space, maybe it could be included in an admonition that's closed by default:

image

image

I have some CSS to center the table, but for some reason I couldn't get it to work in this documentation :(

rasakozi:
Looks good to me! I think I'll play around with the custom CSS once this is released. I don't think that's a blocker

rasakozi (approved):

PR #1204, opened 2025-06-04
Improve doc generator

qkaiser:
Instead of having a custom made `justfile` to generate unblob's handlers markdown documentation, we want this capability exposed by unblob core so that it can be used by unblob power users who have their own collection of handlers.

The `--build-handlers-doc` command line option takes a path as parameter and will write the auto-generated handlers documentation to that location. All the unblob handlers are loaded transparently, including the ones loaded through the plugin manager.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -70,13 +70,9 @@ jobs:
       - name: Setup 3rd party dependencies
         uses: ./.github/actions/setup-dependencies
 
-      - name: Setup just
-        uses: extractions/setup-just@v3
-
       - name: Check - Documentation
         run: |
-          uv sync --group docs
-          uv run just doc generate-handlers-doc
+          uv run unblob --build-handlers-doc docs/handlers.md
           git diff --exit-code docs/handlers.md
 
   run_python_tests:


docs/handlers.md
```diff
@@ -1,52 +1,51 @@
...
     | [`AR`](#ar) | ARCHIVE | :octicons-check-16: |
     | [`ARC`](#arc) | ARCHIVE | :octicons-check-16: |
     | [`ARJ`](#arj) | ARCHIVE | :octicons-check-16: |
-    | [`HP BDL`](#hp-bdl) | ARCHIVE | :octicons-check-16: |
-    | [`INSTAR BNEG`](#instar-bneg) | ARCHIVE | :octicons-check-16: |
-    | [`BZIP2`](#bzip2) | COMPRESSION | :octicons-check-16: |
+    | [`ANDROID EROFS`](#android-erofs) | FILESYSTEM | :octicons-check-16: |
+    | [`ANDROID SPARSE`](#android-sparse) | FILESYSTEM | :octicons-check-16: |
+    | [`AUTEL ECC`](#autel-ecc) | ARCHIVE | :octicons-check-16: |
     | [`CAB`](#cab) | ARCHIVE | :octicons-check-16: |
```
e3krisztian: BZIP2 moved, and also there are other changes later, that would not appear if sorting would be stable.

e3krisztian (changes requested):
The CI fails as well - code needs reformatting to pass the ruff check.

docs/handlers.md
qkaiser: I decided to change sorting from the handler name to the handler's doc name, it provides a better representation. That's why it moved around.

Fixed the code to make ruff happy and added pymdown-extensions to `package.nix` to fix the Nix build.

docs/handlers.md
```diff
@@ -59,17 +58,36 @@
...
+    | [`XIAOMI HDR1`](#xiaomi-hdr1) | ARCHIVE | :octicons-check-16: |
+    | [`XIAOMI HDR2`](#xiaomi-hdr2) | ARCHIVE | :octicons-check-16: |
     | [`YAFFS`](#yaffs) | FILESYSTEM | :octicons-check-16: |
     | [`ZIP`](#zip) | ARCHIVE | :octicons-alert-fill-12: |
-    | [`ZLIB`](#zlib) | COMPRESSION | :octicons-check-16: |
     | [`ZSTD`](#zstd) | COMPRESSION | :octicons-check-16: |
+    | [`BZIP2`](#bzip2) | COMPRESSION | :octicons-check-16: |
+    | [`COMPRESS`](#compress) | COMPRESSION | :octicons-check-16: |
+    | [`MULTI-SEVENZIP`](#multi-sevenzip) | ARCHIVE | :octicons-check-16: |
+    | [`ZLIB`](#zlib) | COMPRESSION | :octicons-check-16: |
```
e3krisztian: I still think this is not the order we want to have, also LZIP above comes after LZO, which is surprising to me.

Looking into the names, some of the `Handler`s are defined with `DOC(name=NAME,..)`, some have explicitly set to the same name, or some variation in capitalization:
- `CPIO (binary)`
- `LZO`
- `Lzip` (lowercase `z`, so it comes later, than all the `LZ*`)
- `XZ`
- `ZSTD`
- `bzip2`
- `compress`
- `multi-sevenzip`
- `zlib` (from `NAME`)

I think if we show a capitalized name in the table, we should also sort by the capitalized name, otherwise we would have surprising inconsistencies like above.

python/unblob/cli.py
```diff
@@ -87,6 +89,57 @@ def show_external_dependencies(
...
+    extra_dir_handlers = plugin_manager.load_dir_handlers_from_plugins()
+    dir_handlers = ctx.params["dir_handlers"] + tuple(extra_dir_handlers)
+
+    all_handlers = handlers + dir_handlers
+
+    slugifier = slugs.slugify(case="lower", percent_encode=True)
+    format_table_headers = """    | Format        | Type                                 | Fully supported?    |\n    | :------------ | :----------------------------------- | :-----------------: |\n"""
+
+    sorted_handlers = sorted(
+        all_handlers, key=lambda handler_class: handler_class.DOC.name
```
e3krisztian: The sorting can be fixed by appending `.upper()` here.

qkaiser: Yep. Done.

docs/handlers.md
qkaiser: Using uppercased names for sorting now.

e3krisztian (approved):
Perfect!

Issue #1216, opened 2025-06-27
landlock sandbox: allow access to /tmp


qkaiser:
Some handlers may require access to `/tmp` in order to work in temporary files (e.g. handler performing intermediate decryption or decompression).

Right now this is blocked by our landlock policy.

Ideally, we would add a `tmp_dir` to `ExtractionConfig`. This temporary directory would be created when unblob is launched and safely deleted at the end of the extraction run.

This way, we can extend the policy this way:

```diff
diff --git a/python/unblob/sandbox.py b/python/unblob/sandbox.py
index 61b02b0..689800f 100644
--- a/python/unblob/sandbox.py
+++ b/python/unblob/sandbox.py
@@ -49,6 +49,12 @@ class Sandbox:
             AccessFS.read("/"),
             # Multiprocessing
             AccessFS.read_write("/dev/shm"),  # noqa: S108
+            # Temporary directory
+            AccessFS.read_write(config.tmp_dir),
+            AccessFS.remove_dir(config.tmp_dir),
+            AccessFS.remove_file(config.tmp_dir),
+            AccessFS.make_dir(config.tmp_dir.parent),
+            AccessFS.remove_dir(config.tmp_dir.parent),
             # Extracted contents
             AccessFS.read_write(config.extract_root),
             AccessFS.remove_dir(config.extract_root),
```

One open question is how we could expose that temporary directory to handlers. If it's not possible without introducing breaking changes, we can also choose to work within `/tmp`.

vlaci:
We probably can create and remove the tmp directory outside the sandbox, reducing the number of rules needed.

Probably we should set related environment variables to this directory, so traditional `mktemp` and similar calls work as is. It looks like, there is no one standard variable for this. For example, nix sets the following environment variables for build/shell usage:

- `TMP`
- `TMPDIR`
- `TEMP`
- `TEMPDIR`

martonilles:
Agree with @vlaci I would create a temp director for unblob to operate on and handle the cleanup in unblob as well. All handlers&co can use temp files or directories within the unblob temp file, which should be accessibale with sandbox rules

PR #1222, opened 2025-07-01
feat(processing): add unblob dedicated temporary directory handling

qkaiser:
Draft implementation for #1216

It fails on `test_archive_success` because we cannot instantiate `ExtractionConfig` with an arbitrary tmp directory right now.

Some unit testing is also required for proper coverage.

### DIFF
--- a/python/unblob/processing.py
+++ b/python/unblob/processing.py
@@ -1,4 +1,3 @@
-import atexit
 import multiprocessing
 import os
 import shutil
@@ -104,21 +103,7 @@ class ExtractionConfig:
     dir_handlers: DirectoryHandlers = BUILTIN_DIR_HANDLERS
     verbose: int = 1
     progress_reporter: type[ProgressReporter] = NullProgressReporter
-    tmp_dir: Path = attrs.field(
-        factory=lambda: Path(tempfile.mkdtemp(prefix="unblob-tmp-"))
-    )
-
-    def __attrs_post_init__(self):
-        atexit.register(self._cleanup_tmp_dir)
-
-    def _cleanup_tmp_dir(self):
-        if isinstance(self.tmp_dir, Path) and self.tmp_dir.exists():
-            try:
-                shutil.rmtree(self.tmp_dir)
-            except Exception as e:
-                logger.warning(
-                    "Failed to clean up tmp_dir", tmp_dir=self.tmp_dir, exc_info=e
-                )
+    tmp_dir: Path = attrs.field(factory=lambda: Path(tempfile.gettempdir()))
 
     def _get_output_path(self, path: Path) -> Path:
         """Return path under extract root."""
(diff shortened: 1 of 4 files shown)


python/unblob/processing.py
```diff
@@ -100,6 +103,21 @@ class ExtractionConfig:
     dir_handlers: DirectoryHandlers = BUILTIN_DIR_HANDLERS
     verbose: int = 1
     progress_reporter: type[ProgressReporter] = NullProgressReporter
+    tmp_dir: Path = attrs.field(factory=lambda: Path(tempfile.mkdtemp(prefix="unblob-tmp-")))
+
+    def __attrs_post_init__(self):
+        """Set environment variables so all subprocesses and handlers use our temp dir"""
+        for var in ("TMP", "TMPDIR", "TEMP", "TEMPDIR"):
+            os.environ[var] = self.tmp_dir.as_posix()
+        atexit.register(self._cleanup_tmp_dir)
```
vlaci: This is problematic for programs using unblob as library, as these variables remain set in the hosting process's environment. We should set and reset them in a tighter scope.

For example, `Processor.process_task` looks like a fine candidate, as it by default will spawn child processes, not poisoning the caller's environment at all (given they use `process_num > 1`).

Another possibility is doing this at `process_file` level, but it still affects the hosting process unfortunately, but at least we can restore the variables upon returning from the function.

qkaiser: Done. Also implemented tests in `test_necessary_resources_can_be_created_in_sandbox`

tests/test_cli.py
```diff
@@ -1,5 +1,6 @@
 from collections.abc import Iterable
 from pathlib import Path
+import tempfile
```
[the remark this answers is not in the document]
qkaiser: Fixed

python/unblob/processing.py
e3krisztian: I do not see the original comment being resolved in the code - below I can find the env variables being set, but not restored, and here `atexit.register` still being used.

Have you pushed your changes?

I am also not sure how it would landlock to a `/tmp/call-specific-tempdir` if used as a library on the second call with another `ExtractionConfig`.

One solution I can think of is `process_file` forking in case of landlock, setting the environment variables in the fork, and cleaning up the temporary directory when the fork exits. Since the environment variables would be set in the child, they need no restoring.

(actually multiprocessing should be used somehow, as https://docs.python.org/3/library/os.html#os.fork has many problems and known not to work on macos)

python/unblob/processing.py
```diff
@@ -244,11 +262,17 @@ def __init__(self, config: ExtractionConfig):
     def process_task(self, task: Task) -> TaskResult:
         result = TaskResult(task)
         try:
+            self._set_tmp_dir()
             self._process_task(result, task)
```
e3krisztian:
```py
with temp_dir(self._config.tmp_dir.as_posix()):
```

Where `temp_dir` is:
```py
@contextmanager
def temp_dir(path):
    """Context manager that temporarily sets all temp directory env vars."""
    temp_vars = ("TMP", "TMPDIR", "TEMP", "TEMPDIR")
    saved = {}
    try:
        for var in temp_vars:
            saved[var] = os.environ.get(var)
            os.environ[var] = str(path)
        yield
    finally:
        for var, original in saved.items():
            if original is None:
                os.environ.pop(var, None)
            else:
                os.environ[var] = original
```

qkaiser: Done. Using `tmp_` rather than `temp_` to match with `tmp_dir`.

python/unblob/processing.py
qkaiser: @e3krisztian do I still need to do something here ?

e3krisztian: This implicit atexit-cleanup pattern still bothers me, and potentially anger provoking when the directory is unintentionally removed.

I would prefer a breaking change, except that is ruled out.

There is another solution: we could make the default '/tmp' (or the first value set for any of the TMP... variables), which should work backward compatibly. The only problem left is the cleanup, which would attempt to remove '/tmp' now.

For cleanup we could introduce another temp-dir for sub-processes:
- make a default for `config.tmp_dir` based on TMP... variables with fallback to `/tmp` 
- use another sub-directory (created with `mkdtemp(dir=config.tmp_dir)`) of `config.tmp_dir` as temporary directory for running tasks, immediately cleaning up the sub-directory when exiting the task (we could add both the `mkdtemp` and the cleanup to the `tmp_dir` context manager)
- do not clean up `config.tmp_dir` here: it is pre-existing, should be cleaned up at caller site if set explicitly (e.g. in `unblob.cli.cli`)

e3krisztian: @vlaci raised a concern, that we should consider doing this only in a more limited context inside `Command` extractors, as only external commands use tempfiles.

PR #1237, opened 2025-08-08
docs: Updated the uv command for installing Python dependencies.

n-ll1:
The `uv install` command does not exist. Replacing with the appropriate command.


qkaiser:
@n-ll1 makes sense, thanks. Can you rebase so that there is a single commit ?

n-ll1:
I struggled with Git, but I think I finally got what it should be. Sorry about that!

qkaiser (approved):

PR #1240, opened 2025-08-13
fix(doc): only generate documentation for handler with a valid DOC attribute

qkaiser:
Some unblob developers may not want to expose some of their handlers through the auto-generated documentation. To do that, they simply do not set the DOC attribute.

### DIFF
--- a/python/unblob/cli.py
+++ b/python/unblob/cli.py
@@ -106,7 +106,10 @@ def build_handlers_doc(
     extra_dir_handlers = plugin_manager.load_dir_handlers_from_plugins()
     dir_handlers = ctx.params["dir_handlers"] + tuple(extra_dir_handlers)
 
-    all_handlers = handlers + dir_handlers
+    # we only want handler with a valid DOC entry
+    all_handlers = [
+        handler for handler in handlers + dir_handlers if handler.DOC is not None
+    ]
 
     slugifier = slugs.slugify(case="lower", percent_encode=True)
     format_table_headers = """    | Format        | Type                                 | Fully supported?    |\n    | :------------ | :----------------------------------- | :-----------------: |\n"""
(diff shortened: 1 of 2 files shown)


qkaiser:
@e3krisztian not sure whether we should check whether `DOC` is set, or whether we should enforce the presence of `DOC` and check if it's `None`.

e3krisztian:
> @e3krisztian not sure whether we should check whether `DOC` is set, or whether we should enforce the presence of `DOC` and check if it's `None`.

I would prefer to keep `DOC` mandatory, as that would help catch accidental omission of documentation early.
Requiring setting it to `None` would also document the explicit intent to exclude the handler from documentation.

qkaiser:
> > @e3krisztian not sure whether we should check whether `DOC` is set, or whether we should enforce the presence of `DOC` and check if it's `None`.
> 
> I would prefer to keep `DOC` mandatory, as that would help catch accidental omission of documentation early. Requiring setting it to `None` would also document the explicit intent to exclude the handler from documentation.

Done.

qkaiser:
Forgot we're still supporting python 3.9 while using 3.10 typing annotations. Fixed.

python/unblob/cli.py
```diff
@@ -106,7 +106,10 @@ def build_handlers_doc(
     extra_dir_handlers = plugin_manager.load_dir_handlers_from_plugins()
     dir_handlers = ctx.params["dir_handlers"] + tuple(extra_dir_handlers)
 
-    all_handlers = handlers + dir_handlers
+    # we only want handler with a valid DOC entry
+    all_handlers = filter(
+        lambda handler: handler.DOC is not None, handlers + dir_handlers
+    )
```
e3krisztian: This is fine here I think (looking at how it is used below), but `filter` returns a *filter object*, that can be read through only once:

> Return an iterator yielding those items of iterable for which function(item) is true

so I would use a list-comprehension as that is an easier construct, and also does not have this potential future mine.

qkaiser: Done.

PR #1241, opened 2025-08-13
fix(unblob): unnecessary carving of padding in qnx elf file

elektrischermoench:
Some QNX binaries are padded with a 4kib of null bytes. There is no reason to carve them.
So this fix will look into the elf binary sections to identify if it a QNX binary and carve the file with padding included.


python/unblob/handlers/executable/elf.py
```diff
@@ -371,6 +374,59 @@ def get_upx_end_offset(self, file: File, start_offset: int, end_offset: int) ->
...
+            # Get the section name from the string table
+            name_offset = section_header.sh_name
+            name_end_offset = shstrtab_content.find(b"\x00", name_offset)
+            if name_end_offset == -1:
+                name_end_offset = len(shstrtab_content)
+            section_name = shstrtab_content[name_offset:name_end_offset].decode(
+                "utf-8", errors="ignore"
+            )
+
+            if section_name.startswith(QNX_SECTION_NAME_PREFIX):
```
e3krisztian: I find it hard to understand when can -1 be a `find` result, and if it is handled properly.

Actually we do not need to properly parse and decode the section name at all, we need only to match against its prefix. In fact interpreting strings less would yield a more straight-forward code, that avoids `find`, but it depends on the fact, that the utf-8 encoding of the prefix `QNX_` is the same as its ascii encoding, so matching bytes directly would also work.

```suggestion
            maybe_section_name_prefix = shstrtab_content[name_offset:name_offset + len(QNX_SECTION_NAME_PREFIX_BYTES)]

            if maybe_section_name_prefix == QNX_SECTION_NAME_PREFIX_BYTES:
```

----
The comment above `# Get the section name from the string table` should also be updated to `# Check section name prefix`

e3krisztian (commented):
Overall looks good, however I would like to have qnx test data to verify against, if possible.

python/unblob/handlers/executable/elf.py
elektrischermoench: Good point added your changes. 

Unfortunately QNX binaries and even the Software Development Platform (SDP) are proprietary we can not just use them for testing. QNX 8 has been opened up for non commercial use, but not sure if we can use it.

Issue #1180, opened 2025-04-22
landlock sandboxing leads to PermissionError on files created from stdout


qkaiser:
**Describe the bug**

If the first layer of your firmware is made of a format for which the extractor is a `Command` creating a file in the extraction directory from `stdout`, you'll hit a `PermissionError` triggered by landlock sandboxing:

```
unblob -e /tmp/out -d 2 -f -k -vvv ../fw-repo/65e24c2e-ce84-4250-9a6a-5141cba736b5.bin 
--snip--
2025-04-22 11:01.06 [debug    ] Calculating chunk for pattern match handler=xz pid=16048 real_offset=0x0 start_offset=0x0
2025-04-22 11:01.06 [debug    ] Found valid chunk              chunk=0x0-0x1da7aefc handler=xz pid=16048
2025-04-22 11:01.06 [debug    ] Ended searching for chunks     all_chunks=[0x0-0x1da7aefc] pid=16048
2025-04-22 11:01.06 [debug    ] Removed inner chunks           outer_chunk_count=1 pid=16048 removed_inner_chunk_count=0
2025-04-22 11:01.06 [debug    ] Running extract command        command='7z x -y /home/quentin/fw-repo/65e24c2e-ce84-4250-9a6a-5141cba736b5.bin -so' pid=16048
2025-04-22 11:01.06 [error    ] Unknown error happened while extracting chunk pid=16048
Traceback (most recent call last):
--snip--
PermissionError: [Errno 13] Permission denied: '/tmp/out/65e24c2e-ce84-4250-9a6a-5141cba736b5.bin_extract/xz.uncompressed'
--snip--
PermissionError: [Errno 13] Permission denied: '/tmp/out/65e24c2e-ce84-4250-9a6a-5141cba736b5.bin_extract'
```

This only happens if the extraction directory does not exists yet.

**To Reproduce**

```
rm -rf /tmp/out
unblob -vvv -e /tmp/out -f -k random.bin.xz
# you should see a PermissionError
rm -rf /tmp/out # cleanup
mkdir -p /tmp/out
unblob -vvv -e /tmp/out -f -k random.bin.xz
# you should not see a PermissionError
```

**Expected behavior**

It should not trigger a `PermissionError`, regardless of whether or not the extraction directory exists prior to unblob launching.

PR #1248, opened 2025-08-26
fix: ensure output directories are created before extraction

dennisliuu:
## Summary
I have noticed this issue #1180 and have the same permission problem.
These changes add two lines to guarantee that the required output directories exist before extraction begins.

## Changes
Added `outdir.mkdir(parents=True, exist_ok=True)` and `extract_root.mkdir(parents=True, exist_ok=True)` to ensure both `outdir` and `extract_root` directories are created if they do not already exist.


python/unblob/extractors/command.py
```diff
@@ -34,6 +34,7 @@ def __init__(self, executable, *args, stdout: Optional[str] = None):
         self._stdout = stdout
 
     def extract(self, inpath: Path, outdir: Path):
+        outdir.mkdir(parents=True, exist_ok=True)
```
qkaiser: Do we need this too to fix #1180 ? Seems overkill to me, but maybe you have a sample that requires this ?

dennisliuu: Hi qkaiser,
No, there is no need. I've removed it in the latest commit. Thanks.

qkaiser (approved):

PR #1254, opened 2025-09-02
move e2fsprogs dependency from fork to upstream

qkaiser:
Our [debugfs fix](https://github.com/tytso/e2fsprogs/commit/56a444e898be418afd4bd11f2a6950dce7e4ae99) is present in the [debian trixie version](https://metadata.ftp-master.debian.org/changelogs//main/e/e2fsprogs/e2fsprogs_1.47.2-3_changelog) and in latest [Ubuntu](https://launchpad.net/ubuntu/+source/e2fsprogs). We can therefore go back to installing it from upstream rather than relying on our own fork.

Which means we can revert c13a0532cc54a3119578136ccf0c63917ea81928 and move to Debian Trixie for the Docker container.

### DIFF
--- a/.github/actions/setup-dependencies/action.yml
+++ b/.github/actions/setup-dependencies/action.yml
@@ -19,7 +19,12 @@ runs:
   steps:
     - if: inputs.install-test-deps == 'true'
       name: Install 3rd party dependencies
-      run: sudo ./install-deps.sh
+      run: |
+        if [ "$(id -u)" -eq 0 ]; then
+          ./install-deps.sh
+        else
+          sudo ./install-deps.sh
+        fi
       shell: bash
 
     - name: Setup sccache # for speeding up Rust builds
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -82,24 +82,72 @@ jobs:
     name: Run tests (Python)
     needs: [check_pre_commit, check_pyright]
     runs-on: ubuntu-latest
+    container:
+      image: ubuntu:25.10
     strategy:
       matrix:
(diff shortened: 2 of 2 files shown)


qkaiser:
`ubuntu-latest` at Github is actually `Ubuntu Noble` (24.04) which is still on e2fsprogs `1.47.0-2` while we need `1.47.2-2`.

e3krisztian:
It is very strange, that the failure https://github.com/onekey-sec/unblob/actions/runs/17400208989/job/49391774625?pr=1254#step:5:1047 happens only in one of the python environments, not all of them.
If it were an rdump-only problem, it should be a problem with all of the supported python versions.

qkaiser:
> It is very strange, that the failure https://github.com/onekey-sec/unblob/actions/runs/17400208989/job/49391774625?pr=1254#step:5:1047 happens only in one of the python environments, not all of them. If it were an rdump-only problem, it should be a problem with all of the supported python versions.

It's affecting all of them, it's just that the others got cancelled. If a failure is observed in one of the tests, the others are cancelled.

vlaci:
> `ubuntu-latest` at Github is actually `Ubuntu Noble` (24.04) which is still on e2fsprogs `1.47.0-2` while we need `1.47.2-2`.

we may run the tests inside docker or force install e2fsprogs from debian

qkaiser:
Chose to run tests in docker container with Ubuntu 25.10 but building dev dependencies (taplo, specifically) is .. difficult.

qkaiser:
We need this https://github.com/onekey-sec/unblob/pull/1368 first

PR #1255, opened 2025-09-03
feat(handler): support NSIS Installers

### LINKED ISSUE #1249: Support NSIS Installers
I'd like to work on adding support for NSIS Installers to unblob.

An example binary:

https://prdownloads.sourceforge.net/nsis/nsis-3.11-setup.exe?download

```
file nsis-3.11-setup.exe
nsis-3.11-setup.exe: PE32 executable (GUI) Intel 80386, for MS Windows, Nullsoft Installer self-extracting archive, 5 sections
```

These can be extracted via `7z`:

```
$ 7z x -oextracted nsis-3.11-setup.exe

7-Zip [64] 16.02 : Copyright (c) 1999-2016 Igor Pavlov : 2016-05-21 p7zip Version 16.02 (locale=en_US.UTF-8,Utf16=on,HugeFiles=on,64 bits,8 CPUs Intel(R) Core(TM) i7-8665U CPU @ 1.90GHz (806EC),ASM,AES-NI)

Scanning the drive for archives:
1 file, 1564991 bytes (1529 KiB)

Extracting archive: nsis-3.11-setup.exe
--      
Path = nsis-3.11-setup.exe
Type = Nsis
Physical Size = 1564991
Method = LZMA:23
Solid = +
Headers Size = 207454
Embedded Stub Size = 53248
SubType = NSIS-3 Unicode BadCmd=11

Everything is Ok

Files: 445
Size:       7336667
Compressed: 1564991
```

With output:

```
$ tree extracted/ | head -n 20
extracted/
├── $PLUGINSDIR
│   ├── modern-header.bmp
│   ├── modern-wizard.bmp
│   ├── nsDialogs.dll
│   └── System.dll
├── Bin
│   ├── GenPat.exe
│   ├── MakeLangId.exe
│   ├── makensis.exe
│   ├── RegTool-x86.bin
│   ├── zip2exe.exe
│   └── zlib1.dll
├── Contrib
│   ├── Graphics
│   │   ├── Checks
│   │   │   ├── big.bmp
│   │   │   ├── classic.bmp
│   │   │   ├── classic-cross.bmp
│   │   │   ├── colorful.bmp
```

Based on my understanding of the repo so far, I think I would want to create a new handler:

```
python/unblob/handlers/executable/pe.py
```

And then mirror some of the structure from `elf.py`, replacing upx with NSIS.

Before I attempt this, I'd like to understand whether I need to implement all the header parsing as is done in `elf.py`. It seems like I can let `lief` do the heavy lifting and then simply detect whether the the PE looks like a NSIS installer before invoking `7z`.

Thoughts?

jcrussell:
Searches for "Nullsoft" in the manifest to avoid false positives. Possibly too strict.

Fixes #1249

### DIFF
--- a/python/unblob/handlers/executable/pe.py
+++ b/python/unblob/handlers/executable/pe.py
@@ -1,0 +1,126 @@
+import io
+import struct
+from pathlib import Path
+from typing import Optional
+
+import lief
+from structlog import get_logger
+
+from unblob.extractors.command import Command
+
+from ...models import (
+    Extractor,
+    ExtractResult,
+    File,
+    Handler,
+    HandlerDoc,
+    HandlerType,
+    HexString,
+    Reference,
+    ValidChunk,
+)
+
+lief.logging.disable()
+
+logger = get_logger()
+
+
+class PEExtractor(Extractor):
+    def extract(self, inpath: Path, outdir: Path) -> Optional[ExtractResult]:
+        binary = lief.PE.parse(inpath)
+        if binary and self.is_nsis(binary):
+            return Command("7z", "x", "-y", "{inpath}", "-o{outdir}").extract(
+                inpath, outdir
+            )
+        return None
+
+    def is_nsis(self, binary: lief.PE.Binary) -> bool:
+        # Test if binary appears to be a Nullsoft Installer self-extracting archive
+        # see https://github.com/file/file/blob/7ed3febfcd616804a2ec6495b3e5f9ccb6fc5f8f/magic/Magdir/msdos#L383
+
+        if binary.has_resources:
+            resource_manager = binary.resources_manager
+            if (
+                isinstance(resource_manager, lief.PE.ResourcesManager)
+                and resource_manager.has_manifest
+            ):
+                manifest = (
+                    resource_manager.manifest
+                    if isinstance(resource_manager.manifest, str)
+                    else resource_manager.manifest.decode(errors="ignore")
+                )
+                if "Nullsoft.NSIS.exehead" in manifest:
+                    return True
+        return False
+
+
+class PEHandler(Handler):
+    NAME = "pe"
+
+    PATTERNS = [
+        HexString(
+            """
+            // MZ header
+            4d 5a
+            """
+        ),
+        HexString(
+            """
+            // PE header
+            50 45 00 00
+            """
+        ),
+    ]
+
+    EXTRACTOR = PEExtractor()
+
+    DOC = HandlerDoc(
+        name="pe",
+        description="The PE (Portable Executable) is a binary file format used for executable code on 32-bit and 64-bit Windows operating systems as well as in UEFI environments.",
+        handler_type=HandlerType.EXECUTABLE,
+        vendor="Microsoft",
+        references=[
+            Reference(
+                title="PE Format",
+                url="https://learn.microsoft.com/en-us/windows/win32/debug/pe-format",
+            ),
+            Reference(
+                title="Portable Executable Wikipedia",
+                url="https://en.wikipedia.org/wiki/Portable_Executable",
+            ),
+        ],
+        limitations=[],
+    )
+
+    def calculate_chunk(self, file: File, start_offset: int) -> Optional[ValidChunk]:
+        file.seek(start_offset, io.SEEK_SET)
+
+        binary = lief.PE.parse(file[start_offset:])
+        if not binary:
+            return None
+
+        # Check to see if we can extract the size of the full NSIS Installer by
+        # including the archive size from the NSIS header.
+        if binary.overlay:
+            overlay = bytes(binary.overlay)
+
+            magic_offset = overlay.find(b"NullsoftInst")
+            if magic_offset != -1:
+                header_start = magic_offset - 8
+                if header_start < 0:
+                    # Malformed NSIS header?
+                    return None
+
+                _, _, _, _, archive_size = struct.unpack(
+                    "II12sII", overlay[header_start : header_start + 28]
+                )
+
+                return ValidChunk(
+                    start_offset=start_offset,
+                    end_offset=start_offset + binary.overlay_offset + archive_size,
+                )
+
+        return ValidChunk(
+            start_offset=start_offset,
+            end_offset=start_offset + binary.original_size,
+        )
--- a/docs/handlers.md
+++ b/docs/handlers.md
@@ -48,6 +48,7 @@
     | [`NTFS`](#ntfs) | FILESYSTEM | :octicons-check-16: |
     | [`PAR2 (MULTI-VOLUME)`](#par2-multi-volume) | ARCHIVE | :octicons-check-16: |
     | [`PARTCLONE`](#partclone) | ARCHIVE | :octicons-check-16: |
+    | [`PE`](#pe) | EXECUTABLE | :octicons-check-16: |
     | [`QNAP NAS`](#qnap-nas) | ARCHIVE | :octicons-check-16: |
     | [`RAR`](#rar) | ARCHIVE | :octicons-alert-fill-12: |
     | [`ROMFS`](#romfs) | FILESYSTEM | :octicons-check-16: |
@@ -850,6 +851,23 @@
 
         - [Partclone GitHub Repository](https://github.com/Thomas-Tsai/partclone){ target="_blank" }
         - [Clonezilla Official Documentation](https://clonezilla.org/){ target="_blank" }
+## pe
+
+!!! success "Fully supported"
+
+    === "Description"
+
+        The PE (Portable Executable) is a binary file format used for executable code on 32-bit and 64-bit Windows operating systems as well as in UEFI environments.
+
+        ---
+
+        - **Handler type:** Executable
+        - **Vendor:** Microsoft
+
+    === "References"
+
+        - [PE Format](https://learn.microsoft.com/en-us/windows/win32/debug/pe-format){ target="_blank" }
+        - [Portable Executable Wikipedia](https://en.wikipedia.org/wiki/Portable_Executable){ target="_blank" }
 ## QNAP NAS
 
 !!! success "Fully supported"
--- a/python/unblob/handlers/__init__.py
+++ b/python/unblob/handlers/__init__.py
@@ -39,7 +39,7 @@ from .compression import (
     zlib,
     zstd,
 )
-from .executable import elf
+from .executable import elf, pe
 from .filesystem import (
     cramfs,
     extfs,
@@ -118,6 +118,7 @@ BUILTIN_HANDLERS: Handlers = (
     zstd.ZSTDHandler,
     elf.ELF32Handler,
     elf.ELF64Handler,
+    pe.PEHandler,
     zlib.ZlibHandler,
     engenius.EngeniusHandler,
     ecc.AutelECCHandler,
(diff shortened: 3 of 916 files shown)


python/unblob/handlers/executable/pe.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def calculate_chunk(self, file: File, start_offset: int) -> Optional[ValidChunk]:
+        file.seek(start_offset, io.SEEK_SET)
+
+        binary = lief.PE.parse(file)
+        if not binary:
+            return None
+
+        if not self.is_nsis(binary):
+            return None
```
qkaiser: This check should be moved to a custom extractor, say `PEExtractor` so that if there is a PE file chunk in a file it will be carved out.

In the `PEExtractor` you can check if it's NSIS and do this if it is:

```python
Command("7z", "x", "-y", "{inpath}", "-o{outdir}").extract(inpath, outdir)
 ```

If it's not NSIS, do nothing. Unless you can think of anything to do with "normal" PE ?

File: python/unblob/handlers/executable/pe.py [hunk 2]
```diff
@@ -0,0 +1,87 @@
...
+            ),
+        ],
+        limitations=[],
+    )
+
+    def is_nsis(self, binary: lief.PE.Binary) -> bool:
+        """
+        Test if binary appears to be a Nullsoft Installer self-extracting archive
+
+        TODO: this series of tests is possibly too strict
```
qkaiser: I leave that call to you. I don't know much about NullSoft.

python/unblob/handlers/executable/pe.py
```diff
@@ -0,0 +1,87 @@
...
+        """
+
+        if not binary.has_resources:
+            return False
+
+        if not binary.resources_manager.has_manifest:
+            return False
+
+        return "Nullsoft" in binary.resources_manager.manifest
```
qkaiser: Can be simplified to:


```suggestion
        return binary.has_resources and binary.resources_manager.has_manifest and "Nullsoft" in binary.resources_manager.manifest
```

python/unblob/handlers/executable/pe.py
```diff
@@ -0,0 +1,87 @@
...
+logger = get_logger()
+
+class PEHandler(Handler):
+    NAME = "pe"
+
+    PATTERNS = [
+        HexString(
+            """
+            // MZ header
+            4d 5a
```
qkaiser, line 64: Should we check for the "PE" signature (`50 45 00 00`) too ?

qkaiser:
@jcrussell you should also create integration tests to check that the handler works as expected.

You have to create the following directories:
- ` unblob/tests/integration/executable/pe/__input__`
- `unblob/tests/integration/executable/pe/__output`

I would put the following in the input directory:
- a normal PE file
- a normal PE file with prefix and suffix padding
- a nullsoft PE file
- a nullsoft PE file with prefix and suffix padding

To generate the output directory content, run the following:

```sh
find unblob/tests/integration/executable/pe/__input__ -type f -exec unblob -f -k -e unblob/tests/integration/executable/pe/__output__ {} \;
```

qkaiser:
@jcrussell any update on this ? do you need assistance ?

python/unblob/handlers/executable/pe.py
jcrussell: This sounds good, I update the PR later in the week (sorry for the delay).

qkaiser: np, take your time :)

jcrussell:
> @jcrussell any update on this ? do you need assistance ?

@qkaiser: I believe the code is close to final. Do you mind adding the integration test data? It is easier for me to release code than data. Here's what I have been testing with:

* a nullsoft PE file -> https://prdownloads.sourceforge.net/nsis/nsis-3.11-setup.exe?download
* a normal PE file -> I was testing on a random PE extracted from ^ (specifically, nsis-3.11-setup.exe_extract/Bin/zip2exe.exe`)

Thanks in advance!

python/unblob/handlers/executable/pe.py
```diff
@@ -0,0 +1,104 @@
...
+    def calculate_chunk(self, file: File, start_offset: int) -> Optional[ValidChunk]:
+        file.seek(start_offset, io.SEEK_SET)
+
+        binary = lief.PE.parse(file[start_offset:])
+        if not binary:
+            return None
+
+        return ValidChunk(
+            start_offset=start_offset,
+            end_offset=start_offset + binary.original_size,
```
qkaiser, line 125: `original_size` is the file size on disk, not the actual PE file size. Samples with suffix are carved with the suffix, which is incorrect. I'm looking into it.

qkaiser:
@jcrussell had to figure out how to handle LFS on forks, looks like it's okay now. Made some adjustments to keep pyright happy given LIEF's ability to return completely different types for the same object.

We need to fix the way the end offset is calculated, it'll probably be based on sections size and header size. Without unblob considers everything after the PE as part of the PE chunk.

jcrussell:
Thanks for moving this along!

> We need to fix the way the end offset is calculated, it'll probably be based on sections size and header size. Without unblob considers everything after the PE as part of the PE chunk.

I started looking into this:

```
>>> pe = lief.PE.parse("tests/integration/executable/pe/__input__/nsis-3.11-setup.exe")
>>> pe.original_size
1564991
>>> pe.sizeof_headers
1024
>>> sum([s.sizeof_raw_data for s in pe.sections])
52224
>>> sum([v.size for v in pe.data_directories])
19456
```

Found this script that dumps a bunch of info, going to try a more complete look at all the parts tomorrow.

https://github.com/lief-project/LIEF/blob/main/api/python/examples/pe_reader.py

jcrussell:
This works for (some) non-NSIS PEs but trims off the data that NSIS adds after the PE that contains what we actually want to extract. The "trimmed" data is not recognized by any handler. It seems like we need to detect if it's a NSIS installer in`calculate_chunk` to see if we need to increase the size for the NSIS data.

```
        size = sum([s.sizeof_raw_data for s in binary.sections]) + binary.sizeof_headers

        return ValidChunk(
            start_offset=start_offset,
            end_offset=start_offset + size
        )   
```

jcrussell:
@qkaiser: sorry for the delay, think this is finally fixed:

```bash
$ ls tests/integration/executable/pe/__output__/nsis-3.11-setup.exe.padded_extract/
0-16.padding  1565007-1565023.padding  16-1565007.pe  16-1565007.pe_extract
$ xxd tests/integration/executable/pe/__output__/nsis-3.11-setup.exe.padded_extract/0-16.padding 
00000000: 0000 0000 0000 0000 0000 0000 0000 0000  ................
$ xxd tests/integration/executable/pe/__output__/nsis-3.11-setup.exe.padded_extract/1565007-1565023.padding 
00000000: 0000 0000 0000 0000 0000 0000 0000 0000  ................
$ sha1sum tests/integration/executable/pe/__output__/nsis-3.11-setup.exe.padded_extract/16-1565007.pe
a64bbad73d4638d668ffdbd0887be7d6528d6a9d  tests/integration/executable/pe/__output__/nsis-3.11-setup.exe.padded_extract/16-1565007.pe
$ sha1sum tests/integration/executable/pe/__input__/nsis-3.11-setup.exe 
a64bbad73d4638d668ffdbd0887be7d6528d6a9d  tests/integration/executable/pe/__input__/nsis-3.11-setup.exe
```

qkaiser:
@jcrussell sorry about the delay, plan is to finalize this next week

python/unblob/handlers/executable/pe.py
```diff
@@ -0,0 +1,126 @@
...
+class PEHandler(Handler):
+    NAME = "pe"
+
+    PATTERNS = [
+        HexString(
+            """
+            // MZ header
+            4d 5a
+            """
+        ),
```
qkaiser, line 66: I'm worried this will lead to many false positive matches being fed to lief, which will have a significant performance impact. One way we limit impact with those small patterns is this:


```suggestion
        Regex("^\x4d\x5a"),
```

Forces the matched pattern to be at offset 0 of the file.

python/unblob/handlers/executable/pe.py
```diff
@@ -0,0 +1,126 @@
...
+        HexString(
+            """
+            // PE header
+            50 45 00 00
```
qkaiser, line 70: Are there fields other than the magic we can do some decent pattern matching on ?

python/unblob/handlers/executable/pe.py
```diff
@@ -0,0 +1,126 @@
...
+            Reference(
+                title="Portable Executable Wikipedia",
+                url="https://en.wikipedia.org/wiki/Portable_Executable",
+            ),
+        ],
+        limitations=[],
+    )
+
+    def calculate_chunk(self, file: File, start_offset: int) -> Optional[ValidChunk]:
+        file.seek(start_offset, io.SEEK_SET)
```
qkaiser, line 96: Not needed.


```suggestion
```

python/unblob/handlers/executable/pe.py
```diff
@@ -0,0 +1,126 @@
...
+            magic_offset = overlay.find(b"NullsoftInst")
+            if magic_offset != -1:
+                header_start = magic_offset - 8
+                if header_start < 0:
+                    # Malformed NSIS header?
+                    return None
+
+                _, _, _, _, archive_size = struct.unpack(
+                    "II12sII", overlay[header_start : header_start + 28]
+                )
```
qkaiser, line 116: What are those fields exactly ? If you turn  this `Handler` into a `StructHandler` you can define `C_DEFINITIONS`, a `HEADER_STRUCT` and use the `parse_header` function to get a dissect.cstruct Structure with named fields.

python/unblob/handlers/executable/pe.py
```diff
@@ -0,0 +1,126 @@
...
+        binary = lief.PE.parse(file[start_offset:])
+        if not binary:
+            return None
+
+        # Check to see if we can extract the size of the full NSIS Installer by
+        # including the archive size from the NSIS header.
+        if binary.overlay:
+            overlay = bytes(binary.overlay)
+
+            magic_offset = overlay.find(b"NullsoftInst")
```
qkaiser, line 107: How big is the overlay ? Is there a way you can derive the start offset of that overlay ? If so you could call find on `file` which is mmap'ed. It should be faster. Also if you know the magic value is within the first X bytes of the overlay, I would set an upper bound on the content that is searched through.

qkaiser: So I would do something like this:

```suggestion
        magic_offset = file[binary.overlay_offset:binary.overlay_offset+SOME_UPPER_BOUND].find(b"NullSoftInst")
```

bonus point if the upper bound is less than a page

python/unblob/handlers/executable/pe.py
```diff
@@ -0,0 +1,126 @@
...
+                isinstance(resource_manager, lief.PE.ResourcesManager)
+                and resource_manager.has_manifest
+            ):
+                manifest = (
+                    resource_manager.manifest
+                    if isinstance(resource_manager.manifest, str)
+                    else resource_manager.manifest.decode(errors="ignore")
+                )
+                if "Nullsoft.NSIS.exehead" in manifest:
+                    return True
```
qkaiser, line 53:
```suggestion
                return "Nullsoft.NSIS.exehead" in manifest
```

python/unblob/handlers/executable/pe.py
qkaiser: @jcrussell did you get the opportunity to look into this ?

qkaiser: Maybe https://github.com/onekey-sec/unblob/pull/1255#issuecomment-3417263183 ?

jcrussell:
@qkaiser: not yet, sorry. hoping to get to it by the end of the month.

PR #1258, opened 2025-09-09
Refactored report format to use Pydantic

### LINKED ISSUE #1250: Support Pydantic models for report format
For the report format it appears that `attrs` is used to define the format of the dumped JSON report. For purposes of ingesting the JSON report generated by `unblob` it would be useful to have a `pydantic` model defined that can support serialization and deserialization of the report. This would be useful for any tools wanting to ingest from the `unblob` JSON report file.

I have implemented a `pydantic` model and can submit a pull request for this, but wanted to gauge interest before doing so to see if this is something the team wanted to maintain moving forward.

garrett-zetier:
Resolves #1250 

# Overview

These changes update the JSON report format (`--report` option) so that `pydantic` models are used to serialize report data to JSON format rather than `attrs`. This allows for easier deserialization in the future, and a Pydantic TypeAdapter is provided to make reading from the JSON files back into Python objects easier. A test has been added to test deserialization.

# Testing

Update the environment to the latest since `pydantic` has been added as a dependency.

The tests have been updated to verify the latest report format, so they can be run to verify these changes. Specifically, the test_models.py (run `uv run pytest tests/test_models.py -v`) verifies both serialization and deserialization.

You can run `unblob` with the `--report` option to manually verify the JSON report. For example, run `uv run unblob --report report.json my.bin`

### DIFF
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -103,11 +103,9 @@ $ cat alpine-report.json
       {
         "path": "/home/walkman/Projects/unblob/demo/alpine-minirootfs-3.16.1-x86_64.tar.gz_extract",
         "depth": 1,
-        "chunk_id": "13590:1",
-        "__typename__": "Task"
+        "chunk_id": "13590:1"
       }
-    ],
-    "__typename__": "TaskResult"
+    ]
   },
   ...
 ]
@@ -143,7 +141,7 @@ $ unblob -vvv unknown-file | grep -C 15 "Entropy distribution"
 2024-10-30 10:52.03 [debug    ] Shannon entropy calculated     block_size=0x20000 highest=99.99 lowest=99.98 mean=99.98 path=unknown-file_extract/0-10485760.unknown pid=1963719 size=0xa00000
 2024-10-30 10:52.03 [debug    ] Chi square probability calculated block_size=0x20000 highest=97.88 lowest=3.17 mean=52.76 path=unknown-file_extract/0-10485760.unknown pid=1963719 size=0xa00000
 2024-10-30 10:52.03 [debug    ] Entropy chart                  chart=
-                              Randomness distribution                           
+                              Randomness distribution
    ┌───────────────────────────────────────────────────────────────────────────┐
 100┤ •• Shannon entropy (%)        •••••••••♰••••••••••••••••••••••••••••••••••│
  90┤ ♰♰ Chi square probability (%)   ♰ ♰ ♰♰♰♰                    ♰    ♰  ♰     │
@@ -157,8 +155,8 @@ $ unblob -vvv unknown-file | grep -C 15 "Entropy distribution"
  10┤       ♰      ♰    ♰  ♰  ♰     ♰♰    ♰         ♰                   ♰♰      │
   0┤                                ♰                                   ♰      │
    └─┬──┬─┬──┬────┬───┬──┬──┬──┬───┬───┬──┬────┬───┬────┬──┬──┬────┬──┬───┬──┬─┘
-   0 2  5 7 11   16  20 23 27 30  34  38 42   47  51   56 60 63   68 71  76 79  
-                                   131072 bytes                                 
+   0 2  5 7 11   16  20 23 27 30  34  38 42   47  51   56 60 63   68 71  76 79
+                                   131072 bytes
  path=unknown-file_extract/0-10485760.unknown pid=1963719
 2024-10-30 10:52.03 [info     ] Extracting unknown chunk       chunk=0xc96196-0x1696196 path=unknown-file_extract/13197718-23683478.unknown pid=1963719
 2024-10-30 10:52.03 [debug    ] Carving chunk                  path=unknown-file_extract/13197718-23683478.unknown pid=1963719
@@ -166,7 +164,7 @@ $ unblob -vvv unknown-file | grep -C 15 "Entropy distribution"
 2024-10-30 10:52.03 [debug    ] Shannon entropy calculated     block_size=0x20000 highest=99.99 lowest=99.98 mean=99.98 path=unknown-file_extract/13197718-23683478.unknown pid=1963719 size=0xa00000
 2024-10-30 10:52.03 [debug    ] Chi square probability calculated block_size=0x20000 highest=99.03 lowest=0.23 mean=42.62 path=unknown-file_extract/13197718-23683478.unknown pid=1963719 size=0xa00000
 2024-10-30 10:52.03 [debug    ] Entropy chart                  chart=
-                              Randomness distribution                           
+                              Randomness distribution
    ┌───────────────────────────────────────────────────────────────────────────┐
 100┤ •• Shannon entropy (%)        •••••••••••••••••••••♰••••••••••••••••••••••│
  90┤ ♰♰ Chi square probability (%)         ♰           ♰♰            ♰         │
@@ -180,8 +178,8 @@ $ unblob -vvv unknown-file | grep -C 15 "Entropy distribution"
  10┤     ♰                ♰    ♰       ♰ ♰  ♰ ♰ ♰♰   ♰ ♰♰     ♰♰ ♰♰   ♰  ♰ ♰   │
   0┤                                           ♰ ♰    ♰♰          ♰       ♰♰   │
    └─┬──┬─┬──┬────┬───┬──┬──┬──┬───┬───┬──┬────┬───┬────┬──┬──┬────┬──┬───┬──┬─┘
-   0 2  5 7 11   16  20 23 27 30  34  38 42   47  51   56 60 63   68 71  76 79  
-                                   131072 bytes 
+   0 2  5 7 11   16  20 23 27 30  34  38 42   47  51   56 60 63   68 71  76 79
+                                   131072 bytes
 ```
 
 ### Skip extraction with file magic
--- a/package.nix
+++ b/package.nix
@@ -86,6 +86,7 @@ python3.pkgs.buildPythonApplication {
     python3.pkgs.lz4 # shadowed by pkgs.lz4
     plotext
     pluggy
+    pydantic
     pyfatfs
     pymdown-extensions
     pyperscan
--- a/python/unblob/file_utils.py
+++ b/python/unblob/file_utils.py
@@ -491,15 +491,7 @@ class FileSystem:
         self.root = root.resolve()
         self.problems = []
 
-    def record_problem(
-        self,
-        problem: Union[
-            ExtractionProblem,
-            LinkExtractionProblem,
-            PathTraversalProblem,
-            SpecialFileExtractionProblem,
-        ],
-    ):
+    def record_problem(self, problem: ExtractionProblem):
         self.problems.append(problem)
         problem.log_with(logger)
 


tests/test_models.py
```diff
@@ -230,16 +295,16 @@ def test_exotic_command_output(self):
...
         decoded_report = json.loads(json_text)
 
         assert decoded_report == {
-            "__typename__": "ExtractCommandFailedReport",
+            "report_type": "ExtractCommandFailedReport",
             "command": "dump all bytes",
             "exit_code": 1,
             "severity": "WARNING",
-            "stderr": "stdout is pretty strange ;)",
+            "stderr": "b'stdout is pretty strange ;)'",
```
qkaiser: Something's wrong with the encoding here somehow.

python/unblob/report.py
```diff
@@ -265,7 +272,7 @@ class ExtractionProblem(Report):
 
     problem: str
     resolution: str
-    path: Optional[str] = None
+    path: str | None = None
```
qkaiser: I'm not sure this change will work. Unblob still supports Python 3.9 which does not implement this typing notation yet.

python/unblob/report.py
```diff
@@ -204,51 +215,47 @@ def lowest(self):
...
-@final
-@attrs.define(kw_only=True, frozen=True)
-class UnknownChunkReport(Report):
+class UnknownChunkReport(ReportBase):
     id: str
     start_offset: int
     end_offset: int
     size: int
-    randomness: Optional[RandomnessReport]
+    randomness: RandomnessReport | None
```
qkaiser: python 3.9 probably won't like this.


```suggestion
    randomness: Optional[RandomnessReport]
```

python/unblob/report.py
```diff
@@ -29,114 +21,134 @@ class Severity(Enum):
...
+    report_type: Literal["ExtractCommandFailedReport"] = "ExtractCommandFailedReport"
+
+    # In case there is any strange encoding in stdout/stderr, convert them to str when serializing
+    @field_serializer("stdout")
+    def stdout_to_str(self, v: bytes, _info):
+        return str(v)
 
+    @field_serializer("stderr")
+    def stderr_to_str(self, v: bytes, _info):
+        return str(v)
```
qkaiser: Ok this is the cause of the `b"..."` in the JSON report earlier.

I would go for something like this:

```suggestion
    # In case there is any strange encoding in stdout/stderr, convert them to str when serializing
    @field_serializer("stdout")
    def stdout_to_str(self, v: bytes, _info):
        return v.decode("utf-8", errors="surrogateescape")

    @field_serializer("stderr")
    def stderr_to_str(self, v: bytes, _info):
        return v.decode("utf-8", errors="surrogateescape")
```

You could also use `replace` instead of `surrogateescape`.

But also, I don't understand why you need to convert `bytes` to `str` for a property that is of type `bytes` in the first place ?

python/unblob/report.py
```diff
@@ -29,114 +21,134 @@ class Severity(Enum):
...
+class UnknownErrorBase(ErrorReportBase):
     """Describes an exception raised during file processing."""
 
-    severity: Severity = attrs.field(default=Severity.ERROR)
-    exception: Union[str, Exception] = attrs.field(  # pyright: ignore[reportGeneralTypeIssues]
-        converter=_convert_exception_to_str
-    )
-    """Exceptions are also formatted at construct time.
+    severity: Severity = Severity.ERROR
+    exception: str | Exception
```
qkaiser: typing notation not ok for python 3.9

qkaiser (commented):
first quick pass over the code, some of my colleagues will also have a look :)

qkaiser:
Code looks ok, let's run the pipeline to see what fails.

python/unblob/report.py
garrett-zetier: Ah, I had updated this based on one of the pre-commit tools. I can roll this back and use `typing.Optional[]` instead.

python/unblob/report.py
garrett-zetier: This suggestion makes sense. Thanks! The issue with Pydantic is that when dealing with raw `bytes` data it may not correctly handle encode/decode since it is written specifically for handling JSON data which doesn't handle `bytes`. When handling raw byte type data, you need to provide a serializer to show Pydantic how to handle the data.

python/unblob/report.py
garrett-zetier: The `ruff` formatting checker was recommending using `|` annotation but I updated this to support Python 3.9.

python/unblob/report.py
garrett-zetier: Updated type to support Python 3.9.

python/unblob/report.py
garrett-zetier: Updated type to support Python 3.9.

python/unblob/report.py
garrett-zetier: I looked at this again but couldn't find anything that works. Using `v.decode("utf-8", errors="replace")` results in unexpected serialized bytes data. Pydantic struggles to handle `bytes` attributes when the fields contain non UTF-8 data. I'll keep looking to see if I can find something else that works other than what's currently implemented.

tests/test_models.py
garrett-zetier: Since `stderr` is specified as `bytes` should we expect the decoded result to also be bytes?

vulture_whitelist.py
qkaiser, line 1: Some things should be added to this whitelist:

```
python/unblob/models.py:291: unused variable 'ReportModelAdapter' (60% confidence)
python/unblob/report.py:29: unused variable 'report_type' (60% confidence)
python/unblob/report.py:38: unused variable 'model_config' (60% confidence)
python/unblob/report.py:42: unused method 'model_post_init' (60% confidence)
python/unblob/report.py:42: unused variable '__context' (100% confidence)
python/unblob/report.py:56: unused variable 'report_type' (60% confidence)
python/unblob/report.py:65: unused variable 'report_type' (60% confidence)
python/unblob/report.py:76: unused variable 'report_type' (60% confidence)
python/unblob/report.py:89: unused variable 'report_type' (60% confidence)
python/unblob/report.py:92: unused method 'stdout_to_str' (60% confidence)
python/unblob/report.py:96: unused method 'stderr_to_str' (60% confidence)
python/unblob/report.py:104: unused variable 'report_type' (60% confidence)
python/unblob/report.py:112: unused variable 'report_type' (60% confidence)
python/unblob/report.py:123: unused variable 'report_type' (60% confidence)
python/unblob/report.py:132: unused variable 'report_type' (60% confidence)
python/unblob/report.py:141: unused variable 'report_type' (60% confidence)
python/unblob/report.py:151: unused variable 'report_type' (60% confidence)
python/unblob/report.py:176: unused variable 'report_type' (60% confidence)
python/unblob/report.py:201: unused variable 'report_type' (60% confidence)
python/unblob/report.py:221: unused variable 'report_type' (60% confidence)
python/unblob/report.py:232: unused variable 'report_type' (60% confidence)
python/unblob/report.py:241: unused variable 'report_type' (60% confidence)
python/unblob/report.py:246: unused variable 'report_type' (60% confidence)
python/unblob/report.py:255: unused variable 'report_type' (60% confidence)
python/unblob/report.py:300: unused variable 'report_type' (60% confidence)
python/unblob/report.py:305: unused variable 'report_type' (60% confidence)
python/unblob/report.py:317: unused variable 'report_type' (60% confidence)
python/unblob/report.py:326: unused variable 'report_type' (60% confidence)
```

python/unblob/report.py
```diff
@@ -275,9 +282,27 @@ def log_with(self, logger):
...
+    - unsafe symlinks pointing outside of extraction directory
+
+    """
+
+    report_type: Literal["ExtractionProblem"] = "ExtractionProblem"
+
+
+class PathTraversalProblem(ExtractionProblemBase):
     extraction_path: str
+    report_type: Literal["PathTraversalProblem"] = "PathTraversalProblem"
```
qkaiser: You probably need to adapt typing for `record_problem` in `python/unblob/file_utils`:

```
  /home/runner/work/unblob/unblob/python/unblob/file_utils.py:518:33 - error: Argument of type "PathTraversalProblem" cannot be assigned to parameter "problem" of type "ExtractionProblem" in function "record_problem"
    "PathTraversalProblem" is not assignable to "ExtractionProblem" (reportArgumentType)
```

python/unblob/report.py
```diff
@@ -287,18 +312,49 @@ def log_with(self, logger):
         )
 
 
-@attrs.define(kw_only=True, frozen=True)
-class LinkExtractionProblem(ExtractionProblem):
+class LinkExtractionProblem(ExtractionProblemBase):
     link_path: str
+    report_type: Literal["LinkExtractionProblem"] = "LinkExtractionProblem"
```
qkaiser: You probably need to adapt typing for `record_problem` in `python/unblob/file_utils`:

```
   /home/runner/work/unblob/unblob/python/unblob/file_utils.py:580:29 - error: Argument of type "LinkExtractionProblem" cannot be assigned to parameter "problem" of type "ExtractionProblem" in function "record_problem"
    "LinkExtractionProblem" is not assignable to "ExtractionProblem" (reportArgumentType)
/home/runner/work/unblob/unblob/python/unblob/file_utils.py:635:21 - error: Argument of type "LinkExtractionProblem" cannot be assigned to parameter "problem" of type "ExtractionProblem" in function "record_problem"
    "LinkExtractionProblem" is not assignable to "ExtractionProblem" (reportArgumentType)
  /home/runner/work/unblob/unblob/python/unblob/file_utils.py:641:37 - error: Argument of type "LinkExtractionProblem" cannot be assigned to parameter "problem" of type "ExtractionProblem" in function "record_problem"
    "LinkExtractionProblem" is not assignable to "ExtractionProblem" (reportArgumentType)
```

python/unblob/report.py
```diff
@@ -287,18 +312,49 @@ def log_with(self, logger):
...
 
 
-@attrs.define(kw_only=True, frozen=True)
-class SpecialFileExtractionProblem(ExtractionProblem):
+class SpecialFileExtractionProblem(ExtractionProblemBase):
     mode: int
     device: int
+    report_type: Literal["SpecialFileExtractionProblem"] = (
+        "SpecialFileExtractionProblem"
+    )
```
qkaiser: You probably need to adapt typing for `record_problem` in `python/unblob/file_utils`:

```
  /home/runner/work/unblob/unblob/python/unblob/file_utils.py:573:33 - error: Argument of type "SpecialFileExtractionProblem" cannot be assigned to parameter "problem" of type "ExtractionProblem" in function "record_problem"
    "SpecialFileExtractionProblem" is not assignable to "ExtractionProblem" (reportArgumentType)
```

qkaiser (commented):
Don't forget to run pre-commit before pushing code:

```
pre-commit install
pre-commit run -a
```

You can also check typing by running `pyright` from the repo root.

python/unblob/report.py
qkaiser: @vlaci what's your take here ?

tests/test_models.py
qkaiser: @vlaci what's your take here ?

python/unblob/report.py
garrett-zetier: Actually I have a better approach to this test since I've taken another look. The goal of the test is to encode and decode, making sure the original data with strange bytes is preserved. I have refactored the test to do this using Pydantic methods. These are the same encode/decode methods we use for dumping the JSON report, and for the new test I added that verifies proper decoding of a JSON report.

Please take a look at the latest implementation and let me know what you think. I have updated the Pydantic model to use `base64` so that we can safely handle any bytes data. Since this is all handled under the hood, the user won't have to do anything different during encode/decode.

tests/test_models.py
garrett-zetier: I have updated the Pydantic model to use `base64` so that we can safely handle any bytes data. Since this is all handled under the hood, the user won't have to do anything different during encode/decode. See also the refactored `test_exotic_command_output` in test_models.py for an example.

vulture_whitelist.py
garrett-zetier: Thanks for the tips! I was trying to figure out how to fix some of the vulture checks. The latest commit has these resolved through whitelisting.

python/unblob/report.py
garrett-zetier: The typing has been updated in `record_problem()`.

python/unblob/report.py
garrett-zetier: The typing has been updated in `record_problem()`.

python/unblob/report.py
garrett-zetier: The typing has been updated in `record_problem()`.

vlaci:
I went over the changes and I am mostly happy with them, good job!

The base64 process output encoding is cool, as it is more portable than python's string escapes.

I disliked that we need to fight with the type checker about the discriminator values introducing a bit convoluted inheritance hierarchy. My suggestion would be to use callable discriminators eliminating the need for the `report_type` field on the models themselves.

I've played around with the idea on how it would look like here: https://github.com/wilypence/unblob/compare/1250-pydantic-report...onekey-sec:unblob:pull/1250/review

garrett-zetier:
I like your implementation of the report models much better. It is much cleaner. Let me know if you want me to continue with those changes on my fork.

One other thing that seems to be missing is that the CI tests are failing because the `pydantic` package is missing from some list of dependencies. I'm not sure where this needs to be added (package.nix?).

vlaci:
> I like your implementation of the report models much better. It is much cleaner. Let me know if you want me to continue with those changes on my fork.

If you are up to it, then please do so. If you say that you've done enough, that's fine as well, we can take over. 

> One other thing that seems to be missing is that the CI tests are failing because the `pydantic` package is missing from some list of dependencies. I'm not sure where this needs to be added (package.nix?).

Yes, it is missing from the `dependencies` list in `package.nix`.

garrett-zetier:
@vlaci I'll take a look and finish it up. Thank you guys for assisting!

garrett-zetier:
I believe everything should be addressed. All tests and pipelines should pass now.

tests/test_report.py
```diff
@@ -377,7 +377,7 @@ def container_task_results(
                 blob_id=chunk_id,
             ),
             reports=[
-                StatReport(
+                StatReport.model_construct(
```
e3krisztian, line 380: Adding `.model_construct`-s looks quite noisy and verbose - are these really needed?

garrett-zetier: The `model_construct` is necessary in places where incompatible types are being used. Since we are constructing objects defined as Pydantic models, they will be validated against the defined types. Since the tests are using placeholder fields like `ANY` for testing, we need to use `model_construct` to construct the models without validating.

vlaci: I was thinking about this as well, but I think this is the easiest solution and it is constrained to the test scope, so I am okay with it.

P.S. we could use something like [dirty-equals](https://github.com/samuelcolvin/dirty-equals) for test doubles, but it is an additional dependency and would have very limited improvement over the current solution.

garrett-zetier:
Thanks everyone for their assistance and time reviewing! I am looking forward to these changes in a future release.

garrett-zetier:
I have rebased from the latest on the upstream main branch. Is there anywhere I should document usage of the updated report with Pydantic?

vlaci:
> I have rebased from the latest on the upstream main branch. Is there anywhere I should document usage of the updated report with Pydantic?

We don't publish API docs, and this change is API-only, so I don't think that this change needs extra documentation

PR #1284, opened 2025-10-17
Potential fix for code scanning alert no. 316: Incomplete ordering

qkaiser:
Potential fix for [https://github.com/onekey-sec/unblob/security/code-scanning/316](https://github.com/onekey-sec/unblob/security/code-scanning/316)

The best way to ensure complete ordering is to use the `functools.total_ordering` decorator on the `YAFFSEntry` class. This decorator automatically fills in the missing rich comparison methods if at least `__eq__` and one of `__lt__`, `__le__`, `__gt__`, or `__ge__` are defined. Since `YAFFSEntry` already defines `__eq__`, `__lt__`, and `__gt__`, it's safe to decorate it with `@functools.total_ordering`. This change should be made directly above the class definition, after importing `functools` in the file if it’s not already imported.

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._

### DIFF
--- a/python/unblob/handlers/filesystem/yaffs.py
+++ b/python/unblob/handlers/filesystem/yaffs.py
@@ -204,18 +204,6 @@ class YAFFSEntry:
     st_mtime: int = attrs.field(default=0)
     st_ctime: int = attrs.field(default=0)
 
-    def __lt__(self, other):
-        return self.object_id < other.object_id
-
-    def __gt__(self, other):
-        return self.object_id > other.object_id
-
-    def __eq__(self, other):
-        return self.object_id == other.object_id
-
-    def __hash__(self):
-        return hash(self.object_id)
-
     def __str__(self):
         return f"{self.object_id}: {self.name}"
 


e3krisztian:
I was looking at the code and if comparison is used at all, but could not find any trivial places (explicit use of `<` and `>`, `sorted`, `max`, `min`) that do not use a `key` parameter for comparisons.

<details>
<summary>Also checked with CC, which also found no use</summary>

● Based on my analysis, I can see that `YAFFSEntry` comparison is NOT directly used in the current code. The `@functools.total_ordering` decorator was added as part of the CodeQL fix, but looking at the code:

  1. No direct sorting of `YAFFSEntry` objects: The code sorts data_chunks (which are `YAFFSChunk` objects, not `YAFFSEntry` objects) on lines 597 and 680.
  2. Tree operations use string IDs: The file entries tree uses `str(entry.object_id)` as node identifiers (lines 439, 443-444, 458-460, 467, 490-491), not the entry objects themselves for comparison.
  3. The comparison methods exist but appear unused: The `__lt__`, `__gt__`, `__eq__` methods defined on lines 209-216 compare based on object_id, but they don't seem to be actively invoked by the current code.

  Why was `@functools.total_ordering` added?

  Looking at the git commit message, CodeQL reported an "incomplete ordering" issue. This is a defensive fix - even though the comparison isn't currently used, the class defines some comparison methods (`__lt__`, `__gt__`, `__eq__`), which creates an incomplete ordering. The `@functools.total_ordering` decorator automatically generates the missing comparison methods (`__le__`, `__ge__`, `__ne__`) based on the ones defined, ensuring complete ordering semantics.

  Summary:

  The `YAFFSEntry` comparison methods (python/unblob/handlers/filesystem/yaffs.py:209-216) define ordering by `object_id`, but this ordering doesn't appear to be actively used in the current codebase. The `@functools.total_ordering` decorator was added as a defensive measure to ensure complete ordering semantics, preventing potential issues if the class is ever used in sorting operations in the future.
</details>

So instead of moving forward to total_ordering I propose to simplify the code and remove all the ordering methods.

PR #1306, opened 2025-11-21
docs/extractors.md: add erofs and partclone

ffontaine:
Update docs/extractors.md with:
 - erofs which is supported since version 25.4.14 and commit 5f50cd2d7e22116874f36deaaead96da52a6786c
 - partclone which is supported since version 25.5.26 and commit e7413cb9590e5dd65e4e4fbf42a16ffc8edc8d5e

Minium versions have been set to latest published version (older versions probably also work but no tests were made)

While at it, also fix a typo: img2simg -> simg2img

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -43,11 +43,11 @@ wheels = [
 
 [[package]]
 name = "attrs"
-version = "25.3.0"
+version = "25.4.0"
 source = { registry = "https://pypi.org/simple" }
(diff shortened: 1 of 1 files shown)


qkaiser:
Something went wrong when pushing back to your branch, probably because you used your `main`. Your commit is still there, being handled in https://github.com/onekey-sec/unblob/pull/1309 :)

PR #1320, opened 2025-12-08
add usedforsecurity=False to md5/sha1 for FIPS compliance

zetier-md:
Small changes to comply with the FIPS standard: updates calls to `hashlib` constructors to include `usedforsecurity=False`, explicitly marking MD5/SHA1 usage as non-cryptographic. This enables unblob to run on FIPS-restricted systems.
- MD5 updated to comply with FIPS 140-2
- SHA1 updated to comply with FIPS 140-3

These changes do not have a functional impact on the hash output or behavior.

### DIFF
--- a/python/unblob/handlers/archive/par2.py
+++ b/python/unblob/handlers/archive/par2.py
@@ -60,7 +60,9 @@ class MultiVolumePAR2Handler(DirectoryHandler):
                 packet_content = f.read(
                     header.packet_length - len(header) + offset_to_recovery_id
                 )
-                packet_checksum = hashlib.md5(packet_content).digest()  # noqa: S324
+                packet_checksum = hashlib.md5(
+                    packet_content, usedforsecurity=False
+                ).digest()
 
                 if packet_checksum != header.md5_hash:
                     return False
--- a/python/unblob/report.py
+++ b/python/unblob/report.py
@@ -171,8 +171,8 @@ class HashReport(ReportBase):
     @classmethod
     def from_path(cls, path: Path):
         chunk_size = 1024 * 64
-        md5 = hashlib.md5()  # noqa: S324
-        sha1 = hashlib.sha1()  # noqa: S324
+        md5 = hashlib.md5(usedforsecurity=False)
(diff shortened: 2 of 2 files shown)


python/unblob/handlers/archive/par2.py
```diff
@@ -60,7 +60,7 @@ def is_valid_header(self, file_paths: list) -> bool:
                 packet_content = f.read(
                     header.packet_length - len(header) + offset_to_recovery_id
                 )
-                packet_checksum = hashlib.md5(packet_content).digest()  # noqa: S324
+                packet_checksum = hashlib.md5(packet_content, usedforsecurity=False).digest()  # noqa: S324
```
qkaiser:
```suggestion
                packet_checksum = hashlib.md5(packet_content, usedforsecurity=False).digest() 
```

python/unblob/report.py
```diff
@@ -171,8 +171,8 @@ class HashReport(ReportBase):
     @classmethod
     def from_path(cls, path: Path):
         chunk_size = 1024 * 64
-        md5 = hashlib.md5()  # noqa: S324
-        sha1 = hashlib.sha1()  # noqa: S324
+        md5 = hashlib.md5(usedforsecurity=False)  # noqa: S324
+        sha1 = hashlib.sha1(usedforsecurity=False)  # noqa: S324
```
qkaiser:
```suggestion
        md5 = hashlib.md5(usedforsecurity=False) 
        sha1 = hashlib.sha1(usedforsecurity=False)
```

python/unblob/handlers/archive/par2.py
zetier-md: removed the explicit ignores in 185334b488615fc70423dcde8f66e5531cd12df5

python/unblob/report.py
zetier-md: removed the explicit ignores in 185334b488615fc70423dcde8f66e5531cd12df5

qkaiser:
squashed commits and fixed wording, kept authorship. On its way to the merge train.

Issue #794, opened 2024-03-01
Add option to remove archive files when extraction was successful


Feel-C:
**Reduce the size of _extract directory**
I am using unblob on large files (5Go to 8Go)  and disk usage is very high : original file + _extract directory + log 
The _extrract directory contains all original archive files and _extract result.

**Add an option to remove all achive files (.gz, .zi, .tgz, ... ) in the _extract directory**
When the extraction of a file is sucessfull, the original file (.zip, .xz, .tgz, ...) is deleted in the _extract directory

qkaiser:
We already discussed this idea at https://github.com/onekey-sec/unblob/discussions/687. It's not on our roadmap right now, but we'll definitely take your request into account since it indicates that more and more users are expecting this feature.

qkaiser:
If you want to give it a try, we had a branch that did exactly what you want: https://github.com/onekey-sec/unblob/pull/766

PR #1323, opened 2025-12-12
feat: add temporary file cleanup options

qkaiser:
- add temporary file deletion mode to `ExtractionConfig` with handler filter, source protection and whole-file-only logic
- expose CLI flags `--delete-temporary-files`/`--temporary-file-handler` and document usage
- delete carved temps after successful extraction; keep failed cases intact
- add integration-style cleanup test and CLI option parsing test

Related to https://github.com/onekey-sec/unblob/issues/794 and https://github.com/onekey-sec/unblob/discussions/687

### DIFF
--- a/docs/guide.md
+++ b/docs/guide.md
@@ -244,6 +244,11 @@ Options:
   --report PATH                   File to store metadata generated during the
                                   extraction process (in JSON format).
   -k, --keep-extracted-chunks     Keep extracted chunks
+  --delete-extracted-files TEXT   Delete fully extracted intermediate files
+                                  (whole-file chunks only). Use
+                                  'selected:<handler1,handler2>' (comma-
+                                  separated) to limit deletions to specific
+                                  handlers.  [default: none]
   -v, --verbose                   Verbosity level, counting, maximum level: 3
                                   (use: -v, -vv, -vvv)
   --show-external-dependencies    Shows commands needs to be available for
--- a/python/unblob/cli.py
+++ b/python/unblob/cli.py
@@ -34,6 +34,7 @@ from .processing import (
     DEFAULT_PROCESS_NUM,
     DEFAULT_SKIP_EXTENSION,
     DEFAULT_SKIP_MAGIC,
+    ExtractedFileDeletionMode,
     ExtractionConfig,
     process_file,
 )
@@ -50,6 +51,36 @@ def restore_cursor():
 
 def get_version():
     return version("unblob")
+
+
+def _parse_delete_extracted_files(
+    _ctx: click.Context,
+    _param: click.Option,
+    value: str,
+) -> tuple[ExtractedFileDeletionMode, tuple[str, ...]]:
+    normalized = value.strip().lower()
+    if normalized == ExtractedFileDeletionMode.NONE.value:
+        return (ExtractedFileDeletionMode.NONE, ())
+    if normalized == ExtractedFileDeletionMode.ALL.value:
+        return (ExtractedFileDeletionMode.ALL, ())
+
+    prefix = f"{ExtractedFileDeletionMode.SELECTED.value}:"
+    if normalized.startswith(prefix):
+        handlers_raw = value[len(prefix) :].strip()
+        if not handlers_raw:
+            raise click.BadParameter(
+                "Provide at least one handler name with 'selected:<handler1,handler2>'."
+            )
+        handlers = tuple(
+            handler.strip() for handler in handlers_raw.split(",") if handler.strip()
+        )
+        if not handlers:
+            raise click.BadParameter(
+                "Provide at least one handler name with 'selected:<handler1,handler2>'."
+            )
+        return (ExtractedFileDeletionMode.SELECTED, handlers)
+
+    raise click.BadParameter("Use 'none', 'all', or 'selected:<handler1,handler2>'.")
 
 
 def show_version(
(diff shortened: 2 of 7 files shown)


docs/guide.md
```diff
@@ -244,6 +244,13 @@ Options:
   --report PATH                   File to store metadata generated during the
                                   extraction process (in JSON format).
   -k, --keep-extracted-chunks     Keep extracted chunks
+  --delete-temporary-files [none|some|all]
```
martonilles: I would combine these two options, like 
```
--delete-temporary-files none|all|selected:handler-names
```

python/unblob/processing.py
```diff
@@ -660,6 +678,9 @@ def _extract_chunk(
         fix_extracted_directory(extract_dir, self.result)
         delete_empty_extract_dir(extract_dir)
 
+        if extraction_successful:
+            self._delete_temporary_file_if_needed(carved_path, chunk)
```
martonilles: the carved_path can already be deleted with remove_extracted_input on line 665-667

also if it is a whole file and called directly from `_process_chunks` this would work, but if it is called for not whole file _carve_then_extract_chunks then the `carved_path` would be a carved chunk, not the original file.

I think the intention is always to delete the original file, I would move this logic into end of `_process_chunks` and delete the `file` when need to. Good question what happens if a file contains multiple chunks and `some` is used, then all chunks needs to be specified in handler or just enough one?

docs/guide.md
martonilles: Also not sure about the name "temporary file" these are actually output of a previous extraction and not just chunks, right?

maybe calling it delete-extracted-files would be better?

qkaiser: How would you provide "selected:handler-names" ? comma separated list ?

qkaiser: > maybe calling it delete-extracted-files would be better?

agree

python/unblob/processing.py
qkaiser: did my best to simplify the whole thing, let me know what you think

python/unblob/cli.py
```diff
@@ -298,6 +329,14 @@
     show_default=True,
     help="Keep extracted chunks",
 )
+@click.option(
+    "--delete-extracted-files",
+    default=ExtractedFileDeletionMode.NONE.value,
+    show_default=True,
+    help="Delete fully extracted temporary files. Use 'selected:<handler1,handler2>' "
+    "to restrict deletions to specific handlers.",
+    callback=lambda ctx, param, value: _parse_delete_extracted_files(ctx, param, value),
```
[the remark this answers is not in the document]
qkaiser: good bot

python/unblob/processing.py
qkaiser: > what happens if a file contains multiple chunks and some is used, then all chunks needs to be specified in handler or just enough one ?

Current implementation only covers scenarios where a file is made of a single chunk. We can improve it later on with the following approach:
- if a file is made of multiple valid chunks only (no unknown chunks, no padding chunks)
- if each and every chunk extraction is successful
- if the `all` policy is chosen, or if each and every chunk format is in the `selected` policy set
- then delete that file

The restriction on valid chunks could be relaxed to also accept padding chunks (but definitely not unknown chunks).

python/unblob/processing.py
```diff
@@ -668,6 +687,49 @@ def _extract_chunk(
...
+
+        if self.task.depth == 0:
+            return
+
+        if not delete_candidate_path.exists() or delete_candidate_path.is_dir():
+            return
+
+        try:
+            delete_candidate_path.unlink()
+            logger.debug(
```
martonilles, line 713: wondering if we should create a report when a file is deleted? Otherwise we would have Stat/Hash/Magic Reports, but the file will be missing. I would add a report to record that the file was indeed deleted

python/unblob/cli.py
```diff
@@ -298,6 +329,14 @@ def __init__(
     show_default=True,
     help="Keep extracted chunks",
 )
+@click.option(
+    "--delete-extracted-files",
+    default=ExtractedFileDeletionMode.NONE.value,
+    show_default=True,
+    help="Delete fully extracted temporary files. Use 'selected:<handler1,handler2>' "
```
martonilles: I would call them "intermediate" 

Also would add a comment that only full chunk files are deleted, so if a file has two chunks and both extracted we wont be deleting it

python/unblob/processing.py
```diff
@@ -633,17 +648,21 @@ def _extract_chunk(
...
 
         extraction_reports = []
         try:
             if result := chunk.extract(carved_path, extract_dir):
                 extraction_reports.extend(result.reports)
 
+            extraction_successful = not any(
+                isinstance(report, ErrorReport) for report in extraction_reports
+            )
+
```
martonilles: if there is an exception on 670/672 we wont be handling it here

why we need the `extraction_successful` variable we can just check for error reports when returning?

python/unblob/cli.py
qkaiser: See 9d19358

python/unblob/processing.py
qkaiser: See be4abe26729ea1a31616dbe0afeb58a98cdb90c1

python/unblob/processing.py
qkaiser: See e78f35f66c314e96b026592add9be3e3d4693ae3

qkaiser:
Squashed, rebased, resolved.

martonilles (approved):
Looks good, the multi-chunk extracted case can be handled later.

Issue #1328, opened 2025-12-21
New pydantic types break support for custom extraction reports


matutter:
**Describe the bug**

The new usage of Pydantic in the `ChunkReport` eliminates the ability to use custom extractors with unblob that return custom reports. The `Report` union is too strict to allow for extensions. This tagged union makes it impossible to include sub-classes of `ReportBase` which are not defined as part of that union.

There are simple solutions for this that will allow for end-users to extend the Report types. Since the `ExtractionConfig` contains the `handlers` list - I assume that the intent is to allow for custom handlers to be defined outside of the unblob package to be used, and that would include support for custom reports as well.

Please see my fork for a solution: https://github.com/matutter/unblob/tree/bug/unblob-1328

**To Reproduce**

Due to the nature of the issue - excuse the vagueness of this.

1. Create an `Extractor` subclass with an `extract(...) -> ExtractResult` method that returns results with reports that are not defined in the `Report` union.
2. Run unblob such that the custom handler is provided to the `ExtractionConfig`.
4. See error (below) ...

<details>
<summary>Error details</summary>

```
  File "/workspaces/platform/libs/unpack/.venv/lib/python3.13/site-packages/unblob/processing.py", line 525, in process
    self._process_chunks(file, outer_chunks, unknown_chunks)
    ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspaces/platform/libs/unpack/.venv/lib/python3.13/site-packages/unblob/processing.py", line 554, in _process_chunks
    self._extract_chunk(
    ~~~~~~~~~~~~~~~~~~~^
        carved_path,
        ^^^^^^^^^^^^
    ...<3 lines>...
        remove_extracted_input=False,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/workspaces/platform/libs/unpack/.venv/lib/python3.13/site-packages/unblob/processing.py", line 658, in _extract_chunk
    self.result.add_report(chunk.as_report(extraction_reports))
                           ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^
  File "/workspaces/platform/libs/unpack/.venv/lib/python3.13/site-packages/unblob/models.py", line 153, in as_report
    return ChunkReport(
        id=self.id,
    ...<5 lines>...
        extraction_reports=extraction_reports,
    )
[...]

PR #1334, opened 2025-12-28
feat(reporting): support plugin-registered report types

qkaiser:
Implemented a report registry and plugin hook so external packages can register Report subclasses, with runtime validation of those reports.

This allows advanced users of unblob have their own handlers and extractors yield custom reports without raising pydantic validation exceptions (which was happening before) while keeping strong validation of these custom reports through pydantic.

Documentation has been edited to cover how users can define and register custom reports, mirroring our handler/extractor documentation.

Related to https://github.com/onekey-sec/unblob/issues/1328

### DIFF
--- a/python/unblob/models.py
+++ b/python/unblob/models.py
@@ -8,7 +8,7 @@ from pathlib import Path
 from typing import Generic, Optional, TypeVar, Union
 
 import attrs
-from pydantic import BaseModel, TypeAdapter
+from pydantic import BaseModel, TypeAdapter, field_validator
 from structlog import get_logger
 
 from .file_utils import Endian, File, InvalidInputFormat, StructParser
@@ -22,6 +22,7 @@ from .report import (
     RandomnessReport,
     Report,
     UnknownChunkReport,
+    validate_report_list,
 )
 
 logger = get_logger()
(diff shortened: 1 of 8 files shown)


python/unblob/models.py
```diff
@@ -267,7 +273,11 @@ def register(self, result: TaskResult):
 
     def to_json(self, indent="  "):
         return json.dumps(
-            [result.model_dump(mode="json") for result in self.results], indent=indent
+            [
+                result.model_dump(mode="json", serialize_as_any=True)
```
e3krisztian, line 277: Why do we have to relax the serialization with `serialize_as_any`?

qkaiser: Without it pydantic serializes the declared field type (`Report`) instead of the subclass. This means it simply serializes the `__typename__` without the fields we need. That breaks deserialization because required fields are missing. `serialize_as_any=True` tells Pydantic to serialize the actual object type, so subclass fields are included.

python/unblob/report.py
```diff
@@ -295,40 +316,73 @@ def log_with(self, logger):
...
+    if report_type is None:
+        raise ValueError(f"Unknown report type: {typename}")
+    return report_type.model_validate(report)
+
+
+def validate_report_list(value: Any) -> list[Report]:
+    if value is None:
+        return []
+    if isinstance(value, tuple):
+        value = list(value)
```
e3krisztian: Why do we need these None and tuple special cases?

qkaiser: I think this is leftover code from debugging sessions, I don't see a reason we would need those. Especially since we bail out if the value is not a list. I'll remove it.

qkaiser: Done

PR #1345, opened 2026-01-13
feat: add no-sandbox opt-out option

qkaiser:
Added a no_sandbox opt-out flag to the CLI so the Landlock sandbox can be bypassed when requested, while remaining on by default.
    
The CLI now accepts --no-sandbox. If set, process_file is called directly instead of called within a dedicated thread locked down by Landlock.

### DIFF
--- a/python/unblob/cli.py
+++ b/python/unblob/cli.py
@@ -285,6 +285,12 @@ class UnblobContext(click.Context):
     help="Only carve chunks and skip further extraction",
 )
 @click.option(
+    "--no-sandbox",
+    is_flag=True,
+    show_default=True,
+    help="Disable Landlock sandboxing (useful for breakpoint-based debugging).",
+)
+@click.option(
     "-k",
     "--keep-extracted-chunks",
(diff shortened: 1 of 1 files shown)


python/unblob/sandbox.py
```diff
@@ -93,6 +94,9 @@ def _run_in_thread(callback, *args, **kwargs):
         return result  # pyright: ignore[reportReturnType]
 
     def _try_enter_sandbox(self):
+        if self._config.no_sandbox:
+            logger.info("Sandboxing disabled via configuration, skipping.")
+            return
```
martonilles: I would actually skip sandboxing in `run` so no thread would be started as that effectively kills debuggers

even easier would be to do it in `cli.py:cli`

qkaiser: Yeah, I heavily simplified the implementation. We don't need to wire it to `ExtractionConfig` in the first place since users of the unblob API can call `process_file` outside of the sandbox anyway.

PR #1348, opened 2026-01-15
feat(handler): add support for DEAFBEAD files

jstucke:
Adds support for DEAFBEAD archive files as found e.g. in D-Link DSL-500G and DSL-504G firmware images.
Unblob already unpacks the contents from those archives, because the files are compressed with gzip, but the folder structure and file names get lost in the process.

### DIFF
--- a/python/unblob/handlers/archive/dlink/deafbead.py
+++ b/python/unblob/handlers/archive/dlink/deafbead.py
@@ -1,0 +1,110 @@
+from __future__ import annotations
+
+import gzip
+import io
+from pathlib import Path, PureWindowsPath
+
+from unblob.file_utils import Endian, File, FileSystem, InvalidInputFormat, StructParser
+from unblob.models import (
+    Extractor,
+    HandlerDoc,
+    HandlerType,
+    HexString,
+    StructHandler,
+    ValidChunk,
+)
+
+C_DEFINITIONS = r"""
+    typedef struct deafbead_header {
+        uint32 magic;           /* "DE AF BE AD" */
+    } deafbead_header_t;
+
+    typedef struct deafbead_dir {
+        uint8  magic;           /* "86" */
+        uint16 name_len;
+        char   name[name_len];
+    } deafbead_dir_t;
+
+    typedef struct deafbead_file {
+        uint8  magic;           /* "87" */
+        uint16 name_len;
+        char   name[name_len];
+        uint32 file_size;
+        char   file_contents[file_size];
+    } deafbead_file_t;
+"""
+DIR_MAGIC = b"\x86"
+FILE_MAGIC = b"\x87"
+VALID_MAGICS = {DIR_MAGIC, FILE_MAGIC}
+HEADER_LEN = 4
+
+
+class DeafBeadExtractor(Extractor):
+    def __init__(self):
+        self._struct_parser = StructParser(C_DEFINITIONS)
+
+    def extract(self, inpath: Path, outdir: Path):
+        fs = FileSystem(outdir)
+        with File.from_path(inpath) as file:
+            file.seek(HEADER_LEN)
+            while (magic := file.read(1)) in VALID_MAGICS:
+                file.seek(-1, io.SEEK_CUR)  # go back to read the full struct
+                if magic == DIR_MAGIC:
+                    self._handle_dir(file, fs)
+                elif magic == FILE_MAGIC:
+                    self._handle_file(file, fs)
+
+    def _handle_dir(self, file: File, fs: FileSystem):
+        dir_header = self._struct_parser.parse("deafbead_dir_t", file, Endian.LITTLE)
+        fs.mkdir(self._convert_path(dir_header.name))
+
+    def _handle_file(self, file: File, fs: FileSystem):
+        file_header = self._struct_parser.parse("deafbead_file_t", file, Endian.LITTLE)
+        try:
+            decompressed = gzip.decompress(file_header.file_contents)
+            fs.write_bytes(self._convert_path(file_header.name), decompressed)
+        except gzip.BadGzipFile as error:
+            raise InvalidInputFormat("Invalid GZIP file") from error
+
+    @staticmethod
+    def _convert_path(path_entry: bytes) -> Path:
+        decoded_path = path_entry.decode("utf-8", errors="replace")
+        if "\\" in decoded_path:  # windows path => convert slashes
+            return Path(PureWindowsPath(decoded_path).as_posix())
+        return Path(decoded_path)
+
+
+class DeafBeadHandler(StructHandler):
+    NAME = "deafbead"
+    PATTERNS = [HexString("DE AF BE AD (86 | 87)")]
+
+    DOC = HandlerDoc(
+        name="D-Link DEAFBEAD",
+        description="Archive files as found in D-Link DSL-500G and DSL-504G firmware images.",
+        handler_type=HandlerType.ARCHIVE,
+        vendor="D-Link",
+        references=[],
+        limitations=[],
+    )
+
+    C_DEFINITIONS = C_DEFINITIONS
+    HEADER_STRUCT = "deafbead_header_t"
+    EXTRACTOR = DeafBeadExtractor()
+
+    def calculate_chunk(self, file: File, start_offset: int) -> ValidChunk | None:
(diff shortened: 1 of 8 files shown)


python/unblob/handlers/archive/dlink/deafbead.py
```diff
@@ -0,0 +1,113 @@
...
+        file_path.write_bytes(decompressed)
+
+    @staticmethod
+    def _combine_path(base: Path, rel_path: str) -> Path:
+        if "\\" in rel_path:  # the archive may contain windows paths => convert slashes
+            rel_path = PureWindowsPath(rel_path).as_posix()
+        return base / rel_path
+
+
+class DeafBeadHandler(Handler):
```
qkaiser: Should be a `StructHandler` since you define a `C_DEFINITIONS` and `HEADER_STRUCT`.


```suggestion
class DeafBeadHandler(StructHandler):
```

python/unblob/handlers/archive/dlink/deafbead.py
```diff
@@ -0,0 +1,113 @@
...
+                    break  # found end of file
+                if magic not in {DIR_MAGIC, FILE_MAGIC}:
+                    break
+                name_len = struct.unpack("<H", file.read(2))[0]
+                current_offset += 1 + 2 + name_len
+                file.seek(current_offset)
+                if magic == FILE_MAGIC:
+                    file_size = struct.unpack("<I", file.read(4))[0]
+                    current_offset += 4 + file_size
+                    file.seek(current_offset)
```
qkaiser: I would probably keep looping by checking magic.

Considering that we define ` VALID_MAGICS` like:

```python
VALID_MAGICS = [DIR_MAGIC, FILE_MAGIC]
```

You can do something along those lines:

```suggestion
            magic = file.read(1)
            while magic in VALID_MAGICS:
                file.seek(-1, io.SEEK_SET)
                if magic == DIR_MAGIC:
                    header = self.c_parser_le.deafbead_dir_t(file)
                else:
                    header = self.c_parser_le.deafbead_file_t(file)
                magic = file.read(1)
```

The `dissect.cstruct` parser will seek to the end of each entry on each loop.

You need to set the Handler as a `StructHandler` for `c_parser_le` to be exposed.

python/unblob/handlers/archive/dlink/deafbead.py
```diff
@@ -0,0 +1,113 @@
...
+        except struct.error:
+            return None
```
qkaiser: Don't catch exceptions here, let it be handled by unblob. Also I don't think you'll see these once the code stabilizes, looks like code added when debugging cstruct parsing :)

python/unblob/handlers/archive/dlink/deafbead.py
```diff
@@ -0,0 +1,113 @@
...
+
+        if current_offset > start_offset + HEADER_LEN:
+            return ValidChunk(start_offset=start_offset, end_offset=current_offset)
+        return None
```
qkaiser: If you follow what I commented above, you should be able to do something like:


```suggestion
        return ValidChunk(start_offset=start_offset, end_offset=file.tell())
```

python/unblob/handlers/archive/dlink/deafbead.py
```diff
@@ -0,0 +1,113 @@
...
+    EXTRACTOR = DeafBeadExtractor()
+
+    def calculate_chunk(self, file: File, start_offset: int) -> ValidChunk | None:
+        """DEAFBEAD format from D-Link DSL 50X routers.
+
+        structure: magic 0xDEAFBEAD followed by file or folder entries:
+        folder: 0x86 + uint16 (folder name length) + char[name_len] (folder name)
+        file: 0x87 + uint16 (file name length) + char[name_len] (file name)
+                   + uint32 (file size) + byte[file_size] (contents)
+        """
```
qkaiser: You can drop the comment. Explanations can go into the commit, but usually C structures you put in C_DEFINITIONS are quite self-explanatory.

python/unblob/handlers/archive/dlink/deafbead.py
```diff
@@ -0,0 +1,113 @@
...
+            while True:
+                if not (magic := file.read(1)):
+                    break  # EOF
+                file.seek(-1, io.SEEK_CUR)  # go back to read the full struct
+                if magic == DIR_MAGIC:
+                    self._handle_dir(file, outdir)
+                elif magic == FILE_MAGIC:
+                    self._handle_file(file, outdir)
+                else:
+                    break
```
qkaiser: Use the  `FileSystem` API so you have free protection against path traversals.

python/unblob/handlers/archive/dlink/deafbead.py
```diff
@@ -0,0 +1,113 @@
...
+            decompressed = gzip.decompress(file_header.file_contents)
+        except gzip.BadGzipFile as error:
+            raise InvalidInputFormat("Invalid GZIP file") from error
+        file_path.write_bytes(decompressed)
+
+    @staticmethod
+    def _combine_path(base: Path, rel_path: str) -> Path:
+        if "\\" in rel_path:  # the archive may contain windows paths => convert slashes
+            rel_path = PureWindowsPath(rel_path).as_posix()
+        return base / rel_path
```
qkaiser: This is vulnerable to path traversal. If I create a firmware with a name like `../../../etc/passwd`:

```
>>> from pathlib import Path
>>> base = Path("/tmp/out/test")
>>> rel_path = "../../../etc/passwd"
>>> base / rel_path
PosixPath('/tmp/out/test/../../../etc/passwd')
```

python/unblob/handlers/archive/dlink/deafbead.py
jstucke: I changed it so that it uses `FileSystem`

python/unblob/handlers/archive/dlink/deafbead.py
jstucke: see above

python/unblob/handlers/archive/dlink/deafbead.py
jstucke: removed the comment

python/unblob/handlers/archive/dlink/deafbead.py
jstucke: I changed it as suggested, but I needed to add a `-= 1` for the case where the file does not end after this chunk

python/unblob/handlers/archive/dlink/deafbead.py
jstucke: I removed the error handling removed but I added some checks for invalid values

python/unblob/handlers/archive/dlink/deafbead.py
jstucke: changed (mostly) as suggested

python/unblob/handlers/archive/dlink/deafbead.py
jstucke: replaced with the correct base class

jstucke:
> Thanks for contributing ! I left a few comments to improve readability / security. I'll probably get back to it over the weekend.
> 
> Note: we're quite opinionated about the Walrus operator internally so expect comments about it from the team :p

Hi and thank you for the detailed review. I changed all the parts you commented on and added some tests for invalid headers. If you don't like the walrus operator I can change it (IMHO here it helps) :sweat_smile:

martonilles:
Great that you have added a test sample, but can you please also add at least in the commit message, how the test sample was created (eg: a command line and tool you used, etc.)

jstucke:
> Great that you have added a test sample, but can you please also add at least in the commit message, how the test sample was created (eg: a command line and tool you used, etc.)

Since I didn't find any tools and the format is simple enough, I somewhat "frankensteined" the test sample together. I could write a packer script (but it would just be the extractor in reverse).

Of course there are "real" samples like these ones: https://files.dlink.com.au/products/DSL-504G/REV_D/Firmware/ (but I don't expect we can include them here)

qkaiser:
@jstucke took the liberty to:
- fix typing issues identified by CI
- add the missing DOC field
- generate doc with `uv run unblob --build-handlers-doc docs/handlers.md`

Code looks good. If you agree with the DOC field, I'll approve.

PR #1352, opened 2026-01-23
feat(handler): add support for squashfs v4 broadcom variants

jstucke:
Adds support for unpacking broadcom variants of squashfs v4 images (e.g. found in D-Link DSL-2640B and Asus DSL-N12HP)

- [x] needs https://github.com/onekey-sec/sasquatch/pull/34 to work

PR #1358, opened 2026-01-27
chore(deps): upgrade sasquatch to 4.5.1-6.

qkaiser:


qkaiser:
> looks ok, but the nix based dependency is not ugpraded, not sure if it needs to be handled here though

will be handled in https://github.com/onekey-sec/unblob/pull/1352 once https://github.com/NixOS/nixpkgs/pull/484387 is in

PR #1359, opened 2026-01-29
feat(handler): add support for minix filesystems

jstucke:
* little endian test files were created with mkfs.minix
  * since there does not seem to be a tool to create big endian MINIX filesystems, those test files were created by byte swapping the little endian test files
* known firmware samples with MINIX FS:
  * Trendnet IP cameras, e.g. TV-IP110W and TV-IP422WN (both MINIX FS v1 LE with 30 byte filenames)
  * Netgear switches, e.g. FSM7326P and GSM7248 (both MINIX FS v2 BE with 30 byte filenames)
  * I have not encountered MINIX v3 filesystems in firmware images, but adding it anyway was little overhead


jstucke:
hmm I noticed that I should probably add a few sanity checks to `calculate_chunk` to avoid false positives

python/unblob/handlers/filesystem/minixfs.py
```diff
@@ -0,0 +1,363 @@
...
+            if index == 0:
+                break
+            if indirectness > 0:
+                zone_pointers = self._get_zone_pointers(index)
+                yield from self._read_zones(zone_pointers, indirectness - 1)
+            else:
+                yield self._read_zone_data(index)
+
+    def _read_directory(self, inode) -> Iterator[cstruct]:
+        for entry in chunked(self._read_file_data(inode), self.dir_entry_size):
```
qkaiser: We have `iterate_file` in `file_utils` that could help you if you're doing what I think you're doing. Lots of things are already implemented in `file_utils`, so you probably don't need the `more-itertools` dependency either.

python/unblob/handlers/filesystem/minixfs.py
```diff
@@ -0,0 +1,363 @@
...
+    VERSION = 1
+    BLOCK_SIZE = 1_024
+
+    @staticmethod
+    def get_dependencies(*_):
+        return []
+
+    def __init__(self):
+        super().__init__()
+        self.EXTRACTOR = MinixFSExtractor(self)
```
qkaiser: I understand the attempt at generalization but I don't think this `self` passing is a good design. I'll try to come up with something better when I'm thru the review.

python/unblob/handlers/filesystem/minixfs.py
```diff
@@ -0,0 +1,363 @@
...
+
+    def _get_zone_count(self, header) -> int:
+        return header.s_nzones
+
+    def calculate_chunk(self, file: File, start_offset: int) -> ValidChunk | None:
+        file.seek(start_offset)
+        header = self.parse_header(file, self.ENDIANNESS)
+
+        # the superblock is preceded by a 1 KiB boot sector (also part of the FS; can be empty)
+        start_offset -= SUPERBLOCK_OFFSET
```
qkaiser: You can set `PATTERN_MATCH_OFFSET = - SUPERBLOCK_OFFSET` in the handler definition and unblob will magically consider that your start offset is where your pattern match + that value. You just need to take it into account in your parsing, but this will be transparent :)

With this you won't need lines 147-148

python/unblob/handlers/filesystem/minixfs.py
```diff
@@ -0,0 +1,363 @@
...
+        if start_offset < 0:
+            return None
+
+        zone_count = self._get_zone_count(header)
+        if zone_count != 0:
+            block_size = header.s_blocksize if self.VERSION == 3 else self.BLOCK_SIZE
```
qkaiser: I would probably define `BLOCK_SIZE` at the file level rather than within the handler. It affects more than the handler.

python/unblob/handlers/filesystem/minixfs.py
```diff
@@ -0,0 +1,363 @@
...
+        file.seek(start_offset)
+        header = self.parse_header(file, self.ENDIANNESS)
+
+        # the superblock is preceded by a 1 KiB boot sector (also part of the FS; can be empty)
+        start_offset -= SUPERBLOCK_OFFSET
+        if start_offset < 0:
+            return None
+
+        zone_count = self._get_zone_count(header)
+        if zone_count != 0:
```
qkaiser: Since zone_count is in the header, you can write patterns that do not match on headers with a null zone count. This way you don't need to get it here or check the value since it won't be matched in the first place.

python/unblob/handlers/filesystem/minixfs.py
```diff
@@ -0,0 +1,363 @@
...
+        HexString("[16] (68 | 78) 24 (00 | 01 | 02) 00 [4] 00 00 00 00"),
+    ]
+
+
+class MinixFSv2BEHandler(_MinixFSv2HandlerBase):
+    NAME = "minix_fs_v2_be"
+    ENDIANNESS = Endian.BIG
+    PATTERNS = [
+        HexString("[16] 24 (68 | 78) 00 (00 | 01 | 02) [4] 00 00 00 00"),
+    ]
```
qkaiser: You can probably generalize BE and LE handlers into one. Look at how it's done for squashfs. In a nutshell, you can define a `get_endian` function that looks at the beginning of your match and tells you whether it's LE or BE.

python/unblob/handlers/filesystem/minixfs.py
```diff
@@ -0,0 +1,363 @@
...
+        if lower_magic in {0x7F, 0x68}:
+            return 14  # v1/v2
+        if lower_magic in {0x8F, 0x78}:
+            return 30  # v1/v2
+        if lower_magic == 0x5A:
+            return 60  # v3
+        raise InvalidInputFormat(f"Invalid magic: {self.superblock.s_magic:x}")
+
+    def _read_zone_data(self, zone_index: int) -> bytes:
+        self.file.seek(zone_index * self.zone_size)
```
qkaiser: use explicit whence:


```suggestion
        self.file.seek(zone_index * self.zone_size, io.SEEK_SET)
```

python/unblob/handlers/filesystem/minixfs.py
```diff
@@ -0,0 +1,363 @@
...
+            else self.superblock.s_blocksize
+        )
+        self.imap_offset = 2 * self.block_size
+        self.zmap_offset = (
+            self.imap_offset + self.superblock.s_imap_blocks * self.block_size
+        )
+        self.inode_offset = (
+            self.zmap_offset + self.superblock.s_zmap_blocks * self.block_size
+        )
+        self.zone_size = self.block_size << self.superblock.s_log_zone_size
```
qkaiser: May want to "cast" it to uint32.

python/unblob/handlers/filesystem/minixfs.py
```diff
@@ -0,0 +1,363 @@
...
+            return 30  # v1/v2
+        if lower_magic == 0x5A:
+            return 60  # v3
+        raise InvalidInputFormat(f"Invalid magic: {self.superblock.s_magic:x}")
+
+    def _read_zone_data(self, zone_index: int) -> bytes:
+        self.file.seek(zone_index * self.zone_size)
+        return self.file.read(self.zone_size)
+
+    def _get_zone_pointers(self, zone: int) -> list[int]:
```
qkaiser: You use `zone_index` in `_read_zone_data`:


```suggestion
    def _get_zone_pointers(self, zone_index: int) -> list[int]:
```

python/unblob/handlers/filesystem/minixfs.py
```diff
@@ -0,0 +1,363 @@
...
+        data = self._read_zone_data(zone)
+        ptr_fmt = "H" if self.zone_ptr_size == 2 else "I"
+        count = self.zone_size // self.zone_ptr_size
+        endianness_fmt = "<" if self._handler.ENDIANNESS == Endian.LITTLE else ">"
+        return list(struct.unpack(f"{endianness_fmt}{count}{ptr_fmt}", data))
+
+    def _read_file_data(self, inode: cstruct) -> bytes:
```
qkaiser: Not sure about the `cstruct` typing here.

python/unblob/handlers/filesystem/minixfs.py
```diff
@@ -0,0 +1,363 @@
...
+            elif self._is_symlink(entry_inode):
+                contents = self._read_file_data(entry_inode)
+                link_target = contents.decode("utf-8", errors="replace")
+                fs.create_symlink(Path(link_target), Path(entry_path))
+
+            elif self._is_dir(entry_inode):
+                fs.mkdir(Path(entry_path), parents=True, exist_ok=True)
+                self.extract(fs, entry_inode, entry_path)
+            # there can be other special file types (e.g. /dev/tty) but we ignore them
+            # here, since we can't extract them
```
qkaiser: What about FIFOs ? Block or character devices ? We have methods to create them in `Filesystem`

python/unblob/handlers/filesystem/minixfs.py
```diff
@@ -0,0 +1,363 @@
...
+                raise InvalidInputFormat("Root entries should be directories")
+
+        for entry in self._read_directory(inode):
+            if entry.name in (b".", b"..") or entry.inode < 1:
+                continue
+            entry_path = f"{path}/{entry.name.decode('utf-8', errors='replace')}"
+            entry_inode = self._read_inode(entry.inode)
+
+            if self._is_file(entry_inode):
+                fs.write_bytes(Path(entry_path), self._read_file_data(entry_inode))
```
qkaiser: I'd recommend using `write_chunks` and implement `_read_file_data` as a generator that yields chunks of bytes. This way we don't have to load large files in memory.

qkaiser:
@jstucke your recent contributions are highly appreciated. Keep them coming ! :)

python/unblob/handlers/filesystem/minixfs.py
```diff
@@ -0,0 +1,363 @@
...
+
+        # the superblock is preceded by a 1 KiB boot sector (also part of the FS; can be empty)
+        start_offset -= SUPERBLOCK_OFFSET
+        if start_offset < 0:
+            return None
+
+        zone_count = self._get_zone_count(header)
+        if zone_count != 0:
+            block_size = header.s_blocksize if self.VERSION == 3 else self.BLOCK_SIZE
+            end_offset = start_offset + block_size * zone_count
```
qkaiser: This ignores `s_log_zone_size` even though the on-disk zone size can be larger than a block. On Minix images with `s_log_zone_size` > 0, this underestimates the filesystem size so the handler will carve a truncated chunk and extraction will miss trailing data. The end offset should be based on `block_size << s_log_zone_size`.

python/unblob/handlers/filesystem/minixfs.py
jstucke: nice, changed

python/unblob/handlers/filesystem/minixfs.py
jstucke: moved it to a constant

python/unblob/handlers/filesystem/minixfs.py
jstucke: Is there some better way to do this? I removed it for the time being

python/unblob/handlers/filesystem/minixfs.py
jstucke: Good idea, merged them and removed the base classes

jstucke:
> @jstucke your recent contributions are highly appreciated. Keep them coming ! :)

Haha thank you for taking the time for a thorough review. I added some validity checks and tests with invalid headers. I also found the [original tool](https://github.com/Stichting-MINIX-Research-Foundation/minix/tree/master/minix/usr.sbin/mkfs.mfs) from the minix project and was also able to create samples with block size  > 1024 and even log_zone_size != 0 (which is not possible with the mkfs.minix included in linux), but sadly I am unable to mount the filesystems and therefore could not add any files. The validity checks seem to work, though. There is are some images on the [minix download page](https://wiki.minix3.org/doku.php?id=www:download:start) which uses v3 with blocksize 4096 and it seems to be unpacked without problems.

I will probably fix the remaining issues tomorrow.

python/unblob/handlers/filesystem/minixfs.py
jstucke: How would I do that? There is the [not (~) syntax](https://yara.readthedocs.io/en/stable/writingrules.html#hexadecimal-strings) in YARA but this does not seem to work. At least I get an error when using it in a pattern:
```
>           raise InvalidHexString(str(e)) from e
E           unblob.parser.InvalidHexString: No terminal matches '~' in the current parser context, at line 1 col 44
E           
E           ] (7f | 8f) 13 (00 | 01 | 02) 00 [2] 00 ~00 00 00
E                                                   ^
E           Expected one of: 
E           	* START_ANCHOR
E           	* SECONDNIBLE
E           	* JUMP
E           	* END_ANCHOR
E           	* LPAR
E           	* WILDCARD
E           	* RANGE_JUMP
E           	* RPAR
E           	* FIRSTNIBLE
E           	* LITERAL
E           	* ALTERNATIVE_SEPARATOR
E           
E           Previous tokens: Token('LITERAL', '00')
```

python/unblob/handlers/filesystem/minixfs.py
jstucke: I changed it, so that `_read_file_data` streams the file data instead (and renamed it to `_stream_file_data`)

python/unblob/handlers/filesystem/minixfs.py
jstucke: fixed

python/unblob/handlers/filesystem/minixfs.py
jstucke: changed

python/unblob/handlers/filesystem/minixfs.py
jstucke: changed

python/unblob/handlers/filesystem/minixfs.py
jstucke: I added support for all of them

python/unblob/handlers/filesystem/minixfs.py
jstucke: You are absolutely right, fixed

python/unblob/handlers/filesystem/minixfs.py
jstucke: I changed `_read_file_data` to stream chunks of data, but these chunks have block size and not `dir_entry_size`, so `chunked()` is still beneficial (making `_read_file_data` yield chunks of size `dir_entry_size` is a bit much overhead because of the indirect data zone structure). 

 I could replace the chunked with slices. That would remove the dependency, but would not really improve readability:

```python
            for dir_data in chunked(zone_data, self.dir_entry_size):
```

```python
            for index in range(0, len(zone_data), self.dir_entry_size):
                dir_data = zone_data[index : index + self.dir_entry_size]
```

python/unblob/handlers/filesystem/minixfs.py
qkaiser: That's something we may want to fix in unblob itself. Naive approach is to accept everything but 00:

```
( 01 | 02 | 03 | ... | ff)
```

You may have an easier time writing patterns as `Regex` rathern than `HexString` for that purpose.

python/unblob/handlers/filesystem/minixfs.py
qkaiser: Actual type is `Structure`, see https://github.com/fox-it/dissect.cstruct/blob/main/dissect/cstruct/types/structure.py#L384

python/unblob/handlers/filesystem/minixfs.py
```diff
@@ -0,0 +1,416 @@
...
+                inode = self._read_inode(ROOT_INODE_INDEX)
+            except EOFError as error:
+                raise InvalidInputFormat("File system is empty") from error
+            if not self._is_dir(inode):
+                raise InvalidInputFormat("Root entries should be directories")
+
+        for entry in self._read_directory(inode):
+            if entry.name in (b".", b"..") or entry.inode < 1:
+                continue
+            entry_path = f"{path}/{entry.name.decode('utf-8', errors='replace')}"
```
qkaiser: You convert to `str` using a format string, and then you convert it back to `Path` on line 375, 380, 383, 388, 391.

Do this instead and drop the `Path` casting:


```suggestion
            entry_path = path / entry.name.decode('utf-8', errors='replace')
```

python/unblob/handlers/filesystem/minixfs.py
```diff
@@ -0,0 +1,416 @@
...
+        for zone_data in self._stream_file_data(inode):
+            for dir_data in chunked(zone_data, self.dir_entry_size):
+                if len(dir_data) < self.dir_entry_size:
+                    # this should never happen, because zone_size should always be a multiple of dir_entry_size
+                    break
+                yield self.struct_parser.parse(
+                    self._handler.DIR_STRUCT, bytes(dir_data), self.endianness
+                )
+
+    def _read_inode(self, index: int) -> cstruct:
```
qkaiser:
```suggestion
    def _read_inode(self, index: int) -> Structure:
```

python/unblob/handlers/filesystem/minixfs.py
```diff
@@ -0,0 +1,416 @@
...
+        for index in zone_index_list:
+            if index == 0:
+                break
+            if indirectness > 0:
+                zone_pointers = self._get_zone_pointers(index)
+                yield from self._read_zones(zone_pointers, indirectness - 1)
+            else:
+                yield self._read_zone_data(index)
+
+    def _read_directory(self, inode) -> Iterator[cstruct]:
```
qkaiser:
```suggestion
    def _read_directory(self, inode) -> Iterator[Structure]:
```

python/unblob/handlers/filesystem/minixfs.py
```diff
@@ -0,0 +1,416 @@
...
+                    break
+                yield self.struct_parser.parse(
+                    self._handler.DIR_STRUCT, bytes(dir_data), self.endianness
+                )
+
+    def _read_inode(self, index: int) -> cstruct:
+        if not 1 <= index <= self.superblock.s_ninodes:
+            raise InvalidInputFormat(f"Invalid inode number: {index}")
+        offset = self.inode_offset + (index - 1) * self.inode_size
+        self.file.seek(offset)
```
qkaiser: Explicit whence:


```suggestion
        self.file.seek(offset, io.SEEK_SET)
```

python/unblob/handlers/filesystem/minixfs.py
qkaiser: You should define a `_MinixFSExtractor` class that do not override the `__init__` function of `Extractor`. Then you create 3 extractors:
- `MinixFSv1Extractor(_MinixFSExtractor)`
- `MinixFSv2Extractor(_MinixFSExtractor)`
- `MinixFSv3Extractor(_MinixFSExtractor)`

None of these extractors should rely on the handler they're attached to. No `handlers` argument or attribute here.

You should define a `_MinixFSHandler(StructHandler)` that does not override the `__init__` function of `StructHandler`. This generic handler should define:
- `HEADER_STRUCT` as `minix_super_block`

Then, create three different handlers:
- `MinixFSv1Handler(_MinixFSHandler)`
- `MinixFSv2Handler(_MinixFSHandler)`
- `MinixFSv3Handler(_MinixFSHandler)`

Each of these handlers must override the C_DEFINITIONS entry with their own. To do so I would split the C_DEFINITIONS in 3 blocks:
- C_DEFINITIONS_V1 (only `minix_inode`, `minix_super_block`, `minix_dir_entry`)
- C_DEFINITIONS_V2 (only `minix2_inode`, `minix_super_block`, `minix_dir_entry`)
- C_DEFINITIONS_V3  (only `minix_inode`, `minix3_super_block`, `minix3_dir_entry`)

In each of these C definitions, use the same names. Do not use `minix2_inode` vs `minix_inode`. Use generic names without version info. That's the reason we can define `HEADER_STRUCT` in the handler parent class.

With this approach you can get rid of `DIR_STRUCT` and `INODE_STRUCT` since they will be generic. Just use the generic name.

You can also drop the `VERSION` class attribute. The `VERSION` is currently used for two things:
- derive the `block_size`, just check if your header has an `s_blocksize` attribute. If it does, use it otherwise default to `BLOCK_SIZE`
- derive the `inode_size`, this can be done through the cstruct parser that expose its fields size (e.g. `self._struct_parser.cparser_le.minix_inode.size`

Do not define `DOC` in `__init__`, just define it in the class with a version that's not resolved at runtime.

Each handler must set its extractor explicitly like:
- `EXTRACTOR = MinixFSv3Extractor()`

This should make the code easier to follow and reduce complexity.

jstucke: Did you come up with anything? I'm still a bit puzzled how to solve this (since both handler and extractor both need many of the fields including the `C_DEFINITIONS`). Is overriding `Handler.extract` an option? :sweat_smile: 
edit: it seems I did not refresh the page before commenting

python/unblob/handlers/filesystem/minixfs.py
jstucke: I changed it to use Structure (but since that riddled the code with missing attribute warnings, I added more concrete subclasses as type hints (maybe that was overkill, though))

python/unblob/handlers/filesystem/minixfs.py
jstucke: fixed :see_no_evil:

python/unblob/handlers/filesystem/minixfs.py
jstucke: fixed

python/unblob/handlers/filesystem/minixfs.py
jstucke: fixed

python/unblob/handlers/filesystem/minixfs.py
jstucke: changed

python/unblob/handlers/filesystem/minixfs.py
jstucke: I tried writing a `Regex` instead, but it seems that it does not like null bytes:

```python
    @lru_cache
    def build_hyperscan_database(handlers: Handlers) -> StreamDatabase:
        patterns = []
        for handler_class in handlers:
            handler = handler_class()
            for pattern in handler.PATTERNS:
                try:
                    patterns.append(
                        Pattern(
                            pattern.as_regex(),
                            Flag.SOM_LEFTMOST,
                            Flag.DOTALL,
                            tag=handler,
                        )
                    )
                except InvalidHexString as e:
                    logger.error(
                        "Invalid pattern",
                        handler=handler.NAME,
                        pattern=pattern,
                        error=str(e),
                    )
                    raise
>       return StreamDatabase(*patterns)
               ^^^^^^^^^^^^^^^^^^^^^^^^^
E       ValueError: Pattern expression contains NULL byte
```

Maybe it is best to do sanity checking during chunk extraction and change it as soon as it is supported?

python/unblob/handlers/filesystem/minixfs.py
jstucke: I tried to change everything like you suggested. I couldn't get completely rid of VERSION, though.

qkaiser:
@jstucke I'll do some cleanups on this branch sometimes this week, but not sure when. Hang on :)

qkaiser:
Cleaning up the code and doing cross-validation as we speak. Will push my changes when I'm done.

qkaiser:
> Cleaning up the code and doing cross-validation as we speak. Will push my changes when I'm done.

@jstucke Still ongoing, will keep you posted.

qkaiser:
@jstucke to me this is ready but since I was involved in the code I'd prefer someone else from our team to give final approval. Thanks again for your patience.

PR #1360, opened 2026-01-29
feat(models): allow multi-pattern globbing

qkaiser:
Allow Glob to accept multiple filename patterns so directory handlers can match several globs since pathlib's glob lacks alternation support.

Designed this way so we don't introduce breaking API changes, especially in `DirectoryHandler` that already use `Glob`.

I'm working on an internal `DirectoryHandler` that requires this very feature.

### DIFF
--- a/tests/test_models.py
+++ b/tests/test_models.py
@@ -6,6 +6,7 @@ import pytest
 from unblob.file_utils import InvalidInputFormat
 from unblob.models import (
     Chunk,
+    Glob,
     ProcessResult,
     ReportModelAdapter,
     Task,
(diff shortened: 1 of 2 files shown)


tests/test_models.py
```diff
@@ -302,3 +303,24 @@ def test_exotic_command_output(self):
         assert decoded_report == report
 
         decoded_report = json.loads(json_text)
+
+
+class TestGlob:
+    def test_single_pattern(self, tmp_path: Path):
+        (tmp_path / "a.txt").write_text("a", encoding="utf-8")
```
kissgyorgy: you don't need encoding, everything is utf-8 by default

tests/test_models.py
```diff
@@ -302,3 +303,24 @@ def test_exotic_command_output(self):
...
+        (tmp_path / "b.log").write_text("b", encoding="utf-8")
+
+        glob = Glob("*.txt")
+        matches = {path.name for path in glob.get_files(tmp_path)}
```
kissgyorgy: I would simplify this and assert on `Path`
```suggestion
        path_a = tmp_path / "a.txt"
        ....
        assert list(get_files(tmp_path)) == [path_a]
```

tests/test_models.py
```diff
@@ -302,3 +303,24 @@ def test_exotic_command_output(self):
...
+
+        assert matches == {"a.txt"}
+
+    def test_multiple_patterns(self, tmp_path: Path):
+        (tmp_path / "a.txt").write_text("a", encoding="utf-8")
```
kissgyorgy: same as above

tests/test_models.py
```diff
@@ -302,3 +303,24 @@ def test_exotic_command_output(self):
...
+        (tmp_path / "b.log").write_text("b", encoding="utf-8")
+        (tmp_path / "c.bin").write_text("c", encoding="utf-8")
+
+        glob = Glob("*.txt", "*.log")
+        matches = {path.name for path in glob.get_files(tmp_path)}
```
kissgyorgy: same as above

qkaiser:
Wait I can just `touch` instead of writing :)

kissgyorgy (approved):

PR #1368, opened 2026-02-02
chore(deps): split dev dependencies in sub-groups

qkaiser:
We have multiple dev dependencies serving different purposes (testing, type checking, code formatting, file linting, fuzzing). All of these dependencies are built or installed in multiple CI steps that only requires a subset of them.

Introduced the 'uv-groups' argument to the setup-dependencies action. The action now always execute uv with `--no-dev`, and we can provide an explicit dev subgroup so it's installed.

The dev group stays the same.

### DIFF
--- a/.github/actions/setup-dependencies/action.yml
+++ b/.github/actions/setup-dependencies/action.yml
@@ -7,6 +7,10 @@ inputs:
     description: "Wether to install 3rd Party dependencies (for tests)"
     required: false
     default: "true" # unfortunately boolean variables are not supported
+  uv-groups:
+    description: "Space-separated uv dependency groups to install"
+    required: false
+    default: ""
 
 name: "Setup dependencies"
 description: "Install all required dependencies for worflows to run."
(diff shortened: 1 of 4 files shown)


.github/actions/setup-dependencies/action.yml
```diff
@@ -38,7 +42,15 @@ runs:
...
-      run: uv sync
+      run: |
+        set -e
+        group_args=""
+        if [ -n "${{ inputs.uv-groups }}" ]; then
+          for group in ${{ inputs.uv-groups }}; do
+            group_args="$group_args --group ${group}"
+          done
+        fi
+        uv sync --no-dev ${group_args}
```
vlaci, line 53: As this is bash, we can use arrays. Also, empty loops like `for f in ; do ...; done` are legit: 

```suggestion
        set -e
        declare -a group_args=()
        for group in ${{ inputs.uv-groups }}; do
          group_args+=(--group ${group})
        done
        uv sync --no-dev "${group_args[@]}"
```

vlaci: Sigh, I forgot to actually post this comment. No need to change this now, but I think it is good to know about these.

PR #1369, opened 2026-02-02
Upgrade macos runners

qkaiser:
macos-13 runners have been officially deprecated by Github, see https://github.com/actions/runner-images/issues/13046


qkaiser:
That's why our jobs are [failing](https://github.com/onekey-sec/unblob/actions/runs/21590946422/job/62211312017?pr=1368)

PR #1374, opened 2026-02-02
fix(extractor): fix report typing issue with ExtractCommandFailedReport

qkaiser:
The `ExtractCommandFailedReport` class only accepts bytes for its stdout and stderr fields.

The `subprocess.run` call that executes the extraction command returns a `subprocess.CompletedProcess` object, where its stdout and stderr can be either `bytes`, `str`, or `None` depending on how it was called:
> stderr
Captured stderr from the child process. A bytes sequence, or a string if [run()](https://docs.python.org/3/library/subprocess.html#subprocess.run) was called with an encoding, errors, or text=True. None if stderr was not captured.
> stdout
Captured stdout from the child process. A bytes sequence, or a string if [run()](https://docs.python.org/3/library/subprocess.html#subprocess.run) was called with an encoding, errors, or text=True. None if stdout was not captured.

 Source: https://docs.python.org/3/library/subprocess.html#subprocess.CompletedProcess

This discrepancy triggered pydantic validation error, especially in the gzip handler:

```
File
"/nix/store/zgivnxh4vs6cwappddrmqfmbxx8f3l7m-inspector-env/lib/python3.13/site-packages/unblob/handlers/compression/gzip.py", line 92, in extract
    return extractor.extract(inpath, outdir)
           ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^
  File
"/nix/store/zgivnxh4vs6cwappddrmqfmbxx8f3l7m-inspector-env/lib/python3.13/site-packages/unblob/extractors/command.py", line 60, in extract
    error_report = ExtractCommandFailedReport(
        command=command,
    ...&lt;2 lines&gt;...
        exit_code=res.returncode,
    )
  File
"/nix/store/zgivnxh4vs6cwappddrmqfmbxx8f3l7m-inspector-env/lib/python3.13/site-packages/pydantic/main.py", line 250, in __init__
    validated_self = self.__pydantic_validator__.validate_python(data,
self_instance=self)
pydantic_core._pydantic_core.ValidationError: 1 validation error for ExtractCommandFailedReport
stdout
  Input should be a valid bytes [type=bytes_type, input_value=None,
input_type=NoneType]
    For further information visit
https://errors.pydantic.dev/2.12/v/bytes_type
```

That's because this handler, and a few others such as compress, lzma, bzip2, and xz redirects stdout to an output file:

```
grep stdout= python/unblob/handlers/ -r                
python/unblob/handlers/compression/gzip.py:        extractor = Command("7z", "x", "-y", "{inpath}", "-so", stdout=name)
python/unblob/handlers/compression/gzip.py:            "7z", "x", "-p", "-y", "{inpath}", "-so", stdout=name
python/unblob/handlers/compression/compress.py:    EXTRACTOR = Command("7z", "x", "-y", "{inpath}", "-so", stdout="lzw.uncompressed")
python/unblob/handlers/compression/lzma.py:    EXTRACTOR = Command("7z", "x", "-y", "{inpath}", "-so", stdout="lzma.uncompressed")
python/unblob/handlers/compression/bzip2.py:    EXTRACTOR = Command("7z", "x", "-y", "{inpath}", "-so", stdout="bzip2.uncompressed")
python/unblob/handlers/compression/xz.py:    EXTRACTOR = Command("7z", "x", "-y", "{inpath}", "-so", stdout="xz.uncompressed")
```

This leads to `res.stdout` being None since "None if stdout was not captured.".

Some notes:
- We now normalize these fields to bytes to make sure the model is strictly followed.
- This is unrelated to 3513bcc since pydantic models of unblob reports were already validated prior.
- It is still unclear why this was not picked up by our type checker. Ideally this type checking miss should be fixed too. I tried a bunch of things to hint pyright to no avail.

The blast radius of this bug is limited to gzip, compress, lzma, bzip2 and xz handler when the decompression command fails. Still, I'd like to push for a timely review here.

### DIFF
--- a/python/unblob/report.py
+++ b/python/unblob/report.py
@@ -6,18 +6,29 @@ import stat
 import traceback
 from enum import Enum
 from pathlib import Path
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Annotated, Any
 
 if TYPE_CHECKING:
     from collections.abc import Iterable
 
 from pydantic import (
     BaseModel,
+    BeforeValidator,
     ConfigDict,
     computed_field,
     field_serializer,
     field_validator,
 )
+
+
+def ensure_bytes(value: Any) -> bytes:
+    if isinstance(value, bytes):
+        return value
+    if isinstance(value, str):
+        return value.encode()
+    if value is None:
+        return b""
+    raise ValueError(f"Unsupported type in ensure_bytes: {type(value)}")
 
 
 class Report(BaseModel):
@@ -82,8 +93,8 @@ class ExtractCommandFailedReport(ErrorReport):
 
     severity: Severity = Severity.WARNING
     command: str
-    stdout: bytes
-    stderr: bytes
+    stdout: Annotated[bytes, BeforeValidator(ensure_bytes)]
+    stderr: Annotated[bytes, BeforeValidator(ensure_bytes)]
     exit_code: int
 
     # Use base64 to encode and decode bytes data in case there are non-standard characters
--- a/tests/extractors/test_command.py
+++ b/tests/extractors/test_command.py
@@ -72,6 +72,32 @@ def test_command_execution_failure(tmpdir: Path):
     ]
 
 
+def test_command_execution_failure_with_stdout_redirect(tmpdir: Path):
+    outdir = PosixPath(tmpdir)
+    command = Command(
+        "python",
+        "-c",
+        "import sys\n"
+        "sys.stdout.write('stdout')\n"
+        "sys.stderr.write('stderr')\n"
+        "sys.exit(1)",
+        stdout="out.txt",
+    )
+
+    with pytest.raises(ExtractError) as excinfo:
+        command.extract(Path("input"), outdir)
+
+    assert (outdir / "out.txt").read_bytes() == b"stdout"
+    assert list(excinfo.value.reports) == [
+        ExtractCommandFailedReport.model_construct(
+            command=mock.ANY,
+            stdout=b"",
+            stderr=b"stderr",
+            exit_code=1,
+        )
+    ]
+
+
 def test_command_not_found(tmpdir: Path):
     outdir = PosixPath(tmpdir)
     command = Command("this-command-should-not-exist-in-any-system")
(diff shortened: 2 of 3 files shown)


martonilles:
Just wondering, but wouldn't it be better to actually store strings instead of bytes in the report. If I understand correctly it is usually strings and mainly meant for human consumption, so eventually those are converted to string. Not sure we want to open that Pandora's box now though...

qkaiser:
> Just wondering, but wouldn't it be better to actually store strings instead of bytes in the report. If I understand correctly it is usually strings and mainly meant for human consumption, so eventually those are converted to string. Not sure we want to open that Pandora's box now though...

Fair point, but I'd save it for another PR.

python/unblob/extractors/command.py
```diff
@@ -59,8 +67,8 @@ def no_op():
             if res.returncode != 0:
                 error_report = ExtractCommandFailedReport(
```
vlaci, line 60: IMO this retyping should be part of a pydantic before validator[^1] instead.

[^1]: https://docs.pydantic.dev/latest/concepts/validators/#field-validators

qkaiser: YES !

qkaiser: Done

python/unblob/report.py
```diff
@@ -8,20 +8,29 @@
...
     field_validator,
 )
 
 
+def ensure_bytes(value: Any) -> bytes:
+    if isinstance(value, bytes):
+        return value
+    if isinstance(value, str):
+        return value.encode()
+    return b""
```
vlaci: Should we silently convert other non-None values to empty instead of failing or logging an error at least?

vlaci (approved):

python/unblob/report.py
qkaiser: Now raising a `ValueError` if `value` is neither `str`, `bytes` or `None`. Tested in `test_extract_command_failed_invalid`.

qkaiser:
@vlaci do you think we really need a test if the test is [picked up by pyright](https://github.com/onekey-sec/unblob/actions/runs/21831607353/job/62991242368?pr=1374#step:4:33) ?

PR #1378, opened 2026-02-09
Upgrade ruff target from python 3.9 to python 3.10

qkaiser:
We've dropped support for python 3.9 since 4632b05 but forgot to tell Ruff.

Fxing B905 and UP036 with `--unsafe-fixes` kept in their own commits so it's obvious they were not auto-fixed

### DIFF
--- a/python/unblob/handlers/filesystem/romfs.py
+++ b/python/unblob/handlers/filesystem/romfs.py
@@ -1,10 +1,11 @@
+from __future__ import annotations
+
 import io
 import os
 import stat
 import struct
 from enum import IntEnum, unique
 from pathlib import Path
-from typing import Optional
 
 from structlog import get_logger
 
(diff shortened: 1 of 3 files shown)


vlaci:
There are some unfixed parts:

```
python/unblob/models.py:8:from typing import Generic, Optional, TypeVar
python/unblob/models.py:142:    def extract(self, inpath: Path, outdir: Path) -> Optional["ExtractResult"]:
python/unblob/models.py:216:    def extract(self, outdir: Path) -> Optional["ExtractResult"]:
python/unblob/report.py:11:from typing import TYPE_CHECKING, Any, Optional, Union
python/unblob/report.py:49:    exception: Union[str, Exception]
python/unblob/report.py:146:    link_target: Optional[Path]
python/unblob/report.py:236:    randomness: Optional[RandomnessReport]
python/unblob/report.py:240:    def validate_randomness(cls, value: Any) -> Optional[RandomnessReport]:
python/unblob/report.py:288:    path: Optional[str] = None
python/unblob/handlers/filesystem/romfs.py:7:from typing import Optional
python/unblob/handlers/filesystem/romfs.py:89:    parent: Optional["FileHeader"] = None
```

qkaiser:
@vlaci took care of missing parts manually, see 4cd8529

PR #1386, opened 2026-02-13
feat(ai): add AGENTS.md to help coding agents

qkaiser:
Our contributors will have an easier time working in this repo with coding agents if we provide this file :)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,52 @@
+# Repository Guidelines
+
+## Project Structure & Module Organization
+- `python/unblob/`: Python package, CLI, core logic, handlers.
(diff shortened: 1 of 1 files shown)


kissgyorgy:
There are multiple studies, papers and videos recently which measured the performance of agents with and without the AGENTS.md file and they found that not even they make performance (code quality) worse, but even makes the inference cost 20% higher: https://arxiv.org/abs/2602.11988

I would either not include this or craft it manually, very carefully which I'm not sure how to do 😄 
but definitely leave out unhelpful sections completely like `## Coding Style & Naming Conventions`

One thing I noticed might help to list relevant directories/files to guide the agent for faster discovery for a given section of the codebase.

qkaiser:
I'm happy dropping the MR if research results demonstrate it's bad

PR #1392, opened 2026-02-17
chore(deps): migrate from p7zip to 7zip

qkaiser:
Up until now we've been using p7zip with backported patches from Debian but we don't think it provides sufficient security guarantees anymore.

Through Nix derivation we were pulling version 16.02 with backport patches from Debian. In our Dockerfile we were pulling the bookworm version, which is 22.01. This is also problematic since we're not relying on the same version depending on environment.

Given recent vulnerabilities affecting 7zip (CVE-2025-11002 (path traversal), CVE-2023-52168 (heap overflow), CVE-2023-52169 (oob read)), we consider it safer to simply pull from the latest version provided by official maintainers of 7zip.

This partly goes against the rationale of
7adce8e35eea4dbaf2601cc06bc627bcf7a448de, which is related to MBR extraction. This will be handled by having our own internal MBR extractor instead of relying on 7zip, given its behavior.

### DIFF
--- a/overlay.nix
+++ b/overlay.nix
@@ -8,30 +8,21 @@ final: prev:
     nativeCheckInputs = (super.nativeCheckInputs or [ ]) ++ [ final.which ];
   });
 
-  p7zip16 = prev.p7zip.overrideAttrs (super: rec {
-    pname = "p7zip16";
-    version = "16.02";
-    srcs = [
-      (final.fetchurl {
-        url = "mirror://sourceforge/p7zip/p7zip_${version}_src_all.tar.bz2";
-        sha256 = "5eb20ac0e2944f6cb9c2d51dd6c4518941c185347d4089ea89087ffdd6e2341f";
-      })
-      (final.fetchurl {
-        url = "http://deb.debian.org/debian/pool/main/p/p7zip/p7zip_${version}+dfsg-8.debian.tar.xz";
-        sha256 = "sha256-ASF9yhZnrw3kiTWlHcRqrUQubryseZ1xQQG37fllHrU=";
-      })
-    ];
-    sourceRoot = "p7zip_${version}";
-    nativeBuildInputs = (super.nativeBuildInputs or [ ]) ++ [ final.quilt ];
-    prePatch = ''
-      export QUILT_PATCHES=../debian/patches
-      quilt push -a
-    '';
-    # orig had `src` attribute, but we are using `srcs`. This trips a warning.
-    __intentionallyOverridingVersion = true;
-
-    separateDebugInfo = true;
-  });
+  sevenzip =
+    let
+      inherit (final) _7zz;
+      _7z-link = final.runCommand "_7z-link" { } ''
+        mkdir -p $out/bin
+        ln -sfn ${_7zz}/bin/7zz "$out/bin/7z"
+      '';
+    in
+    final.symlinkJoin {
+      name = "sevenzip";
+      paths = [
+        _7zz
+        _7z-link
+      ];
+    };
 
   erofs-utils = prev.erofs-utils.overrideAttrs (_: rec {
     version = "1.8.10";
--- a/install-deps.sh
+++ b/install-deps.sh
@@ -9,13 +9,27 @@ apt-get install --no-install-recommends -y \
     lz4 \
     lziprecover \
     lzop \
-    p7zip-full \
     partclone \
     unar \
     upx \
     xz-utils \
     libmagic1 \
     zstd
+
+case "$(dpkg --print-architecture)" in
+    amd64) sevenzip_arch="x64" ;;
+    arm64) sevenzip_arch="arm64" ;;
+    *)
+        echo "Unsupported architecture for 7-Zip: $(dpkg --print-architecture)" >&2
+        exit 1
+        ;;
+esac
+
+curl -L -o 7zip.tar.xz "https://www.7-zip.org/a/7z2600-linux-${sevenzip_arch}.tar.xz"
+install -d /usr/local/bin
+tar -xf 7zip.tar.xz -C /usr/local/bin 7zz 7zzs
+ln -sf /usr/local/bin/7zz /usr/local/bin/7z
+rm -f 7zip.tar.xz
 
 curl -L -o sasquatch_1.0.deb "https://github.com/onekey-sec/sasquatch/releases/download/sasquatch-v4.5.1-6/sasquatch_1.0_$(dpkg --print-architecture).deb"
 dpkg -i sasquatch_1.0.deb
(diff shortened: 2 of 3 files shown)


overlay.nix
```diff
@@ -8,30 +8,60 @@ final: prev:
...
+          url = "${baseUrl}/7z${versionNoDots}-linux-arm64.tar.xz";
+          sha256 = "0zf8dhbjq74z2jnpbk96wrqi0l31was8iiqfmz9p95mg3453v3xa";
+        }
+      else if final.stdenvNoCC.hostPlatform.isLinux && final.stdenvNoCC.hostPlatform.isx86_64 then
+        final.fetchurl {
+          url = "${baseUrl}/7z${versionNoDots}-linux-x64.tar.xz";
+          sha256 = "1ipdfz225js41iirv9k298754vx6n8zxa47cbwzy9kcjhjjc8kf7";
+        }
+      else
+        throw "sevenzip: unsupported platform ${final.stdenvNoCC.hostPlatform.system}";
```
vlaci: You can simplify this if-chain:
Also note, that we need to switch based on `targetPlatform`, not the host, as in theory one can cross-compile the package.

```suggestion
        baseUrl = "https://www.7-zip.org/a";
        versionNoDots = final.lib.replaceStrings [ "." ] [ "" ] version;
        sources = {
          x86_64-linux = {
            url = "${baseUrl}/7z${versionNoDots}-linux-x64.tar.xz";
            sha256 = "0jf8a688lwlp67jvxg91n8231skv2c53xwgmd3avfb1bnlsafbla";
          };
          aarch64-linux = {
            url = "${baseUrl}/7z${versionNoDots}-linux-arm64.tar.xz";
            sha256 = "0zf8dhbjq74z2jnpbk96wrqi0l31was8iiqfmz9p95mg3453v3xa";
          };
          x86_64-darwin = {
            url = "${baseUrl}/7z${versionNoDots}-mac.tar.xz";
            sha256 = "1ipdfz225js41iirv9k298754vx6n8zxa47cbwzy9kcjhjjc8kf7";
          };
          aarch64-darwin = {
            url = "${baseUrl}/7z${versionNoDots}-mac.tar.xz";
            sha256 = "1ipdfz225js41iirv9k298754vx6n8zxa47cbwzy9kcjhjjc8kf7";
          };
        };
      in
      fetchurl sources."${stdenv.targetPlatform}";
```

Instead of the build-time exception, add a `meta` tag. This will trip the build in eval-time. 

```
  meta = {
    platforms = [
      "x86_64-linux"
      "aarch64-linux"
      "x86_64-darwin"
      "aarch64-darwin"
    ];
  };
```

qkaiser:
workflow got stuck for some reason ...

qkaiser:
It's also getting stuck locally when running `nix flake check -L --option system aarch64-linux --extra-platforms aarch64-linux`, need to investigate what's going on and why it's not happening on x86.

```
quentin   200169  0.7  0.9 2287008 464120 pts/0  Sl+  10:03   0:16 nix flake check -L --option system aarch64-linux --extra-platforms aarch64-linux
quentin   200254  0.1  0.0 227704 19072 ?        Ssl  10:03   0:02 /usr/libexec/qemu-binfmt/aarch64-binfmt-P /nix/store/mxq13s0brxw1qdmh2qghdgd2gbk3hpwj-bash-5.3p9/bin/bash bash -e /nix/store/l622p70vy8k5sh7y5wizi5f2mic6ynpg-source-stdenv.sh /nix/store/shkw4qm9qcw5sc5n1k5jznc83ny02r39-default-builder.sh
quentin   206874  6.0  0.5 609960 289552 ?       Sl   10:09   1:50 /usr/libexec/qemu-binfmt/aarch64-binfmt-P /nix/store/ygzfhw5nxrn7qqfqmz2s9p6cikcjqqkh-python3-3.13.11/bin/python3.13 /nix/store/ygzfhw5nxrn7qqfqmz2s9p6cikcjqqkh-python3-3.13.11/bin/python3.13 -m pytest
```

qkaiser:
`nix flake check -L` runs without errors or issues on `Linux debian 6.1.0-26-arm64 #1 SMP Debian 6.1.112-1 (2024-09-30) aarch64 GNU/Linux`. Rebased to see what's happening, could be related to qemu aarch64 emulation.

qkaiser:
Something's wrong with their macos workers.

overlay.nix
```diff
@@ -8,30 +8,71 @@ final: prev:
...
-    srcs = [
-      (final.fetchurl {
-        url = "mirror://sourceforge/p7zip/p7zip_${version}_src_all.tar.bz2";
-        sha256 = "5eb20ac0e2944f6cb9c2d51dd6c4518941c185347d4089ea89087ffdd6e2341f";
-      })
-      (final.fetchurl {
-        url = "http://deb.debian.org/debian/pool/main/p/p7zip/p7zip_${version}+dfsg-8.debian.tar.xz";
-        sha256 = "sha256-ASF9yhZnrw3kiTWlHcRqrUQubryseZ1xQQG37fllHrU=";
-      })
+  sevenzip = final.stdenvNoCC.mkDerivation rec {
```
vlaci: Since last week, 7z is updated to 26.00 in nixpkgs as well

https://github.com/NixOS/nixpkgs/blob/master/pkgs/by-name/_7/_7zz/package.nix

qkaiser: Excellent point. Done.

overlay.nix
```diff
@@ -8,29 +8,11 @@ final: prev:
...
-    sourceRoot = "p7zip_${version}";
-    nativeBuildInputs = (super.nativeBuildInputs or [ ]) ++ [ final.quilt ];
-    prePatch = ''
-      export QUILT_PATCHES=../debian/patches
-      quilt push -a
+  sevenzip = prev._7zz.overrideAttrs (oldAttrs: {
+    pname = "sevenzip";
+    postInstall = (oldAttrs.postInstall or "") + ''
+      ln -sfn 7zz "$out/bin/7z"
     '';
```
vlaci: This invalidates the previous derivation for the sake of adding one symlink.
See the logs proving that we need to recompile _7zz: https://github.com/onekey-sec/unblob/actions/runs/22766754946/job/66036503844?pr=1392#step:6:531

For better cache-reuse we can do some trickery layering a new derivation on top of the original one:


```suggestion
  sevenzip =
    let
      inherit (final) _7zz;
      _7z-link = final.runCommand "_7z-link" { } ''
        mkdir -p $out/bin
        ln -sfn ${_7zz}/bin/7zz "$out/bin/7z"
      '';
    in
    final.symlinkJoin {
      name = "sevenzip";
      paths = [
        _7zz
        _7z-link
      ];
    };
```

vlaci: or you can just do every link creation in the runCommand call, that works as well

vlaci (approved):

PR #1393, opened 2026-02-17
fix(doc): collapse repeated separators in fragment identifiers

qkaiser:
Some `HandlerDoc` name may be set to things like `Format A | Format B` or `Format A / Format A variant`, which leads to fragment identifier set to `#format-a--format-a-variant` while it should be set to `#format-a-format-a-variant`.

Fixed by collapsing repeating `-` character using regular expression matching.


rasakozi:
lgtm!

Issue #1175, opened 2025-04-15
dissect.cstruct size field shadowing


qkaiser:
I'd like to add a semgrep scan in pre-commit and CI to make sure that we're not using fields named `size` in our C definitions consumed by `StructParser`.

It's not causing problems right now, but I recently discovered that having a `size` field in the struct shadows the `size` attribute of the generated object, which is set to the size of that object. It can become a problem when using that `size` attribute in complex parsing situations. Just want to remove that footgun.

Here's an example:

```
>>> cstruct().load("typedef struct something { uint32 magic; uint32 size; }").something(b"AAAABBBB").size
1111638594
>>> cstruct().load("typedef struct something { uint32 magic; uint32 my_size; }").something(b"AAAABBBB").size
8
```

PR #1395, opened 2026-02-19
feat: added d-link alpha encimg archive handler

jstucke:
Some D-Link devices manufactured by Alpha Networks mangle key and IV of the AES encryption by XORing them with the device name. Every device has a unique device name, key and IV value. Since the first few blocks of the encrypted firmware are constant across versions, it is possible to write signatures even for the encrypted firmware images. The key/IV values originally stem from GPL source releases of D-Link.

References:
- https://github.com/openwrt/openwrt/pull/13930
- https://github.com/devttys0/delink/blob/main/src/encimg.rs
- https://thottysploity.github.io/posts/dir-842/

There are more devices that use this format (e.g. DIR 822 C1, 865L, 880L, 885L, 890L and 895L), but the keys were not released AFAIK (you can retrieve the values from a "middle firmware" image as described in the blog post, though)

### DIFF
--- a/python/unblob/handlers/archive/dlink/alpha_encimg.py
+++ b/python/unblob/handlers/archive/dlink/alpha_encimg.py
@@ -1,0 +1,344 @@
+from __future__ import annotations
+
+import io
+import os
+from dataclasses import dataclass
+from pathlib import Path
+from typing import TYPE_CHECKING
+
+from cryptography.hazmat.backends import default_backend
+from cryptography.hazmat.primitives.ciphers import (
+    Cipher,
+    algorithms,
+    modes,
+)
+
+from unblob._rust import math_tools
+from unblob.file_utils import Endian, File, FileSystem, InvalidInputFormat, StructParser
+from unblob.models import (
+    Extractor,
+    ExtractResult,
+    HandlerDoc,
+    HandlerType,
+    HexString,
+    Reference,
+    StructHandler,
+    ValidChunk,
+)
+
+if TYPE_CHECKING:
+    from collections.abc import Iterable
+
+    from cryptography.hazmat.primitives.ciphers import CipherContext
+
+C_DEFINITIONS = r"""
+    typedef struct dlink_alpha_header {
+        uint32 magic1;
+        uint32 signature_len;
+        uint32 padding;
+        char   signature[signature_len];    /* "signature=XXXXXX_dlink.YYYY_ZZZZZZ\0" */
+        // where XXXXXX seems to be the Alpha Networks model number, YYYY the year
+        // and ZZZZZZ the D-Link model number
+        uint32 magic2;
+        uint32 signature_len2;
+        uint32 fsize;                       /* size of the firmware image (without 128 Bytes header) */
+    } dlink_alpha_header_t;
+
+    // see https://github.com/openwrt/firmware-utils/blob/master/src/mkwrggimg.c
+    typedef struct wrgg03_header {
+        char        signature[32];    /* signature without "signature=" */
+        uint32      magic1;
+        uint32      magic2;
+        char        version[16];
+        char        model[16];
+        uint32      flag[2];
+        uint32  	reserve[2];
+        char        buildno[16];
+        uint32  	fsize;
+        uint32  	offset;
+        char		devname[32];
+        char		digest[16];
+    } wrgg03_header_t;
+"""
+XOR_RANGE = 0xFC
+UNENCRYPTED_ENTROPY_THRESHOLD = 3
+
+
+@dataclass
+class EncParams:
+    signature: bytes
+    key: bytes
+    iv: bytes
+
+    @property
+    def mangled_key(self):
+        return self._mangle(self.signature, self.key)
+
+    @property
+    def mangled_iv(self):
+        return self._mangle(self.signature, self.iv)
+
+    @staticmethod
+    def _mangle(image_sign: bytes, data: bytes) -> bytes:
+        sign_len = len(image_sign)
+        return bytes(
+            data_byte ^ ((i + 1) % XOR_RANGE) ^ image_sign[i % sign_len]
+            for i, data_byte in enumerate(data)
+        )
+
+
+@dataclass
+class EncParamsV2(EncParams):
+    key: bytes = b"oVhq0hvXHdfaGFLdubM4/QvuVHdKee7v"
+    iv: bytes = b"0BO5nlYankuVBe4s"
+
+
+ENC_START_TO_PARAMS = {
+    # DAP-1665 B
+    bytes.fromhex("35 66 6f 68"): EncParams(
+        signature=b"wapac25_dlink.2015_dap1665",
+        key=b"EfCHXytwsC6F0zsedwZc+9vDbCjE3ge4",
+        iv=b"ggPy917jwESpnfXm",
+    ),
+    # DAP-1720 A
+    bytes.fromhex("68 01 cc fb"): EncParams(
+        signature=b"wapac28_dlink.2015_dap1720",
+        key=b"qBiz6o/1RVQTtJBd3FS7FDbqogE8yoBm",
+        iv=b"EfDMqWWxHCOhEqgY",
+    ),
+    # DIR-822 C1
+    bytes.fromhex("df 8c 39 0d"): EncParams(
+        signature=b"wrgac43s_dlink.2015_dir822c1",
+        key=b"KNpsEntCcsep1jdFIs3wnXySKRGNCGmf",
+        iv=b"uph587JdKHrtAUlr",
+    ),
+    # DIR-842 C1 / C2
+    bytes.fromhex("f5 2a a0 b4"): EncParams(
+        signature=b"wrgac65_dlink.2015_dir842",
+        key=b"xQYoRZeD726UAbRb846kO7TeNw8eZa6u",
+        iv=b"zufEbNF3kUafxFiE",
+    ),
+    # DIR-842 C3 (same as C1 except for the "EU" at the end of the signature)
+    bytes.fromhex("21 dd da 00"): EncParams(
+        signature=b"wrgac65_dlink.2015_dir842EU",
+        key=b"xQYoRZeD726UAbRb846kO7TeNw8eZa6u",
+        iv=b"zufEbNF3kUafxFiE",
+    ),
+    # DIR-850L A1
+    bytes.fromhex("e3 13 00 5b"): EncParams(
+        signature=b"wrgac05_dlob.hans_dir850l",
+        key=b"BIuS1CVMEQG+0pUeE99jnR+vLlLd9unr",
+        iv=b"f3+odwHhmJL1ceW1",
+    ),
+    # DIR-850L B1
+    bytes.fromhex("0a 14 e4 24"): EncParams(
+        signature=b"wrgac25_dlink.2013gui_dir850l",
+        key=b"qQehHMEmEPQ5izL+cabn8bNHZXHjkp6W",
+        iv=b"Mmb+IKQgnO8OuF4b",
+    ),
+    # DIR-859 A
+    bytes.fromhex("4c 1b 95 af"): EncParams(
+        signature=b"wrgac37_dlink.2013gui_dir859",
+        key=b"KY0H9R2PDL3eu1J4uCVd1CK7BJ7vF1kc",
+        iv=b"qbStAzIRvWeQHz5U",
+    ),
+    # FixMe: add more entries when they are found
+}
+
+
+def decrypt_chunk(encrypted: bytes, enc_params: EncParams) -> bytes:
+    decryptor = _get_decryptor(enc_params)
+    return decryptor.update(encrypted) + decryptor.finalize()
+
+
+def decrypt_file(
+    file: File, enc_params: EncParams, chunk_size: int = 16
+) -> Iterable[bytes]:
+    decryptor = _get_decryptor(enc_params)
+    while chunk := file.read(chunk_size):
+        yield decryptor.update(chunk)
+    yield decryptor.finalize()
+
+
+def _get_decryptor(enc_params: EncParams) -> CipherContext:
+    cipher = Cipher(
+        algorithms.AES(enc_params.mangled_key),
+        modes.CBC(enc_params.mangled_iv),
+        backend=default_backend(),
+    )
+    return cipher.decryptor()
+
+
+class AlphaEncimgExtractorV1(Extractor):
+    def extract(self, inpath: Path, outdir: Path) -> ExtractResult | None:
+        fs = FileSystem(outdir)
+        with File.from_path(inpath) as file:
+            enc_magic = file.read(4)
+            enc_params = ENC_START_TO_PARAMS.get(enc_magic)
+            if not enc_params:
+                raise InvalidInputFormat("Device not supported")
+            file.seek(0, io.SEEK_SET)
+            out_path = Path(f"{enc_params.signature.decode()}.bin")
+            fs.write_chunks(out_path, decrypt_file(file, enc_params))
+        return ExtractResult(reports=fs.problems)
+
+
+class AlphaEncimgExtractorV2(Extractor):
+    def extract(self, inpath: Path, outdir: Path) -> ExtractResult | None:
+        struct_parser = StructParser(C_DEFINITIONS)
+        fs = FileSystem(outdir)
+        with File.from_path(inpath) as file:
+            header = struct_parser.parse("wrgg03_header_t", file, Endian.BIG)
+            enc_params = EncParamsV2(signature=header.signature.rstrip(b"\0"))
+            file.seek(AlphaEncimgV2Handler.HEADER_SIZE, io.SEEK_SET)
+            out_path = Path(f"{enc_params.signature.decode()}.bin")
+            fs.write_chunks(out_path, decrypt_file(file, enc_params))
+        return ExtractResult(reports=fs.problems)
+
+
+class AlphaEncimgHandler(StructHandler):
+    NAME = "alpha_encimg_v1"
+    PATTERNS = [
+        HexString("35 66 6f 68 ef 1a fe 1f 34 ef 4f 11 21 05 4e be"),  # DAP-1665 B
+        HexString("68 01 cc fb ad 6b a0 ba 33 04 b0 9c bb 48 b0 27"),  # DAP-1720 A
+        HexString("df 8c 39 0d 22 b4 dc 29 fb 4e bf db e8 e1 8b fb"),  # DIR-822 C1
+        HexString("f5 2a a0 b4 92 53 bf ef f8 21 a6 2e 28 a7 39 8b"),  # DIR-842 C1 / C2
+        HexString("21 dd da 00 99 d8 87 a9 d5 2d 7e ff 3b 58 70 a6"),  # DIR-842 C3
+        HexString("e3 13 00 5b 76 df 0b e8 83 24 5a 42 ff 91 2d 3d"),  # DIR-850L A1
+        HexString("0a 14 e4 24 ff 0f b4 d7 53 66 a0 b0 72 fe ab df"),  # DIR-850L B1
+        HexString("4c 1b 95 af 93 72 5f 81 03 03 96 4d dd 76 01 74"),  # DIR-859 A
+        HexString("98 9c 3c 2c 6b 19 1b 25 6e 03 ee 0f 72 5c 6c a0"),  # DIR-865L A
+        HexString("3d e6 02 63 3f c1 85 e7 d2 bf 64 af 29 57 3a bd"),  # DIR-868L A
+        HexString("8e 69 57 e7 76 09 d7 94 47 75 78 a2 4a 1f c9 b2"),  # DIR-880L A
+        HexString("92 61 58 58 14 db bb 3b e5 a5 f3 e7 10 9c a2 0b"),  # DIR-885L A
+        HexString("cb c3 a7 4c 50 0f 42 43 a5 d9 7c a5 25 6b cd ba"),  # DIR-890L A
+        HexString("91 5a c9 a2 11 ff aa 6d b0 12 e8 8d 2c 3c 23 cb"),  # DIR-895L A
+    ]
+    EXTRACTOR = AlphaEncimgExtractorV1()
+    C_DEFINITIONS = C_DEFINITIONS
+    DOC = HandlerDoc(
+        name="D-Link Alpha encimg V1 Firmware",
+        description=(
+            "Encrypted firmware images found in D-Link DIR devices manufactured by Alpha Networks."
+            "Uses AES-256-CBC encryption with device-specific keys."
+        ),
+        handler_type=HandlerType.ARCHIVE,
+        vendor="D-Link",
+        references=[
+            Reference(
+                title="OpenWRT forum",
+                url="https://forum.openwrt.org/t/adding-openwrt-support-for-d-link-dir-x1860-mt7621-mt7915-ax1800/106500",
+            ),
+            Reference(
+                title="delink tool",
+                url="https://github.com/devttys0/delink/blob/main/src/encimg.rs",
+            ),
+        ],
+        limitations=[],
+    )
+    HEADER_STRUCT = "dlink_alpha_header_t"
+    HEADER_SIZE = 0x80
+    _SIGNATURE_PREFIX = b"signature="
+    MAGIC = {0x5EA3A417}
+
+    def _get_header(self, file: File):
+        encrypted_header = file.read(self.HEADER_SIZE)
+        enc_params = ENC_START_TO_PARAMS.get(encrypted_header[:4])
+        if not enc_params:
+            raise InvalidInputFormat("Device not supported")
+
+        decrypted_header = decrypt_chunk(encrypted_header, enc_params)
+        header = self._struct_parser.parse(
+            self.HEADER_STRUCT, decrypted_header, Endian.BIG
+        )
+        self._validate(enc_params, file, header)
+        return header
+
+    def calculate_chunk(self, file: File, start_offset: int) -> ValidChunk | None:
+        header = self._get_header(file)
+        end_offset = start_offset + self.HEADER_SIZE + header.fsize
+        return ValidChunk(start_offset=start_offset, end_offset=end_offset)
+
+    def _validate(self, enc_params: EncParams, file: File, header):
+        if header.magic1 not in self.MAGIC:
+            raise InvalidInputFormat(f"Invalid magic: {header.magic1}")
+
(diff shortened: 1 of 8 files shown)


python/unblob/handlers/archive/dlink/alpha_encimg.py
```diff
@@ -0,0 +1,288 @@
...
+    )
+    return cipher.decryptor()
+
+
+class AlphaEncimgExtractor(Extractor):
+    def extract(self, inpath: Path, outdir: Path) -> ExtractResult | None:
+        fs = FileSystem(outdir)
+
+        with File.from_path(inpath) as file:
+            file.seek(0, io.SEEK_SET)
```
qkaiser: Not needed.

python/unblob/handlers/archive/dlink/alpha_encimg.py
```diff
@@ -0,0 +1,288 @@
...
+        key=b"KY0H9R2PDL3eu1J4uCVd1CK7BJ7vF1kc",
+        iv=b"qbStAzIRvWeQHz5U",
+    ),
+    # DAP-2610
+    bytes.fromhex("8e a3 0e 0a"): EncParams(
+        signature=b"wapac30_dkbs_dap2610",
+        key=b"oVhq0hvXHdfaGFLdubM4/QvuVHdKee7v",
+        iv=b"0BO5nlYankuVBe4s",
+    ),
+    # FixMe: add missing devices
```
qkaiser: yes ?

python/unblob/handlers/archive/dlink/alpha_encimg.py
```diff
@@ -0,0 +1,285 @@
...
+        enc_params = ENC_START_TO_PARAMS.get(encrypted_header[:4])
+        if not enc_params:
+            raise InvalidInputFormat("Device not supported")
+
+        try:
+            decrypted_header = decrypt_chunk(encrypted_header, enc_params)
+            header = self._struct_parser.parse(
+                self.HEADER_STRUCT, decrypted_header, Endian.BIG
+            )
+        except Exception as err:
```
qkaiser: This is too generic. What kind of exception are you really trying to catch ?

python/unblob/handlers/archive/dlink/alpha_encimg.py
```diff
@@ -0,0 +1,285 @@
...
+
+    C_DEFINITIONS = C_DEFINITIONS
+    HEADER_STRUCT = "dlink_alpha_header_t"
+    HEADER_SIZE = 0x80
+    _SIGNATURE_PREFIX = b"signature="
+
+    def calculate_chunk(self, file: File, start_offset: int) -> ValidChunk | None:
+        encrypted_header = file.read(self.HEADER_SIZE)
+        if len(encrypted_header) < self.HEADER_SIZE:
+            raise InvalidInputFormat("File too small")
```
qkaiser: You can drop this. Your matching patterns make sure we have a file of at least 16 bytes, which gets us to header parsing. If `dissect.cstruct` can't read enough bytes for the given struct, it will raise an exception that will get caught downstream by unblob itself.

python/unblob/handlers/archive/dlink/alpha_encimg.py
```diff
@@ -0,0 +1,285 @@
...
+            )
+        except Exception as err:
+            raise InvalidInputFormat("Failed to decrypt header") from err
+
+        self._validate(enc_params, file, header)
+
+        end_offset = start_offset + self.HEADER_SIZE + header.size
+        return ValidChunk(start_offset=start_offset, end_offset=end_offset)
+
+    @classmethod
```
qkaiser: Why do you need it to be a classmethod ?

python/unblob/handlers/archive/dlink/alpha_encimg.py
```diff
@@ -0,0 +1,285 @@
...
+    PATTERNS = [
+        HexString("8e a3 0e 0a 7a c2 40 b3 bc 33 9e 3c 13 43 08 02"),  # DAP-2610 A
+    ]
+    HEADER_SIZE = 0x90
+    PATTERN_MATCH_OFFSET = -HEADER_SIZE
+    HEADER_STRUCT = "dlink_alpha2_header_t"
+    _SIGNATURE_PREFIX = b""
+
+    def calculate_chunk(self, file: File, start_offset: int) -> ValidChunk | None:
+        file.seek(start_offset, io.SEEK_SET)
```
qkaiser: Not needed.


```suggestion
```

python/unblob/handlers/archive/dlink/alpha_encimg.py
```diff
@@ -0,0 +1,285 @@
...
+        header = self.parse_header(file, Endian.LITTLE)
+
+        enc_params = ENC_START_TO_PARAMS.get(file.read(4))
```
qkaiser: What if enc_params is `None` ?

python/unblob/handlers/archive/dlink/alpha_encimg.py
```diff
@@ -0,0 +1,285 @@
...
+                title="delink tool",
+                url="https://github.com/devttys0/delink/blob/main/src/encimg.rs",
+            ),
+        ],
+        limitations=[],
+    )
+
+    C_DEFINITIONS = C_DEFINITIONS
+    HEADER_STRUCT = "dlink_alpha_header_t"
+    HEADER_SIZE = 0x80
```
qkaiser, line 240: The `signature_len` value is fixed for all headers ? It never changes ?

python/unblob/handlers/archive/dlink/alpha_encimg.py
jstucke: I mean I have more (at least the ones listed in the description) ... I'm just not sure about the legal implications of being the one to release them into the wild

python/unblob/handlers/archive/dlink/alpha_encimg.py
jstucke: Good question but hard to say for sure. At least for the samples that I looked at. There are more than just the DAP-2610 A that use this format (e.g. DAP-2230 and DAP-2310), but it is the only one I found at the moment that also uses this encryption. Since it is the first field of the header, it would probably be somewhat unpractical for it to have variable length, though.

python/unblob/handlers/archive/dlink/alpha_encimg.py
jstucke: Good point. Added a check analogously to the other handler

python/unblob/handlers/archive/dlink/alpha_encimg.py
jstucke: removed

python/unblob/handlers/archive/dlink/alpha_encimg.py
jstucke: well it does not _need_ to be a classmethod :sweat_smile: 
since it is only called from `calculate_chunk` there is no real benefit from it, so I converted it to a regular method

python/unblob/handlers/archive/dlink/alpha_encimg.py
jstucke: this is more of a debugging artifact, so I removed it

python/unblob/handlers/archive/dlink/alpha_encimg.py
jstucke: removed

python/unblob/handlers/archive/dlink/alpha_encimg.py
qkaiser: up to you, but i would leave the fixme line then if you don't add them. We can cross that bridge when/if we look into these ourselves.

jstucke: I added the FixMe back and I'll also leave this convenient script here for future reference:

<details>

```python
import argparse
from subprocess import check_call
from pathlib import Path
import sys
import re

OUTPUT_BASE = """
=== ENC_START_TO_PARAMS ===
    # {model}
    bytes.fromhex("{short_pattern}"): EncParams(
        signature=b"{signature}",
        key=b"{key}",
        iv=b"{iv}",
    ),

=== PATTERN ===
        HexString("{pattern}"),  # {model}
"""
ENCIMG_REGEX = re.compile(rb"[a-zA-Z0-9+/]{76}\n[a-zA-Z0-9+/]{51}")


def _parse_args() -> argparse.Namespace:
    parser = argparse.ArgumentParser()
    parser.add_argument(
        "middle_fw",
        help="The middle firmware image that introduces encryption",
        type=Path,
    )
    parser.add_argument("enc_fw", help="An encrypted firwmare image", type=Path)
    return parser.parse_args()


def main():
    args = _parse_args()
    assert args.middle_fw.is_file(), f"middle_fw file {args.file} not found"
    assert args.enc_fw.is_file(), f"enc_fw file {args.file} not found"
    check_call(f"unblob -f {args.middle_fw}", shell=True)
    extraction_folder = args.middle_fw.parent / f"{args.middle_fw.name}_extract"
    assert extraction_folder.is_dir(), "extraction not successful"
    try:
        encimg_file = list(extraction_folder.glob("**/encimg"))[0]
    except IndexError:
        print("encimg file not found in extracted firmware")
        sys.exit(1)
    matches = ENCIMG_REGEX.findall(encimg_file.read_bytes())
    if len(matches) != 1:
        print("key/iv not found")
        sys.exit(1)
    key, iv = matches[0][0x20:0x40], matches[0][0x60:0x70]
    try:
        image_sign_file = list(extraction_folder.glob("**/image_sign"))[0]
    except IndexError:
        print("image_sign file not found in extracted firmware")
        sys.exit(1)
    signature = image_sign_file.read_text().rstrip("\n")
    model = signature.split("_")[-1].upper()
    model = model[:3] + "-" + model[3:]
    with args.enc_fw.open("rb") as fp:
        pattern = fp.read(16)

    print(
        OUTPUT_BASE.format(
            pattern=pattern.hex(" "),
            short_pattern=pattern[:4].hex(" "),
            model=model,
            signature=signature,
            key=key.decode(),
            iv=iv.decode(),
        )
    )


if __name__ == "__main__":
    main()
```

</details>

python/unblob/handlers/archive/dlink/alpha_encimg.py
```diff
@@ -0,0 +1,280 @@
...
+def decrypt_file(
+    file: File, enc_params: EncParams, chunk_size: int = 16
+) -> Iterable[bytes]:
+    decryptor = _get_decryptor(enc_params)
+    while chunk := file.read(chunk_size):
+        yield decryptor.update(chunk)
+    yield decryptor.finalize()
+
+
+def _get_decryptor(enc_params: EncParams) -> AEADDecryptionContext:
```
qkaiser:
```
/home/runner/work/unblob/unblob/python/unblob/handlers/archive/dlink/alpha_encimg.py
  /home/runner/work/unblob/unblob/python/unblob/handlers/archive/dlink/alpha_encimg.py:161:12 - error: Type "CipherContext" is not assignable to return type "AEADDecryptionContext"
    "CipherContext" is not assignable to "AEADDecryptionContext" (reportReturnType)
```


```suggestion
def _get_decryptor(enc_params: EncParams) -> CipherContext:
```

python/unblob/handlers/archive/dlink/alpha_encimg.py
```diff
@@ -0,0 +1,280 @@
...
+                title="OpenWRT forum",
+                url="https://forum.openwrt.org/t/adding-openwrt-support-for-d-link-dir-x1860-mt7621-mt7915-ax1800/106500",
+            ),
+            Reference(
+                title="delink tool",
+                url="https://github.com/devttys0/delink/blob/main/src/encimg.rs",
+            ),
+        ],
+        limitations=[],
+    )
```
qkaiser, line 238: Don't forget to generate documentation: `uv run unblob --build-handlers-doc docs/handlers.md`

python/unblob/handlers/archive/dlink/alpha_encimg.py
```diff
@@ -0,0 +1,280 @@
...
+        if header.size > file.size():
+            raise InvalidInputFormat("Invalid file size")
+
+        if header.size % 16:
+            raise InvalidInputFormat(
+                f"Firmware size not aligned to 16 bytes: {header.size}"
+            )
+
+
+class AlphaEncimgV2Handler(AlphaEncimgHandler):
```
qkaiser, line 280: I would add a specific `DOC` field to differentiate between v1 and v2 like we do for other handlers.

jstucke: I added a doc field also for v2

python/unblob/handlers/archive/dlink/alpha_encimg.py
jstucke: done

python/unblob/handlers/archive/dlink/alpha_encimg.py
jstucke: fixed

jstucke:
Hmm somehow `test_missing_handlers_integrations_tests` seems to expect the folder structure
```
tests/integration/archive/dlink/
├── alpha_encimg
│   ├── alpha_encimg_v2
│   │   ├── __input__
│   │   └── __output__
│   ├── __input__
│   └── __output__
```

instead of 

```
tests/integration/archive/dlink/
├── alpha_encimg
│   ├── __input__
│   └── __output__
├── alpha_encimg_v2
│   ├── __input__
│   └── __output__
```

Why though? :sweat_smile:  Should I maybe rename the first handler to "alpha_encimg_v1"?

jstucke:
I just found some relevant infos and references which I would like to add. Will set the PR to draft while doing so.

jstucke:
I found some documentation which also describes the header format of "variant 2": https://openwrt.org/toh/d-link/d-link_dap_series_of_business_access_points#old_generation_dap-2xxxdap-3xxx_built_by_alpha_networks

I also found some more encrypted samples of the DAP-2610 and noticed that the first encrypted blocks are not constant in this variant, so I had to change how the parameters are matched to the files for this variant.

jstucke:
I have good news and bad news:
- I have found another key inside a GPL release (DAP-2682) and added it
- variant 2 (WRGG03) seems to appear in big and little endian and I had to change some things to make it work
- you should probably look at the code again, because a lot has changed (sorry about that!)

jstucke:
Sorry again, this was the last fix I swear :sweat_smile: 
Key/IV of variant 2 seem to be constant (didn't notice that the keys were the same earlier, sorry!). It should unpack all (encrypted) samples now (DAP 2610, 2680 and 2682)

python/unblob/handlers/archive/dlink/alpha_encimg.py
```diff
@@ -0,0 +1,325 @@
...
+    # Unlike the variant above, this one uses an unencrypted WRGG03 header.
+    # The encryption, which starts directly after the header, is the same, though.
```
qkaiser: This is part of the description, you can drop this.


```suggestion
```

python/unblob/handlers/archive/dlink/alpha_encimg.py
```diff
@@ -0,0 +1,325 @@
...
+    MAGIC = {0x20080321, 0x21030820}
+
+    def _get_header(self, file: File):
+        current_offset = file.tell()
+        header = self.parse_header(file, Endian.LITTLE)
+        if not self._size_is_plausible(header.size):
+            # there are little and big endian devices that use this format, but there seems to be
+            # only one mandatory integer header field: size (sadly we cannot write signatures for that)
+            file.seek(current_offset, io.SEEK_SET)
+            header = self.parse_header(file, Endian.BIG)
```
qkaiser: Can't you read the header signature to detect whether it's big-endian or little-endian ? If you really want to use the size, I would parse it both as little and big endian, then keep the header with the smallest size between the two.

Edit: you can use the magic1 field to detect whether it's big-endian or little-endian, you have both in `MAGIC`.

python/unblob/handlers/archive/dlink/alpha_encimg.py
```diff
@@ -0,0 +1,325 @@
...
+            uint32 signature_len;
+            uint32 padding;
+            char   signature[signature_len];    /* "signature=XXXXXX_dlink.YYYY_ZZZZZZ\0" */
+            // where XXXXXX seems to be the Alpha Networks model number, YYYY the year
+            // and ZZZZZZ the D-Link model number
+            uint32 magic2;
+            uint32 signature_len2;
+            uint32 size;                        /* size of the firmware image (without 128 Bytes header) */
+        } dlink_alpha_header_t;
+    """
```
qkaiser: I would move this into the `WRGG_HEADER` constant and rename that constant to `C_DEFINITIONS`.

python/unblob/handlers/archive/dlink/alpha_encimg.py
```diff
@@ -0,0 +1,325 @@
...
+            out_path = Path(f"{enc_params.signature.decode()}.bin")
+            fs.write_chunks(out_path, decrypt_file(file, enc_params))
+
+
+class AlphaEncimgExtractorV2(Extractor):
+    def extract(self, inpath: Path, outdir: Path) -> ExtractResult | None:
+        struct_parser = StructParser(WRGG_HEADER)
+        fs = FileSystem(outdir)
+        with File.from_path(inpath) as file:
+            header = struct_parser.parse("wrgg03_header_t", file, Endian.BIG)
```
qkaiser, line 191: Could be little-endian, no ?

python/unblob/handlers/archive/dlink/alpha_encimg.py
```diff
@@ -0,0 +1,325 @@
...
+        fs = FileSystem(outdir)
+        with File.from_path(inpath) as file:
+            enc_magic = file.read(4)
+            enc_params = ENC_START_TO_PARAMS.get(enc_magic)
+            if not enc_params:
+                raise InvalidInputFormat("Device not supported")
+            file.seek(0, io.SEEK_SET)
+            out_path = Path(f"{enc_params.signature.decode()}.bin")
+            fs.write_chunks(out_path, decrypt_file(file, enc_params))
+
```
qkaiser, line 184:
```suggestion
        return ExtractResult(reports=fs.problems)
```

python/unblob/handlers/archive/dlink/alpha_encimg.py
```diff
@@ -0,0 +1,325 @@
...
+    def extract(self, inpath: Path, outdir: Path) -> ExtractResult | None:
+        struct_parser = StructParser(WRGG_HEADER)
+        fs = FileSystem(outdir)
+        with File.from_path(inpath) as file:
+            header = struct_parser.parse("wrgg03_header_t", file, Endian.BIG)
+            enc_params = EncParamsV2(signature=header.signature.rstrip(b"\0"))
+            file.seek(AlphaEncimgV2Handler.HEADER_SIZE, io.SEEK_SET)
+            out_path = Path(f"{enc_params.signature.decode()}.bin")
+            fs.write_chunks(out_path, decrypt_file(file, enc_params))
+
```
qkaiser, line 197:
```suggestion
        return ExtractResult(reports=fs.problems)
```

python/unblob/handlers/archive/dlink/alpha_encimg.py
jstucke: The only field that seems to have different endianness is the size field. Here are two examples (size is at offset 0x68):

<details>

DAP-2610:

```
00000000: 7761 7061 6333 305f 646b 6273 5f64 6170  wapac30_dkbs_dap
00000010: 3236 3130 0000 0000 0000 0000 0000 0000  2610............
00000020: 2103 0820 2103 0820 7632 2e30 3172 3036  !.. !.. v2.01r06
00000030: 3773 3833 3731 3400 6461 7032 3631 3000  7s83714.dap2610.
00000040: 0000 0000 0000 0000 0000 0000 0000 0000  ................
00000050: 0000 0000 0000 0000 6a61 6b61 0000 0000  ........jaka....
00000060: 0000 0000 0000 0000 20d0 9f00 0000 0000  ........ .......
00000070: 2f64 6576 2f6d 7464 626c 6f63 6b2f 3800  /dev/mtdblock/8.
00000080: 0000 0000 0000 0000 0000 0000 0000 0000  ................
00000090: f9ef 92a7 b633 78d9 065c 2641 cfd7 fe67  .....3x..\&A...g
```

DAP-2680:

```
00000000: 7761 7061 6333 365f 646b 6273 5f64 6170  wapac36_dkbs_dap
00000010: 3236 3830 0000 0000 0000 0000 0000 0000  2680............
00000020: 2103 0820 2103 0820 7631 3030 7263 3031  !.. !.. v100rc01
00000030: 3176 3233 3136 7337 6461 7032 3638 3000  1v2316s7dap2680.
00000040: 0000 0000 0000 0000 0000 0000 0000 0000  ................
00000050: 0000 0000 0000 0000 6934 3462 0000 0000  ........i44b....
00000060: 0000 0000 0000 0000 00ae 3030 0000 0000  ..........00....
00000070: 2f64 6576 2f6d 7464 626c 6f63 6b2f 3100  /dev/mtdblock/1.
00000080: 0000 0000 0000 0000 0000 0000 0000 0000  ................
00000090: 4ab9 ecf3 e557 84a1 4649 ca60 e023 26ca  J....W..FI.`.#&.
```

</details>

...which means sadly we cannot use the magic field. Keeping the header with the smaller size should work, though.

python/unblob/handlers/archive/dlink/alpha_encimg.py
jstucke: Yes, but since we only use `header.signature` here, it does not make a difference

jstucke:
There was another unexpected problem: with the more general pattern for v2, unencrypted samples were also matched and "decrypted" (rendering them unusable). I added some validation to prevent this.

python/unblob/handlers/archive/dlink/alpha_encimg.py
```diff
@@ -0,0 +1,341 @@
...
+    typedef struct dlink_alpha_header {
+        uint32 magic1;
+        uint32 signature_len;
+        uint32 padding;
+        char   signature[signature_len];    /* "signature=XXXXXX_dlink.YYYY_ZZZZZZ\0" */
+        // where XXXXXX seems to be the Alpha Networks model number, YYYY the year
+        // and ZZZZZZ the D-Link model number
+        uint32 magic2;
+        uint32 signature_len2;
+        uint32 size;                        /* size of the firmware image (without 128 Bytes header) */
```
qkaiser: We recommend against using `size` because it shadows cstruct's `size` attribute that returns the header size:

```suggestion
        uint32 fsize;                        /* size of the firmware image (without 128 Bytes header) */
```

python/unblob/handlers/archive/dlink/alpha_encimg.py
```diff
@@ -0,0 +1,341 @@
...
+    typedef struct wrgg03_header {
+        char        signature[32];    /* signature without "signature=" */
+        uint32      magic1;
+        uint32      magic2;
+        char        version[16];
+        char        model[16];
+        uint32      flag[2];
+        uint32  	reserve[2];
+        char        buildno[16];
+        uint32  	size;
```
qkaiser: We recommend against using `size` because it shadows cstruct's `size` attribute that returns the header size:

```suggestion
        uint32 fsize;                        /* size of the firmware image (without 128 Bytes header) */
```

python/unblob/handlers/archive/dlink/alpha_encimg.py
```diff
@@ -0,0 +1,341 @@
...
+        decrypted_header = decrypt_chunk(encrypted_header, enc_params)
+        header = self._struct_parser.parse(
+            self.HEADER_STRUCT, decrypted_header, Endian.BIG
+        )
+        self._validate(enc_params, file, header)
+        return header
+
+    def calculate_chunk(self, file: File, start_offset: int) -> ValidChunk | None:
+        header = self._get_header(file)
+        end_offset = start_offset + self.HEADER_SIZE + header.size
```
qkaiser:
```suggestion
        end_offset = start_offset + self.HEADER_SIZE + header.fsize
```

python/unblob/handlers/archive/dlink/alpha_encimg.py
```diff
@@ -0,0 +1,341 @@
...
+        if header.signature.rstrip(b"\0") != expected_signature:
+            raise InvalidInputFormat(f"Invalid signature {header.signature}")
+
+        if header.size > file.size():
+            raise InvalidInputFormat("Invalid file size")
+
+        if header.size % 16:
+            raise InvalidInputFormat(
+                f"Firmware size not aligned to 16 bytes: {header.size}"
+            )
```
qkaiser: Let's not shadow the `size` attribute.

Also, the header size field is documented as the firmware size without the 0x80/0xA0 header, but the validation only checks `header.size > file.size()`. This allows a corrupted size slightly smaller than the full file size (e.g., payload size = file size − 1) to pass even though `header.size + header.fsize` exceeds the file length, yielding a chunk end offset beyond EOF and potentially truncating or over-reading during later processing. This should
validate against the remaining file length (e.g., header.size + header.fsize > file.size() - start_offset).

```suggestion
        if header.size + header.fsize > file.size() - start_offset:
            raise InvalidInputFormat("Invalid file size")

        if header.fsize % 16:
            raise InvalidInputFormat(
                f"Firmware size not aligned to 16 bytes: {header.fsize}"
            )
```

python/unblob/handlers/archive/dlink/alpha_encimg.py
qkaiser: Related - https://github.com/onekey-sec/unblob/issues/1175

python/unblob/handlers/archive/dlink/alpha_encimg.py
jstucke: changed

python/unblob/handlers/archive/dlink/alpha_encimg.py
jstucke: changed

python/unblob/handlers/archive/dlink/alpha_encimg.py
jstucke: changed

python/unblob/handlers/archive/dlink/alpha_encimg.py
jstucke: Used 

```python
remaining = file.size() - file.tell()
        if header.fsize > remaining:
```

instead of `if header.size + header.fsize > file.size() - start_offset` since `start_offset` is not part of the validation function (and I would rather not drag it through)

qkaiser:
@jstucke thanks for your time and effort, I know it wasn't an easy review. It's on its way to the merge train, hopefully Github infra won't crash on us again.

PR #1396, opened 2026-02-20
feat(handler): Add support for QNAP networking device firmware

gardar:
QNAP networking devices use the same PC1 cipher as NAS firmware but derive the decryption key directly from the last 74 bytes of the encrypted firmware itself rather than a hardcoded secret prefix. 
Refactors shared cipher and struct definitions into _qnap.py.

### DIFF
--- a/python/unblob/handlers/archive/qnap/qnap_networking.py
+++ b/python/unblob/handlers/archive/qnap/qnap_networking.py
@@ -1,0 +1,65 @@
+from unblob.file_utils import File
+from unblob.models import (
+    Endian,
+    HandlerDoc,
+    HandlerType,
+    HexString,
+    Reference,
+    StructHandler,
+    ValidChunk,
+)
+
+from ._qnap import C_DEFINITIONS, FOOTER_LEN, NAS_DEVICE_ID_PREFIX, QnapExtractor
+
+
+class QnapNetworkingExtractor(QnapExtractor):
+    def _get_secret(self, header) -> str:
+        return header.device_id.rstrip(b"\x00").decode("ascii")
+
+
+class QnapNetworkingHandler(StructHandler):
+    NAME = "qnap_networking"
+
+    PATTERNS = [
+        HexString("69 63 70 6e 61 73"),  # "icpnas" footer signature
+    ]
+    C_DEFINITIONS = C_DEFINITIONS
+    HEADER_STRUCT = "qnap_header_t"
+    EXTRACTOR = QnapNetworkingExtractor()
+
+    DOC = HandlerDoc(
+        name="QNAP Networking",
+        description=(
+            "QNAP networking device firmware encrypted with the PC1 cipher. "
+            "The encryption key is self-describing: it is stored as the "
+            "device_id in the 74-byte 'icpnas' footer appended to the image, "
+            "unlike NAS firmware which uses a shared secret prefix."
+        ),
+        handler_type=HandlerType.ARCHIVE,
+        vendor="QNAP",
+        references=[
+            Reference(
+                title="Pwn2Own Ireland 2024: QNAP Qhora-322",
+                url="https://neodyme.io/en/blog/pwn2own-2024_qhora/",
+            ),
+            Reference(
+                title="QNAP firmware encryption/decryption (PC1)",
+                url="https://gist.github.com/galaxy4public/0420c7c9a8e3ff860c8d5dce430b2669",
+            ),
+        ],
+        limitations=[],
+    )
+
+    def calculate_chunk(self, file: File, start_offset: int) -> ValidChunk | None:
+        if start_offset != file.size() - FOOTER_LEN:
+            return None
+        header = self.parse_header(file, endian=Endian.LITTLE)
+
+        if header.encrypted_len == 0 or header.encrypted_len > start_offset:
+            return None
+
+        device_id = header.device_id.rstrip(b"\x00").decode("ascii", errors="replace")
+        if device_id.upper().startswith(NAS_DEVICE_ID_PREFIX):
+            return None
+
+        return ValidChunk(start_offset=0, end_offset=start_offset + FOOTER_LEN)
--- a/docs/handlers.md
+++ b/docs/handlers.md
@@ -53,6 +53,7 @@
     | [`PAR2 (MULTI-VOLUME)`](#par2-multi-volume) | ARCHIVE | :octicons-check-16: |
     | [`PARTCLONE`](#partclone) | ARCHIVE | :octicons-check-16: |
     | [`QNAP NAS`](#qnap-nas) | ARCHIVE | :octicons-check-16: |
+    | [`QNAP NETWORKING`](#qnap-networking) | ARCHIVE | :octicons-check-16: |
     | [`RAR`](#rar) | ARCHIVE | :octicons-alert-fill-12: |
     | [`ROMFS`](#romfs) | FILESYSTEM | :octicons-check-16: |
     | [`SQUASHFS (V1)`](#squashfs-v1) | FILESYSTEM | :octicons-check-16: |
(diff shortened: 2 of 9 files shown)


python/unblob/handlers/archive/qnap/qnap_networking.py
```diff
@@ -0,0 +1,106 @@
...
+class QnapNetworkingExtractor(Extractor):
+    """Extractor for encrypted QNAP networking device firmware.
+
+    Unlike NAS firmware, the encryption key is self-describing: it is stored
+    as the device_id field in the 74-byte footer appended to the image.
+
+    Known examples:
+        Qhora-322  -> key "QHora-322"
+        Qhora-301W -> key "HORA_301W"
+    """
```
qkaiser: No need for docstring, put this into the commit message :)

python/unblob/handlers/archive/qnap/qnap_networking.py
```diff
@@ -0,0 +1,106 @@
...
+            with outpath.open("wb") as outfile:
+                for chunk in iterate_file(file, 0, header.encrypted_len, 1024):
+                    outfile.write(cryptor.decrypt_chunk(chunk))
+                for chunk in iterate_file(
+                    file,
+                    header.encrypted_len,
+                    eof - FOOTER_LEN - header.encrypted_len,
+                    1024,
+                ):
+                    outfile.write(chunk)
```
qkaiser: This is more in line with how we design extraction these days, feel free to adapt the NAS extractor too:

```suggestion
    def extract(self, inpath: Path, outdir: Path):
        fs = FileSystem(outdir)
        with File.from_path(inpath) as infile, fs.open(f"{inpath.name}.decrypted", "wb+") as outfile:
            infile.seek(-FOOTER_LEN, io.SEEK_END)
            header = self._struct_parser.parse("qnap_header_t", infile, Endian.LITTLE)
            secret = header.device_id.rstrip(b"\x00").decode("ascii")
            cryptor = Cryptor(secret)
            for chunk in iterate_file(infile, 0, header.encrypted_len, 1024):
                outfile.write(cryptor.decrypt_chunk(chunk))
            for chunk in iterate_file(
                infile,
                header.encrypted_len,
                infile.size() - FOOTER_LEN - header.encrypted_len,
                1024,
            ):
                outfile.write(chunk)
        return ExtractResult(reports=fs.problems)
```

python/unblob/handlers/archive/qnap/qnap_networking.py
```diff
@@ -0,0 +1,106 @@
...
+        references=[
+            "https://neodyme.io/en/blog/pwn2own-2024_qhora/",
+            "https://gist.github.com/galaxy4public/0420c7c9a8e3ff860c8d5dce430b2669",
+        ],
+        limitations=[],
+    )
+
+    def calculate_chunk(self, file: File, start_offset: int) -> ValidChunk | None:
+        # start_offset points at the "icpnas" magic inside the footer.
+        # Validate it sits exactly at file_size - FOOTER_LEN to reject stray matches.
```
qkaiser: No need for this.


```suggestion
```

python/unblob/handlers/archive/qnap/qnap_networking.py
```diff
@@ -0,0 +1,106 @@
...
+        file.seek(0, io.SEEK_END)
+        file_size = file.tell()
+        if start_offset != file_size - FOOTER_LEN:
+            return None
```
qkaiser: There's a size method on `File`.

```suggestion
        if start_offset != file.size() - FOOTER_LEN:
            return None
```

python/unblob/handlers/archive/qnap/qnap_networking.py
```diff
@@ -0,0 +1,106 @@
...
+
+        struct_parser = StructParser(C_DEFINITIONS)
+        file.seek(start_offset)
```
qkaiser: No need to seek, cursor is right there when `calculate_chunk` is called (of course you need to get rid of previous seeks per my previous comment).

python/unblob/handlers/archive/qnap/qnap_networking.py
```diff
@@ -0,0 +1,106 @@
...
+        header = struct_parser.parse("qnap_header_t", file, Endian.LITTLE)
+
+        if header.encrypted_len == 0 or header.encrypted_len > start_offset:
+            return None
+
+        # NAS firmware device_id begins with "QNAPNAS" — let QnapHandler own those.
+        device_id = header.device_id.rstrip(b"\x00").decode("ascii", errors="replace")
+        if device_id.upper().startswith("QNAPNAS"):
```
qkaiser: I would move the "QNAPNAS" string to a constant in `_qnap.py` and use the constant in all QNAP related code.

python/unblob/handlers/archive/qnap/qnap_networking.py
```diff
@@ -0,0 +1,106 @@
...
+            return None
+
+        return ValidChunk(start_offset=0, end_offset=file_size)
```
qkaiser: This just considers the whole file as the chunk. From a cursory read, I suppose the `end_offset` is actually placed at `start_offset + len(header)` since we match on the "header" (actually footer) magic and parse it at the end of the file.

For the `start_offset` it's a bit trickier, probably why we matched on the encrypted gzip bytes in the NAS handler. Looks like it's not doable here unless you collect them for all devices.

qkaiser:
Documentation looks ok, don't forget to re-generate it with `uv run unblob --build-handlers-doc docs/handlers.md`

python/unblob/handlers/archive/qnap/qnap_networking.py
```diff
@@ -0,0 +1,68 @@
...
+            ),
+        ],
+        limitations=[],
+    )
+
+    def calculate_chunk(self, file: File, start_offset: int) -> ValidChunk | None:
+        if start_offset != file.size() - FOOTER_LEN:
+            return None
+
+        header = StructParser(C_DEFINITIONS).parse("qnap_header_t", file, Endian.LITTLE)
```
qkaiser: You can use StructHandler for this:

```suggestion
class QnapNetworkingHandler(StructHandler):
    NAME = "qnap_networking"

    PATTERNS = [
        HexString("69 63 70 6e 61 73"),  # "icpnas" footer signature
    ]
    C_DEFINITIONS = C_DEFINITIONS
    HEADER_STRUCT = "qnap_header_t"
    EXTRACTOR = QnapNetworkingExtractor()

    DOC = HandlerDoc(
        name="QNAP Networking",
        description=(
            "QNAP networking device firmware encrypted with the PC1 cipher. "
            "The encryption key is self-describing: it is stored as the "
            "device_id in the 74-byte 'icpnas' footer appended to the image, "
            "unlike NAS firmware which uses a shared secret prefix."
        ),
        handler_type=HandlerType.ARCHIVE,
        vendor="QNAP",
        references=[
            Reference(
                title="Pwn2Own Ireland 2024: QNAP Qhora-322",
                url="https://neodyme.io/en/blog/pwn2own-2024_qhora/",
            ),
            Reference(
                title="QNAP firmware encryption/decryption (PC1)",
                url="https://gist.github.com/galaxy4public/0420c7c9a8e3ff860c8d5dce430b2669",
            ),
        ],
        limitations=[],
    )

    def calculate_chunk(self, file: File, start_offset: int) -> ValidChunk | None:
        if start_offset != file.size() - FOOTER_LEN:
            return None
        header = self.parse_header(file, endian=Endian.LITTLE)
```

qkaiser:
@gardar there are a few things you need to fix:
- run `pre-commit run -a` and commit the changes, also make sure pre-commit hook is active so you don't end up in the same situation with future contributions
- add integration tests, can be a synthetic samples it does not need to be an official image. Test sample must be placed under `tests/integration/archive/qnap/qnap_networking/__input__`, and you can generate the output with `uv run unblob -vvv -f -k -e tests/integration/archive/qnap/qnap_networking/__output__ tests/integration/archive/qnap/qnap_networking/__input__/sample.bin`. Make sure tests are not failing by running `uv run pytest tests -v` for the whole suite, `uv run pytest tests -v -k qnap_networking` to filter yours
- rebase your branch against upstream main so history is linear

gardar:
> @gardar there are a few things you need to fix:
> 
>     * run `pre-commit run -a` and commit the changes, also make sure pre-commit hook is active so you don't end up in the same situation with future contributions
> 
>     * add integration tests, can be a synthetic samples it does not need to be an official image. Test sample must be placed under `tests/integration/archive/qnap/qnap_networking/__input__`, and you can generate the output with `uv run unblob -vvv -f -k -e tests/integration/archive/qnap/qnap_networking/__output__ tests/integration/archive/qnap/qnap_networking/__input__/sample.bin`. Make sure tests are not failing by running `uv run pytest tests -v` for the whole suite, `uv run pytest tests -v -k qnap_networking` to filter yours
> 
>     * rebase your branch against upstream main so history is linear

Should be all sorted now!

PR #1465, opened 2026-04-03
Memory improvements

qkaiser:
Reduced peak memory usage across several handlers by replacing large in-memory reads with chunked processing. cramfs, trx, shrs, par2, and romfs now validate or extract data with streaming reads, so large attacker-controlled regions no longer get copied wholesale into Python memory.

### DIFF
--- a/python/unblob/handlers/filesystem/cramfs.py
+++ b/python/unblob/handlers/filesystem/cramfs.py
@@ -1,9 +1,10 @@
 import binascii
+import io
 import struct
 
 from unblob.extractors import Command
 
-from ...file_utils import Endian, convert_int32, get_endian
+from ...file_utils import get_endian, iterate_file
 from ...models import (
     File,
     HandlerDoc,
@@ -16,6 +17,8 @@ from ...models import (
 
 CRAMFS_FLAG_FSID_VERSION_2 = 0x00000001
 BIG_ENDIAN_MAGIC = 0x28_CD_3D_45
+FSID_CRC_OFFSET = 32
+FSID_CRC_SIZE = 4
 
 
 def swap_int32(i):
@@ -71,7 +74,7 @@ class CramFSHandler(StructHandler):
         header = self.parse_header(file, endian)
         valid_signature = header.signature == b"Compressed ROMFS"
 
-        if valid_signature and self._is_crc_valid(file, start_offset, header, endian):
+        if valid_signature and self._is_crc_valid(file, start_offset, header):
             return ValidChunk(
                 start_offset=start_offset,
                 end_offset=start_offset + header.fs_size,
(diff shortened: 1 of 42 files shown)


python/unblob/handlers/filesystem/cramfs.py
```diff
@@ -83,17 +84,20 @@ def _is_crc_valid(
...
+        header_bytes = bytearray(file.read(36))
+        header_bytes[32:36] = b"\x00\x00\x00\x00"
+        computed_crc = binascii.crc32(header_bytes)
+
+        for chunk in iterate_file(file, start_offset + 36, header.fs_size - 36):
+            computed_crc = binascii.crc32(chunk, computed_crc)
+
         # some vendors like their CRC's swapped, don't ask why
-        return header_crc == computed_crc or header_crc == swap_int32(computed_crc)
+        return header.fsid_crc == computed_crc or header.fsid_crc == swap_int32(
```
vlaci, line 109: It'd worth to add to the commit message, that endian parameter is no longer used because the CRC endiannes may not match with it(?)

tests/handlers/archive/test_shrs.py
```diff
@@ -0,0 +1,29 @@
...
+    def __init__(self, payload: bytes):
+        self.file_size = len(payload)
+        self.file_size_no_padding = len(payload)
+        self.encrypted_digest = hashlib.sha512(payload).digest()
+
+    def __len__(self):
+        return 0
+
+
+def test_is_valid_header_hashes_incrementally():
```
vlaci: how did you decide what handlers to add tests for and what not?

python/unblob/handlers/filesystem/romfs.py
```diff
@@ -115,14 +116,13 @@ def valid_checksum(self) -> bool:
         finally:
             self.file.seek(current_position, io.SEEK_SET)
 
-    @property
-    def content(self) -> bytes:
-        """Returns the file content. Applicable to files and symlinks."""
+    def read_content(self) -> bytes:
+        """Read inode content for small metadata payloads."""
```
vlaci: I cannot really pinpoint it, but find these doc string rewrites and commit messages in general jarring somehow. It is like the LLM really want to come up with a reason for changes but it just misses the mark.

python/unblob/handlers/filesystem/cramfs.py
```diff
@@ -83,17 +84,20 @@ def _is_crc_valid(
...
-        header_crc = convert_int32(crc_bytes, endian)
-        content[32:36] = b"\x00\x00\x00\x00"
-        computed_crc = binascii.crc32(content)
+
+        file.seek(start_offset, io.SEEK_SET)
+        header_bytes = bytearray(file.read(36))
+        header_bytes[32:36] = b"\x00\x00\x00\x00"
+        computed_crc = binascii.crc32(header_bytes)
+
+        for chunk in iterate_file(file, start_offset + 36, header.fs_size - 36):
```
vlaci: probably it'd be a great time to add and use `FSID_CRC_OFFSET = 32` and `FSID_CRC_SIZE = 4` constants.

python/unblob/handlers/filesystem/cramfs.py
qkaiser: Done.

tests/handlers/archive/test_shrs.py
qkaiser: This is a leftover, the LLM generated a bunch of tests that do not bring value since we're already testing that through heavy integration tests. I removed most but forgot this one.

python/unblob/handlers/filesystem/romfs.py
qkaiser: Likely cause is ruff complaining with `D401` (imperative docstrings), leading it to rewriting the whole thing while it just needed to turn `Returns the file content` into `Return the file content`.

Fixed it.

python/unblob/handlers/filesystem/cramfs.py
qkaiser: Good point. Done.

python/unblob/handlers/filesystem/cramfs.py
```diff
@@ -83,17 +86,26 @@ def _is_crc_valid(
...
+        header_bytes = bytearray(file.read(FSID_CRC_OFFSET + FSID_CRC_SIZE))
+        header_bytes[FSID_CRC_OFFSET : FSID_CRC_OFFSET + FSID_CRC_SIZE] = (
+            b"\x00\x00\x00\x00"
+        )
+        computed_crc = binascii.crc32(header_bytes)
+
+        for chunk in iterate_file(
+            file,
+            start_offset + FSID_CRC_OFFSET + FSID_CRC_SIZE,
+            header.fs_size - FSID_CRC_OFFSET + FSID_CRC_SIZE,
```
elektrischermoench: Dont we need:

```
for chunk in iterate_file(
    file,
    start_offset + FSID_CRC_OFFSET + FSID_CRC_SIZE,
    header.fs_size - (FSID_CRC_OFFSET + FSID_CRC_SIZE),
):
```

It is evaluated left to right so would be  `(fs_size - 32) + 4 = fs_size - 28` otherwise. The header without CRC is 32, right? So it should be 36:
`fs_size - (32 + 4) = fs_size - 36` or did I misunderstand this?

qkaiser: Good catch. We never hit it because our samples were not suffixed with padding. See https://github.com/onekey-sec/unblob/pull/1465/commits/a5c533c5d4aa9730c5854b899850e5d27e0ddd0e

elektrischermoench (approved):

PR #1479, opened 2026-04-14
fix(handlers): make UnblobTarInfo work with python 3.13.13+

qkaiser:
In CPython **3.13.13**, tarfile changed its internal parsing path¹: `TarInfo.fromtarfile()` now goes through a new` _fromtarfile()` helper, which calls `cls._frombuf(...)` instead of `cls.frombuf(...)`.

Our workaround in `python/unblob/handlers/archive/_safe_tarfile.py` only overrode `frombuf()`, so on **3.13.13** it was silently bypassed and stdlib parsing came back.

We patched `UnblobTarInfo` to override `_frombuf()` as well, and made `frombuf()` delegate to it so both **3.13.12** and **3.13.13+** use the same GNU-prefix fix.

Also added a direct regression test in for `_frombuf()` so future CPython refactors won't bypass the workaround unnoticed.

1 - https://github.com/python/cpython/commit/ae99fe3a33b43e303a05f012815cef60b611a9c7

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -590,9 +590,9 @@ checksum = "69cdb34c158ceb288df11e18b4bd39de994f6657d83847bdffdbd7f346754b0f"
 
 [[package]]
 name = "rand"
-version = "0.10.0"
+version = "0.10.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


elektrischermoench (approved):
LGTM

PR #1480, opened 2026-04-16
feat(handler): add extraction for UFS1, UFS2 & Solaris UFS1

Corv0O:
Add extraction support for the UFS family of filesystems:

- **UFS1** (FreeBSD / NetBSD / OpenBSD)
- **UFS2** (FreeBSD / NetBSD / OpenBSD)
- **Solaris UFS1** (Solaris / OpenSolaris / illumos)

Introduce more C struct into the parser from linux kernel, FreeBSD & sleuthkit to match data in the filesystem.

Separate each family in a parser and handler class to respect each specificity.

For each UFS, the extractor can extract all type of files, even empty ones. 

## Notes

- UFS1 & UFS2 use different inode  structure but can share the same directory entry structure apart of ancient UFS1 on FreeBSD 4.2 4.3 where a different structures is required.

FreeBSD UFS1 and Solaris UFS1 share the same magic but differ in two places:

- `dirent` layout
- `fs_nrpos` is always equals to `8` on Solaris-based distros and is variable on FreeBSD (most of the time `fs_nrpos` == `1`)

# Source : 

https://cgit.freebsd.org/src/tree/sys/ufs/ffs/fs.h
https://github.com/sleuthkit/sleuthkit/blob/develop/tsk/fs/ffs.cpp#L29
https://github.com/sleuthkit/sleuthkit/blob/develop/tsk/fs/tsk_ffs.h
https://github.com/torvalds/linux/blob/master/fs/ufs/ufs_fs.h#L717

### DIFF
--- a/docs/handlers.md
+++ b/docs/handlers.md
@@ -61,6 +61,7 @@
     | [`QNX DEFLATE`](#qnx-deflate) | COMPRESSION | :octicons-check-16: |
     | [`RAR`](#rar) | ARCHIVE | :octicons-alert-fill-12: |
     | [`ROMFS`](#romfs) | FILESYSTEM | :octicons-check-16: |
+    | [`SOLARIS_UFS1`](#solaris_ufs1) | FILESYSTEM | :octicons-check-16: |
     | [`SQUASHFS (V1)`](#squashfs-v1) | FILESYSTEM | :octicons-check-16: |
     | [`SQUASHFS (V2)`](#squashfs-v2) | FILESYSTEM | :octicons-check-16: |
     | [`SQUASHFS (V2-NON-STANDARD)`](#squashfs-v2-non-standard) | FILESYSTEM | :octicons-check-16: |
@@ -1086,6 +1087,23 @@
 
         - [RomFS Documentation](https://www.kernel.org/doc/html/latest/filesystems/romfs.html){ target="_blank" }
         - [RomFS Wikipedia](https://en.wikipedia.org/wiki/Romfs){ target="_blank" }
+## solaris_ufs1
+
+!!! success "Fully supported"
+
+    === "Description"
+
+        Solaris UFS1 is the variant of UFS1 used by Oracle Solaris and illumos-based systems such as OpenIndiana and OmniOS. It shares the same overall design as FreeBSD UFS1 but keeps an older on-disk convention inherited from early Unix, making it incompatible with FreeBSD's variant despite the shared magic number.
+
+        ---
+
+        - **Handler type:** FileSystem
+        
+
+    === "References"
+
+        - [Unix File System Wikipedia](https://en.wikipedia.org/wiki/Unix_File_System){ target="_blank" }
+        - [Oracle Solaris](https://en.wikipedia.org/wiki/Oracle_Solaris){ target="_blank" }
 ## SquashFS (v1)
 
 !!! success "Fully supported"
--- a/docs/handlers.md
+++ b/docs/handlers.md
@@ -77,8 +77,8 @@
     | [`TAR (USTAR)`](#tar-ustar) | ARCHIVE | :octicons-check-16: |
     | [`UBI`](#ubi) | FILESYSTEM | :octicons-check-16: |
     | [`UBIFS`](#ubifs) | FILESYSTEM | :octicons-check-16: |
-    | [`UFS1`](#ufs1) | FILESYSTEM | :octicons-alert-fill-12: |
-    | [`UFS2`](#ufs2) | FILESYSTEM | :octicons-alert-fill-12: |
+    | [`UFS1`](#ufs1) | FILESYSTEM | :octicons-check-16: |
+    | [`UFS2`](#ufs2) | FILESYSTEM | :octicons-check-16: |
     | [`UZIP`](#uzip) | COMPRESSION | :octicons-check-16: |
     | [`XAMARIN COMPRESSED ASSEMBLIES`](#xamarin-compressed-assemblies) | EXECUTABLE | :octicons-check-16: |
     | [`XIAOMI HDR1`](#xiaomi-hdr1) | ARCHIVE | :octicons-check-16: |
@@ -1358,7 +1358,7 @@
         - [UBIFS Wikipedia](https://en.wikipedia.org/wiki/UBIFS){ target="_blank" }
 ## ufs1
 
-!!! warning "Partially supported"
+!!! success "Fully supported"
 
     === "Description"
 
@@ -1372,17 +1372,13 @@
     === "References"
 
         - [Unix File System Wikipedia](https://en.wikipedia.org/wiki/Unix_File_System){ target="_blank" }
-
-    === "Limitations"
-
-        - File extraction is not yet supported, only carving of the filesystem is currently available.
 ## ufs2
 
-!!! warning "Partially supported"
+!!! success "Fully supported"
 
     === "Description"
 
-        UFS2 (Unix File System 2) is an extended version of UFS1 supported by Unix-like operating systems such as FreeBSD and Solaris. It introduces 64-bit block addressing, extended file attributes, and improved performance over UFS1, while retaining the hierarchical tree structure and inodes for file metadata and data block management.
+        UFS2 (Unix File System 2) is an extended version of UFS1 supported by Unix-like operating systems such as FreeBSD. It introduces 64-bit block addressing, extended file attributes, and improved performance over UFS1, while retaining the hierarchical tree structure and inodes for file metadata and data block management.
 
         ---
 
@@ -1392,10 +1388,6 @@
     === "References"
 
         - [Unix File System Wikipedia](https://en.wikipedia.org/wiki/Unix_File_System){ target="_blank" }
-
-    === "Limitations"
-
-        - File extraction is not yet supported, only carving of the filesystem is currently available.
 ## UZIP
 
 !!! success "Fully supported"
--- a/python/unblob/handlers/filesystem/ufs.py
+++ b/python/unblob/handlers/filesystem/ufs.py
@@ -1,9 +1,18 @@
 import io
+import stat
+import struct
+from collections.abc import Iterator
+from pathlib import Path
 
 from unblob.file_utils import (
+    Endian,
+    FileSystem,
     InvalidInputFormat,
+    StructParser,
 )
 from unblob.models import (
+    Extractor,
+    ExtractResult,
     File,
     HandlerDoc,
     HandlerType,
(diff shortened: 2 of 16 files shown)


python/unblob/handlers/filesystem/ufs.py
```diff
@@ -184,12 +192,306 @@
...
+
+    def extract(self, inpath: Path, outdir: Path):
+        fs = FileSystem(outdir)
+        with File.from_path(inpath) as f:
+            parser = self.parser(f, self.sb_offset)
+            parser.walk_extract(fs, UFS_ROOT_INO, Path("/"))
+        return ExtractResult(reports=fs.problems)
+
+
+class UFSBaseHandler(StructHandler):
```
qkaiser: This can stay private:


```suggestion
class _UFSBaseHandler(StructHandler):
```

python/unblob/handlers/filesystem/ufs.py
```diff
@@ -234,20 +537,56 @@ class UFS1Handler(_UFSHandler):
...
+            Reference(
+                title="Oracle Solaris",
+                url="https://en.wikipedia.org/wiki/Oracle_Solaris",
+            ),
         ],
+        limitations=[],
     )
 
     def get_block_size(self, header) -> int:
         return header.fs_size
```
qkaiser: It breaks the whole diff since you inserted it in the middle of two pre-existing handler. Please move it after the `UFS2Handler` definition. This will make reviewing and history cleaner.

python/unblob/handlers/filesystem/ufs.py
```diff
@@ -234,20 +537,56 @@ class UFS1Handler(_UFSHandler):
...
+        return header.fs_size
+
+    def is_valid_header(self, header) -> bool:
+        # fs_nrpos=8 indicates Solaris-flavor UFS1
+        if header.fs_nrpos == 8:
+            return False
+        return super().is_valid_header(header)
+
+
+class SolarisHandler(UFSBaseHandler):
```
qkaiser: Split your changes in separate commits:
- adding the extractor for UFS1/UFS2
- adding the Solaris Handler
- the modifications to the extractor to support Solaris

python/unblob/handlers/filesystem/ufs.py
```diff
@@ -184,12 +192,306 @@
...
+
+
+class FileType(IntEnum):
+    FIFO = 0o010000
+    CHR = 0o020000
+    DIR = 0o040000
+    BLK = 0o060000
+    REG = 0o100000
+    LNK = 0o120000
+    SCK = 0o140000
```
qkaiser: These are standard, you can use `stat` I think.

python/unblob/handlers/filesystem/ufs.py
```diff
@@ -184,12 +192,306 @@
...
+            case FileType.LNK:
+                fs.create_symlink(src=Path(self.read_symlink(inode)), dst=path)
+            case FileType.FIFO:
+                fs.mkfifo(path)
+            case FileType.SCK:
+                fs.mknod(path, mode=inode.ui_mode)
+            case FileType.CHR | FileType.BLK:
+                fs.mknod(
+                    path, mode=inode.ui_mode, device=self.get_direct_blocks(inode)[0]
+                )
```
qkaiser: Just use `stat`:

```suggestion
         file_type = stat.S_IFMT(inode.ui_mode)

        match file_type:
            case stat.S_IFDIR:
                fs.mkdir(path, exist_ok=True)
                for child_ino, name, _ in self.parse_dentries(inode):
                    if name in (".", ".."):
                        continue
                    self.walk_extract(fs, child_ino, path / name)
            case stat.S_IFREG:
                fs.write_chunks(path, self.read_file_content(inode))
            case stat.S_IFLNK:
                fs.create_symlink(src=Path(self.read_symlink(inode)), dst=path)
            case stat.S_IFIFO:
                fs.mkfifo(path)
            case stat.S_IFSOCK:
                fs.mknod(path, mode=inode.ui_mode)
            case stat.S_IFCHR | stat.S_IFBLK:
                fs.mknod(
                    path, mode=inode.ui_mode, device=self.get_direct_blocks(inode)[0]
                )
```

python/unblob/handlers/filesystem/ufs.py
```diff
@@ -184,12 +192,306 @@
...
+            uint16 d_namlen;
+            char d_name[d_namlen];
+        }
+
 """
 
 MAGIC_OFFSET = 0x55C  #  relative to SB_OFFSET
+UFS_ROOT_INO = 2
+
+S_IFMT = 0o170000  # FileType mask
```
qkaiser: Not needed.

python/unblob/handlers/filesystem/ufs.py
```diff
@@ -14,6 +21,7 @@
 )
 
 UFS_C_DEFINITION = """
+        #define UFS_MAXNAMLEN    255
```
qkaiser: Unused.

python/unblob/handlers/filesystem/ufs.py
```diff
@@ -184,12 +192,306 @@
...
+
+
+class UFSParser:
+    INODE_STRUCT: str
+    INODE_SIZE: int
+    PTR_SIZE: int
+
+    def __init__(self, file: File, sb_offset: int):
+        self.file = file
+        self._struct_parser = StructParser(UFS_C_DEFINITION).cparser_le
```
qkaiser:
```suggestion
        self._struct_parser = StructParser(UFS_C_DEFINITION)
```

python/unblob/handlers/filesystem/ufs.py
```diff
@@ -184,12 +192,306 @@
...
+        self.file.seek(sb_offset, io.SEEK_SET)
+        self.sb = self._struct_parser.ufs_superblock_t(self.file)
```
qkaiser:
```suggestion
        self.super_block = self._struct_parser.parse("ufs_superblock_t", self.file, Endian.LITTLE)
```

python/unblob/handlers/filesystem/ufs.py
```diff
@@ -184,12 +192,306 @@
...
+                    path, mode=inode.ui_mode, device=self.get_direct_blocks(inode)[0]
+                )
+
+    def parse_dentries(self, inode):
+        for chunk in self.read_file_content(inode):
+            offset = 0
+            while offset < len(chunk):
+                entry = self._struct_parser.ufs_dirent(chunk[offset:])
+                if entry.d_reclen == 0:
+                    break
```
qkaiser, line 364: What does it mean when reclen is 0 ?

python/unblob/handlers/filesystem/ufs.py
```diff
@@ -184,12 +192,306 @@
...
+        self.file.seek(offset, io.SEEK_SET)
+        return self.inode_parser(self.file)
+
+    def frag_to_bytes(self, frag_addr: int) -> int:
+        return frag_addr * self.sb.fs_fsize
+
+    def cg_start(self, cg: int) -> int:
+        return cg * self.sb.fs_fpg
+
+    def get_direct_blocks(self, inode):
```
qkaiser:
```suggestion
    def get_direct_blocks(self, inode) -> list[int]:
```

python/unblob/handlers/filesystem/ufs.py
```diff
@@ -184,12 +192,306 @@
...
+        return inode.ui_u2.ui_addr.ui_db
+
+    def get_indirect_blocks(self, inode):
```
qkaiser:
```suggestion
    def get_indirect_blocks(self, inode) -> list[int]:
```

python/unblob/handlers/filesystem/ufs.py
```diff
@@ -184,12 +192,306 @@
...
+                if data_addr == 0:
+                    yield b"\x00" * to_read
+                else:
+                    self.file.seek(self.frag_to_bytes(data_addr), io.SEEK_SET)
+                    yield self.file.read(to_read)
+                remaining -= to_read
+
+    def read_block_pointers(self, blk_addr: int) -> list[int]:
+        self.file.seek(self.frag_to_bytes(blk_addr), io.SEEK_SET)
+        data = self.file.read(self.sb.fs_bsize)
```
qkaiser: Can we make sure that `fs_bsize` is not larger than some upper bound ? Maybe during initialization ?

python/unblob/handlers/filesystem/ufs.py
```diff
@@ -184,12 +192,306 @@
...
+        if inode.ui_blocks == 0:
+            return bytes(inode.ui_u2.ui_symlink[: inode.ui_size]).decode(
+                "utf-8", errors="replace"
+            )
+        chunk = next(self.read_file_content(inode))
+        return chunk[: inode.ui_size].decode("utf-8", errors="replace")
+
+    def read_inode(self, ino_number: int):
+        cg = ino_number // self.sb.fs_ipg
+        index = ino_number % self.sb.fs_ipg
```
qkaiser: Renaming the fields to their full name would help the comprehension here. What does `ipg` stands for ?

python/unblob/handlers/filesystem/ufs.py
```diff
@@ -184,12 +192,306 @@
...
+            + index * self.INODE_SIZE
+        )
+        self.file.seek(offset, io.SEEK_SET)
+        return self.inode_parser(self.file)
+
+    def frag_to_bytes(self, frag_addr: int) -> int:
+        return frag_addr * self.sb.fs_fsize
+
+    def cg_start(self, cg: int) -> int:
+        return cg * self.sb.fs_fpg
```
qkaiser: Renaming the fields to their full name would help the comprehension here. What does `fpg` stands for ?

python/unblob/handlers/filesystem/ufs.py
```diff
@@ -184,12 +192,306 @@
             uint32  fs_magic;
             uint8   fs_space[1];
 } ufs_superblock_t;
+
+        #define UFS_NDADDR  12
+        #define UFS_NINDIR  3
+
+        /* UFS1 FreeBSD & Solaris */
+        struct ufs1_inode {
```
qkaiser, line 213: All the `ui_` prefix for `unsigned int` are polluting the code in python. I'd get rid of it.

python/unblob/handlers/filesystem/ufs.py
```diff
@@ -184,12 +192,306 @@
...
+            uint32     ui_ctimensec;   /*  72: Last inode change time. */
+            uint32     ui_birthnsec;   /*  76: Inode creation time. */
+            uint32     ui_gen;         /*  80: Generation number. */
+            uint32     ui_kernflags;   /*  84: Kernel flags. */
+            uint32     ui_flags;       /*  88: Status flags (chflags). */
+            uint32     ui_extsize;     /*  92: External attributes block. */
+            uint64     ui_extb[UFS_NXADDR];/*  96: External attributes block. */
+            union {
+                struct {
+                    uint64     ui_db[UFS_NDADDR]; /* 112: Direct disk blocks. */
```
qkaiser:
```suggestion
                    uint64     direct_blocks[UFS_NDADDR]; /* 112: Direct disk blocks. */
```

python/unblob/handlers/filesystem/ufs.py
```diff
@@ -184,12 +192,306 @@
...
+                    uint64     ui_ib[UFS_NINDIR];/* 208: Indirect disk blocks.*/
```
qkaiser:
```suggestion
                    uint64     indirect_blocks[UFS_NINDIR];/* 208: Indirect disk blocks.*/
```

python/unblob/handlers/filesystem/ufs.py
```diff
@@ -184,12 +192,306 @@
...
+                uint32	ui_inumber;		/*  0x4 lsf: inode number */
+                uint32	ui_author;		/*  0x4 GNU HURD: author */
+            } ui_u1;
+            uint64	ui_size;		/*  0x8 */
+            struct ufs_timeval ui_atime;	/* 0x10 access */
+            struct ufs_timeval ui_mtime;	/* 0x18 modification */
+            struct ufs_timeval ui_ctime;	/* 0x20 creation */
+            union {
+                struct {
+                    uint32	ui_db[UFS_NDADDR];/* 0x28 data blocks */
```
qkaiser:
```suggestion
                    uint32	direct_blocks[UFS_NDADDR];/* 0x28 data blocks */
```

python/unblob/handlers/filesystem/ufs.py
```diff
@@ -184,12 +192,306 @@
...
+                    uint32	ui_ib[UFS_NINDIR];/* 0x58 indirect blocks */
```
qkaiser:
```suggestion
                    uint32	indirect_blocks[UFS_NINDIR];/* 0x58 indirect blocks */
```

python/unblob/handlers/filesystem/ufs.py
qkaiser: Same remark for `cg`, which can probably be replaced by `cylinder_group` everywhere.

python/unblob/handlers/filesystem/ufs.py
```diff
@@ -184,12 +192,306 @@
...
+        index = ino_number % self.sb.fs_ipg
+        offset = (
+            self.frag_to_bytes(self.cg_start(cg) + self.sb.fs_iblkno)
+            + index * self.INODE_SIZE
+        )
+        self.file.seek(offset, io.SEEK_SET)
+        return self.inode_parser(self.file)
+
+    def frag_to_bytes(self, frag_addr: int) -> int:
+        return frag_addr * self.sb.fs_fsize
```
qkaiser: Technically this is "fragment index to offset" no ?

python/unblob/handlers/filesystem/ufs.py
```diff
@@ -184,12 +192,306 @@
...
+        for chunk in self.read_indirect_db(inode, remaining):
+            yield chunk
+
+    def read_direct_db(self, inode):
+        for addr in self.get_direct_blocks(inode):
+            if addr == 0:
+                yield b"\x00" * self.sb.fs_bsize
+            else:
+                self.file.seek(self.frag_to_bytes(addr), io.SEEK_SET)
+                yield self.file.read(self.sb.fs_bsize)
```
qkaiser: Is it a known mechanism to emit sparse data when the address is 0x0 ? If not, get rid of it:


```suggestion
            if addr:
                self.file.seek(self.frag_to_bytes(addr), io.SEEK_SET)
                yield self.file.read(self.sb.fs_bsize)
```

python/unblob/handlers/filesystem/ufs.py
```diff
@@ -184,12 +192,306 @@
...
+                offset += entry.d_reclen
+                if entry.d_ino == 0:
+                    continue
+                yield (
+                    entry.d_ino,
+                    entry.d_name.decode("utf-8", errors="replace"),
+                    entry.d_type,
+                )
+
+    def read_file_content(self, inode):
```
qkaiser:
```suggestion
    def read_file_content(self, inode) -> bytes:
```

python/unblob/handlers/filesystem/ufs.py
```diff
@@ -184,12 +192,306 @@
...
+    def parse_dentries(self, inode):
+        for chunk in self.read_file_content(inode):
+            offset = 0
+            while offset < len(chunk):
+                entry = self._struct_parser.ufs_dirent(chunk[offset:])
+                if entry.d_reclen == 0:
+                    break
+                offset += entry.d_reclen
+                if entry.d_ino == 0:
+                    continue
```
qkaiser: Why do we skip these ?

python/unblob/handlers/filesystem/ufs.py
```diff
@@ -184,12 +192,306 @@
...
+                yield (
+                    entry.d_ino,
+                    entry.d_name.decode("utf-8", errors="replace"),
+                    entry.d_type,
```
qkaiser: I think you never actually use this value when returned by `parse_dentries`. See line 336.

python/unblob/handlers/filesystem/ufs.py
```diff
@@ -184,12 +192,306 @@
...
+                fs.write_chunks(path, self.read_file_content(inode))
+            case FileType.LNK:
+                fs.create_symlink(src=Path(self.read_symlink(inode)), dst=path)
+            case FileType.FIFO:
+                fs.mkfifo(path)
+            case FileType.SCK:
+                fs.mknod(path, mode=inode.ui_mode)
+            case FileType.CHR | FileType.BLK:
+                fs.mknod(
+                    path, mode=inode.ui_mode, device=self.get_direct_blocks(inode)[0]
```
qkaiser: The device value is not stored by the inode when it's a SOCK ?

python/unblob/handlers/filesystem/ufs.py
```diff
@@ -184,12 +192,306 @@
...
+    INODE_STRUCT: str
+    INODE_SIZE: int
+    PTR_SIZE: int
+
+    def __init__(self, file: File, sb_offset: int):
+        self.file = file
+        self._struct_parser = StructParser(UFS_C_DEFINITION).cparser_le
+        self.file.seek(sb_offset, io.SEEK_SET)
+        self.sb = self._struct_parser.ufs_superblock_t(self.file)
+        self.inode_parser = getattr(self._struct_parser, self.INODE_STRUCT)
```
qkaiser: I don't like this approach. You use the inode parser at a single location.

python/unblob/handlers/filesystem/ufs.py
```diff
@@ -184,12 +192,306 @@
...
+
+    def read_inode(self, ino_number: int):
+        cg = ino_number // self.sb.fs_ipg
+        index = ino_number % self.sb.fs_ipg
+        offset = (
+            self.frag_to_bytes(self.cg_start(cg) + self.sb.fs_iblkno)
+            + index * self.INODE_SIZE
+        )
+        self.file.seek(offset, io.SEEK_SET)
+        return self.inode_parser(self.file)
```
qkaiser:
```suggestion
        return self._struct_parser.parse(self.INODE_STRUCT, self.file, Endian.LITTLE)
```

qkaiser (commented):
Your code looks good, but its legibility can be improved by:
- getting rid of type prefixes in the C definitions
- simplify how calls to the parser are made
- use `stat` instead of a custom enum
- use full variable names instead of acronyms
- add typing info so pyright can help us spot potential issues

python/unblob/handlers/filesystem/ufs.py
Corv0O: `d_reclen` (directory record length) is the size of the entire directory in bytes, if it is at 0 it means that they are no more valid entries in this directory. without this condition, `offset += entry.rec_len` would lead to an infinite loop

python/unblob/handlers/filesystem/ufs.py
Corv0O: when `d_ino == 0` it means that the directory entry has been deleted, we skip because there's nothing to extract.

python/unblob/handlers/filesystem/ufs.py
Corv0O: This handles sparse files a Unix optimization where contiguous runs of zeroes are never written to disk, saving space. In UFS, a block pointer of 0 means it's a hole. 

POSIX requires holes to read as zeroes. We can't skip them because `read_file_content` tracks a remaining byte counter across all blocks if we don't yield data for a hole, the offset shifts and every subsequent block ends up at the wrong position in the extracted file.

python/unblob/handlers/filesystem/ufs.py
qkaiser: Make it explicit then, define a constant `DELETED_INO = 0` and do:

```suggestion
                if entry.d_ino == DELETED_INO:
                    continue
```

python/unblob/handlers/filesystem/ufs.py
Corv0O: they have no device number for a SOCK . Only S_IFCHR (character), S_IFBLK (block) devices store a device in the direct block pointer area.

python/unblob/handlers/filesystem/ufs.py
Corv0O: inodes per group

python/unblob/handlers/filesystem/ufs.py
Corv0O: frag per group

python/unblob/handlers/filesystem/ufs.py
```diff
@@ -201,6 +516,8 @@ def get_block_size(self, header) -> int:
     def is_valid_header(self, header) -> bool:
         return (
             header.fs_fsize > 0
+            and header.fs_bsize > 0
+            and header.fs_bsize <= 65536
```
qkaiser, line 509: Is this constant documented somewhere by FreeBSD/UFS ? Is it the known max block size ?

python/unblob/handlers/filesystem/ufs.py
```diff
@@ -467,6 +475,22 @@ class UFS2Parser(UFSParser):
...
+            while offset < len(chunk):
+                entry = self._struct_parser.parse(
+                    "old_ufs_dirent", chunk[offset:], Endian.LITTLE
+                )
+                if entry.d_reclen == 0:
+                    break
+                offset += entry.d_reclen
+                if entry.d_ino == DELETED_INO:
+                    continue
+                yield entry.d_ino, entry.d_name.decode("utf-8", errors="replace")
```
qkaiser: If you add an `DIRENT_STRUCT` to the `UFSParser` with a default value of "ufs_dirent" and overload it in `SolarisUFS1Parser` with "old_ufs_dirent", you can get rid of this function entirely if you make sure it uses this attribute when calling the struct parser.

python/unblob/handlers/filesystem/ufs.py
Corv0O: it's a  constant from FreeBSD  
https://cgit.freebsd.org/src/tree/sys/sys/param.h line 273

qkaiser: Then define it as such in your code. E.g.`MAX_BLOCK_SIZE = 65536`

python/unblob/handlers/filesystem/ufs.py
```diff
@@ -44,152 +53,434 @@
...
+
+class UFS2Parser(UFSParser):
+    INODE_STRUCT = "ufs2_inode_t"
+    INODE_SIZE = 256
+    PTR_SIZE = 8
+
+
+class UFSExtractor(Extractor):
+    def __init__(self, parser, sb_offset):
+        self.parser = parser
```
qkaiser: It's a type when you pass it, and you can enforce it through type hints:

```python
def __init__(self, parser: type[UFSParser], superblock_offset: int):
    self.parser_type = parser_type
    self.superblock_offset = superblock_offset
```

python/unblob/handlers/filesystem/ufs.py
```diff
@@ -44,152 +53,434 @@
...
+        offset = (
+            self.frag_to_offset(
+                self.cylinder_group_start(cylinder_group) + self.super_block.fs_iblkno
+            )
+            + index * self.INODE_SIZE
+        )
+        self.file.seek(offset, io.SEEK_SET)
+        return self._struct_parser.parse(self.INODE_STRUCT, self.file, Endian.LITTLE)
+
+    def frag_to_offset(self, frag_addr: int) -> int:
```
qkaiser: Use `fragment_index` instead of `frag_addr` if it's indeed a fragment index.

python/unblob/handlers/filesystem/ufs.py
```diff
@@ -44,152 +53,434 @@
...
+                to_read = min(self.super_block.fs_bsize, remaining)
+                if data_addr == 0:
+                    # Sparse file: unallocated block reads as zeroes
+                    yield b"\x00" * to_read
+                else:
+                    self.file.seek(self.frag_to_offset(data_addr), io.SEEK_SET)
+                    yield self.file.read(to_read)
+                remaining -= to_read
+
+    def read_block_pointers(self, blk_addr: int) -> list[int]:
```
qkaiser: Is it a block address or a block index ?

python/unblob/handlers/filesystem/ufs.py
```diff
@@ -44,152 +53,434 @@
...
+            to_read = min(len(chunk), remaining)
+            yield chunk[:to_read]
+            remaining -= to_read
+            if remaining <= 0:
+                return
+        for chunk in self.read_indirect_blocks(inode, remaining):
+            yield chunk
+
+    def read_direct_blocks(self, inode) -> Iterator[bytes]:
+        for addr in self.get_direct_blocks(inode):
```
qkaiser: address or index ?

python/unblob/handlers/filesystem/ufs.py
```diff
@@ -547,6 +547,12 @@ class UFS1Handler(_UFSBaseHandler):
     def get_block_size(self, header) -> int:
         return header.fs_size
 
+    def is_valid_header(self, header) -> bool:
+        # fs_nrpos=8 indicates Solaris-flavor UFS1
+        if header.fs_nrpos == 8:
+            return False
```
qkaiser: Maybe you can disambiguate through the pattern rather than in this function ? What are valid values for  `fs_nrpos` that are not 0x8 ?

docs/handlers.md
```diff
@@ -77,8 +78,8 @@
     | [`TAR (USTAR)`](#tar-ustar) | ARCHIVE | :octicons-check-16: |
     | [`UBI`](#ubi) | FILESYSTEM | :octicons-check-16: |
     | [`UBIFS`](#ubifs) | FILESYSTEM | :octicons-check-16: |
-    | [`UFS1`](#ufs1) | FILESYSTEM | :octicons-alert-fill-12: |
-    | [`UFS2`](#ufs2) | FILESYSTEM | :octicons-alert-fill-12: |
+    | [`UFS1`](#ufs1) | FILESYSTEM | :octicons-check-16: |
+    | [`UFS2`](#ufs2) | FILESYSTEM | :octicons-check-16: |
```
qkaiser, line 82: Doc modifications do not belong to this commit. Move them to where you actually change the HandlerDoc content for UFS1, UFS2, and UFS1 Solaris.

qkaiser (commented):
Small stuff, otherwise looks good. I think you should merge the last two commits into one (Solaris + disambiguation).

python/unblob/handlers/filesystem/ufs.py
```diff
@@ -218,11 +512,12 @@ def calculate_chunk(self, file: File, start_offset: int) -> ValidChunk | None:
         return ValidChunk(start_offset=start_offset, end_offset=end_offset)
 
 
-class UFS1Handler(_UFSHandler):
+class UFS1Handler(_UFSBaseHandler):
     NAME = "ufs1"
-    PATTERNS = [HexString("54 19 01 00 00")]  # UFS1 fs_magic + null fs_space
```
qkaiser, line 223: why delete the comment ?

Corv0O: my bad !

qkaiser (approved):

PR #1490, opened 2026-04-20
feat(handler): add Mach-O executable handler

elektrischermoench:
Adds a handler for Mach-O binaries (32/64-bit, little/big-endian). Chunk boundaries are computed by walking load commands and tracking the highest segment file extent, so the handler works correctly for both single-arch and fat binaries embedded in firmware images.

### DIFF
--- a/python/unblob/handlers/executable/macho.py
+++ b/python/unblob/handlers/executable/macho.py
@@ -1,0 +1,256 @@
+import io
+from enum import Enum, IntEnum
+from pathlib import Path
+
+import attrs
+
+from unblob.file_utils import (
+    Endian,
+    File,
+    FileSystem,
+    InvalidInputFormat,
+    StructParser,
+)
+from unblob.models import (
+    ExtractResult,
+    HandlerDoc,
+    HandlerType,
+    HexString,
+    Reference,
+    StructHandler,
+    ValidChunk,
+)
+
+
+class MachOMagic(bytes, Enum):
+    """Mach-O magics as they appear on disk (first 4 bytes).
+
+    Fat headers are always big-endian; a thin magic is byte-swapped ("cigam")
+    when the binary's endianness is big.
+    """
+
+    FAT = b"\xca\xfe\xba\xbe"  # fat / universal binary
+    THIN_64_LE = b"\xcf\xfa\xed\xfe"  # 64-bit little-endian (arm64, x86_64)
+    THIN_32_LE = b"\xce\xfa\xed\xfe"  # 32-bit little-endian (armv7, x86)
+    THIN_64_BE = b"\xfe\xed\xfa\xcf"  # 64-bit big-endian
+    THIN_32_BE = b"\xfe\xed\xfa\xce"  # 32-bit big-endian
+
+
+# Header sizes in bytes
+_HEADER_SIZE_32 = 28
+_HEADER_SIZE_64 = 32  # adds a 4-byte reserved field
+
+# Maps (cputype, cpusubtype) to a human-readable name for fat binary slice filenames
+_CPU_ARCH_NAMES: dict[tuple[int, int], str] = {
+    (0xC, 5): "arm_v4t",
+    (0xC, 6): "arm_v6",
+    (0xC, 9): "arm_v7",
+    (0xC, 11): "arm_v7s",
+    (0xC, 12): "arm_v7k",
+    (0x100000C, 0): "arm64",
+    (0x100000C, 1): "arm64_v8",
+    (0x100000C, 2): "arm64e",
+    (0x7, 3): "i386",
+    (0x1000007, 3): "x86_64",
+    (0x12, 0): "ppc",
+    (0x1000012, 0): "ppc64",
+}
+
+# Single source of truth for every Mach-O structure, shared by the handler's struct parser
+# and the fat binary extractor below.
+C_DEFINITIONS = r"""
+    typedef struct macho_header {
+        uint32 magic;
+        uint32 cputype;
+        uint32 cpusubtype;
+        uint32 filetype;
+        uint32 ncmds;
+        uint32 sizeofcmds;
+        uint32 flags;
+    } macho_header_t;
+
+    typedef struct macho_load_command {
+        uint32 cmd;
+        uint32 cmdsize;
+    } macho_load_command_t;
+
+    // Covers both LC_SEGMENT (32-bit) and LC_SEGMENT_64 via a union. Only the fields up to
+    // filesize are modelled (all we need to compute file extents), so the struct is 56 bytes
+    // and never reads past the smaller 32-bit segment command.
+    typedef struct macho_segment_command {
+        uint32 cmd;
+        uint32 cmdsize;
+        char   segname[16];
+        union {
+            struct {
+                uint32 vmaddr;
+                uint32 vmsize;
+                uint32 fileoff;
+                uint32 filesize;
+            } bits32;
+            struct {
+                uint64 vmaddr;
+                uint64 vmsize;
+                uint64 fileoff;
+                uint64 filesize;
+            } bits64;
+        } seg;
+    } macho_segment_command_t;
+
+    typedef struct fat_header {
+        uint32 magic;      // 0xcafebabe, always big-endian
+        uint32 nfat_arch;  // number of architecture slices
+    } fat_header_t;
+
+    typedef struct fat_arch {
+        uint32 cputype;    // CPU architecture type
+        uint32 cpusubtype; // CPU subtype
+        uint32 offset;     // file offset to this architecture's Mach-O slice
+        uint32 slice_size; // size of this architecture's Mach-O slice
+        uint32 align;      // slice alignment as a power of 2
+    } fat_arch_t;
+"""
+
+_parser = StructParser(C_DEFINITIONS)
+
+
+def _cpu_arch_name(cputype: int, cpusubtype: int) -> str:
+    return _CPU_ARCH_NAMES.get(
+        (cputype, cpusubtype), f"cputype_{cputype:#x}_{cpusubtype:#x}"
+    )
+
+
+class LoadCommandType(IntEnum):
+    SEGMENT = 0x1
+    SEGMENT_64 = 0x19
+
+
+@attrs.define(repr=False)
+class MachOChunk(ValidChunk):
+    """A fat (universal) Mach-O binary whose architecture slices are carved out on extraction."""
+
+    def extract(self, inpath: Path, outdir: Path) -> ExtractResult | None:
+        fs = FileSystem(outdir)
+        with File.from_path(inpath) as file:
+            fat_hdr = _parser.parse("fat_header_t", file, Endian.BIG)
+            architectures = [
+                _parser.parse("fat_arch_t", file, Endian.BIG)
+                for _ in range(fat_hdr.nfat_arch)
+            ]
+            for arch in architectures:
+                if arch.slice_size == 0:
+                    continue
+                cpu_name = _cpu_arch_name(arch.cputype, arch.cpusubtype)
+                fs.carve(
+                    Path(f"arch_{cpu_name}.macho"), file, arch.offset, arch.slice_size
+                )
+        return ExtractResult(reports=fs.problems)
+
+
+class MachOHandler(StructHandler):
+    NAME = "macho"
+
+    EXTRACTOR = None
+
+    PATTERNS = [
+        # Thin Mach-O: magic (4) + [24] wildcards = full 28-byte common header guaranteed present,
+        # avoiding false positives from partial matches and removing the need for truncation checks.
+        HexString("CF FA ED FE [24]"),  # 64-bit little-endian (arm64, x86_64)
+        HexString("CE FA ED FE [24]"),  # 32-bit little-endian (armv7, x86)
+        HexString("FE ED FA CF [24]"),  # 64-bit big-endian
+        HexString("FE ED FA CE [24]"),  # 32-bit big-endian
+        # Fat (universal) binary:
+        # CA FE BA BE              — fat magic (always big-endian)
+        # 00 00 00 (01-08)         — nfat_arch 1-8; limits to <=8 architectures which
+        #                            disambiguates from Java class files (major version ≥ 45)
+        HexString("CA FE BA BE 00 00 00 (01|02|03|04|05|06|07|08)"),
+    ]
+
+    C_DEFINITIONS = C_DEFINITIONS
+    HEADER_STRUCT = "macho_header_t"
+
+    DOC = HandlerDoc(
+        name="Mach-O",
+        description="Mach-O (Mach Object) is the native executable and library binary format used by macOS, iOS, tvOS, and watchOS. It encodes load commands describing memory segments, dynamic libraries, and code signing, and supports multiple CPU architectures via fat/universal binaries.",
+        handler_type=HandlerType.EXECUTABLE,
+        vendor="Apple",
+        references=[
+            Reference(
+                title="Mach-O Programming Topics",
+                url="https://developer.apple.com/library/archive/documentation/DeveloperTools/Conceptual/MachOTopics/0-Introduction/introduction.html",
+            ),
+        ],
+        limitations=[],
+    )
+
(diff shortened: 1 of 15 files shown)


python/unblob/handlers/executable/macho.py
```diff
@@ -0,0 +1,124 @@
...
+                end_offset = max(end_offset, start_offset + fileoff + filesize)
+        pos += cmdsize
+    return end_offset
+
+
+def _calculate_end_offset(file: File, start_offset: int) -> int:
+    file.seek(start_offset)
+    magic_bytes = file.read(4)
+    if len(magic_bytes) < 4:
+        raise InvalidInputFormat("Mach-O magic truncated")
```
qkaiser: This is impossible to trigger since you matched on at least 4 bytes through `PATTERNS`.

python/unblob/handlers/executable/macho.py
```diff
@@ -0,0 +1,124 @@
...
+            ),
+        ],
+        limitations=[],
+    )
+
+    def calculate_chunk(self, file: File, start_offset: int) -> ValidChunk | None:
+        try:
+            end_offset = _calculate_end_offset(file, start_offset)
+        except struct.error as e:
+            raise InvalidInputFormat(f"Mach-O parse error: {e}") from e
```
qkaiser: Make sure you match the whole header through `PATTERNS` so you don't need to take care of this.

python/unblob/handlers/executable/macho.py
```diff
@@ -0,0 +1,124 @@
...
+    file.seek(start_offset)
+    hdr = file.read(hdr_size)
+    if len(hdr) < hdr_size:
+        raise InvalidInputFormat("Mach-O header truncated")
+
+    ncmds, sizeofcmds = struct.unpack_from(f"{endian}II", hdr, 16)
+    lc_end = start_offset + hdr_size + sizeofcmds
+
+    file.seek(start_offset + hdr_size)
+    lc_data = file.read(sizeofcmds)
```
qkaiser: Why don't you use a `StructParser` to parse a well defined header ?

python/unblob/handlers/executable/macho.py
```diff
@@ -0,0 +1,124 @@
...
+        if cmd == _LC_SEGMENT_64:
+            fileoff, filesize = struct.unpack_from(f"{endian}QQ", lc_data, pos + 40)
+            if filesize > 0:
+                end_offset = max(end_offset, start_offset + fileoff + filesize)
+        elif cmd == _LC_SEGMENT:
+            fileoff, filesize = struct.unpack_from(f"{endian}II", lc_data, pos + 32)
+            if filesize > 0:
+                end_offset = max(end_offset, start_offset + fileoff + filesize)
+        pos += cmdsize
+    return end_offset
```
qkaiser: Use `StructParser` and a well defined struct for load commands so the code is easier to read.

python/unblob/handlers/executable/macho.py
```diff
@@ -0,0 +1,124 @@
...
+logger = get_logger()
+
+# Mach-O magic values (read as little-endian uint32)
+_MAGIC_64_LE = 0xFEEDFACF  # bytes: CF FA ED FE
+_MAGIC_32_LE = 0xFEEDFACE  # bytes: CE FA ED FE
+_MAGIC_64_BE = 0xCFFAEDFE  # bytes: FE ED FA CF
+_MAGIC_32_BE = 0xCEFAEDFE  # bytes: FE ED FA CE
+
+_LC_SEGMENT = 0x1
+_LC_SEGMENT_64 = 0x19
```
qkaiser: You probably want an `IntEnum` for load command types.

qkaiser (commented):
You can merge both commits. It would be nice to have one sample per match (LE32, BE32, LE64, BE64), and samples for fat Mach-O.

python/unblob/handlers/executable/macho.py
```diff
@@ -0,0 +1,265 @@
...
+                title="Mach-O Programming Topics",
+                url="https://developer.apple.com/library/archive/documentation/DeveloperTools/Conceptual/MachOTopics/0-Introduction/introduction.html",
+            ),
+        ],
+        limitations=[],
+    )
+
+    def calculate_chunk(self, file: File, start_offset: int) -> ValidChunk | None:
+        magic = int.from_bytes(file.read(4), "little")
+        file.seek(start_offset)
```
qkaiser: Not needed.

```suggestion
```

python/unblob/handlers/executable/macho.py
```diff
@@ -0,0 +1,265 @@
...
+            Reference(
+                title="Mach-O Programming Topics",
+                url="https://developer.apple.com/library/archive/documentation/DeveloperTools/Conceptual/MachOTopics/0-Introduction/introduction.html",
+            ),
+        ],
+        limitations=[],
+    )
+
+    def calculate_chunk(self, file: File, start_offset: int) -> ValidChunk | None:
+        magic = int.from_bytes(file.read(4), "little")
```
qkaiser: There's a file util function named `convert_int32` for this. Also if you store `_MAGIC_FAT` as bytes you don't need to do the conversion.

python/unblob/handlers/executable/macho.py
```diff
@@ -0,0 +1,265 @@
...
+        if magic in (_MAGIC_64_LE, _MAGIC_32_LE):
+            endian = Endian.LITTLE
+            is_64bit = magic == _MAGIC_64_LE
+        elif magic in (_MAGIC_64_BE, _MAGIC_32_BE):
+            endian = Endian.BIG
+            is_64bit = magic == _MAGIC_64_BE
+        else:
+            raise InvalidInputFormat(f"Unknown Mach-O magic: {magic:#010x}")
+
+        return self._calculate_thin_chunk(file, start_offset, endian, is_64bit=is_64bit)
```
qkaiser: A `match` clause might be in order here.

python/unblob/handlers/executable/macho.py
```diff
@@ -0,0 +1,265 @@
...
+
+    def _calculate_thin_chunk(
+        self, file: File, start_offset: int, endian: Endian, *, is_64bit: bool
+    ) -> ValidChunk:
+        header = self.parse_header(file, endian)
+
+        # 64-bit headers have an extra 4-byte reserved field after flags
+        hdr_size = _HEADER_SIZE_64 if is_64bit else _HEADER_SIZE_32
+        if is_64bit:
+            file.seek(start_offset + hdr_size)
```
qkaiser: Always use explicit `whence`:

```suggestion
            file.seek(start_offset + hdr_size, io.SEEK_SET)
```

python/unblob/handlers/executable/macho.py
```diff
@@ -0,0 +1,265 @@
...
+                file.seek(lc_start)
+                seg = self._struct_parser.parse(
+                    "macho_segment_command_32_t", file, endian
+                )
+                if seg.filesize > 0:
+                    end_offset = max(
+                        end_offset, start_offset + seg.fileoff + seg.filesize
+                    )
+
+            file.seek(lc_start + lc.cmdsize)
```
qkaiser:
```suggestion
            file.seek(lc_start + lc.cmdsize, io.SEEK_SET)
```

python/unblob/handlers/executable/macho.py
```diff
@@ -0,0 +1,265 @@
...
+            char   segname[16];
+            uint64 vmaddr;
+            uint64 vmsize;
+            uint64 fileoff;
+            uint64 filesize;
+            uint32 maxprot;
+            uint32 initprot;
+            uint32 nsects;
+            uint32 flags;
+        } macho_segment_command_64_t;
```
qkaiser: You could have a single structure with a union to cover both 32 and 64 bits variants no ?

python/unblob/handlers/executable/macho.py
```diff
@@ -0,0 +1,265 @@
...
+                    )
+            elif lc.cmd == LoadCommandType.SEGMENT:
+                file.seek(lc_start)
+                seg = self._struct_parser.parse(
+                    "macho_segment_command_32_t", file, endian
+                )
+                if seg.filesize > 0:
+                    end_offset = max(
+                        end_offset, start_offset + seg.fileoff + seg.filesize
+                    )
```
qkaiser, line 236: If you use a `union` in your struct, you don't need to parse twice. Just parse once and access the correct member of your union based on the `cmd` attribute.

python/unblob/handlers/executable/macho.py
```diff
@@ -0,0 +1,265 @@
...
+@attrs.define(repr=False)
+class MachoChunk(ValidChunk):
+    def extract(self, inpath: Path, outdir: Path):
+        with File.from_path(inpath) as file:
+            magic = int.from_bytes(file.read(4), "little")
+            file.seek(0)
+
+            if magic != _MAGIC_FAT:
+                # Thin Mach-O: nothing to extract; raising ExtractError preserves the carved file
+                raise ExtractError
```
qkaiser: I would encode `_MAGIC_FAT` as bytes and compare against a slice of the file so you don't need to seek back.

Like:

```suggestion
            if file[0:len(_MAGIC_FAT)] != _MAGIC_FAT:
                # Thin Mach-O: nothing to extract; raising ExtractError preserves the carved file
                raise ExtractError
```

qkaiser: Woud make sense to add an explicit error message to the `ExtractError`.

python/unblob/handlers/executable/macho.py
```diff
@@ -0,0 +1,265 @@
...
+
+            fat_hdr = _fat_parser.parse("fat_header_t", file, Endian.BIG)
+            arches = [
+                _fat_parser.parse("fat_arch_t", file, Endian.BIG)
+                for _ in range(fat_hdr.nfat_arch)
+            ]
+            outdir.mkdir(parents=True, exist_ok=False)
```
qkaiser: unblob takes care of it, not needed:

```suggestion
```

python/unblob/handlers/executable/macho.py
```diff
@@ -0,0 +1,265 @@
...
+
+            if magic != _MAGIC_FAT:
+                # Thin Mach-O: nothing to extract; raising ExtractError preserves the carved file
+                raise ExtractError
+
+            fat_hdr = _fat_parser.parse("fat_header_t", file, Endian.BIG)
+            arches = [
+                _fat_parser.parse("fat_arch_t", file, Endian.BIG)
+                for _ in range(fat_hdr.nfat_arch)
+            ]
```
qkaiser:
```suggestion
            architectures = [
                _fat_parser.parse("fat_arch_t", file, Endian.BIG)
                for _ in range(fat_hdr.nfat_arch)
            ]
```

python/unblob/handlers/executable/macho.py
```diff
@@ -0,0 +1,265 @@
...
+            outdir.mkdir(parents=True, exist_ok=False)
+            for i, arch in enumerate(arches):
+                if arch.slice_size == 0:
+                    continue
+                cpu_name = _cpu_arch_name(arch.cputype, arch.cpusubtype)
+                slice_path = outdir / f"arch_{i}_{cpu_name}.macho"
+                carve(slice_path, file, arch.offset, arch.slice_size)
```
qkaiser: Why the `i` index ? Is it possible to have different slices with the exact same CPU target within a Mach-O ?

python/unblob/handlers/executable/macho.py
```diff
@@ -0,0 +1,265 @@
...
+    } fat_header_t;
+
+    typedef struct fat_arch {
+        uint32 cputype;
+        uint32 cpusubtype;
+        uint32 offset;
+        uint32 slice_size;
+        uint32 align;
+    } fat_arch_t;
+""")
```
qkaiser: These are already defined in your handler struct parser. Centralize your definitions in a single C_DEFINITIONS constant and use it where needed (`_fat_parser`, `MachOHandler`).

python/unblob/handlers/executable/macho.py
elektrischermoench: You are right about this. It will be rejected by the apple toolchain like lipo. But it could be present in a handcrafted or malicious one. But there is no need to protect against definition of carve already is secure against with open("xb") I missed that and overengineered.

python/unblob/handlers/executable/macho.py
```diff
@@ -0,0 +1,251 @@
...
+            case m if m in (_MAGIC_64_LE, _MAGIC_32_LE):
+                endian = Endian.LITTLE
+                is_64bit = m == _MAGIC_64_LE
+            case m if m in (_MAGIC_64_BE, _MAGIC_32_BE):
+                endian = Endian.BIG
+                is_64bit = m == _MAGIC_64_BE
+            case _:
+                raise InvalidInputFormat(f"Unknown Mach-O magic: {magic:#010x}")
+
+        return self._calculate_thin_chunk(file, start_offset, endian, is_64bit=is_64bit)
```
qkaiser: You'll need to define `_MAGIC_FAT` as an int, but you can do something super explicit and easier to read with:

```suggestion
        magic = convert_int32(file[start_offset:start_offset+4], Endian.LITTLE)
        match magic:
            case _MAGIC_FAT:
                return self._calculate_fat_chunk(file, start_offset)    
            case _MAGIC_THIN_64_LE:
                return self._calculate_thin_chunk(file, start_offset, Endian.LITTLE, is_64bit=True)
            case _MAGIC_THIN_32_LE:
                return self._calculate_thin_chunk(file, start_offset, Endian.LITTLE, is_64bit=False)
            case _MAGIC_THIN_64_BE:
                return self._calculate_thin_chunk(file, start_offset, Endian.BIG, is_64bit=True)
            case _MAGIC_THIN_32_BE:
                return self._calculate_thin_chunk(file, start_offset, Endian.BIG, is_64bit=False)
            case _:
                raise InvalidInputFormat(f"Unknown Mach-O magic: {magic:#010x}")
```

python/unblob/handlers/executable/macho.py
```diff
@@ -0,0 +1,251 @@
...
+
+# Thin Mach-O magic values as little-endian uint32 (file-order bytes shown per line)
+_MAGIC_64_LE = 0xFEEDFACF  # CF FA ED FE — 64-bit little-endian (arm64, x86_64)
+_MAGIC_32_LE = 0xFEEDFACE  # CE FA ED FE — 32-bit little-endian (armv7, x86)
+_MAGIC_64_BE = 0xCFFAEDFE  # FE ED FA CF — 64-bit big-endian
+_MAGIC_32_BE = 0xCEFAEDFE  # FE ED FA CE — 32-bit big-endian
+
+# Fat (universal) binary magic — always big-endian in the file. Stored as bytes so it can
+# be compared against the file's leading bytes directly, without reading and seeking back.
+_MAGIC_FAT = bytes.fromhex("CAFEBABE")  # CA FE BA BE
```
qkaiser:
```suggestion
_MAGIC_FAT = 0xCAFEBABE
```

python/unblob/handlers/executable/macho.py
```diff
@@ -0,0 +1,251 @@
...
+            architectures = [
+                _parser.parse("fat_arch_t", file, Endian.BIG)
+                for _ in range(fat_hdr.nfat_arch)
+            ]
+            for arch in architectures:
+                if arch.slice_size == 0:
+                    continue
+                cpu_name = _cpu_arch_name(arch.cputype, arch.cpusubtype)
+                slice_path = outdir / f"arch_{cpu_name}.macho"
+                carve(slice_path, file, arch.offset, arch.slice_size)
```
qkaiser: You should use `Filesystem`

qkaiser (commented):
~Looks good to me but your integration tests output is missing I think. Please add it before we launch the pipeline.~
Oh wait, so the file are not carved/extracted because they don' t have prefix, suffix. Correct ?

elektrischermoench:
> ~Looks good to me but your integration tests output is missing I think. Please add it before we launch the pipeline.~ Oh wait, so the file are not carved/extracted because they don' t have prefix, suffix. Correct ?

Yes.

qkaiser (approved):

PR #1492, opened 2026-04-20
feat(handler): add Apple IMG4/IM4P handler

elektrischermoench:
Adds a handler for Apple IMG4 containers (DER-encoded ASN.1). Supports both 0x82 (16-bit length) and 0x83 (24-bit length) DER header variants and extracts the raw IM4P payload.

### DIFF
--- a/python/unblob/handlers/archive/apple/img4.py
+++ b/python/unblob/handlers/archive/apple/img4.py
@@ -1,0 +1,133 @@
+from pathlib import Path
+
+from unblob.file_utils import File, FileSystem, InvalidInputFormat, iterate_patterns
+from unblob.models import (
+    Extractor,
+    ExtractResult,
+    HandlerDoc,
+    HandlerType,
+    HexString,
+    Reference,
+    StructHandler,
+    ValidChunk,
+)
+
+# IM4P (Image4 Payload) DER structure inside an IMG4 container:
+#   SEQUENCE
+#     IA5String "IM4P"       (image type tag)
+#     IA5String <name>       (component name, e.g. "illb")
+#     OCTET STRING <payload> (compressed or raw binary)
+_IM4P_MAGIC = b"IM4P"
+_IA5STRING_TAG = 0x16
+_OCTET_STRING_TAG = 0x04
+
+
+def _read_der_length(file: File, offset: int) -> tuple[int, int]:
+    """Return (value, header_len) for the DER length field starting at offset."""
+    file.seek(offset)
+    first = file.read(1)[0]
+    if first < 0x80:  # short form: the byte itself is the length
+        return first, 1
+    num_bytes = first & 0x7F  # long form: low 7 bits count the length bytes
+    return int.from_bytes(file.read(num_bytes), "big"), 1 + num_bytes
+
+
+class IM4PExtractor(Extractor):
+    def extract(self, inpath: Path, outdir: Path) -> ExtractResult:
+        fs = FileSystem(outdir)
+
+        with File.from_path(inpath) as file:
+            for pos in iterate_patterns(file, _IM4P_MAGIC):
+                # pos is the start of the "IM4P" IA5String content (4 bytes).
+                p = pos + len(_IM4P_MAGIC)
+
+                # component name IA5String: tag, length, name bytes
+                file.seek(p)
+                if file.read(1)[0] != _IA5STRING_TAG:
+                    continue
+                p += 2 + file.read(1)[0]
+
+                # payload OCTET STRING: tag, DER length, payload bytes
+                file.seek(p)
+                if file.read(1)[0] != _OCTET_STRING_TAG:
+                    continue
+                size, header_len = _read_der_length(file, p + 1)
+                payload_start = p + 1 + header_len
+
+                available = file.size() - payload_start
+                fs.carve(
+                    Path(f"{inpath.stem}.bin"),
+                    file,
+                    payload_start,
+                    min(size, available),
+                )
+                break
+
+        return ExtractResult(reports=fs.problems)
+
+
+class IMG4Handler(StructHandler):
+    NAME = "img4"
+
+    PATTERNS = [
+        # 30          — DER SEQUENCE tag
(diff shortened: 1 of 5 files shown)


python/unblob/handlers/archive/apple/img4.py
```diff
@@ -0,0 +1,108 @@
...
+                fs.write_bytes(Path("im4p_payload.bin"), data[magic_idx + 12 :])
+
+        return ExtractResult(reports=fs.problems)
+
+
+class IMG4Handler(Handler):
+    NAME = "img4"
+    PATTERNS = [
+        HexString("30 82 ?? ?? 16 04 49 4D 34 50"),
+        HexString("30 83 ?? ?? ?? 16 04 49 4D 34 50"),
```
qkaiser, line 84: What does it represent ? Is it unique to img4 or will we match on everything DER encoded ?

python/unblob/handlers/archive/apple/img4.py
```diff
@@ -0,0 +1,108 @@
...
+                url="https://github.com/apple-oss-distributions/libimg4",
+            ),
+        ],
+        limitations=[
+            "Only IM4P payload extraction is supported; full IMG4 manifests are not verified"
+        ],
+    )
+
+    def __init__(self):
+        self._struct_parser = StructParser(IMG4_C_DEFINITIONS)
```
qkaiser: Make `IMG4Handler` inherit from `StructHandler` it's way simpler.

python/unblob/handlers/archive/apple/img4.py
```diff
@@ -0,0 +1,108 @@
...
+                total_size = 4 + header.length
+            elif l_type == 0x83:
+                file.seek(start_offset)
+                header = self._struct_parser.parse(
+                    "der_header_83_t", file, endian=Endian.BIG
+                )
+                actual_len = (header.len_hi << 16) | header.len_lo
+                total_size = 5 + actual_len
+            else:
+                return None
```
qkaiser: I'm pretty sure you can decode DER streams using standard libs.

python/unblob/handlers/archive/apple/img4.py
```diff
@@ -0,0 +1,108 @@
...
+
+            file.seek(0, io.SEEK_END)
+            file_size = file.tell()
```
qkaiser: There's file.size()

python/unblob/handlers/archive/apple/img4.py
```diff
@@ -0,0 +1,108 @@
...
+
+            if start_offset + total_size > file_size:
+                return None
```
qkaiser: You don't need to check yourself, unblob will complain if the chunk overflows.

python/unblob/handlers/archive/apple/img4.py
```diff
@@ -0,0 +1,108 @@
...
+
+            return ValidChunk(
+                start_offset=start_offset, end_offset=start_offset + total_size
+            )
+        finally:
+            file.seek(current_pos)
```
qkaiser: You don't need that.

python/unblob/handlers/archive/apple/img4.py
```diff
@@ -0,0 +1,108 @@
...
+
+class IM4PExtractor(Extractor):
+    def extract(self, inpath: Path, outdir: Path) -> ExtractResult:
+        fs = FileSystem(outdir)
+
+        with File.from_path(inpath) as file:
+            data = file.read()
+            magic_idx = data.find(b"IM4P")
+            if magic_idx != -1:
+                fs.write_bytes(Path("im4p_payload.bin"), data[magic_idx + 12 :])
```
qkaiser: So we're simply carving everything but the header ? Also you should use  `fs.carve` instead of loading the whole thing in memory and writing it with  `write_bytes`.

python/unblob/handlers/archive/apple/img4.py
elektrischermoench: Kind of specific, see explanations in the comment

python/unblob/handlers/archive/apple/img4.py
```diff
@@ -0,0 +1,106 @@
...
+            length = int.from_bytes([header.b2, header.b3], "big")
+            total_size = 4 + length
+        elif header.length_type == 0x83:
+            length = int.from_bytes([header.b2, header.b3, header.b4], "big")
+            total_size = 5 + length
+        else:
+            raise InvalidInputFormat(
+                f"IMG4: unexpected DER length type {header.length_type:#x}"
+            )
+
```
qkaiser, line 130:
```suggestion
        if header.length_type == 0x82:
            length = int.from_bytes([header.b2, header.b3], "big")
        elif header.length_type == 0x83:
            length = int.from_bytes([header.b2, header.b3, header.b4], "big")
        else:
            raise InvalidInputFormat(
                f"IMG4: unexpected DER length type {header.length_type:#x}"
            )
        total_size = len(header) + length
```

python/unblob/handlers/archive/apple/img4.py
```diff
@@ -0,0 +1,106 @@
...
+                url="https://github.com/apple-oss-distributions/libimg4",
+            ),
+        ],
+        limitations=[
+            "Only IM4P payload extraction is supported; full IMG4 manifests are not verified"
+        ],
+    )
+
+    def calculate_chunk(self, file: File, start_offset: int) -> ValidChunk | None:
+        header = self.parse_header(file)
```
qkaiser, line 118: You could use an asn1 parser like `asn1crypto`, this way you don't need to reinvent the wheel to parse the payload size.

qkaiser:
@elektrischermoench if you touch this code, I recommend you squash both commits into one

elektrischermoench:
Rebased onto latest `main` (was 92 commits behind, base was the Mach-O merge `aa7e12cd`). Still a single commit, no conflicts. Handler docs regenerated and in sync. @qkaiser could you take another look when you get a chance?

PR #1493, opened 2026-04-20
feat(handler): add dyld shared cache multi-file handler

elektrischermoench:
Adds a directory handler for split dyld shared cache layouts (dyld_shared_cache_* glob). Groups the primary cache file with its .01, .atlas, .dylddata, .symbols and .dyldlinkedit siblings into a unified logical view for extraction.

### DIFF
--- a/python/unblob/handlers/archive/apple/dyld.py
+++ b/python/unblob/handlers/archive/apple/dyld.py
@@ -1,0 +1,68 @@
+from pathlib import Path
+
+from structlog import get_logger
+
+from unblob.models import (
+    DirectoryHandler,
+    Glob,
+    HandlerDoc,
+    HandlerType,
+    MultiFile,
+    Reference,
+)
+
+logger = get_logger()
+
+_MAGIC_PREFIX = b"dyld_v1 "
+
+
+class MultifileDyldCacheHandler(DirectoryHandler):
+    NAME = "multifile_dyld_cache"
+
+    EXTRACTOR = None
+    PATTERN = Glob("dyld_shared_cache_*")
+
+    DOC = HandlerDoc(
+        name="dyld Shared Cache",
+        description="The dyld shared cache is a pre-linked collection of system dynamic libraries used by macOS and iOS to accelerate application launch. Modern caches are split across multiple files with suffixes such as .01, .symbols, .atlas, .dylddata, and .dyldlinkedit.",
+        handler_type=HandlerType.ARCHIVE,
+        vendor="Apple",
+        references=[
+            Reference(
+                title="dyld - Apple open-source dynamic linker",
+                url="https://github.com/apple-oss-distributions/dyld",
+            ),
+        ],
+        limitations=[],
+    )
+
+    def calculate_multifile(self, file: Path) -> MultiFile | None:
+        if file.suffix:
+            return None
+
+        try:
+            with file.open("rb") as f:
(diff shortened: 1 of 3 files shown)


python/unblob/handlers/archive/apple/dyld.py
```diff
@@ -0,0 +1,77 @@
...
+class MultifileDyldCacheHandler(DirectoryHandler):
+    NAME = "multifile_dyld_cache"
+
+    EXTRACTOR = None
+    PATTERN = Glob("dyld_shared_cache_*")
+
+    DOC = HandlerDoc(
+        name="dyld Shared Cache",
+        description="The dyld shared cache is a pre-linked collection of system dynamic libraries used by macOS and iOS to accelerate application launch. Modern caches are split across multiple files with suffixes such as .01, .symbols, .atlas, .dylddata, and .dyldlinkedit.",
+        handler_type=HandlerType.FILESYSTEM,
```
qkaiser:
```suggestion
        handler_type=HandlerType.ARCHIVE,
```

python/unblob/handlers/archive/apple/dyld.py
```diff
@@ -0,0 +1,77 @@
...
+        ],
+        limitations=[],
+    )
+
+    def calculate_multifile(self, file: Path) -> MultiFile | None:
+        if any(
+            file.name.endswith(s)
+            for s in [".01", ".atlas", ".dylddata", ".symbols", ".dyldlinkedit"]
+        ):
+            return None
```
qkaiser: You're checking for what it should not ends with. Are you looking for a file with a specific extension ? If so this check could be simplified.

python/unblob/handlers/archive/apple/dyld.py
```diff
@@ -0,0 +1,77 @@
...
+
+        try:
+            with file.open("rb") as f:
+                if not f.read(8).startswith(_MAGIC_PREFIX):
+                    return None
+        except Exception:
+            return None
```
qkaiser: Broad exception handling is bad :)

python/unblob/handlers/archive/apple/dyld.py
```diff
@@ -0,0 +1,77 @@
...
+
+        base_name = file.name
+        for suffix in [".01", ".atlas", ".dylddata", ".symbols", ".dyldlinkedit"]:
+            if base_name.endswith(suffix):
+                base_name = base_name[: -len(suffix)]
+                break
```
qkaiser: You can use `.stem` and `.suffix` from pathlib's Path here.

python/unblob/handlers/archive/apple/dyld.py
```diff
@@ -0,0 +1,77 @@
...
+        )
+
+        if not siblings:
+            return None
+
+        logger.info(
+            "creating unified dyld cache view", main=file.name, parts=len(siblings) + 1
+        )
+
+        return MultiFile(name=f"{base_name}.unified", paths=[file, *siblings])
```
qkaiser: What's the purpose of this handler really ? I don't see an extractor that does something with these files so I'm a bit confused.

elektrischermoench: It is just a directory handler to collect the parts and mark them as a multi file the content can be extracted with other handlers if this is done.

PR #1494, opened 2026-04-20
feat(handler): add APFS container handler

elektrischermoench:
Adds a handler for Apple File System (APFS) container images. Parses the NX superblock to determine container geometry (block size and block count) and delegates extraction to 7zz.

### DIFF
--- a/python/unblob/handlers/filesystem/apfs.py
+++ b/python/unblob/handlers/filesystem/apfs.py
@@ -1,0 +1,101 @@
+import io
+
+from unblob.extractors import Command
+from unblob.file_utils import InvalidInputFormat
+from unblob.models import (
+    File,
+    HandlerDoc,
+    HandlerType,
+    HexString,
+    Reference,
+    StructHandler,
+    ValidChunk,
+)
+
+_MIN_BLOCK_SIZE = 512
+_MAX_BLOCK_SIZE = 1024 * 1024  # 1 MB sanity cap
+
+
+def _fletcher64_valid(block: bytes) -> bool:
+    mod = 0xFFFFFFFF
+    sum1 = sum2 = 0
+    for offset in range(8, len(block), 4):
+        sum1 = (sum1 + int.from_bytes(block[offset : offset + 4], "little")) % mod
+        sum2 = (sum2 + sum1) % mod
+    check1 = mod - ((sum1 + sum2) % mod)
+    check2 = mod - ((sum1 + check1) % mod)
+    expected = check1 | (check2 << 32)
+    return expected == int.from_bytes(block[0:8], "little")
+
+
+class APFSHandler(StructHandler):
+    NAME = "apfs"
+
+    PATTERNS = [
+        # Match the superblock's object header, not just the "NXSB" magic, so we don't
+        # fire on every occurrence of that string:
+        #   01 00 00 80    o_type = 0x80000001 (NX_SUPERBLOCK | PHYSICAL flag), little-endian
+        #   ?? ?? ?? ??    o_subtype
+        #   4E 58 53 42    nx_magic "NXSB"
+        HexString("01 00 00 80 ?? ?? ?? ?? 4E 58 53 42"),
+    ]
+
+    # The match starts at o_type, 24 bytes into the superblock; shift start back to block 0.
+    PATTERN_MATCH_OFFSET = -24
+
+    # APFS container superblock layout (all fields little-endian):
+    #   offset  0: obj_phys_t (32 bytes) — Fletcher-64 checksum + oid + xid + type + subtype
+    #   offset 32: nx_magic    uint32  "NXSB" (0x4253584e)
+    #   offset 36: nx_block_size  uint32
+    #   offset 40: nx_block_count uint64
+    C_DEFINITIONS = r"""
+        typedef struct apfs_nx_superblock {
+            char    o_cksum[8];     // Fletcher-64 checksum
+            uint64  o_oid;          // object identifier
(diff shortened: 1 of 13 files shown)


python/unblob/handlers/archive/apple/apfs.py
```diff
@@ -0,0 +1,77 @@
...
+
+class APFSHandler(Handler):
+    NAME = "apfs"
+
+    PATTERNS = [HexString("4e 58 53 42")]  # "NXSB"
+
+    # NXSB sits 32 bytes into the container superblock; shift start back to block 0.
+    PATTERN_MATCH_OFFSET = -_NXSB_MAGIC_OFFSET
+
+    EXTRACTOR = Command("7zz", "x", "{inpath}", "-o{outdir}")
```
qkaiser: Does it support APFS ? Don't we use "7z" instead of "7zz" ?

python/unblob/handlers/archive/apple/apfs.py
qkaiser, line 1: Should be moved from "archive" to "filesystem".

python/unblob/handlers/archive/apple/apfs.py
```diff
@@ -0,0 +1,77 @@
...
+        if block_count == 0:
+            raise InvalidInputFormat("APFS block_count is zero")
+
+        end_offset = start_offset + block_size * block_count
+        logger.debug(
+            "APFS container geometry",
+            block_size=block_size,
+            block_count=block_count,
+            end_offset=end_offset,
+        )
```
qkaiser: Not needed.

python/unblob/handlers/archive/apple/apfs.py
```diff
@@ -0,0 +1,77 @@
...
+        if len(raw) < 12:
+            raise InvalidInputFormat("APFS superblock truncated")
+
+        block_size = struct.unpack_from("<I", raw, 0)[0]
+        block_count = struct.unpack_from("<Q", raw, 4)[0]
+
+        if not (_MIN_BLOCK_SIZE <= block_size <= _MAX_BLOCK_SIZE):
+            raise InvalidInputFormat(f"APFS block_size out of range: {block_size:#x}")
+        if block_count == 0:
+            raise InvalidInputFormat("APFS block_count is zero")
```
qkaiser: Just use `StructHandler` with a well defined header for the container superblock.

python/unblob/handlers/archive/apple/apfs.py
elektrischermoench: Ah damn you are right. Yes it supports APFS but not all of them. recoveryOS and some encrypted ones are not supported yet.

python/unblob/handlers/filesystem/apfs.py
```diff
@@ -0,0 +1,75 @@
...
+_MAX_BLOCK_SIZE = 1024 * 1024  # 1 MB sanity cap
+
+
+class APFSHandler(StructHandler):
+    NAME = "apfs"
+
+    PATTERNS = [HexString("4e 58 53 42")]  # "NXSB"
+
+    # NXSB sits 32 bytes into the container superblock; shift start back to block 0.
+    PATTERN_MATCH_OFFSET = -_NXSB_MAGIC_OFFSET
```
qkaiser: It's only used once so no need to use a constant

python/unblob/handlers/filesystem/apfs.py
```diff
@@ -0,0 +1,86 @@
...
+            raise InvalidInputFormat(
+                f"APFS superblock o_type mismatch: {header.o_type:#x}"
+            )
+
+        if not (_MIN_BLOCK_SIZE <= header.nx_block_size <= _MAX_BLOCK_SIZE):
+            raise InvalidInputFormat(
+                f"APFS block_size out of range: {header.nx_block_size:#x}"
+            )
+        if header.nx_block_count == 0:
+            raise InvalidInputFormat("APFS block_count is zero")
```
qkaiser: Move this to `is_valid_header`

python/unblob/handlers/filesystem/apfs.py
```diff
@@ -0,0 +1,86 @@
...
+    PATTERN_MATCH_OFFSET = -_NXSB_MAGIC_OFFSET
+
+    # APFS container superblock layout (all fields little-endian):
+    #   offset  0: obj_phys_t (32 bytes) — Fletcher-64 checksum + oid + xid + type + subtype
+    #   offset 32: nx_magic    uint32  "NXSB" (0x4253584e)
+    #   offset 36: nx_block_size  uint32
+    #   offset 40: nx_block_count uint64
+    C_DEFINITIONS = r"""
+        typedef struct apfs_nx_superblock {
+            char    o_cksum[8];     // Fletcher-64 checksum
```
qkaiser, line 53: Do we want to validate this checksum ?

python/unblob/handlers/filesystem/apfs.py
```diff
@@ -0,0 +1,86 @@
...
+            Reference(
+                title="Apple File System Reference",
+                url="https://developer.apple.com/support/downloads/Apple-File-System-Reference.pdf",
+            ),
+        ],
+        limitations=[],
+    )
+
+    def valid_header(self, header) -> bool:
+        return header.o_type == _NX_OBJ_TYPE_NX_SUPERBLOCK
```
qkaiser: This can be enforced through `PATTERNS` by placing `80 00 00 01` before `4e 58 53 42`. My endianness may be wrong, to be checked.

PR #1495, opened 2026-04-21
feat(handler): add support for Moxa FRM archives

jstucke:
Adds support for Moxa FRM archives.
This format is used in Moxa firmware of NPort, MGate and MiiNePort devices (e.g. NPort IA5000).
These archives seem to contain files for a web server (optionally GZIP compressed).

Samples can be found here: https://www.moxa.com/en/support/product-support/software-and-documentation

### DIFF
--- a/python/unblob/handlers/archive/moxa/frm.py
+++ b/python/unblob/handlers/archive/moxa/frm.py
@@ -1,0 +1,148 @@
+from __future__ import annotations
+
+import gzip
+import io
+from enum import IntEnum
+from pathlib import Path
+
+from unblob.file_utils import Endian, File, FileSystem, InvalidInputFormat, StructParser
+from unblob.models import (
+    Extractor,
+    HandlerDoc,
+    HandlerType,
+    HexString,
+    StructHandler,
+    ValidChunk,
+)
+
+C_DEFINITIONS = r"""
+    typedef struct frm_container_header {
+        char    magic[4];       /* *FRM */
+        uint32  unknown1;       /* version maybe, only 00 00 00 01 observed */
+        uint32  total_length;   /* the size of the whole container (should equal the sum of the section sizes + header_length) */
+        uint16  header_length;  /* only 60 00 observed -> size 0x60 = 96 */
+        uint16  section_count;  /* only 02 00 observed -> 2 section entries (webserver FS and FW binary) */
+        uint8   unknown2[48];   /* optional unknown entries; empty in some samples */
+    } frm_container_header_t;
+
+    typedef struct frm_section_entry {
+        uint32  type;   /* probably type; values 1 (FW binary) and 2 (webserver FS) observed */
+        uint32  offset;
+        uint32  length;
+        uint32  unknown;
+    } frm_section_entry_t;
+
+    typedef struct frm_fs_header {
+        char    device_name[32];
+        uint8   unknown1[4];        /* version maybe? */
+        uint32  timestamp;          /* creation time (UNIX timestamp) */
+        uint32  unknown2;
+        uint32  unknown3;           /* maybe checksum? */
+        uint32  file_table_offset;  /* offset of the file table; only 0x100 observed -> usually starts at 0x160 */
+        uint32  file_table_length;  /* size of the file table (= file_header_length * file_count) */
+        uint16  file_header_length; /* size of each entry; only 0x40 (64) observed */
+        uint16  file_count;         /* the number of table entries (files) */
+        uint32  data_length;        /* the length of the file content data */
+        uint16  unknown4;
+        uint16  file_count2;        /* for some devices (e.g. Nport 5600) this field contains the file count and file_count is 0 */
+        uint32  file_table_length2;
+        uint32  unknown5;
+    } frm_fs_header_t;
+
+    typedef struct frm_file_header {
+        char    name[48];
+        uint8   unknown[8];    /* could be the creation time */
+        uint32  file_length;
+        uint32  data_offset;
+    } frm_file_header_t;
+"""
+
+MAGIC = b"*FRM"
+GZIP_MAGIC = bytes.fromhex("1f 8b")
+
+
+class SectionTypes(IntEnum):
+    FW_BINARY = 1
+    FILESYSTEM = 2
+
+
+class MoxaFRMExtractor(Extractor):
+    def __init__(self):
+        self._struct_parser = StructParser(C_DEFINITIONS)
+
+    def extract(self, inpath: Path, outdir: Path):
+        fs = FileSystem(outdir)
+        with File.from_path(inpath) as file:
+            header = self._struct_parser.parse(
+                "frm_container_header_t", file, Endian.LITTLE
+            )
+            sections = self._parse_section_table(file, header.section_count)
+            for section in sections:
+                self._extract_section(section, file, fs)
+
+    def _parse_section_table(self, file: File, section_count: int):
+        return [
+            self._struct_parser.parse("frm_section_entry_t", file, Endian.LITTLE)
+            for _ in range(section_count)
+        ]
+
+    def _extract_section(self, section, file: File, fs: FileSystem):
+        match section.type:
+            case SectionTypes.FW_BINARY:
+                fs.carve(Path("firmware.bin"), file, section.offset, section.length)
+            case SectionTypes.FILESYSTEM:
+                self._extract_fs_section(file, fs, section.offset)
+            case _:
+                raise InvalidInputFormat(f"Unknown section type: {section.type}")
+
+    def _extract_fs_section(self, file: File, fs: FileSystem, section_offset: int):
+        file.seek(section_offset, io.SEEK_SET)
+        fs_header = self._struct_parser.parse("frm_fs_header_t", file, Endian.LITTLE)
+
+        file_table_offset = section_offset + fs_header.file_table_offset
+        for index in range(fs_header.file_count or fs_header.file_count2):
+            entry_offset = file_table_offset + index * fs_header.file_header_length
+            file.seek(entry_offset, io.SEEK_SET)
+            entry = self._struct_parser.parse("frm_file_header_t", file, Endian.LITTLE)
+
+            name = bytes(entry.name).rstrip(b"\x00")
+            if not name:
+                continue
+
+            file.seek(section_offset + entry.data_offset, io.SEEK_SET)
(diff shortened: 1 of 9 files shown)


python/unblob/handlers/archive/moxa/frm.py
```diff
@@ -0,0 +1,167 @@
...
+    C_DEFINITIONS = C_DEFINITIONS
+    HEADER_STRUCT = "frm_container_header_t"
+    EXTRACTOR = MoxaFRMExtractor()
+
+    def calculate_chunk(self, file: File, start_offset: int) -> ValidChunk | None:
+        header = self.cparser_le.frm_container_header_t(file)
+        if header.section_count == 0:
+            raise InvalidInputFormat(f"Invalid section count: {header.section_count}")
+        if header.header_length < 16:
+            raise InvalidInputFormat(f"Invalid header length: {header.header_length}")
```
qkaiser: Can be enforced through pattern matching in `PATTERNS` rather than checked after the fact.

python/unblob/handlers/archive/moxa/frm.py
```diff
@@ -0,0 +1,167 @@
...
+        references=[],
+        limitations=[],
+    )
+
+    C_DEFINITIONS = C_DEFINITIONS
+    HEADER_STRUCT = "frm_container_header_t"
+    EXTRACTOR = MoxaFRMExtractor()
+
+    def calculate_chunk(self, file: File, start_offset: int) -> ValidChunk | None:
+        header = self.cparser_le.frm_container_header_t(file)
```
qkaiser: You're in a `StructHandler` with a defined `HEADER_STRUCT`, you can call  `parse_header`.

python/unblob/handlers/archive/moxa/frm.py
```diff
@@ -0,0 +1,167 @@
...
+    PATTERNS = [HexString("2A 46 52 4D 00 00 00 01")]
+
+    DOC = HandlerDoc(
+        name="Moxa FRM",
+        description=(
+            "Firmware container format used in Moxa firmware (e.g. NPort, MGate and MiiNePort devices)."
+        ),
+        handler_type=HandlerType.ARCHIVE,
+        vendor="Moxa",
+        references=[],
```
qkaiser, line 136: No public information about this format ?

python/unblob/handlers/archive/moxa/frm.py
```diff
@@ -0,0 +1,167 @@
...
+    } frm_file_header_t;
+"""
+
+MAGIC = b"*FRM"
+GZIP_MAGIC = bytes.fromhex("1f 8b")
+
+
+class SectionTypes:
+    FW_BINARY = 1
+    FILESYSTEM = 2
```
qkaiser: Should be an `IntEnum`

python/unblob/handlers/archive/moxa/frm.py
```diff
@@ -0,0 +1,167 @@
...
+            self._parse_section_table(file, fs)
+
+    def _parse_section_table(self, file: File, fs: FileSystem):
+        while True:
+            try:
+                section = self._struct_parser.parse(
+                    "frm_section_entry_t", file, Endian.LITTLE
+                )
+            except EOFError:
+                return
```
qkaiser: Shouldn't you parse `section_count` sections ?

python/unblob/handlers/archive/moxa/frm.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+class MoxaFRMExtractor(Extractor):
+    def __init__(self):
+        self._struct_parser = StructParser(C_DEFINITIONS)
+
+    def extract(self, inpath: Path, outdir: Path):
+        fs = FileSystem(outdir)
+        with File.from_path(inpath) as file:
+            self._struct_parser.parse("frm_container_header_t", file, Endian.LITTLE)
```
qkaiser: I would save the header somewhere, can be useful in `_parse_section_table` for example.

python/unblob/handlers/archive/moxa/frm.py
```diff
@@ -0,0 +1,167 @@
...
+        while True:
+            try:
+                section = self._struct_parser.parse(
+                    "frm_section_entry_t", file, Endian.LITTLE
+                )
+            except EOFError:
+                return
+            match section.type:
+                case SectionTypes.FW_BINARY:
+                    with _restore_offset(file):
```
qkaiser: I would collect the sections in a list, then carve them. This way you don't need this.

python/unblob/handlers/archive/moxa/frm.py
```diff
@@ -0,0 +1,167 @@
...
+        # some paths seem to use URL encoding (but with ~ instead of %)
+        out_path = Path(unquote(dir_name.replace("~", "%")))
+        # some (but usually not all) files are GZIP compressed
+        if raw[:2] != GZIP_MAGIC:
+            fs.write_bytes(out_path, raw)
+            return
+        try:
+            fs.write_bytes(out_path, gzip.decompress(raw))
+        except gzip.BadGzipFile as error:
+            raise InvalidInputFormat(f"Invalid GZIP: {dir_name!r}") from error
```
qkaiser: This could be handled by the gzip handler down the line. This would simplify `_extract_fs_section` since you could simply call `fs.carve`.

python/unblob/handlers/archive/moxa/frm.py
jstucke: That is more or less the same problem as with https://github.com/onekey-sec/unblob/pull/1359#discussion_r2745355967, right? (since I mostly want to check that the fields are not null)
Is there a better way to do this now?

python/unblob/handlers/archive/moxa/frm.py
jstucke: Not from Moxa as far as I know. The handler is mainly based on reversing samples. I did not find a lot of information. There is [this GitHub repo](https://github.com/seqred-s-a/NP5200A-RE) but with hardly any information. [This blog] is about a different device with a different kind of firmware (encrypted linux/squashfs).

python/unblob/handlers/archive/moxa/frm.py
jstucke: fixed

python/unblob/handlers/archive/moxa/frm.py
jstucke: changed

python/unblob/handlers/archive/moxa/frm.py
jstucke: Whoops I actually wanted to do just that but somehow lost track of it :sweat_smile: (which is also the reason for the lonely `self._struct_parser.parse("frm_container_header_t", file, Endian.LITTLE)` line)

python/unblob/handlers/archive/moxa/frm.py
jstucke: see above

python/unblob/handlers/archive/moxa/frm.py
jstucke: I thought about that, too. But without the decompression, the directory structure doesn't "line up" (e.g. the files referenced in the html templates are not at their expected location, because they are instead in a subfolder `<filename>_extract/<filename>`  which would make it harder to work with the unpacked result)

jstucke:
I found some samples (Nport 5600 series) with a slightly different FS header structure and made some adjustments. 
I also removed the filename "URL unquoting" because the files seem to be referenced with the quoted names in the html templates.

jstucke:
Do you think it would make sense to put each section in a different subfolder, so the contents don't get mixed up?

python/unblob/handlers/archive/moxa/frm.py
```diff
@@ -0,0 +1,154 @@
...
+        if file_contents[:2] == GZIP_MAGIC:
+            with suppress(gzip.BadGzipFile):
+                file_contents = gzip.decompress(file_contents)
+        fs.write_bytes(file_path, file_contents)
+
+
+class MoxaFRMHandler(StructHandler):
+    NAME = "moxa_frm"
+
+    PATTERNS = [HexString("2A 46 52 4D 00 00 00 01")]
```
qkaiser: According to your comment on lines 24-25 you could use this and get rid of checks on 148-151

```suggestion
    PATTERNS = [HexString("2A 46 52 4D 00 00 00 01 [4] 60 00 02 00")]
```

python/unblob/handlers/archive/moxa/frm.py
qkaiser: > Is there a better way to do this now?

Not yet, would be nice tho. See my other comment.

python/unblob/handlers/archive/moxa/frm.py
```diff
@@ -0,0 +1,154 @@
...
+            file.seek(section_offset + entry.data_offset, io.SEEK_SET)
+            raw = file.read(entry.file_length)
+            self._write_file(fs, Path(name.decode("ascii", errors="replace")), raw)
+
+    @staticmethod
+    def _write_file(fs: FileSystem, file_path: Path, file_contents: bytes):
+        # some (but usually not all) files are GZIP compressed
+        if file_contents[:2] == GZIP_MAGIC:
+            with suppress(gzip.BadGzipFile):
+                file_contents = gzip.decompress(file_contents)
```
qkaiser: if you get `BadGzipFile` it usually means it's reading too far and fail when trying to read another gzip compressed stream after the first one. There's a single stream so it's safer to use `zlib`.

Ideally you want streamed decompression with a `zlib.decompressobj` eating from `iterate_file`.

jstucke: Now that you mention it, the decompression works without any errors (now) with the firmware samples that I have, so we can probably just remove the error handling (and let unblob handle all other exceptions)

python/unblob/handlers/archive/moxa/frm.py
jstucke: changed

qkaiser:
@jstucke you have to build the docs manually

```
uv run unblob --build-handlers-doc docs/handlers.md
```

I rebased your branch so pull first.

jstucke:
> @jstucke you have to build the docs manually

Sorry I should know this by now :sweat_smile:

qkaiser (approved):

PR #1496, opened 2026-04-23
feat(handler) add support for RPM package

### LINKED ISSUE #1489: Format Support - RPM Package
## Summary

Unblob currently handles RPM packages by carving two chunks:

* unknown chunk, which is the RPM header and metadata
* XZ compressed chunk, which is the RPM CPIO archive

This means that for every RPM file, we store at least twice the size since we keep the original RPM (since we have an unknown chunk) and we keep intermediate files (such as the XZ compressed CPIO).

Ideally the RPM should be directly decompressed and extracted to the extraction directory, without unknown chunks nor intermediate representation.

### Basic example

Download a Rocky Linux, Fedora, or RHEL ISO to have a large collection of RPM samples.

Corv0O:
RPM (Red Hat Package Manager) packages are standard binary or source files (.rpm) used for installing, updating, and managing software on Linux distributions like RHEL, CentOS, Fedora, and openSUSE. They contain compressed software, configuration files, and metadata (dependencies, version info)

# Format 
RPM files are encoded in **big-endian**

### Lead (96 bytes)

Magic : ED AB EE DB
Major, minor : (03, 04) format version
Type : 0 = binary package, 1 = source
Arch, OS : target architecture and OS identifiers
Name : 66-byte null-padded package name
Signature type : 5 for the header-based signatures used by modern RPMs

### Signature header (16 bytes)
Magic : 8E AD E8
Version : 1 byte
Reserved : 4 bytes
nindex : number of index entries (4 bytes)
hsize : size of the data section (4 bytes)

Each index entry has the shape { tag, type, offset, count }, each field is a uint32. 

### Main header
Same structure as the signature header, but carries metadata tags about the package. The one we need:

- **RPMTAG_PAYLOADCOMPRESSOR** (1125)  C string naming the payload compressor: gzip, bzip2, xz, zstd. If this tag is absent, the payload is uncompressed (ufdio).
- 
### Payload
A CPIO archive  compressed with the algorithm named by PAYLOADCOMPRESSOR. Starts right after the main header's data section and runs to EOF.

So i had to change the behavior of `CPIO.py` to be able to handle stream input, otherwise it would be inefficient to keep all the information in memory.

# Sample : 
We can get grab rpm files over the internet from the [Fedora mirror](https://src.fedoraproject.org/)

or we can build ourselves with `rpmbuild` by : 
- defining a `.spec` file containing data, and instruction to recreate files, [information here](https://rpm-packaging-guide.github.io/#working-with-spec-files)
- using this rpm command with the spec file `rpmbuild -bb --define "_topdir $PWD" --define "_binary_payload <compression>" <your-file>.spec`

`_binary_payload` values: `w9.gzdio` (gzip) · `w9.bzdio` (bzip2) · `w2.xzdio `(xz) · `w19.zstdio` (zstd) · `w0.ufdio `(uncompressed).

# Reference : 
https://rpm-packaging-guide.github.io/#what-is-an-rpm
https://refspecs.linuxbase.org/LSB_3.1.0/LSB-Core-generic/LSB-Core-generic/pkgformat.html
https://rpm.org/docs/6.0.x/man/rpm-payloadflags.7

### DIFF
--- a/docs/handlers.md
+++ b/docs/handlers.md
@@ -62,6 +62,7 @@
     | [`QNX DEFLATE`](#qnx-deflate) | COMPRESSION | :octicons-check-16: |
     | [`RAR`](#rar) | ARCHIVE | :octicons-alert-fill-12: |
     | [`ROMFS`](#romfs) | FILESYSTEM | :octicons-check-16: |
+    | [`RPM`](#rpm) | ARCHIVE | :octicons-check-16: |
     | [`SOLARIS_UFS1`](#solaris_ufs1) | FILESYSTEM | :octicons-check-16: |
     | [`SQUASHFS (V1)`](#squashfs-v1) | FILESYSTEM | :octicons-check-16: |
     | [`SQUASHFS (V2)`](#squashfs-v2) | FILESYSTEM | :octicons-check-16: |
@@ -1103,6 +1104,23 @@
 
         - [RomFS Documentation](https://www.kernel.org/doc/html/latest/filesystems/romfs.html){ target="_blank" }
         - [RomFS Wikipedia](https://en.wikipedia.org/wiki/Romfs){ target="_blank" }
+## RPM
+
+!!! success "Fully supported"
+
+    === "Description"
+
+        RPM (Red Hat Package Manager) is a package archive format used by Red Hat-based Linux distributions. An RPM file contains metadata (signature, header) and a compressed cpio archive as payload.
+
+        ---
+
+        - **Handler type:** Archive
+        - **Vendor:** Red Hat
+
+    === "References"
+
+        - [RPM File Format](https://rpm-software-management.github.io/rpm/manual/format.html){ target="_blank" }
+        - [RPM Package Manager](https://en.wikipedia.org/wiki/RPM_Package_Manager){ target="_blank" }
 ## solaris_ufs1
 
 !!! success "Fully supported"
--- a/docs/handlers.md
+++ b/docs/handlers.md
@@ -66,6 +66,7 @@
     | [`QNX DEFLATE`](#qnx-deflate) | COMPRESSION | :octicons-check-16: |
     | [`RAR`](#rar) | ARCHIVE | :octicons-alert-fill-12: |
     | [`ROMFS`](#romfs) | FILESYSTEM | :octicons-check-16: |
+    | [`RPM`](#rpm) | ARCHIVE | :octicons-check-16: |
     | [`SOLARIS_UFS1`](#solaris_ufs1) | FILESYSTEM | :octicons-check-16: |
     | [`SQUASHFS (V1)`](#squashfs-v1) | FILESYSTEM | :octicons-check-16: |
     | [`SQUASHFS (V2)`](#squashfs-v2) | FILESYSTEM | :octicons-check-16: |
@@ -1180,6 +1181,23 @@
 
         - [RomFS Documentation](https://www.kernel.org/doc/html/latest/filesystems/romfs.html){ target="_blank" }
         - [RomFS Wikipedia](https://en.wikipedia.org/wiki/Romfs){ target="_blank" }
+## RPM
+
+!!! success "Fully supported"
+
+    === "Description"
+
+        RPM (Red Hat Package Manager) is a package archive format used by Red Hat-based Linux distributions. An RPM file contains metadata (signature, header) and a compressed cpio archive as payload.
+
+        ---
+
+        - **Handler type:** Archive
+        - **Vendor:** Red Hat
+
+    === "References"
+
+        - [RPM File Format](https://rpm.org/docs/4.20.x/manual/format_v4.html){ target="_blank" }
+        - [RPM Package Manager](https://en.wikipedia.org/wiki/RPM_Package_Manager){ target="_blank" }
 ## solaris_ufs1
 
 !!! success "Fully supported"
--- a/python/unblob/handlers/archive/cpio.py
+++ b/python/unblob/handlers/archive/cpio.py
@@ -27,12 +27,14 @@ from ...models import (
     Reference,
     ValidChunk,
 )
+from ...report import ExtractionProblem
 
 logger = get_logger()
 
 CPIO_TRAILER_NAME = "TRAILER!!!"
 MAX_LINUX_PATH_LENGTH = 0x1000
 
+C_TRAILER = 0o00000
 C_ISBLK = 0o60000
 C_ISCHR = 0o20000
 C_ISDIR = 0o40000
@@ -43,6 +45,7 @@ C_ISCTG = 0o110000
 C_ISREG = 0o100000
 
 C_FILE_TYPES = (
+    C_TRAILER,
     C_ISBLK,
     C_ISCHR,
     C_ISDIR,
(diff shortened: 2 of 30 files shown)


python/unblob/handlers/archive/rpm.py
```diff
@@ -0,0 +1,239 @@
...
+        ],
+        limitations=[],
+    )
+
+    def is_valid_header(self, header) -> bool:
+        return (
+            header.signature_type == 5
+            and header.type in (0, 1)
+            and bytes(header.name).endswith(b"\x00")
+        )
```
qkaiser: All of these can be enforced by pattern matching.

```
ED AB EE DB (03 | 04) ?? ( 00 | 01 ) [67] 00
```

Anything else you can validate through pattern  such as `signature_type`, `os_num`  or `reserved` ?

python/unblob/handlers/archive/rpm.py
```diff
@@ -0,0 +1,239 @@
...
+        if size_entry is None:
+            raise InvalidInputFormat("RPM signature header missing SIZE / LONGSIZE tag")
+
+        file.seek(data_offset + size_entry.offset, io.SEEK_SET)
+        if size_is_long:
+            return self.cparser_be.uint64(file)
+        return self.cparser_be.uint32(file)
+
+    def calculate_chunk(self, file: File, start_offset: int) -> ValidChunk | None:
+        file.seek(start_offset, io.SEEK_SET)
```
qkaiser: Not needed.

python/unblob/handlers/archive/rpm.py
```diff
@@ -0,0 +1,239 @@
...
+            RPM_SIGNATURE_ALIGNMENT,
+        )
+
+        end_offset = (
+            start_offset
+            + RPM_LEAD_SIZE
+            + RPM_HEADER_INTRO_SIZE
+            + sig_body_size
+            + payload_size
+        )
```
qkaiser: How about this ?

```python
end_offset = (
    file.tell()
    + sig_body_size
    + payload_size
)
```

python/unblob/handlers/archive/rpm.py
```diff
@@ -0,0 +1,239 @@
...
+    )
+
+    def is_valid_header(self, header) -> bool:
+        return (
+            header.signature_type == 5
+            and header.type in (0, 1)
+            and bytes(header.name).endswith(b"\x00")
+        )
+
+    def read_payload_size(self, file, sig_header) -> int:
```
qkaiser: missing type hint for file

python/unblob/handlers/archive/rpm.py
```diff
@@ -0,0 +1,239 @@
...
+        } rpm_cstring_t;
+    """
+
+
+class RPMExtractor(Extractor):
+    def __init__(self):
+        self._struct_parser = StructParser(RPM_C_DEFINITION)
+
+    def extract(self, inpath: Path, outdir: Path) -> ExtractResult:
+        logger.info("RPMExtractor running", input=str(inpath))
```
qkaiser: Not needed

python/unblob/handlers/archive/rpm.py
```diff
@@ -0,0 +1,239 @@
...
+        fs = FileSystem(outdir)
+        with File.from_path(inpath) as file:
+            payload_offset, compressor_type = self.parse_headers(file)
+            file.seek(payload_offset, io.SEEK_SET)
+            decoder = self._open_payload_stream(file, compressor_type)
+            try:
+                PortableASCIIParser.parse_and_extract(decoder, fs)
+            finally:
+                if decoder is not file:
+                    decoder.close()
```
qkaiser: Did you try with a context manager ?

```python
with self._open_payload_stream(file, compressor_type) as decoder:
    PortableASCIIParser.parse_and_extract(decoder, fs)
```

python/unblob/handlers/archive/rpm.py
```diff
@@ -0,0 +1,239 @@
...
+        if compressor == "gzip":
+            return gzip.GzipFile(fileobj=file, mode="rb")
+        if compressor == "bzip2":
+            return bz2.BZ2File(file, mode="rb")  # pyright: ignore[reportArgumentType]
+        if compressor in ("xz", "lzma"):
+            return lzma.LZMAFile(file, mode="rb")  # pyright: ignore[reportArgumentType]
+        if compressor == "zstd":
+            return pyzstd.ZstdFile(file, mode="rb")  # pyright: ignore[reportArgumentType]
+        if compressor == "none":
+            return file
```
qkaiser: maybe  a `match` statement would be cleaner

python/unblob/handlers/archive/rpm.py
```diff
@@ -0,0 +1,239 @@
...
+        raise InvalidInputFormat(f"Unsupported RPM payload compressor: {compressor}")
+
+    def parse_headers(self, file) -> tuple[int, str]:
```
qkaiser: missing type hint for file

python/unblob/handlers/archive/cpio.py
```diff
@@ -122,84 +123,187 @@ class CPIOEntry:
     path: Path
 
 
+@attrs.define
```
qkaiser: We use dataclasses, not attrs

qkaiser (commented):
Please split your work in separate commits:
- 1 commit for CPIO, maybe more
- 1 commit for RPM

I see a lot of movement in CPIO. It would be easier to follow if each change was atomic and linked to a commit with a complete description. Even more when it looks like a complete refactor.

python/unblob/handlers/archive/rpm.py
Corv0O: it doesn't work with the samples, got unknown chunks

Corv0O: found something simpler 
```suggestion
        rpm_header = self.cparser_be.rpm_header_t(file)
        sig_body_start = file.tell()
        payload_size = self.read_payload_size(file, rpm_header)
        
        end_offset = sig_body_start + sig_body_size + payload_size
```

python/unblob/handlers/archive/rpm.py
Corv0O: The documentation indicate that `signature_type` shall be 5, i  already used it on the is_valid_header function but it can add to the hexstring i think.

For `os_num`, the value can change but i can't prove it by finding documentation about it, the only information i got is that Linux = 1. by using hexdump on my samples, i usually got 1 but on src package and other that i downloaded on the fedora mirror, i got 00 for `osnum` not a strong candidate i think.  

`reserved` seems to be untouched in all the samples i got, we can probably use it for the hardening.

python/unblob/handlers/archive/cpio.py
```diff
@@ -122,84 +123,187 @@ class CPIOEntry:
...
+            if entry.is_trailer:
+                discard(stream, content_padding)
+                break
 
-            padded_header_size = self._pad_header(header, c_namesize)
-            current_offset += padded_header_size
+            path = Path(entry.filename)
 
-            tmp_filename = self.file.read(c_namesize)
+            if path.name in ("", "."):
```
qkaiser: What about `".."`, should we care ?

python/unblob/handlers/archive/cpio.py
```diff
@@ -122,84 +123,187 @@ class CPIOEntry:
...
+                discard(stream, entry.c_filesize + content_padding)
+                continue
 
-            # heuristics 2: check that filename is null-byte terminated
-            if not tmp_filename.endswith(b"\x00"):
-                raise InvalidInputFormat(
-                    "CPIO entry filename is not null-byte terminated"
-                )
+            if not stat.S_ISDIR(entry.c_mode):
+                fs.unlink(path)
```
qkaiser: Why ?

python/unblob/handlers/archive/cpio.py
```diff
@@ -122,84 +123,187 @@ class CPIOEntry:
...
+
+        return CPIOEntryHeader(
+            header=header,
+            filename=filename,
+            c_filesize=c_filesize,
+            c_namesize=c_namesize,
+            c_mode=c_mode,
+            padded_header_size=padded_header_size,
+            is_trailer=False,
+        )
```
qkaiser: This can be simplified with an if/else followed by a single return statement.

python/unblob/handlers/archive/cpio.py
```diff
@@ -122,84 +123,187 @@ class CPIOEntry:
...
+
+
+def stream_chunks(stream, size: int, bufsize: int = DEFAULT_BUFSIZE):
+    remaining = size
+    while remaining > 0:
+        chunk = stream.read(min(bufsize, remaining))
+        if not chunk:
+            raise InvalidInputFormat("Unexpected EOF inside CPIO entry body")
+        remaining -= len(chunk)
+        yield chunk
```
qkaiser: There's `iterate_file`, why do you need this ?

python/unblob/handlers/archive/cpio.py
```diff
@@ -122,84 +123,187 @@ class CPIOEntry:
...
+
+
+def discard(stream, size: int, bufsize: int = DEFAULT_BUFSIZE):
```
qkaiser: What's the point of this ? Can't you just seek past what you need to discard ?

python/unblob/handlers/archive/cpio.py
Corv0O: was present before my commit in the handler, i should've kept also the comments on top of it : 
```
# There are cases where CPIO archives have duplicated entries
# We then unlink the files to overwrite them and avoid an error.
```

python/unblob/handlers/archive/rpm.py
```diff
@@ -0,0 +1,241 @@
...
+    def __init__(self, file: File, start_offset: int):
+        self._file = file
+        self._start_offset = start_offset
+        self._struct_parser = StructParser(C_DEFINITIONS)
+
+        self._lead: Any | None = None
+        self._signature: RPMParser.HeaderSection | None = None
+        self._main_header: RPMParser.HeaderSection | None = None
+        self._package_size: int | None = None
+        self.compressor: str = "none"
```
qkaiser, line 117: Valid RPM v3/v4 payloads are gzip by default, but `RPMParser` initializes `self.compressor` to "none" and only overrides it when `RPMTAG_PAYLOADCOMPRESSOR` is present. On legacy packages that omit this tag,  `extract()` will pass the still-compressed payload straight into `PortableASCIIParser`, which immediately raises InvalidInputFormat instead of extracting the archive.

python/unblob/handlers/archive/rpm.py
```diff
@@ -0,0 +1,241 @@
...
+
+class RPMExtractor(Extractor):
+    def extract(self, inpath: Path, outdir: Path) -> ExtractResult:
+        fs = FileSystem(outdir)
+        with File.from_path(inpath) as file:
+            parser = RPMParser(file, 0)
+            parser.parse()
+            file.seek(parser.payload_offset, io.SEEK_SET)
+            with RPMExtractor.open_payload_stream(file, parser.compressor) as decoder:
+                PortableASCIIParser.parse_and_extract(decoder, fs)
```
qkaiser: RPM 4.12+ uses the stripped-down 07070X CPIO variant when the package contains files larger than 4 GiB. `extract()` always dispatches the payload to `PortableASCIIParser.parse_and_extract()`, which only understands the normal 070701/070702 headers, so these otherwise valid RPMs will fail on the first entry during extraction.

See https://github.com/rpm-software-management/rpm/blob/master/lib/cpio.cc#L51

We probably want to extend the CPIO handler to support this "stripped" signature.

python/unblob/handlers/archive/cpio.py
```diff
@@ -460,6 +460,57 @@ def valid_checksum(self, header, start_offset: int) -> bool:
...
 
+class StrippedCPIOParser(CPIOParserBase):
+    """a stripped CPIO variant (magic 07070X) used in RPM 4.12+."""
+
+    _PAD_ALIGN = 4
+    _MAGIC = b"07070X"
+    _HEADER_SIZE = 14  # 6 magic + 8 fx
+
+    @classmethod
+    def extract(cls, stream, fs: FileSystem, entries):  # noqa: C901
```
qkaiser: Why did you create this function ? How is this different from `dump_entries` in `CPIOParserBase` ? Try identifying what really needs to change. From what I see the only difference is when identifying the trailer entry, which we may adjust elsewhere.

python/unblob/handlers/archive/rpm.py
```diff
@@ -156,6 +186,53 @@ def _read_cstring_entry(self, section: HeaderSection, entry: Any) -> str:
...
+        return [
+            (
+                Path(dirnames[dirindexes[i]] + file_names[i]),
+                modes[i],
+                file_sizes[i],
+                links[i] or None,
+                rdevs[i],
+            )
+            for i in range(len(file_names))
+        ]
```
qkaiser: Do we really need this here ? Is it specific to RPM or to CPIO ? Can't we collect entries in CPIO like "classic" CPIO handlers ?

python/unblob/handlers/archive/rpm.py
```diff
@@ -168,16 +245,29 @@ def end_offset(self) -> int:
...
+            stripped_entries = (
+                parser.build_stripped_entries() if parser.has_stripped_payload else None
+            )
             file.seek(parser.payload_offset, io.SEEK_SET)
             with RPMExtractor.open_payload_stream(file, parser.compressor) as decoder:
-                PortableASCIIParser.parse_and_extract(decoder, fs)
+                if stripped_entries is not None:
+                    StrippedCPIOParser.extract(decoder, fs, stripped_entries)
+                else:
+                    PortableASCIIParser.parse_and_extract(decoder, fs)
```
qkaiser: There's a responsibility issue here, you should not have to provide stripped_entries from RPM to CPIO code. Unless I'm missing something major, which may be the case :)

python/unblob/handlers/archive/cpio.py
```diff
@@ -139,11 +139,11 @@ class CPIOParserBase:
     HEADER_STRUCT: str
     _STRUCT_PARSER = StructParser(C_DEFINITIONS)
 
-    def __init__(self, file: File, start_offset: int):
+    def __init__(self, file: File, start_offset: int, entries: list | None = None):
         self.file = file
         self.start_offset = start_offset
         self.end_offset = -1
-        self.entries = []
+        self.entries = entries if entries is not None else []
```
qkaiser: Why don't you default to an empty list ? No need for `None` and conditionals here.

python/unblob/handlers/archive/rpm.py
```diff
@@ -168,16 +245,29 @@ def end_offset(self) -> int:
...
         with File.from_path(inpath) as file:
             parser = RPMParser(file, 0)
             parser.parse()
+            stripped_entries = (
+                parser.build_stripped_entries() if parser.has_stripped_payload else None
+            )
             file.seek(parser.payload_offset, io.SEEK_SET)
             with RPMExtractor.open_payload_stream(file, parser.compressor) as decoder:
-                PortableASCIIParser.parse_and_extract(decoder, fs)
+                if stripped_entries is not None:
```
qkaiser: You can do:

```python
if parser.has_stripped_payload:
    StrippedCPIOParser(decoder, 0, parser.build_stripped_entries()).dump_entries(fs)
```

python/unblob/handlers/archive/cpio.py
```diff
@@ -122,84 +122,167 @@ class CPIOEntry:
     path: Path
 
 
+@attrs.define
+class CPIOEntryHeader:
```
qkaiser: What's the difference between `CPIOEntry` and `CPIOEntryHeader` ? Why do you need both ?

python/unblob/handlers/archive/cpio.py
```diff
@@ -122,84 +122,167 @@ class CPIOEntry:
...
+            filename=filename,
+            c_filesize=c_filesize,
+            c_namesize=c_namesize,
+            c_mode=c_mode,
+            padded_header_size=padded_header_size,
+            is_trailer=is_trailer,
+        )
+
+    @classmethod
+    def parse_and_extract(cls, stream, fs: FileSystem):  # noqa: C901
```
qkaiser: This function is the functional equivalent of `parse` from line 207 to 219, and the equivalent of `dump_entries`  from line 220 to 253. Why do you need this function ? Adapt the code from `parse` to handle streaming input instead of rewriting a whole thing here.

python/unblob/handlers/archive/cpio.py
```diff
@@ -460,6 +460,64 @@ def valid_checksum(self, header, start_offset: int) -> bool:
...
+            elif stat.S_ISCHR(mode) or stat.S_ISBLK(mode) or stat.S_ISSOCK(mode):
+                fs.mknod(path, mode=mode & 0o777, device=rdev)
+                stream.seek(size, io.SEEK_CUR)
+            else:
+                logger.warning("unknown file type in stripped CPIO archive")
+                stream.seek(size, io.SEEK_CUR)
+
+            stream.seek(content_padding, io.SEEK_CUR)
+
+
```
qkaiser: I don't think you need a custom `dump_entries` function here. The generic `dump_entries` loop over `self.entries` which you populate through `__init__`.

python/unblob/handlers/archive/rpm.py
qkaiser, line 328: You should probably return a list of `CPIOEntry` or `CPIOEntryHeader` instead of a long tuple. You can enforce it using typing hints in the CPIO handler code.

python/unblob/handlers/archive/cpio.py
Corv0O: got an error from ruff, it doesn't like the use of mutable reference as a default parameter

python/unblob/handlers/archive/cpio.py
Corv0O: `CPIOEntry` is used by existing parser,  in my case i did want to minimize change to the legacy code.

 for the use of the streaming parser, it was needed to simplify data accessing by returning the `dissect.cstruct` and other information for each entry that `CPIOEntry` doesn't.

qkaiser: What if you just extend `CPIOEntry` with the information you need  instead of creating a new class ?

python/unblob/handlers/archive/cpio.py
Corv0O: the generic `dump_entries` relies on `fs.carves` and other seekable access. 
the RPM payload is a decompressor stream and can't use the same handling, it need to seek after the next entry  after each operation.

also the stripped CPIO case require a different handling in the beginning (from line 478 to 493), the rest could be unified in a function.

Corv0O:
Just resolved merge conflict on  `__init__.py ` because of the new sbfh handler :)

qkaiser:
@Corv0O any plans to revisit this ? If not I'll take over.

qkaiser:
> @Corv0O any plans to revisit this ? If not I'll take over.

We're officially taking over.

docs/handlers.md
```diff
@@ -1180,6 +1181,23 @@
...
+        RPM (Red Hat Package Manager) is a package archive format used by Red Hat-based Linux distributions. An RPM file contains metadata (signature, header) and a compressed cpio archive as payload.
+
+        ---
+
+        - **Handler type:** Archive
+        - **Vendor:** Red Hat
+
+    === "References"
+
+        - [RPM File Format](https://rpm-software-management.github.io/rpm/manual/format.html){ target="_blank" }
```
0roman: i am getting 404 on https://rpm-software-management.github.io/rpm/manual/format.html

python/unblob/handlers/archive/cpio.py
```diff
@@ -114,133 +117,150 @@
...
+            or stat.S_ISBLK(entry.mode)
+            or stat.S_ISSOCK(entry.mode)
+        ):
+            fs.mknod(entry.path, mode=entry.mode & 0o777, device=entry.rdev)
+            self.file.seek(entry.size, io.SEEK_CUR)
+        else:
+            logger.warning("unknown file type in CPIO archive")
+            self.file.seek(entry.size, io.SEEK_CUR)
+
+    def record_checksum_mismatch(
```
0roman: it's more a verification. I would name the function `verify_checksum_mismatch`

0roman (approved):

qkaiser (approved):

PR #1504, opened 2026-05-07
fix(erofs): allow build_time_nsec to be zero + end_offset calculation fix

elektrischermoench:
EROFS images produced by certain build systems set build_time_nsec to 0, which is valid per spec. The is_valid_header check incorrectly required build_time_nsec > 0, causing these images to fall through as unknown chunks rather than being extracted.

The end_offset calculation omitted start_offset, producing a wrong chunk boundary whenever EROFS is embedded at a non-zero offset in a larger image.


qkaiser:
@elektrischermoench please split your fixes in two separate commits:
- nsec 0
- offset calculation fix

qkaiser (approved):

PR #1507, opened 2026-05-11
feat(handler) add Airoha firmware handler

Corv0O:
Airoha BT firmware is the container format used by Airoha Bluetooth audio SoCs found in earbuds and headphones (Sony WH/WF families, AirReps, etc.). Wraps either as raw, LZMA-compressed, or LZMA-then-AES-encrypted firmware blob.

# Format
All fields are **little-endian**.

Airhoha firmware doesn't have a **magic** field, but we can use the fixed header with multiple constant to detect it

### Fixed header (256 bytes)
file_checksum (32 bytes) SHA256 of the rest of the file
padding (224 bytes) 0xFF
TLV chain (starts at offset 0x100)

Chain ends when type == 0xFFFF. 
The first TLV is always `BASIC_INFO`.

### TLV type used in the handler : 
`BASIC_INFO` : compression_type,  integrity_check_type, firmware_offset, firmware_size
`MOVER_INFO` : number_of_sections, then N × {source_offset, decompressed_size, dest_offset}

### Compression types
`NONE` : raw firmware
`LZMA` : .lzma stream
`LZMA_AES` :  AES-128-CBC over LZMA

# Samples
We can get samples from this repo : https://github.com/lzghzr/MDR_Proxy

Unencrypted (LZMA): [HDX-2952](github.com/lzghzr/MDR_Proxy/blob/master/firmware/2952/MDRID295200/HDX-2952_Earbuds_V1007_FOTA.18a70d16ae4e655ad71257bd0cc798e21d3b71fc.bin) / [HDX-2957 files](https://github.com/lzghzr/MDR_Proxy/blob/master/firmware/2957/MDRID295700/HDX-2957_V1007_FOTA.166253013da804a1d35c3e6bc1fc0230753dd394.bin) Non encrypted Airoha files are rare.
In the same repo, there is also encrypted file from various devices but without the key & iv, we can't extract anything

# References
https://github.com/ramikg/airoha-firmware-parser
https://insinuator.net/2025/06/airoha-bluetooth-security-vulnerabilities/ 
https://github.com/lzghzr/MDR_Proxy

### DIFF
--- a/python/unblob/handlers/archive/airoha.py
+++ b/python/unblob/handlers/archive/airoha.py
@@ -1,0 +1,184 @@
+import io
+import lzma
+from enum import IntEnum
+from pathlib import Path
+
+from unblob.file_utils import (
+    Endian,
+    File,
+    FileSystem,
+    InvalidInputFormat,
+    StructParser,
+)
+from unblob.models import (
+    Extractor,
+    ExtractResult,
+    HandlerDoc,
+    HandlerType,
+    HexString,
+    Reference,
+    StructHandler,
+    ValidChunk,
+)
+from unblob.report import ExtractionProblem
+
+C_DEFINITIONS = r"""
+    typedef struct airoha_header {
+        uint8  file_checksum[32];           /* SHA256 of the rest of the file */
+        uint8  padding[224];                /* 0xFF padding */
+        uint16 basic_info_tlv_type;         /* 0x11 BASIC_INFO */
+        uint16 basic_info_tlv_length;       /* 10 */
+        uint8  compression_type;            /* 0=NONE, 1=LZMA, 2=LZMA_AES */
+        uint8  integrity_check_type;        /* 0=CRC32, 1=SHA256, 2=SHA256_RSA */
+        uint32 firmware_offset;
+        uint32 firmware_size;
+    } airoha_header_t;
+
+    typedef struct airoha_tlv_header {
+        uint16 tlv_type;
+        uint16 tlv_length;
+    } airoha_tlv_header_t;
+
+    typedef struct airoha_mover_info_header {
+        uint32 number_of_sections;
+    } airoha_mover_info_header_t;
+
+    typedef struct airoha_section {
+        uint32 source_offset;               /* offset in decompressed stream */
+        uint32 decompressed_size;
+        uint32 dest_offset;                 /* in-memory load address */
+    } airoha_section_t;
+"""
+
+PRELUDE_SIZE = 256
+PATTERN_FF_LEN = 16
+
+
+class CompressionType(IntEnum):
+    NONE = 0
+    LZMA = 1
+    LZMA_AES = 2
+
+
+class TlvType(IntEnum):
+    MOVER_INFO = 0x12
+
+
+class AirohaExtractor(Extractor):
+    def __init__(self):
+        self._struct_parser = StructParser(C_DEFINITIONS)
+
+    def _read_sections(self, file: File) -> list:
+        tlv = self._struct_parser.parse("airoha_tlv_header_t", file, Endian.LITTLE)
+        while tlv.tlv_type != TlvType.MOVER_INFO:
+            file.seek(tlv.tlv_length, io.SEEK_CUR)
+            tlv = self._struct_parser.parse("airoha_tlv_header_t", file, Endian.LITTLE)
+
+        mover = self._struct_parser.parse(
+            "airoha_mover_info_header_t", file, Endian.LITTLE
+        )
+        return [
+            self._struct_parser.parse("airoha_section_t", file, Endian.LITTLE)
+            for _ in range(mover.number_of_sections)
+        ]
+
+    def extract(self, inpath: Path, outdir: Path) -> ExtractResult:
+        fs = FileSystem(outdir)
+        with File.from_path(inpath) as file:
+            header = self._struct_parser.parse("airoha_header_t", file, Endian.LITTLE)
+            sections = self._read_sections(file)
+            if not sections:
+                raise InvalidInputFormat("Airoha file has no MOVER_INFO sections")
+            base_address = min(section.dest_offset for section in sections)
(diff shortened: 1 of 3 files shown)


python/unblob/handlers/archive/airoha.py
```diff
@@ -0,0 +1,199 @@
...
+    NAME = "airoha"
+
+    PATTERNS = [
+        HexString(
+            """
+                FF FF FF FF FF FF FF FF
+                FF FF FF FF FF FF FF FF
+                11 00 0A 00
+                ( 00 | 01 | 02 )
+                ( 00 | 01 | 02 )
```
qkaiser: Would be nice to have comments indicating what these fields are.

python/unblob/handlers/archive/airoha.py
```diff
@@ -0,0 +1,199 @@
...
+            """
+        ),
+    ]
+    PATTERN_MATCH_OFFSET = -(PRELUDE_SIZE - PATTERN_FF_LEN)
+
+    C_DEFINITIONS = C_DEFINITIONS
+    HEADER_STRUCT = "airoha_header_t"
+    EXTRACTOR = AirohaExtractor()
+
+    DOC = HandlerDoc(
```
qkaiser, line 150:
```suggestion
    NAME = "airoha"
    PATTERNS = [
        HexString(
            """
                FF FF FF FF FF FF FF FF
                FF FF FF FF FF FF FF FF
                11 00 0A 00
                ( 00 | 01 | 02 )
                ( 00 | 01 | 02 )
            """
        ),
    ]
    PATTERN_MATCH_OFFSET = -(PRELUDE_SIZE - PATTERN_FF_LEN)
    C_DEFINITIONS = C_DEFINITIONS
    HEADER_STRUCT = "airoha_header_t"
    EXTRACTOR = AirohaExtractor()
    DOC = HandlerDoc(
```

python/unblob/handlers/archive/airoha.py
```diff
@@ -0,0 +1,199 @@
...
+                    "airoha_mover_info_header_t", file, Endian.LITTLE
+                )
+                if mover.number_of_sections > 0:
+                    section = self._struct_parser.parse(
+                        "airoha_section_t", file, Endian.LITTLE
+                    )
+                    base_address = section.dest_offset
+                break
+            file.seek(tlv.tlv_length, io.SEEK_CUR)
+        return base_address
```
qkaiser:
```suggestion
    def _find_base_address(self, file: File) -> int | None:
        tlv = self._struct_parser.parse("airoha_tlv_header_t", file, Endian.LITTLE)
        while tlv.tlv_type != TlvType.MOVER_INFO:
            file.seek(tlv.tlv_length, io.SEEK_CUR)
            tlv = self._struct_parser.parse("airoha_tlv_header_t", file, Endian.LITTLE)

        mover = self._struct_parser.parse(
            "airoha_mover_info_header_t", file, Endian.LITTLE
        )
        if mover.number_of_sections > 0:
            section = self._struct_parser.parse(
                "airoha_section_t", file, Endian.LITTLE
             )
             return int(section.dest_offset)
        return None
```

python/unblob/handlers/archive/airoha.py
```diff
@@ -0,0 +1,199 @@
...
+                    )
+                case CompressionType.NONE:
+                    fs.carve(
+                        Path(firmware_name),
+                        file,
+                        header.firmware_offset,
+                        header.firmware_size,
+                    )
+
+        return ExtractResult(reports=fs.problems)
```
qkaiser, line 130:
```suggestion
        return ExtractResult(reports=fs.problems)
```

python/unblob/handlers/archive/airoha.py
```diff
@@ -0,0 +1,199 @@
...
+                            )
+                        ),
+                    )
+                case CompressionType.NONE:
+                    fs.carve(
+                        Path(firmware_name),
+                        file,
+                        header.firmware_offset,
+                        header.firmware_size,
+                    )
```
qkaiser, line 128: You're dumping the firmware "as-is", but what is needed is a reconstruction of the file as if it was loaded in memory. Similar to what you did for Tesla Wall connector.

Your own description states that:
> MOVER_INFO : number_of_sections, then N × {source_offset, decompressed_size, dest_offset}

So you should create a file that contains each section. The section with the lowest dest_offset goes first (this is your offset 0), and subsequent sections are placed at dest_offset - load_address in the file.

python/unblob/handlers/archive/airoha.py
```diff
@@ -0,0 +1,204 @@
...
+
+        for chunk in iterate_file(file, header.firmware_offset, header.firmware_size):
+            data = decompressor.decompress(chunk)
+            keep_from = max(0, sec.source_offset - pos)
+            keep_until = min(len(data), end - pos)
+            if keep_until > keep_from:
+                yield data[keep_from:keep_until]
+            pos += len(data)
+            if pos >= end:
+                return
```
qkaiser: `sec` has no meaning. Should be `section`.

python/unblob/handlers/archive/airoha.py
```diff
@@ -0,0 +1,204 @@
...
+            if not sections:
+                raise InvalidInputFormat("Airoha file has no MOVER_INFO sections")
+            base_address = min(sec.dest_offset for sec in sections)
+            firmware_name = Path(f"firmware_{base_address:08x}.bin")
+
+            with fs.open(firmware_name, "wb+") as outfile:
+                for sec in sections:
+                    outfile.seek(sec.dest_offset - base_address, io.SEEK_SET)
+                    chunks = (
+                        self._iter_decompress_section(file, header, sec)
```
qkaiser: You're decompressing the whole file for every section. It would be easier to decompress once and yield decompressed section content in order. Note: no need for temp files.

qkaiser (approved):

PR #1518, opened 2026-05-22
feat(handler) add LVM2 support

Corv0O:
LVM2 (Logical Volume Manager 2) is Linux's volume management layer for grouping physical volumes (PVs) into volume groups (VGs) that expose logical volumes (LVs) as resizable virtual block devices. Each PV pairs a text-format metadata area (describing the VG and its LVs) with a data area holding LV contents as fixed-size physical extents.

all the **useful information** can be found [here](https://github.com/libyal/libvslvm/blob/main/documentation/Logical%20Volume%20Manager%20(LVM)%20format.asciidoc), thanks to  Joachim Metz who wrote this wonderful documentation. :)

## Format

All binary fields are **little-endian**.

Each Physical volume (PV) self-describes through structures placed at the start of the device.

#### Label header (sector 1, 32 bytes)

- signature (8 bytes): `LABELONE`
- sector (8 bytes): sector where this label lives (usually 1)
- crc, header_size (4 + 4 bytes)
- type (8 bytes): `LVM2 001`

#### PV header (immediately after label header)

- uuid (32 bytes ASCII, dash-less; the text metadata stores it dash-separated)
- device_size (8 bytes)
- Variable-length list of data area descriptors, terminated by an all-zero descriptor
- Variable-length list of metadata area descriptors, same termination

#### Metadata area header

Found at the offset given by the first metadata area descriptor. Carries four raw location descriptors (`locns[0..3]`), points to the text metadata.

#### Text metadata

Describes the VG in ASCII text, its `physical_volumes` (each entry carries the same UUID as the PV header, dash-separated), and its `logical_volumes`. Each LV is one or more **segments**, where a segment is a mapping `{ start_extent, extent_count, type, stripe_count, stripes = [pv_name, pe_index, ...] }`.

#### Data area

Physical extents (size = `extent_size * SECTOR_SIZE`) holding LV contents. Each linear segment maps `[start_extent, extent_count)` of an LV to `[pe_index, pe_index + extent_count)` on the named PV.

## Handler ability

- **Single-PV VGs** are extracted as complete logical volumes.
- **Multi-PV VGs** produce one partial LV image per PV chunk. The data is preserved across all PV extractions, but combining the partials into a single LV image is left to the user for now.
- **Only linear segments** (striped) are supported, because they're the only shape where the LV's bytes map to a single contiguous range on a single PV.  Other shapes either: split data across multiple PVs and would need cross-chunk reassembly.

### Samples

Generated locally with `losetup` + LVM2 tools, the steps are also described in the documentation mentioned before.

## References 

- https://github.com/libyal/libvslvm/blob/main/documentation/Logical%20Volume%20Manager%20(LVM)%20format.asciidoc
- https://sourceware.org/git/?p=lvm2.git

### DIFF
--- a/python/unblob/handlers/filesystem/lvm.py
+++ b/python/unblob/handlers/filesystem/lvm.py
@@ -1,0 +1,291 @@
+import io
+from pathlib import Path
+
+from unblob.file_utils import (
+    Endian,
+    FileSystem,
+    InvalidInputFormat,
+    StructParser,
+    iterate_file,
+)
+from unblob.models import (
+    Extractor,
+    ExtractResult,
+    File,
+    HandlerDoc,
+    HandlerType,
+    HexString,
+    Reference,
+    StructHandler,
+    ValidChunk,
+)
+from unblob.report import ExtractionProblem
+
+C_DEFINITION = r"""
+        typedef struct lvm_label_header {
+            char     signature[8];   // "LABELONE"
+            uint64   sector;         // sector number of this label header, usually 1
+            uint32   crc;            // CRC of fields below to end of sector
+            uint32   header_size;    // size of this header; pv_header follows immediately
+            char     type[8];        // "LVM2 001"
+        } lvm_label_header_t;
+
+        typedef struct lvm_pv_header {
+            char     uuid[32];       // PV UUID, ASCII
+            uint64   device_size;    // PV size in bytes
+        } lvm_pv_header_t;
+
+        typedef struct lvm_data_area_descriptor {
+            uint64   area_offset;    // relative to start of the PV
+            uint64   area_size;      // 0 = unbounded
+        } lvm_data_area_descriptor_t;
+
+        typedef struct lvm_raw_location_descriptor {
+            uint64   data_offset;    // relative to start of metadata area
+            uint64   data_size;
+            uint32   crc;
+            uint32   flags;          // 0x1 = ignored
+        } lvm_raw_location_descriptor_t;
+
+        typedef struct lvm_metadata_area_header {
+            uint32                          crc;
+            char                            magic[16];   // "\x20LVM2\x20x[5A%r0N*>" signature
+            uint32                          version;
+            uint64                          offset;      // metadata area offset from PV start
+            uint64                          size;        // metadata area size
+            lvm_raw_location_descriptor_t   locns[4];
+            char                            padding[376];
+        } lvm_metadata_area_header_t;
+
+    """
+
+SECTOR_SIZE = 512  # LVM2 format constant
+LABEL_HEADER_SIZE = 32
+
+
+def parse_lvm_metadata(text: str) -> dict:
+    """Parse LVM2 text metadata into nested dicts.
+
+    Grammar (per libvslvm §5.5):
+        section { ... }                section opens a named scope
+        key = value                    int, "string", or [list]
+        # ...                          comment to end of line
+    Lists may span multiple lines until ']'.
+    """
+    root: dict = {}
+    stack: list[dict] = [root]
+    lines = _clean(text)
+
+    for line in lines:
+        if line == "}":
+            stack.pop()
+        elif line.endswith("{"):
+            section: dict = {}
+            stack[-1][line[:-1].strip()] = section
+            stack.append(section)
+        elif "=" in line:
+            key, _, value = (s.strip() for s in line.partition("="))
+            if value.startswith("[") and "]" not in value:
+                value = _consume_list(value, lines)
+            stack[-1][key] = _parse_value(value)
+
+    return root
+
+
+def _clean(text: str):
+    """Yield non-empty, comment-stripped lines."""
+    for raw in text.splitlines():
+        line = raw.split("#", 1)[0].strip()
+        if line:
+            yield line
+
+
+def _consume_list(first: str, lines) -> str:
+    """Join continuation lines until ']' is seen, return the full list value as one string."""
+    parts = [first]
+    for line in lines:
+        parts.append(line)
+        if "]" in line:
+            break
+    return " ".join(parts)
+
+
+def _parse_value(text: str) -> int | str | list:
+    if text.startswith('"') and text.endswith('"'):
+        return text[1:-1]
+    if text.startswith("[") and text.endswith("]"):
+        items = [chunk.strip() for chunk in text[1:-1].split(",")]
+        return [_parse_value(item) for item in items if item]
+    if text.lstrip("-").isdigit():
+        return int(text)
+    return text
+
+
+class LVM2Extractor(Extractor):
+    def __init__(self):
+        self._struct_parser = StructParser(C_DEFINITION)
+
+    def extract(self, inpath: Path, outdir: Path) -> ExtractResult:
+        fs = FileSystem(outdir)
+        with File.from_path(inpath) as file:
+            file.seek(SECTOR_SIZE, io.SEEK_SET)
+            label = self._struct_parser.parse("lvm_label_header_t", file, Endian.LITTLE)
+
+            file.seek(SECTOR_SIZE + label.header_size, io.SEEK_SET)
+            pv_header = self._struct_parser.parse(
+                "lvm_pv_header_t", file, Endian.LITTLE
+            )
+
+            data_areas = self._read_descriptors(file)
+            metadata_areas = self._read_descriptors(file)
+            mda_offset = metadata_areas[0].area_offset
+
+            file.seek(mda_offset, io.SEEK_SET)
+            mda = self._struct_parser.parse(
+                "lvm_metadata_area_header_t", file, Endian.LITTLE
+            )
+
+            locn = mda.locns[0]
+            file.seek(mda_offset + locn.data_offset, io.SEEK_SET)
+            text = file.read(locn.data_size).decode("utf-8")
+            metadata = parse_lvm_metadata(text)
+
+            vg = self._get_vg(metadata)
+            pv_name = self._get_pv_name(vg, pv_header.uuid.decode("ascii"))
+            extent_bytes = vg["extent_size"] * SECTOR_SIZE
+            pe_start = data_areas[0].area_offset
+
+            for lv_name, lv in vg.get("logical_volumes", {}).items():
+                self._extract_lv(file, fs, lv_name, lv, pv_name, pe_start, extent_bytes)
+
+        return ExtractResult(reports=fs.problems)
+
+    @staticmethod
+    def _get_vg(metadata: dict) -> dict:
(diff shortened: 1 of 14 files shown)


python/unblob/handlers/filesystem/lvm.py
```diff
@@ -0,0 +1,291 @@
...
+    def calculate_chunk(self, file: File, start_offset: int) -> ValidChunk | None:
+        header = self.parse_header(file, Endian.LITTLE)
+
+        if not self.is_valid_header(header, start_offset):
+            raise InvalidInputFormat("Invalid LVM label header.")
+
+        pv_start = start_offset - header.sector * SECTOR_SIZE
+
+        file.seek(start_offset + header.header_size, io.SEEK_SET)
+        pv_header = self.cparser_le.lvm_pv_header_t(file)
```
qkaiser, line 283: if header size is `header.header_size`, then you don't need to seek since you're just past it after your call to `parse_header`.


```suggestion
        pv_header = self.cparser_le.lvm_pv_header_t(file)
```

qkaiser, line 283:
```suggestion
        file.seek(start_offset + header.header_size, io.SEEK_SET)
        pv_header = self.cparser.parse("lvm_pv_header_t", file, Endian.LITTLE)
```

python/unblob/handlers/filesystem/lvm.py
```diff
@@ -0,0 +1,291 @@
...
+        ],
+        limitations=[
+            "Multi-PV volume groups produce one partial LV image per PV chunk. The data is preserved across all extractions, but combining the partials into a single LV image is left to the user.",
+            "Only linear segments (striped with stripe_count=1) are supported. Other segment types (multi-stripe, mirror, raid, thin, snapshot, cache) require cross-PV reassembly or a separate format parser",
+        ],
+    )
+
+    def is_valid_header(self, header, start_offset: int) -> bool:
+        return (
+            LABEL_HEADER_SIZE <= header.header_size <= SECTOR_SIZE
```
qkaiser, line 270: Shouldn't header_size be exactly 32 bytes ? You can check using dissect.cstruct size field itself:

```suggestion
            header.header_size == header.size and header.header_size == LABEL_HEADER_SIZE
```

python/unblob/handlers/filesystem/lvm.py
```diff
@@ -0,0 +1,291 @@
...
+            and header.sector * SECTOR_SIZE <= start_offset
```
qkaiser, line 271: What are you really testing here ? Shouldn't you test that sector is non null ?

python/unblob/handlers/filesystem/lvm.py
```diff
@@ -0,0 +1,291 @@
...
+    StructHandler,
+    ValidChunk,
+)
+from unblob.report import ExtractionProblem
+
+C_DEFINITION = r"""
+        typedef struct lvm_label_header {
+            char     signature[8];   // "LABELONE"
+            uint64   sector;         // sector number of this label header, usually 1
+            uint32   crc;            // CRC of fields below to end of sector
```
qkaiser, line 28: Is this something we can/should check ?

python/unblob/handlers/filesystem/lvm.py
```diff
@@ -0,0 +1,291 @@
...
+    def __init__(self):
+        self._struct_parser = StructParser(C_DEFINITION)
+
+    def extract(self, inpath: Path, outdir: Path) -> ExtractResult:
+        fs = FileSystem(outdir)
+        with File.from_path(inpath) as file:
+            file.seek(SECTOR_SIZE, io.SEEK_SET)
+            label = self._struct_parser.parse("lvm_label_header_t", file, Endian.LITTLE)
+
+            file.seek(SECTOR_SIZE + label.header_size, io.SEEK_SET)
```
qkaiser, line 134: No need to seek, you're just past it after parsing the header.


```suggestion
```

python/unblob/handlers/filesystem/lvm.py
```diff
@@ -0,0 +1,291 @@
...
+
+    def _read_descriptors(self, file: File) -> list:
+        descs = []
+        while True:
+            d = self._struct_parser.parse(
+                "lvm_data_area_descriptor_t", file, Endian.LITTLE
+            )
+            if d.area_offset == 0 and d.area_size == 0:
+                return descs
+            descs.append(d)
```
qkaiser, line 233:
```suggestion
    def _read_descriptors(self, file: File) -> list:
        descriptors = []
        descriptor = self._struct_parser.parse(
            "lvm_data_area_descriptor_t", file, Endian.LITTLE
        )
        while descriptor.area_offset != 0 and descriptor.area_size != 0::
            descriptors.append(descriptor)
            descriptor = self._struct_parser.parse(
                "lvm_data_area_descriptor_t", file, Endian.LITTLE
            )
        return descs
```

python/unblob/handlers/filesystem/lvm.py
```diff
@@ -0,0 +1,291 @@
...
+            label = self._struct_parser.parse("lvm_label_header_t", file, Endian.LITTLE)
+
+            file.seek(SECTOR_SIZE + label.header_size, io.SEEK_SET)
+            pv_header = self._struct_parser.parse(
+                "lvm_pv_header_t", file, Endian.LITTLE
+            )
+
+            data_areas = self._read_descriptors(file)
+            metadata_areas = self._read_descriptors(file)
+            mda_offset = metadata_areas[0].area_offset
```
qkaiser, line 141: Can raise an IndexError if `metadata_areas` is empty.

python/unblob/handlers/filesystem/lvm.py
```diff
@@ -0,0 +1,291 @@
...
+
+            file.seek(mda_offset, io.SEEK_SET)
+            mda = self._struct_parser.parse(
```
qkaiser, line 144: What does mda stand for ?

python/unblob/handlers/filesystem/lvm.py
```diff
@@ -0,0 +1,291 @@
...
+                "lvm_metadata_area_header_t", file, Endian.LITTLE
+            )
+
+            locn = mda.locns[0]
```
qkaiser, line 148: You don't know if locns is empty or not.

python/unblob/handlers/filesystem/lvm.py
```diff
@@ -0,0 +1,291 @@
...
+            file.seek(mda_offset + locn.data_offset, io.SEEK_SET)
+            text = file.read(locn.data_size).decode("utf-8")
```
qkaiser, line 150: How big can `data_size` be ?

python/unblob/handlers/filesystem/lvm.py
```diff
@@ -0,0 +1,291 @@
...
+            metadata = parse_lvm_metadata(text)
+
+            vg = self._get_vg(metadata)
+            pv_name = self._get_pv_name(vg, pv_header.uuid.decode("ascii"))
```
qkaiser, line 154:
```suggestion
            pv_name = self._get_pv_name(vg, pv_header.uuid.decode("ascii", errors="replace"))
```

python/unblob/handlers/filesystem/lvm.py
```diff
@@ -0,0 +1,291 @@
...
+            extent_bytes = vg["extent_size"] * SECTOR_SIZE
+            pe_start = data_areas[0].area_offset
```
qkaiser, line 156: You don't know if `data_areas` is empty or not.

python/unblob/handlers/filesystem/lvm.py
```diff
@@ -0,0 +1,291 @@
...
+    @staticmethod
+    def _get_vg(metadata: dict) -> dict:
+        """Locate the VG block — the only top-level value that is a dict."""
+        for body in metadata.values():
+            if isinstance(body, dict):
+                return body
+        raise InvalidInputFormat("LVM metadata has no volume group block.")
+
+    @staticmethod
+    def _get_pv_name(vg: dict, pv_uuid: str) -> str:
```
qkaiser, line 172: I would use python's uuid library to encode this value instead of a `str`. Makes comparison easier.

python/unblob/handlers/filesystem/lvm.py
```diff
@@ -0,0 +1,291 @@
...
+        fs: FileSystem,
+        lv_name: str,
+        lv: dict,
+        pv_name: str,
+        pe_start: int,
+        extent_bytes: int,
+    ):
+        out_path = Path(f"{lv_name}.img")
+        with fs.open(out_path, "wb+") as outfile:
+            for key, seg in lv.items():
```
qkaiser, line 191:
```suggestion
            for key, segment in lv.items():
```

python/unblob/handlers/filesystem/lvm.py
```diff
@@ -0,0 +1,291 @@
...
+
+def _parse_value(text: str) -> int | str | list:
+    if text.startswith('"') and text.endswith('"'):
+        return text[1:-1]
+    if text.startswith("[") and text.endswith("]"):
+        items = [chunk.strip() for chunk in text[1:-1].split(",")]
+        return [_parse_value(item) for item in items if item]
+    if text.lstrip("-").isdigit():
+        return int(text)
+    return text
```
qkaiser, line 121: I would expect a lot of unit testing to exercise these functions. These are not simple and represent a complex grammar.

qkaiser (commented):
First pass.

python/unblob/handlers/filesystem/lvm.py
Corv0O: metadata_areas, thought it would be explicit  since line 141 have the same kind of naming, and the mda = lvm_metadata_area parsed 

i'll probably rename it

python/unblob/handlers/filesystem/lvm.py
Corv0O: it's seems the UUID is not well formated for the uuid library, LVM2 UUIDs aren't 128 bits RFC 4122 UUID, they are 192 bits identifier. 
source : 
https://unix.stackexchange.com/questions/173722/what-is-the-uuid-format-used-by-lvm

i had a look at [pvdissect](https://wiki.syslinux.org/wiki/index.php?title=Development/LVM_support/pvdissect)
it's handling in a better way than i do,  i'll change it to format the UUID once

qkaiser:
@Corv0O any plans to revisit this ? If not I'll take over.

qkaiser:
> @Corv0O any plans to revisit this ? If not I'll take over.

We're officially taking over.

qkaiser:
Haven't heard from @Corv0O, closing and re-opening elsewhere.

PR #1519, opened 2026-05-24
extfs: handle output paths containing quotes

Gares95:
The `extfs` handler extracts ext2/3/4 images by shelling out to `debugfs -R 'rdump / "{outdir}"' {inpath}`. `debugfs` parses the `-R` request through **libss**, which re-tokenises the string (whitespace splits arguments;
double quotes group whitespace; a literal `"` must be written as `""`). The output directory is interpolated into that quoted request **without escaping**.

During recursive extraction, `{outdir}` includes directory/file names produced by a previous extraction step, which can legally contain a `"`. When it does, the rendered request has an unbalanced quote, so `debugfs` prints:

```
debugfs: Unbalanced quotes in command line
```

…and exits 0 without running `rdump`. Because unblob's `Command` treats a non-zero exit code as the failure signal, this means the nested ext filesystem is not extracted and no extraction error is surfaced for that handler.

This is specific to `extfs.py`: it is the only handler that embeds `{outdir}` inside a tool's quoted sub-language; other handlers pass `{outdir}` as a plain argv token.

## Reproduce

```console
# a tiny populated ext2 image
mkdir d && echo marker > d/marker.txt
dd if=/dev/zero of=nested.ext bs=1M count=2
mke2fs -F -q -d d nested.ext

# place it under a directory whose name contains a double quote, in a tar
mkdir -p 'evil"dir' && cp nested.ext 'evil"dir/nested.ext'
tar -cf bad.tar 'evil"dir'

unblob -e out bad.tar
# -> out/bad.tar_extract/evil"dir/nested.ext is identified as extfs,
#    but out/.../nested.ext_extract is not created.
# Control: the same image under a name without a quote extracts normally.
```

## Fix

Escape the output path for libss before it is placed in the request (a literal `"` becomes `""` inside a double-quoted token). Implemented as a small `Command` subclass so the change is contained to the `extfs` handler:

```python
class ExtFSExtractor(Command):
    """debugfs -R extractor that is safe for output paths with metacharacters."""

    def _make_extract_command(self, inpath: Path, outdir: Path):
        escaped_outdir = Path(str(outdir).replace('"', '""'))
        return super()._make_extract_command(inpath, escaped_outdir)

...
    EXTRACTOR = ExtFSExtractor("debugfs", "-R", 'rdump / "{outdir}"', "{inpath}")
```

(An alternative — running `debugfs` with `cwd=outdir` and a request of `rdump / .`, removing the path from the request entirely — would also work but needs a `cwd` option on `Command`; happy to take that route if preferred.)

## Tests

Added behaviour-pinning unit tests in `tests/handlers/filesystem/test_extfs.py`:
- the handler uses the escaping extractor;
- an `outdir` containing `"` renders a balanced request that round-trips
  (`evil"dir` → `evil""dir` inside the quotes);
- a plain `outdir` is unchanged.

## Verification done locally

- New unit tests: 3 passed.
- End-to-end A/B in a container (e2fsprogs + unblob built from source): the `evil"dir` case now extracts `marker.txt`; the safe-named control is unchanged.
- `ruff check` / `ruff format --check` (pinned `ruff==0.15.13`): clean on both changed files.

### DIFF
--- a/python/unblob/handlers/filesystem/extfs.py
+++ b/python/unblob/handlers/filesystem/extfs.py
@@ -1,3 +1,5 @@
+from pathlib import Path
+
 from structlog import get_logger
 
 from unblob.file_utils import InvalidInputFormat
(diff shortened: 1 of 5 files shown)


qkaiser:
Thanks for the contribution. I placed the reproduction test case in extfs integration tests.

qkaiser (approved):

PR #1522, opened 2026-05-26
reject qnx_deflate block offsets smaller than the header

metsw24-max:
All QNX deflate block sizes are computed as `cmphdr.next - 8`, but `next` is an unchecked uint16. A value below 8 makes the size negative, and `mmap.read()` then slurps the whole remaining file as a single block instead of stopping at the block boundary. Reject `next < 8` in the chunk walk and the extractor.


qkaiser (approved):
Good catch !

qkaiser:
@metsw24-max did you use an LLM or coding agent to find and fix the bug ? If so, can you share details about the models, harnesses, and prompts ? Were you looking into unblob as a whole or into specific handlers ?

I'm asking cause the unit tests looks like what Codex would generate :)

metsw24-max:
@qkaiser 
I used an codex to assist with drafting the changes, but I reviewed and verified everything manually before submitting.

PR #1523, opened 2026-05-27
reject btrfs encoded_write data length underflow

metsw24-max:
replay_encoded_write() derives the DATA length as `cmd_header.data_len - len(command) - len(tlv_type)`. `data_len` is an untrusted uint32 and the command carries a variable-length path TLV, so a small value makes the result negative and `File.read()` (mmap) returns the whole rest of the file straight into the zlib/zstd/LZO decompressor instead of the declared block. The sibling `replay_write()` runs the same subtraction through `iterate_file()`, which skips negatives, so only the encoded-write path over-reads. Reject the underflow before the read.


qkaiser (approved):

qkaiser:
Nice catch. Keep'em coming :)

PR #1524, opened 2026-05-29
reject negative cpio entry name size

metsw24-max:
c_namesize on cpio.py:149 is decoded with decode_int, which accepts a leading minus, so a crafted entry like b"-00001" turns negative and slips past the >MAX and ==0 checks. self.file.read(c_namesize) at line 161 then returns the whole rest of the mmap instead of the filename, over-reading past the entry and desyncing the parser, so reject non-positive sizes alongside the existing bound.

### DIFF
--- a/python/unblob/handlers/archive/cpio.py
+++ b/python/unblob/handlers/archive/cpio.py
@@ -155,6 +155,9 @@ class CPIOParserBase:
             if c_namesize <= 0:
                 raise InvalidInputFormat("CPIO entry filename size is invalid.")
 
+            if c_filesize < 0:
+                raise InvalidInputFormat("CPIO entry file size is invalid.")
+
             padded_header_size = self._pad_header(header, c_namesize)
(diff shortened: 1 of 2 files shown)


qkaiser:
@metsw24-max we should probably check `c_file_size` too

metsw24-max:
Good call, c_filesize goes through the same decode_int so a leading minus sneaks past too and would feed a negative length into the content padding and the carve. Added a < 0 guard for it (0 is still valid for empty files and the trailer) plus a test.

qkaiser (changes requested):
Looks good. Thanks for adding filesize too. Please reword your commits so they follow conventional commits structure.

Something like:

```
fix(handlers): reject negative entry name size in cpio
```

metsw24-max:
Done, reworded both to fix(handlers): ... so they follow conventional commits.

qkaiser (approved):

PR #1525, opened 2026-05-31
reject sparse chunk sizes smaller than the chunk header

metsw24-max:
calculate_chunk in android/sparse.py skips each chunk with file.seek(chunk_header.total_sz - len(chunk_header), io.SEEK_CUR), but total_sz is an attacker-controlled uint32 with no lower bound, so a value below the 12-byte chunk header underflows to a negative SEEK_CUR and walks the parser backwards over the mmap. Reject total_sz smaller than the chunk header before the seek.

### DIFF
--- a/python/unblob/handlers/filesystem/android/sparse.py
+++ b/python/unblob/handlers/filesystem/android/sparse.py
@@ -4,7 +4,7 @@ from structlog import get_logger
 
 from unblob.extractors.command import Command
 
-from ....file_utils import Endian
+from ....file_utils import Endian, InvalidInputFormat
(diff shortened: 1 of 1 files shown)


qkaiser (changes requested):
Looks good. Please follow conventional commits convention with commit titles like:

```
fix(handlers): reject chunk sizes smaller than the chunk header in android sparse
```

metsw24-max:
Done, retitled the commit to fix(handlers): reject chunk sizes smaller than the chunk header in android sparse.

qkaiser (approved):

PR #1526, opened 2026-05-31
reject shrs file_size_no_padding larger than file_size

metsw24-max:
Noticed is_valid_header rewinds with file.seek(-header.file_size_no_padding, io.SEEK_CUR) after reading that many bytes for the digest, but only file_size is range-checked. file_size_no_padding is an unvalidated uint32 from the header, so a value above file_size reads past the firmware and the negative SEEK_CUR walks the mmap-backed File position back before the header. Reject file_size_no_padding > file_size before the read and seek.


qkaiser:
Fair point, and it's addressed in https://github.com/onekey-sec/unblob/pull/1465/changes/ef36e70fdd888eb56c5f7c5ac13ff97ce3a570b4 as part of https://github.com/onekey-sec/unblob/pull/1465

PR #1527, opened 2026-06-01
reject par2 packet length smaller than the header

metsw24-max:
`is_valid_header` reads each packet body as `f.read(header.packet_length - len(header) + offset_to_recovery_id)`, where `packet_length` is an attacker-controlled uint64, `len(header)` is 64 and `offset_to_recovery_id` is 32. Only the magic is checked. A `packet_length` of 31 makes the read length -1, and `read(-1)` slurps the rest of the file past the packet boundary into the md5 (the controlled `md5_hash` then lets it validate); smaller values raise `ValueError`. Reject `packet_length < len(header)` before the read.

### DIFF
--- a/python/unblob/handlers/archive/par2.py
+++ b/python/unblob/handlers/archive/par2.py
@@ -53,6 +53,9 @@ class MultiVolumePAR2Handler(DirectoryHandler):
                 if header.magic != PAR2_MAGIC:
                     return False
 
+                if header.packet_length < len(header):
+                    return False
+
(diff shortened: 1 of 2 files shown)


qkaiser:
Handled in https://github.com/onekey-sec/unblob/pull/1465/changes/6ce633ea15a0d6f99a932b19f74245abac56df86 as part of https://github.com/onekey-sec/unblob/pull/1465/ where a negative `packet_content_length` would lead `iterate_file` to return immediately.

PR #1529, opened 2026-06-04
fix(handlers): reject romfs inode size past end of file

metsw24-max:
**Unbounded inode size in romfs content**
The inode size is read straight from the file header and never checked against the image, so content() reads from the inode data start all the way to EOF, pulling neighbouring inodes into the extracted file, and the following seek(-size) underflows to a negative offset and raises SeekError. Added a bound rejecting any inode whose data runs past the end of the file.

### DIFF
--- a/python/unblob/handlers/filesystem/romfs.py
+++ b/python/unblob/handlers/filesystem/romfs.py
@@ -118,6 +118,8 @@ class FileHeader:
     @property
     def content(self) -> bytes:
         """Returns the file content. Applicable to files and symlinks."""
+        if self.start_offset + self.size > self.file.size():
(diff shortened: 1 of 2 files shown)


qkaiser (approved):

qkaiser:
Having an integration test that exercise this path would be nice.

metsw24-max:
added a couple of tests in test_romfs.py that build an inode whose size runs past EOF and assert content raises, plus the within-bounds case. the integration fixtures are golden-output based so they don't really fit an error path, but happy to add a crafted image there too if you'd prefer.

qkaiser (approved):

PR #1531, opened 2026-06-07
bound yaffs file chunk reads to the page size

metsw24-max:
**Unbounded chunk read in get_file_chunks**

A chunk's `byte_count` is taken from the YAFFS spare tags (a full uint32 on YAFFS2, a 10-bit field on YAFFS1) and never bounded against the page size, so a crafted image can make the slice run from the chunk start to EOF and pull neighbouring chunks into the extracted file. Capped it to one page, which is the most a single chunk can hold; valid images are unaffected.


qkaiser (approved):
I moved your comment to the commit description and rewrote it to follow conventional commit. Also made the test into a single function that tests 3 scenarios: below page size, page size, and above page size.

PR #1536, opened 2026-06-08
fix(handlers): bound xiaomi hdr crc to the image data

metsw24-max:
**CRC range over-read in `_is_crc_valid`**

The checksum size is `end_offset - start_offset + CRC_CONTENT_OFFSET`, but the content starts `CRC_CONTENT_OFFSET` bytes in, so that term should be subtracted rather than added. As written the CRC runs 24 bytes past `end_offset` into whatever follows the image. A standalone file hides this since the extra bytes are past EOF and dropped, but an HDR1/HDR2 image embedded in a larger blob gets checksummed over adjacent trailing bytes and is wrongly rejected, and the read also crosses the validated chunk boundary. Flipped the sign so the CRC covers the image only; both handlers share the helper.

### DIFF
--- a/python/unblob/handlers/archive/xiaomi/hdr.py
+++ b/python/unblob/handlers/archive/xiaomi/hdr.py
@@ -154,7 +154,7 @@ class HDRHandlerBase(StructHandler):
         computed_crc = calculate_crc(
             file,
             start_offset=start_offset + CRC_CONTENT_OFFSET,
-            size=end_offset - start_offset + CRC_CONTENT_OFFSET,
+            size=end_offset - start_offset - CRC_CONTENT_OFFSET,
         )
         return header.crc32 == computed_crc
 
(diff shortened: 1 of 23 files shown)


qkaiser:
Excellent catch. CRC validation worked because our test samples are not padded with data at the end. If it was the case we would have caught this error earlier. 

Did you notice it using real world Xiaomi samples or was it code review only ?

metsw24-max:
Code review only, no real-world sample on hand. I noticed it reading `_is_crc_valid`: `calculate_crc` starts `CRC_CONTENT_OFFSET` into the chunk, so adding that term back overshoots the image by twice the offset instead of trimming it. To confirm it wasn't just a cosmetic off-by-one I built a small synthetic HDR1 and padded it with trailing bytes, which reproduces the wrongful rejection before the fix and validates after. That's the embedded case the new test covers.

qkaiser (approved):
Dropped the unit test and added integration test exercising the CRC bug.

PR #1537, opened 2026-06-09
fix(handlers): correct first yaffs chunk data offset

metsw24-max:
**Off-by-one in the first yaffs chunk offset**

`iterate_over_file` captures `start_offset` before the first page and spare are read, so `parse` derives a data offset of `-(page_size + spare_size)` for the opening chunk while later chunks are correct; a crafted image whose first chunk is a file data chunk then makes `get_file_chunks` slice from the end of the image and pull unrelated trailing bytes into the extracted file. Reading the offset after the page and spare keeps the first chunk consistent with the rest, and valid images are unaffected since their first chunk is an object header whose offset is never used.

### DIFF
--- a/python/unblob/handlers/filesystem/yaffs.py
+++ b/python/unblob/handlers/filesystem/yaffs.py
@@ -232,9 +232,9 @@ def iterate_over_file(
 
     while len(page) == config.page_size and len(spare) == config.spare_size:
         yield (start_offset, page, spare)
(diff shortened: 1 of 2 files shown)


qkaiser (approved):
Bug is there, but I followed another approach: getting the right offset within the inner loop at `iterate_over_file` and getting rid of the substraction.

PR #1539, opened 2026-06-13
fix(handlers): bound moxa frm file read to its section

metsw24-max:
**Unbounded file read in moxa FRM extraction**

`_extract_fs_section` reads each entry with `file.read(entry.file_length)`, and `file_length` is taken straight from the file table with no check against the section, so a crafted oversized value reads past the filesystem section into the following section (the firmware binary) or to EOF and writes those bytes into the extracted file. Bounded the read to the section length, the same boundary already trusted when carving the `FW_BINARY` section, so valid images stay unaffected.

### DIFF
--- a/python/unblob/handlers/archive/moxa/frm.py
+++ b/python/unblob/handlers/archive/moxa/frm.py
@@ -91,14 +91,17 @@ class MoxaFRMExtractor(Extractor):
             case SectionTypes.FW_BINARY:
                 fs.carve(Path("firmware.bin"), file, section.offset, section.length)
             case SectionTypes.FILESYSTEM:
-                self._extract_fs_section(file, fs, section.offset)
+                self._extract_fs_section(file, fs, section.offset, section.length)
             case _:
                 raise InvalidInputFormat(f"Unknown section type: {section.type}")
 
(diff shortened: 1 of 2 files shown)


python/unblob/handlers/archive/moxa/frm.py
```diff
@@ -109,8 +112,17 @@ def _extract_fs_section(self, file: File, fs: FileSystem, section_offset: int):
             if not name:
                 continue
 
-            file.seek(section_offset + entry.data_offset, io.SEEK_SET)
-            raw = file.read(entry.file_length)
+            # A file declared inside the filesystem section must keep its content
+            # within that section. file_length comes straight from the entry, so an
+            # oversized value would otherwise read past the section into the next
+            # section (or to EOF) and write those bytes into the extracted file.
```
qkaiser: This should be in your commit description, not in the code.

metsw24-max: Good point. Dropped the comment and moved that explanation into the fix commit description.

qkaiser:
@metsw24-max  I think you can squash both commits into one.

metsw24-max:
Done, squashed into one commit.

qkaiser (approved):

PR #1540, opened 2026-06-13
fix(handlers): reject unsorted uzip block offset table

metsw24-max:
**Unsorted block offset table in uzip**

The table of contents is read from the header and treated as consecutive `[start, end)` block bounds without checking the offsets ascend, so a descending entry gives a negative `next - current` length that `iterate_file` quietly treats as zero (the block is dropped) and leaves `toc[-1]` no longer marking the end of the block data. The non-decreasing check lives in the header validation and on the extractor path, so a crafted image is rejected rather than silently mis-extracted; samples with a normal sorted table are unaffected.

### DIFF
--- a/python/unblob/handlers/compression/uzip.py
+++ b/python/unblob/handlers/compression/uzip.py
@@ -1,3 +1,4 @@
+import itertools
 import lzma
 import re
 import zlib
@@ -78,6 +79,10 @@ DECOMPRESS_METHOD: dict[bytes, type[Decompressor]] = {
 }
 
 
+def is_toc_sorted(toc) -> bool:
+    return all(current <= nxt for current, nxt in itertools.pairwise(toc))
+
+
 class UZIPExtractor(Extractor):
     def extract(self, inpath: Path, outdir: Path):
         with File.from_path(inpath) as infile:
@@ -135,6 +140,7 @@ class UZIPHandler(StructHandler):
             header.block_count > 0
             and header.block_size > 0
             and header.block_size % 512 == 0
+            and is_toc_sorted(header.toc)
         )
 
     def calculate_chunk(self, file: File, start_offset: int) -> ValidChunk | None:
--- a/tests/handlers/compression/test_uzip.py
+++ b/tests/handlers/compression/test_uzip.py
@@ -1,0 +1,39 @@
+import struct
+
+import pytest
+
+from unblob.file_utils import File, InvalidInputFormat
+from unblob.handlers.compression.uzip import UZIPHandler
+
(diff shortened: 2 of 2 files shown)


python/unblob/handlers/compression/uzip.py
```diff
@@ -78,11 +79,20 @@ class ZSTDDecompressor(Decompressor):
 }
 
 
+def is_toc_sorted(toc) -> bool:
+    """Check that the TOC block boundaries never decrease."""
```
qkaiser: No need. The function name is self explanatory.

python/unblob/handlers/compression/uzip.py
```diff
@@ -78,11 +79,20 @@ class ZSTDDecompressor(Decompressor):
...
+
 class UZIPExtractor(Extractor):
     def extract(self, inpath: Path, outdir: Path):
         with File.from_path(inpath) as infile:
             parser = StructParser(C_DEFINITIONS)
             header = parser.parse(HEADER_STRUCT, infile, Endian.BIG)
+            if not is_toc_sorted(header.toc):
+                raise InvalidInputFormat(
+                    "uzip block offset table is not monotonically increasing"
+                )
```
qkaiser: Since it's checked by the handler, you can never reach this code during extraction. It can be removed.

python/unblob/handlers/compression/uzip.py
metsw24-max: Good point, dropped the docstring.

python/unblob/handlers/compression/uzip.py
metsw24-max: Removed. The check in is_valid_header already rejects the image before extraction runs, so I dropped the direct-extract test that was exercising this path too.

qkaiser (approved):

tests/handlers/compression/test_uzip.py
```diff
@@ -0,0 +1,38 @@
...
+    )
+    # keep toc[-1] within the file so only the ordering check can reject the image
+    return header + b"\x00" * (max(toc) + 64 - len(header))
+
+
+def test_calculate_chunk_accepts_sorted_toc():
+    f = File.from_bytes(build_image([600, 700, 800]))
+    f.seek(0)
+    chunk = UZIPHandler().calculate_chunk(f, 0)
+    assert chunk.end_offset == 800
```
qkaiser, line 31:
```suggestion
    assert chunk is not None
    assert chunk.end_offset == 800
```

metsw24-max: Done, added the None guard so the type narrows before reading end_offset.

PR #1543, opened 2026-06-15
feat(handler): added QCA LZMA handler

jstucke:
Adds a handler for unpacking compressed executables from Qualcomm Atheros powerline firmware images (.nvm) used in devices of various vendors (only the newer ones, as the older ones do not use compression). The images use LZMA compression with a custom header and the decompressed data often contains additional compressed streams of this kind. Some devices known to use this format:
- D-Link COVR-P2502
- AVM/fritz powerline 500E, 510E, 1210, 1220E
- TP-Link TL-PA8010, TL-WPA4220 v4, TL-WPA4530
- TRENDnet TPL-410APK

The test files were created manually (the plc-utils to not contain a tool for this)


python/unblob/handlers/compression/qca_lzma.py
```diff
@@ -0,0 +1,123 @@
...
+    )
+
+    C_DEFINITIONS = C_DEFINITIONS
+    HEADER_STRUCT = "qca_lzma_header_t"
+    EXTRACTOR = QcaLzmaExtractor()
+    PATTERN_MATCH_OFFSET = 8
+
+    def calculate_chunk(self, file: File, start_offset: int) -> ValidChunk | None:
+        header = self.parse_header(file, Endian.LITTLE)
+        end_offset = start_offset + HEADER_SIZE + header.compressed_size
```
qkaiser:
```suggestion
        end_offset = start_offset + header.size + header.compressed_size
```

or


```suggestion
        end_offset = start_offset + len(header) + header.compressed_size
```

python/unblob/handlers/compression/qca_lzma.py
```diff
@@ -0,0 +1,123 @@
...
+
+        if header.compressed_size == 0 or end_offset > file.size():
+            raise InvalidInputFormat(
+                f"Invalid compressed size {header.compressed_size}"
+            )
```
qkaiser: The end offset check is already handled by unblob if you return a ValidChunk with an end offset past EOF.

python/unblob/handlers/compression/qca_lzma.py
```diff
@@ -0,0 +1,123 @@
...
+
+        if (
+            header.decompressed_size == 0
+            or header.decompressed_size < header.compressed_size
+        ):
+            raise InvalidInputFormat(
+                f"Invalid decompressed size {header.decompressed_size}"
+            )
```
qkaiser: I would move this into a dedicated function, something like `valid_header(header) -> bool`

python/unblob/handlers/compression/qca_lzma.py
```diff
@@ -0,0 +1,123 @@
...
+
+        compressed_data = file.read(header.compressed_size)
+        # validate by attempting decompression to avoid FPs
```
qkaiser: No need for this comment. You can put it into the commit description if required.

python/unblob/handlers/compression/qca_lzma.py
```diff
@@ -0,0 +1,123 @@
...
+            {
+                "id": lzma.FILTER_LZMA1,
+                "lc": lc,
+                "lp": 0,
+                "pb": pb,
+                "dict_size": dict_size,
+            }
+        ],
+    )
+    return decompressor.decompress(data, max_length=decomp_size)
```
qkaiser: It's not picked up by unblob's LZMA handler ?

python/unblob/handlers/compression/qca_lzma.py
```diff
@@ -0,0 +1,123 @@
...
+        ):
+            raise InvalidInputFormat(
+                f"Invalid decompressed size {header.decompressed_size}"
+            )
+
+        compressed_data = file.read(header.compressed_size)
+        # validate by attempting decompression to avoid FPs
+        try:
+            _decompress(
+                compressed_data,
```
qkaiser: Is it possible to use streaming decompression so you don't need to read the whole file in memory ? The doc states that `LZMADecompressor`:
> Create a decompressor object, which can be used to decompress data incrementally.

python/unblob/handlers/compression/qca_lzma.py
```diff
@@ -0,0 +1,123 @@
...
+
+class QcaLzmaExtractor(Extractor):
+    def __init__(self):
+        self._struct_parser = StructParser(C_DEFINITIONS)
+
+    def extract(self, inpath: Path, outdir: Path):
+        fs = FileSystem(outdir)
+        with File.from_path(inpath) as file:
+            header = self._struct_parser.parse("qca_lzma_header_t", file, Endian.LITTLE)
+            compressed_data = file.read(header.compressed_size)
```
qkaiser: Same comment about incremental decompression / streaming decompression here. Can we not load the whole thing in memory ?

python/unblob/handlers/compression/qca_lzma.py
```diff
@@ -0,0 +1,123 @@
...
+            decompressed = _decompress(
+                compressed_data,
+                header.decompressed_size,
+                lc=header.literal_context_bits,
+                pb=header.position_bits,
+                dict_size=header.dict_size,
+            )
+            fs.write_bytes(Path("qca_lzma.uncompressed"), decompressed)
```
qkaiser: Maybe use:


```suggestion
            fs.write_bytes(Path(f"{inpath.stem}.uncompressed"), decompressed)
```

python/unblob/handlers/compression/qca_lzma.py
jstucke: `lzma.FORMAT_RAW` does not have a header.
This format does have a header, though (but a custom/proprietary one, unsupported by the built-in lzma package). 

We could alternatively write our own legacy ("ALONE") header in front of the raw LZMA stream and let unblob handle the rest, since we have all the necessary information to do that. Would that be better?

qkaiser: I think so. The whole end offset derivation logic is already present in the LZMA handler.

python/unblob/handlers/compression/qca_lzma.py
jstucke: Yes, it is possible. I changed it so that it decompresses the data chunk-wise.

python/unblob/handlers/compression/qca_lzma.py
jstucke: yes, see above

python/unblob/handlers/compression/qca_lzma.py
jstucke: I moved it into a validation method

python/unblob/handlers/compression/qca_lzma.py
jstucke: removed the check

python/unblob/handlers/compression/qca_lzma.py
jstucke: changed

python/unblob/handlers/compression/qca_lzma.py
jstucke: changed

tests/integration/compression/qca_lzma/__output__/lorem.qca_lzma_extract/0-132.unknown
qkaiser, line 1: Why do you end up with 132 bytes of unknown chunk here ?

jstucke:
rebased on main

python/unblob/handlers/compression/qca_lzma.py
```diff
@@ -0,0 +1,116 @@
...
+        self._struct_parser = StructParser(C_DEFINITIONS)
+
+    def extract(self, inpath: Path, outdir: Path):
+        fs = FileSystem(outdir)
+        with File.from_path(inpath) as file:
+            header = self._struct_parser.parse("qca_lzma_header_t", file, Endian.LITTLE)
+            fs.write_chunks(
+                Path(f"{inpath.stem}.uncompressed"), _decompress(file, header)
+            )
+
```
qkaiser, line 40:
```suggestion
    def extract(self, inpath: Path, outdir: Path):
        fs = FileSystem(outdir)
        with File.from_path(inpath) as file:
            header = self._struct_parser.parse("qca_lzma_header_t", file, Endian.LITTLE)
            fs.write_chunks(
                Path(f"{inpath.stem}.uncompressed"), _decompress(file, header)
            )
        return ExtractResult(reports=fs.problems)

```

tests/integration/compression/qca_lzma/__output__/lorem.qca_lzma_extract/0-132.unknown
jstucke: Good question. The thing is, that in most samples there is another header with some padding in front of this format. This header is probably unrelated, because the samples that are recursively unpacked from other samples of this type do not seem to have this header. 

Should I remove it from the test file?

qkaiser: Let's keep it. I just wanted to know if it was there due to a mistake in this handler :)

python/unblob/handlers/compression/qca_lzma.py
jstucke: changed

qkaiser (approved):

PR #1545, opened 2026-06-16
fix(handlers): skip lzo compressed checksum for stored blocks

metsw24-max:
**Stored lzo blocks carry no compressed checksum**

`LZOHandler.calculate_chunk` walks each data block by skipping `checksum_size + compressed_size`, and it counts a compressed-block checksum whenever the `F_*_C` flag is set. lzop only writes that checksum for blocks it actually compressed; incompressible blocks are stored verbatim (`compressed_size == uncompressed_size`) with no compressed checksum, so the walk skips four bytes too far and loses sync. A trailing stored block then makes the next size read run past the end and a valid stream gets rejected, and with more data after it the chunk boundary settles on an attacker-influenced offset instead.

Limited the checksum accounting to blocks where `compressed_size < uncompressed_size`, which keeps the common compressed-block path untouched and should be the safer default to reason about.

Added a small regression test building a stored block with the compressed-checksum flag set.

### DIFF
--- a/python/unblob/handlers/compression/lzo.py
+++ b/python/unblob/handlers/compression/lzo.py
@@ -146,7 +146,7 @@ class LZOHandler(StructHandler):
             if (
                 header.flags & HeaderFlags.ADLER32_C
                 or header.flags & HeaderFlags.CRC32_C
-            ):
+            ) and compressed_size < uncompressed_size:
                 checksum_size += CHECKSUM_LENGTH
 
             file.seek(checksum_size + compressed_size, io.SEEK_CUR)
--- a/tests/handlers/compression/test_lzo.py
+++ b/tests/handlers/compression/test_lzo.py
@@ -1,0 +1,61 @@
+import struct
+import zlib
+
+from unblob.file_utils import File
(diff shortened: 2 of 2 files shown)


python/unblob/handlers/compression/lzo.py
```diff
@@ -143,10 +143,14 @@ def calculate_chunk(self, file: File, start_offset: int) -> ValidChunk | None:
             ):
                 checksum_size += CHECKSUM_LENGTH
 
+            # The compressed checksum is only written for blocks that were
+            # actually compressed. Incompressible blocks are stored verbatim
+            # (compressed_size == uncompressed_size) and carry no compressed
+            # checksum, so accounting for one here desyncs the block walk.
```
qkaiser: Should be part of commit description, not a comment.

qkaiser (commented):
Good catch !

python/unblob/handlers/compression/lzo.py
metsw24-max: Moved it into the commit body and dropped the inline comment.

qkaiser (approved):

PR #1548, opened 2026-06-16
fix(handlers): bound qnap nas encrypted_len to the payload

metsw24-max:
**Unbounded encrypted_len in the qnap_nas footer scan**

The NAS handler accepts the footer's `encrypted_len` without checking it against the payload that precedes the footer, so a crafted image makes `QnapExtractor` decrypt past the payload into the footer and underflow the trailing plaintext copy length (a value past EOF also raises `SeekError`). The fix applies the same bound the `qnap_networking` handler already enforces, rejecting the image during the footer scan before extraction runs.

### DIFF
--- a/python/unblob/handlers/archive/qnap/qnap_nas.py
+++ b/python/unblob/handlers/archive/qnap/qnap_nas.py
@@ -57,6 +57,9 @@ def _hyperscan_match(
     logger.debug("qnap_header_t", header=header)
 
     if is_valid_header(header):
+        encrypted_data_len = offset - context.start_offset
+        if header.encrypted_len == 0 or header.encrypted_len > encrypted_data_len:
+            return Scan.Continue
         context.end_offset = context.file.tell()
         return Scan.Terminate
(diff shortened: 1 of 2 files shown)


python/unblob/handlers/archive/qnap/qnap_nas.py
```diff
@@ -57,6 +57,13 @@ def _hyperscan_match(
     logger.debug("qnap_header_t", header=header)
 
     if is_valid_header(header):
+        # encrypted_len must stay within the payload preceding this footer,
+        # otherwise QnapExtractor decrypts past it into the footer and the
+        # trailing plaintext copy length underflows. The QnapNetworking
+        # handler enforces the same bound.
```
qkaiser: should be in commit description, not in code

metsw24-max: Moved it into the commit description and dropped the comment. Pushed.

qkaiser (approved):

PR #1550, opened 2026-06-17
fix(handlers): reject ufs inode numbers outside the inode table

metsw24-max:
**Out-of-range inode number in `read_inode`**
The inode number comes straight from directory entries but is never checked against the inode table, so a crafted entry makes `read_inode` seek to an arbitrary offset and parse unrelated bytes as an inode (and a zero `fs_inodes_per_group` divides by zero). Bounded it to the filesystem's inode count before use, matching the guard `minixfs` already has.

### DIFF
--- a/python/unblob/handlers/filesystem/ufs.py
+++ b/python/unblob/handlers/filesystem/ufs.py
@@ -330,6 +330,9 @@ class UFSParser:
         self.super_block = self._struct_parser.parse(
             "ufs_superblock_t", self.file, Endian.LITTLE
         )
+        self.inode_count = (
+            self.super_block.fs_inodes_per_group * self.super_block.fs_ncg
+        )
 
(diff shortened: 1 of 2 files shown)


python/unblob/handlers/filesystem/ufs.py
```diff
@@ -432,6 +432,9 @@ def read_symlink(self, inode) -> str:
         return chunk[: inode.size].decode("utf-8", errors="replace")
 
     def read_inode(self, ino_number: int):
+        inode_count = self.super_block.fs_inodes_per_group * self.super_block.fs_ncg
```
qkaiser: I would store this value as an object attribute during `__init__` instead of recomputing it wheneve we read an inode.

metsw24-max: Good call. Moved it to self.inode_count in __init__ so read_inode just checks against the stored value now, and squashed it back into the single commit.

qkaiser (approved):

PR #1552, opened 2026-06-19
fix(handlers): bound minixfs data zone index to the filesystem

metsw24-max:
**Out-of-range zone index in MinixFS extraction**

`MinixFS._read_zone_data` turns a zone index into a byte offset (`zone_index * zone_size`) with no check that the index falls inside the filesystem's data area. The indices come straight from an inode's `i_zone[]` table and from indirect zone-pointer blocks, so a crafted image can point a regular file (or directory) at a zone below `s_firstdatazone`, which makes extraction read the boot block, superblock, inode bitmap or inode table and write that metadata into the extracted file; an index at or past the zone count instead seeks beyond the image and aborts the extraction. The inode *number* is already range-checked in `_read_inode`, but the zone numbers reached from a valid inode were not.

The guard goes in the single `_read_zone_data` choke point shared by the direct, indirect and directory read paths, rejecting anything outside `s_firstdatazone <= zone < zone_count` (`s_nzones` for v1, `s_zones` for v2/v3). Images whose data zones sit in the data area are unaffected.

Added a unit test that builds a minimal v1 superblock and checks a valid data zone still reads while metadata and out-of-filesystem zones are rejected.

### DIFF
--- a/python/unblob/handlers/filesystem/minixfs.py
+++ b/python/unblob/handlers/filesystem/minixfs.py
@@ -158,6 +158,9 @@ class MinixFS:
         zmap_offset = imap_offset + self.superblock.s_imap_blocks * block_size
         self.inode_offset = zmap_offset + self.superblock.s_zmap_blocks * block_size
         self.zone_size = (block_size << self.superblock.s_log_zone_size) & 0xFF_FF_FF_FF
+        self.zone_count = (
+            self.superblock.s_nzones if self.version == 1 else self.superblock.s_zones
+        )
         self.inode_size = self.struct_parser.cparser_le.minix_inode.size
         self.zone_ptr_size = self.struct_parser.cparser_le.minix_inode.fields[
             "i_zone"
@@ -178,6 +181,8 @@ class MinixFS:
         raise InvalidInputFormat(f"Invalid magic: {self.superblock.s_magic:x}")
 
     def _read_zone_data(self, zone_index: int) -> bytes:
+        if not self.superblock.s_firstdatazone <= zone_index < self.zone_count:
+            raise InvalidInputFormat(f"Zone index out of range: {zone_index}")
         self.file.seek(zone_index * self.zone_size, io.SEEK_SET)
         return self.file.read(self.zone_size)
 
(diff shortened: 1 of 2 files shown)


python/unblob/handlers/filesystem/minixfs.py
```diff
@@ -158,6 +158,10 @@ def __init__(self, file: File, version: int, c_definitions: str):
         zmap_offset = imap_offset + self.superblock.s_imap_blocks * block_size
         self.inode_offset = zmap_offset + self.superblock.s_zmap_blocks * block_size
         self.zone_size = (block_size << self.superblock.s_log_zone_size) & 0xFF_FF_FF_FF
+        self.first_data_zone = self.superblock.s_firstdatazone
```
qkaiser: Not sure we need to assign it to a dedicated attribute. Most of our object attributes are set because they involve some sort of calculation.

python/unblob/handlers/filesystem/minixfs.py
```diff
@@ -178,6 +182,8 @@ def _get_name_len(self) -> int:
         raise InvalidInputFormat(f"Invalid magic: {self.superblock.s_magic:x}")
 
     def _read_zone_data(self, zone_index: int) -> bytes:
+        if not self.first_data_zone <= zone_index < self.zone_count:
```
qkaiser:
```suggestion
        if not self.superblock.s_firstdatazone <= zone_index < self.zone_count:
```

qkaiser (commented):
Make sure your commit message starts with `fix(handlers): `.

python/unblob/handlers/filesystem/minixfs.py
metsw24-max: Agreed, it was just a copy with no calculation behind it. Dropped the attribute and read s_firstdatazone straight off the superblock in the check.

python/unblob/handlers/filesystem/minixfs.py
metsw24-max: Done, applied the suggestion.

metsw24-max:
Reworded the commit so it starts with fix(handlers):. Also dropped the first_data_zone attribute and inlined the superblock read per the review.

qkaiser (approved):

PR #1553, opened 2026-06-20
fix(handlers): bound ufs block pointers to the data area

metsw24-max:
**Unbounded block pointers in the UFS extractor**

Block addresses come straight from inode and indirect-block data but are turned into file offsets and read without checking they fall inside the data area, so a pointer below the first data fragment reads filesystem metadata into the extracted file and one past the fragment count seeks outside the image. Bounded each data-block read to the cylinder group's data range in a shared helper, mirroring the inode-number guard already in `read_inode`. Added a unit test covering an out-of-area pointer, a past-the-end pointer, and a valid block.

### DIFF
--- a/python/unblob/handlers/filesystem/ufs.py
+++ b/python/unblob/handlers/filesystem/ufs.py
@@ -376,23 +376,27 @@ class UFSParser:
 
     def read_file_content(self, inode) -> Iterator[bytes]:
         remaining = inode.size
-        for chunk in self.read_direct_blocks(inode):
-            to_read = min(len(chunk), remaining)
-            yield chunk[:to_read]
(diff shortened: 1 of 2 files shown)


qkaiser:
@metsw24-max  You always provide good PR description. It would be even better if you put those description in your commit message :) This is applicable to all your open PRs.

metsw24-max:
Good point. Folded the description into the commit body here, and I'll do the same across my other open PRs.

qkaiser:
Made some adjustments that came up during review and did some rewording between the handler and extractor so they're in sync. Looks good to me.

qkaiser (approved):

PR #1554, opened 2026-06-20
fix(handlers): bound xiaomi hdr blob size to the signature

metsw24-max:
**Unbounded blob size in HDR extraction**

Each blob's size is read from the image and carved without checking it against the blob region, which ends where the trailing signature block starts, so a blob declaring an oversized length runs past its own data and pulls the signature header and any following blobs into that blob's output file. Bounded each blob's data range to `signature_offset` in the shared parser so both HDR1 and HDR2 reject it; blobs that sit before the signature are unaffected.

### DIFF
--- a/python/unblob/handlers/archive/xiaomi/hdr.py
+++ b/python/unblob/handlers/archive/xiaomi/hdr.py
@@ -122,11 +122,15 @@ class HDRExtractor(Extractor):
             if not is_valid_blob_header(blob_header):
                 raise InvalidInputFormat("Invalid HDR blob header.")
 
+            # file.tell() points to right after the blob_header == start_offset
+            blob_start = file.tell()
+            if blob_start + blob_header.blob_size > header.signature_offset:
+                raise InvalidInputFormat("HDR blob data extends past the signature.")
+
             yield (
                 (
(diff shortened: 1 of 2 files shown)


python/unblob/handlers/archive/xiaomi/hdr.py
```diff
@@ -122,11 +122,19 @@ def parse(self, file: File) -> Iterable[tuple[Path, int, int]]:
             if not is_valid_blob_header(blob_header):
                 raise InvalidInputFormat("Invalid HDR blob header.")
 
+            # file.tell() points to right after the blob_header == start_offset
+            blob_start = file.tell()
+            # Blobs precede the trailing signature block, so a blob whose data
+            # would reach into (or past) the signature region is malformed. Without
+            # this bound an oversized blob_size makes the carve pull the signature
+            # header and any following blobs into this blob's output file.
```
qkaiser: should be part of commit description, not the code

metsw24-max: Good point. Moved the rationale into the commit body and dropped the block comment, leaving just the bound check.

qkaiser (approved):

PR #1555, opened 2026-06-20
fix(handlers): correct engenius chunk end double-counting header

metsw24-max:
**Header size double-counted in the engenius chunk**

`length` is the absolute size of the firmware image: the extractor deciphers the payload over `range(header_end, length)` and the reference tool deciphers from the end of the header up to the end of the file. `calculate_chunk` adds `len(header)` on top of `length`, so the header is counted twice and the chunk runs `len(header)` bytes past the real image. A whole-file image whose `length` equals the file size then ends up past EOF and the chunk is dropped, while an embedded image swallows the bytes that follow it. Switched the end offset to `start_offset + header.length` so the boundary matches what the extractor actually reads.

Added a small unit test exercising the chunk boundary.


qkaiser:
> length is the absolute size of the firmware image

Are you sure ?

> the extractor deciphers the payload over range(header_end, length) and the reference tool deciphers from the end of the header up to the end of the file.

So length is the payload length that follows the header.

> calculate_chunk adds len(header) on top of length, so the header is counted twice and the chunk runs len(header) bytes past the real image.

Why is it counted twice if `length` is the length of the encrypted payload that follows the header ?

metsw24-max:
**`length` is an absolute offset, not a payload count**

The extractor uses it directly as the stop bound:

```python
for offset in range(f.tell(), engenius_header.length):
```

`f.tell()` is the header end (`len(header)`), and the loop runs up to `engenius_header.length`. If `length` were the number of payload bytes, the stop would need to be `f.tell() + length`. Since it's plain `length`, the payload sits in `[len(header), length)` and `length` is the offset of the end of the image, i.e. the image size. That matches the reference decryptor too, which ciphers from the header end to EOF, so `length == EOF`.

`is_valid_header` lines up with this: it rejects `length <= len(header)`, which only makes sense if `length` has to exceed the header itself (an absolute size) rather than being a payload count.

**Why the header ends up counted twice**

The chunk needs to cover `[start, start + length)`, and that one range already contains both the header `[0, len(header))` and the payload `[len(header), length)`. The old `start_offset + len(header) + header.length` adds the header size again on top of `length`, so the chunk runs `len(header)` bytes past the real image. For a whole-file image that pushes `end_offset` past EOF and the finder drops the chunk; for an embedded one it swallows the bytes that follow.

My wording about "the payload length that follows the header" was loose and that's probably where the mix-up crept in. Happy to reword the description if it reads more clearly.

qkaiser:
@metsw24-max you'll need to re-generate the integration tests artifacts with your changes for the pipeline to pass.

metsw24-max:
Done, regenerated the engenius integration output against the fixed bound. The decrypted payload is byte-for-byte identical; the only change is that the chunk now ends at `length` (4096) rather than running to the end of the file, so the 147 trailing bytes the old whole-file chunk used to swallow are now carved as a separate `4096-4243.unknown`. Diff check passes locally.

qkaiser (approved):

PR #1556, opened 2026-06-21
fix(handlers): derive dmg chunk start from all koly offsets

metsw24-max:
dmghandler.calculate_chunk reconstructs the chunk start by subtracting datafork_length and xmllength from the koly trailer position, which only holds when the xml plist sits right after the data fork. a dmg that carries a resource fork, or leaves any gap before the plist, then has its start computed inside the data fork and loses the leading bytes of the image; and since datafork_length comes straight from the trailer, an oversized value walks the start in front of the image so unrelated preceding bytes get pulled into the carved chunk. derive the extent from the furthest structure the koly actually references (data fork, resource fork and xml offsets together with their lengths), the same shape the elf handler already uses, so the chunk covers the whole image whatever the layout.

the integration sample `fruits_rsrc_fork.dmg` carries a genuine resource fork between the data fork and the plist (`RsrcForkOffset`/`RsrcForkLength` set in the koly). without the fix the start lands `RsrcForkLength` bytes inside the image and 7z can't open the carved chunk; with it the chunk spans the whole image and extraction matches `fruits.dmg`.

### DIFF
--- a/python/unblob/handlers/archive/dmg.py
+++ b/python/unblob/handlers/archive/dmg.py
@@ -104,8 +104,18 @@ class DMGHandler(StructHandler):
         # │koly trailer │
         # └─────────────┘
         #
+        # The image begins where the furthest structure the trailer references
+        # ends, walking back from the trailer. Deriving the start from
+        # DataForkLength alone assumes the XML plist sits right after the data
+        # fork; images with a resource fork (or any gap) are then carved from
+        # inside the data fork.
+        image_size = max(
+            header.DataForkOffset + header.DataForkLength,
+            header.RsrcForkOffset + header.RsrcForkLength,
+            header.XMLOffset + header.XMLLength,
+        )
 
         return ValidChunk(
-            start_offset=start_offset - header.XMLLength - header.DataForkLength,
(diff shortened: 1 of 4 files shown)


qkaiser:
@metsw24-max can you generate integration tests samples that contains a resource fork ?

metsw24-max:
Added `fruits_rsrc_fork.dmg`. hdiutil on current macOS only emits UDIF images that keep the block table in the XML plist with the resource-fork fields zeroed, so a sample doesn't fall out of the tool directly. I took the existing `fruits.dmg`, rebuilt its blkx records into a real resource fork, and placed that between the data fork and the plist, with `RsrcForkOffset`/`RsrcForkLength` set in the koly. The data fork and plist are untouched, so extraction is byte-identical to `fruits.dmg` and reuses the same `__output__`.

It exercises the bug end to end: without the patch the carve starts `RsrcForkLength` bytes into the image and 7z can't open the result, with it the chunk spans the whole image and extracts cleanly. Dropped the synthetic unit test in favour of this.

qkaiser (approved):

PR #1561, opened 2026-06-25
feat(handler): TP-Link MiniFS

jstucke:
* adds support for TP-Link MiniFS filesystem format
  * based on the structure described in https://arxiv.org/html/2407.05064v1
  * technically the data is unpacked by unblob without this unpacker (because everything is compressed with LZMA), but all information about directory structure and file names is lost
  * successfully tested with samples of multiple TP-Link routers: 
    * Archer-C20-6.20-TPOS-up-noboot-128B_2024-09-24_16.51.32.bin
    * mr505v1-1.13.3-up-boot_2025-01-23_16.16.14.bin
    * re190v5-up-noboot-01900005-ver1-0-12-P1[20221019-rel57747].bin
    * re450v4-ver1-0-12-P1[20220908-rel64762].bin
    * wa850rev7-up-noboot-ver1-0-10-P1[20211117-rel31538].bin
* the test files were created manually (since there is no known (open source) packer available AFAIK)


python/unblob/handlers/filesystem/minifs.py
```diff
@@ -0,0 +1,176 @@
...
+            + last_chunk.compressed_size,
+        )
+
+
+def _validate_header(header, file: File):
+    if header.file_count == 0 or header.file_count > MAX_FILE_COUNT:
+        raise InvalidInputFormat("Invalid number of files.")
+    if header.first_chunk_len == 0:
+        raise InvalidInputFormat(f"Invalid first chunk size {header.first_chunk_len}.")
+    if header.name_table_len == 0 or header.name_table_len > file.size():
```
qkaiser: Incorrect if your chunk starts after 0:

```suggestion
    if header.name_table_len == 0 or header.name_table_len > file.size() - start_offset:
```

python/unblob/handlers/filesystem/minifs.py
```diff
@@ -0,0 +1,176 @@
...
+        # go to last entry of the file table to find the chunk count...
+        file.seek(
+            header.name_table_len + FILE_TABLE_ENTRY_SIZE * (header.file_count - 1),
+            io.SEEK_CUR,
+        )
+        last_entry = self._struct_parser.parse("file_table_entry_t", file, Endian.BIG)
+        chunk_count = last_entry.chunk_id + 1
+
+        # ...then go to the last entry of the chunk table to find where the final chunk ends
+        file.seek(CHUNK_TABLE_ENTRY_SIZE * (chunk_count - 1), io.SEEK_CUR)
```
qkaiser: you do `chunk_id + 1` on 156, then `chunk_count - 1` on 159.


```suggestion
        # ...then go to the last entry of the chunk table to find where the final chunk ends
        file.seek(CHUNK_TABLE_ENTRY_SIZE * last_entry.chunk_id, io.SEEK_CUR)
```

python/unblob/handlers/filesystem/minifs.py
```diff
@@ -0,0 +1,176 @@
...
+        last_chunk = self._struct_parser.parse("chunk_table_entry_t", file, Endian.BIG)
+        chunk_data_start = file.tell()
+        return ValidChunk(
+            start_offset=start_offset,
+            end_offset=chunk_data_start
+            + last_chunk.data_offset
+            + last_chunk.compressed_size,
+        )
```
qkaiser: Are chunks in the chunk table guaranteed to be ordered by their data offsets + compressed size ?

python/unblob/handlers/filesystem/minifs.py
```diff
@@ -0,0 +1,176 @@
...
+                        decompressed[
+                            entry.chunk_offset : entry.chunk_offset + entry.file_len
+                        ],
+                    )
+
+
+def _group_by_chunk(file_entries: list) -> dict[int, list]:
+    by_chunk = {}
+    for entry in file_entries:
+        by_chunk.setdefault(entry.chunk_id, []).append(entry)
```
qkaiser: So you can have multiple file entries linked to the same chunk id ?

python/unblob/handlers/filesystem/minifs.py
```diff
@@ -0,0 +1,176 @@
...
+            files_by_chunk = _group_by_chunk(file_entries)
+
+            for chunk_id, file_entry_list in sorted(files_by_chunk.items()):
+                decompressed = _decompress_chunk(
+                    file, chunk_data_start, chunk_entries[chunk_id]
+                )
+
+                for entry in file_entry_list:
+                    out_path = Path(_resolve_path(names_data, entry))
+                    fs.mkdir(out_path.parent, parents=True, exist_ok=True)
```
qkaiser: You don't need to take care of that. It's handled by `FileSystem` through `_ensure_parent_dir` :)

python/unblob/handlers/filesystem/minifs.py
```diff
@@ -0,0 +1,176 @@
...
+            chunk_data_start = file.tell()
+            files_by_chunk = _group_by_chunk(file_entries)
+
+            for chunk_id, file_entry_list in sorted(files_by_chunk.items()):
+                decompressed = _decompress_chunk(
+                    file, chunk_data_start, chunk_entries[chunk_id]
+                )
+
+                for entry in file_entry_list:
+                    out_path = Path(_resolve_path(names_data, entry))
```
qkaiser: It would be easier to read if `_resolve_path` would return a `Path` object.

qkaiser (commented):
It may be beneficial to have a MiniFSParser class that takes care of parsing all file entries and all chunk entries + encapsulating `_resolve_path` and `_extract_cstring_at`.

Then use this class in the extractor and in the handler.

What do you think ?

python/unblob/handlers/filesystem/minifs.py
jstucke: Yes, because the chunk ID starts at 0 (so the number of chunks is last_chunk_id + 1) and later we seek the offset of the last entry (so we need to jump number_of_chunks - 1 ahead). Storing "largest_chunk_id" (or something like that) in a variable would make less sense and would be less intuitive IMHO

python/unblob/handlers/filesystem/minifs.py
jstucke: That is a good question. Since it is a proprietary format, I can not definitively answer this question, but it was exactly like that in all the samples that I looked at. So probably? We could also parse all tables and look for the last offset (or biggest ID in the other case) to make sure. What do you think is the better approach?

python/unblob/handlers/filesystem/minifs.py
jstucke: Yes, one chunk consists of one or multiple files. I have not encountered a case where a file was split into multiple chunks, so I ruled out that case.

python/unblob/handlers/filesystem/minifs.py
qkaiser: Yes, it's just that it can be simplified.

python/unblob/handlers/filesystem/minifs.py
qkaiser: I would go with parsing all tables and looking for maximum value. If you move things to a `MiniFSParser` this can be done easily.

python/unblob/handlers/filesystem/minifs.py
qkaiser: Noted. Thanks.

jstucke:
I added a parser class which makes handler and extractor much simpler
Also another rebase on main
...and I added the `ExtractResult` as in #1543

python/unblob/handlers/filesystem/minifs.py
jstucke: changed

python/unblob/handlers/filesystem/minifs.py
jstucke: removed the `mkdir`

python/unblob/handlers/filesystem/minifs.py
jstucke: fixed

python/unblob/handlers/filesystem/minifs.py
```diff
@@ -0,0 +1,216 @@
...
+        return cls(
+            data_offset=entry.data_offset,
+            compressed_size=entry.compressed_size,
+            decompressed_size=entry.decompressed_size,
+            id=chunk_id,
+        )
+
+
+class MiniFSParser:
+    def __init__(self, file: File, header=None):
```
qkaiser: You're never instantiating it with a header.


```suggestion
    def __init__(self, file: File):
```

python/unblob/handlers/filesystem/minifs.py
```diff
@@ -0,0 +1,216 @@
...
+        self._struct_parser = StructParser(C_DEFINITIONS)
+
+        self._start_offset = file.tell()
+        self.header = header or self._struct_parser.parse(
+            "minifs_header_t", file, Endian.BIG
+        )
+        self._validate_header(file)
+
+        names_data = file.read(self.header.name_table_len)
```
qkaiser, line 111: It would be nice to have some kind of sanity check for name_table_len. There's a POSIX limit on file name and file path iirc. We could check that `self.header.name_table_len` is not higher than file count multiplied by this POSIX limit.

`name_table_len` is a uint32, which means you could end up loading 4GB in memory with a malformed value.

python/unblob/handlers/filesystem/minifs.py
```diff
@@ -0,0 +1,216 @@
...
+
+    def _group_files_by_chunk(
+        self, parser: MiniFSParser
+    ) -> dict[int, list[MiniFSFile]]:
+        by_chunk = {}
+        for file in parser.file_entries:
+            by_chunk.setdefault(file.chunk_id, []).append(file)
+        for file_list in by_chunk.values():
+            file_list.sort(key=lambda file: file.chunk_offset)
+        return by_chunk
```
qkaiser: This could be a property of `MiniFSParser`. Something like:

```python
@property
def files_by_chunk(self) -> dict[int, list[MiniFSFile]]:
        by_chunk = {}
        for file in parser.file_entries:
            by_chunk.setdefault(file.chunk_id, []).append(file)
        for file_list in by_chunk.values():
            file_list.sort(key=lambda file: file.chunk_offset)
        return by_chunk
```

python/unblob/handlers/filesystem/minifs.py
```diff
@@ -0,0 +1,216 @@
...
+                    input_file, parser.chunk_data_offset, chunk
+                )
+                for file in files_by_chunk[chunk.id]:
+                    fs.write_bytes(
+                        file.path,
+                        decompressed[
+                            file.chunk_offset : file.chunk_offset + file.length
+                        ],
+                    )
+
```
qkaiser: Missing ExtractResult return clause.

```suggestion
    def extract(self, inpath: Path, outdir: Path) -> None:
        fs = FileSystem(outdir)
        with File.from_path(path=inpath) as input_file:
            parser = MiniFSParser(input_file)
            files_by_chunk = self._group_files_by_chunk(parser)
            for chunk in parser.chunk_entries:
                decompressed = self._decompress_chunk(
                    input_file, parser.chunk_data_offset, chunk
                )
                for file in files_by_chunk[chunk.id]:
                    fs.write_bytes(
                        file.path,
                        decompressed[
                            file.chunk_offset : file.chunk_offset + file.length
                        ],
                    )
        return ExtractResult(reports=fs.problems)
```

python/unblob/handlers/filesystem/minifs.py
```diff
@@ -0,0 +1,216 @@
...
+                decompressed = self._decompress_chunk(
+                    input_file, parser.chunk_data_offset, chunk
+                )
+                for file in files_by_chunk[chunk.id]:
+                    fs.write_bytes(
+                        file.path,
+                        decompressed[
+                            file.chunk_offset : file.chunk_offset + file.length
+                        ],
+                    )
```
qkaiser: If you've got the same file linked to different chunks, you only keep the last chunk since `write_bytes` overwrite everything. Is that possible ? Is that what we want ?

python/unblob/handlers/filesystem/minifs.py
```diff
@@ -0,0 +1,216 @@
...
+        return by_chunk
+
+    @staticmethod
+    def _decompress_chunk(file: File, chunk_data_start: int, chunk) -> bytes:
+        file.seek(chunk_data_start + chunk.data_offset, io.SEEK_SET)
+        raw = file.read(chunk.compressed_size)
+        try:
+            return lzma.decompress(raw, format=lzma.FORMAT_ALONE)
+        except lzma.LZMAError as error:
+            raise InvalidInputFormat("LZMA decompression failed.") from error
```
qkaiser, line 187: Similar to https://github.com/onekey-sec/unblob/pull/1543, it would be nice to use streaming decompression. I know it won't be easy to handle it at caller site, but this will ensure memory consumption stays low.

qkaiser (commented):
Nice ! Some more comments but it's getting into shape :)

python/unblob/handlers/filesystem/minifs.py
jstucke: I don't think that is possible, because each file only has one chunk ID (so each file should only appear once).

python/unblob/handlers/filesystem/minifs.py
jstucke: I'm not sure if it is worth it. Because there are usually multiple files in each chunk, it gets complex (and ugly) fast. This seems to accomplish it, but I'm still not sure if it handles all the edge cases:

```python
    def _extract_chunk_streaming(
        self,
        file: File,
        chunk_data_start: int,
        chunk: MiniFSChunk,
        files: list[MiniFSFile],
        fs: FileSystem,
        window_size: int = 2048,
    ) -> None:
        file.seek(chunk_data_start + chunk.data_offset, io.SEEK_SET)

        decompressor = lzma.LZMADecompressor(format=lzma.FORMAT_ALONE)
        overlap = None
        compressed_remaining = chunk.compressed_size

        def iter_file_chunkwise(remaining: int):
            nonlocal overlap, compressed_remaining

            if overlap:
                # we decompressed more than needed for the last file
                if len(overlap) > remaining:
                    yield overlap[:remaining]
                    overlap = overlap[remaining:]
                    return
                else:
                    yield overlap
                    remaining -= len(overlap)
                    overlap = None

            while remaining > 0:
                if decompressor.eof or compressed_remaining <= 0:
                    raise InvalidInputFormat("Chunk data exhausted.")
                read_size = min(window_size, compressed_remaining)
                compressed_data = file.read(read_size)
                if not compressed_data:
                    raise InvalidInputFormat("Reached EoF during decompression")
                decompressed = decompressor.decompress(compressed_data)
                compressed_remaining -= read_size
                if len(decompressed) < remaining:
                    yield decompressed
                    remaining -= len(decompressed)
                else:
                    yield decompressed[:remaining]
                    if len(decompressed) > remaining:
                        overlap = decompressed[remaining:]
                    break

        for f in files:
            fs.write_chunks(f.path, iter_file_chunkwise(f.length))
```

python/unblob/handlers/filesystem/minifs.py
jstucke: fixed

python/unblob/handlers/filesystem/minifs.py
jstucke: I don't know if POSIX even applies (because it seems to be bare metal), but this is a good idea. The longest path entry in all of my samples was 53 characters long. How about adding a limit of 1024 (that seems somewhat reasonable)?

python/unblob/handlers/filesystem/minifs.py
jstucke: Whoops, at some point I passed the header from the handler but then came to the conclusion that the handler doesn't really need it anymore

python/unblob/handlers/filesystem/minifs.py
jstucke: I put it in `MiniFSParser` but added a class variable, so it does not get computed multiple times in the loop in `extract`

python/unblob/handlers/filesystem/minifs.py
qkaiser: Yeah I was afraid of this kind of complexity at the call site. Let's keep it how it is then.

python/unblob/handlers/filesystem/minifs.py
qkaiser: Sounds good.

python/unblob/handlers/filesystem/minifs.py
qkaiser: So what we can do instead is use `memlimit` when calling `lzma.decompress`:

> The memlimit argument specifies a limit (in bytes) on the amount of memory that the decompressor can use. When this argument is used, decompression will fail with an [LZMAError](https://docs.python.org/3/library/lzma.html#lzma.LZMAError) if it is not possible to decompress the input within the given memory limit.

If you can come up with a decent limit in bytes based on observed samples this would be a nice way of ensuring a crafted image cannot make unblob eats all the RAM.

qkaiser: `memlimit` is working memory, not the output. so setting `max_output` to the chunk's `decompressed_size` could also help, if we bound check that value too.

jstucke: I gathered some more samples. The largest single compressed chunk I found is ~270KB. I was able to extract all samples with 9MB memory limit. I set it to 48 MiB to have a bit of a safety margin.

qkaiser: `max_output` is still missing

jstucke: Oh, sorry I must have missed that. I set the `max_length` to `chunk.decompressed_size` as suggested.

jstucke:
@qkaiser do you have an idea what could have caused the fail of `tests/test_handlers.py::test_all_handlers[archive.dmg]`? Seems kinda unrelated to this PR

qkaiser (approved):

PR #1562, opened 2026-06-25
fix(handlers): reject qca_lzma stream ending before compressed_size

metsw24-max:
**QCA LZMA chunk over-claims past the compressed stream**

The chunk end is taken straight from the header's `compressed_size`, but a raw LZMA1 stream carries its own end marker and can finish earlier, so a crafted image can declare a size that runs past the stream into trailing bytes which then get carved into the chunk and hidden from the scan while decompression still succeeds. The handler now tracks the bytes the decompressor actually consumes and rejects the image when the stream ends before the declared `compressed_size`, so valid images (where the two match) are unaffected.

### DIFF
--- a/python/unblob/handlers/compression/qca_lzma.py
+++ b/python/unblob/handlers/compression/qca_lzma.py
@@ -61,6 +61,16 @@ def _decompress(file: File, header, chunk_size: int = 1_024) -> Iterator[bytes]:
         bytes_read += len(chunk)
         yield decompressor.decompress(chunk)
 
+    # The end offset is derived from compressed_size, so the stream must actually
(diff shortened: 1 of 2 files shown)


qkaiser:
@jstucke thoughts ?

jstucke:
Sounds like a good idea to me. I tested it with some powerline samples and it does not seem to negatively affect extraction.

metsw24-max:
Good to hear the powerline samples come through clean. That's the intent: the check only trips when the stream actually ends before the declared `compressed_size`, so valid images where the consumed length matches the declared size are left untouched.

qkaiser (approved):

PR #1563, opened 2026-06-26
fix(handlers): bound fpkg filename to the file header

metsw24-max:
**Filename overruns the fixed FPKG file header**

FPKG file headers are a fixed `0x1C` bytes and each file's data is read from `entry + header_len`, but the name is parsed as an unbounded NUL terminated string and never checked to terminate inside that header. A crafted entry whose name has no terminator within the header is then read out of the following payload, so the carved data starts inside the name and the entry walk desyncs. Bounded the name to `header_len` so such entries are rejected; images whose names fit the header are unaffected.

### DIFF
--- a/python/unblob/handlers/archive/dlink/fpkg.py
+++ b/python/unblob/handlers/archive/dlink/fpkg.py
@@ -50,6 +50,10 @@ CPKG_HEADER = "cpkg_header_t"
 FILE_HEADER = "file_header_t"
 CPKG_HEADER_SIZE = 12
 FILE_HEADER_SIZE = 0x1C
+# offset of the filename within a file header (header_len + type + unknown + file_size)
+FILE_HEADER_NAME_OFFSET = 12
+# space left for the filename plus its terminator inside the fixed header
+MAX_FILENAME_SIZE = FILE_HEADER_SIZE - FILE_HEADER_NAME_OFFSET
 VALID_MAGICS = {b"FPKG", b"CPKG"}
 
 
@@ -100,6 +104,11 @@ class FPKGParser:
         if file_header.header_len != FILE_HEADER_SIZE:
             raise InvalidInputFormat(
                 f"Invalid file header length: {file_header.header_len}"
+            )
+        if len(file_header.filename) >= MAX_FILENAME_SIZE:
+            raise InvalidInputFormat(
+                "Filename not terminated within file header: "
+                f"{file_header.filename[:MAX_FILENAME_SIZE]}..."
             )
         if not file_header.filename.isascii():
             raise InvalidInputFormat(f"Invalid filename: {file_header.filename}")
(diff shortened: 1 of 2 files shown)


python/unblob/handlers/archive/dlink/fpkg.py
```diff
@@ -101,6 +103,16 @@ def _validate_file_header(file_header) -> None:
             raise InvalidInputFormat(
                 f"Invalid file header length: {file_header.header_len}"
             )
+        # The file data is read from entry start + header_len, so the name must be
+        # NUL terminated inside the fixed header. An unterminated name is otherwise
+        # parsed out of the following file data and desyncs the entry walk.
```
qkaiser: Should be in the commit message

qkaiser (commented):
Looks good

python/unblob/handlers/archive/dlink/fpkg.py
```diff
@@ -101,6 +103,16 @@ def _validate_file_header(file_header) -> None:
...
+        if (
+            FILE_HEADER_NAME_OFFSET + len(file_header.filename)
+            >= file_header.header_len
+        ):
+            raise InvalidInputFormat(
+                f"Filename not terminated within file header: {file_header.filename}"
```
e3krisztian: This would duplicate the hypothetical filename of the crafted entry (which is unlimited) in the error message.

```suggestion
                f"Filename not terminated within file header: {file_header.filename[:MAX_FILENAME_SIZE]}..."
```

python/unblob/handlers/archive/dlink/fpkg.py
e3krisztian: I am not familiar with the D-Link data format, however in `fpkg_header` `name_len` is used for `name`, so maybe strings are not always NUL terminated?

Maybe file names are NUL terminated only if they are shorter than the allocated place?

----

May I ask what triggered the PR?
Was it a D-Link (thought to be damaged) or non-D-Link firmware (hijacked as D-Link), that had problem during the extraction?

python/unblob/handlers/archive/dlink/fpkg.py
metsw24-max: Good point, applied. The name is sliced to MAX_FILENAME_SIZE in the message now so a crafted long name cannot bloat the log.

python/unblob/handlers/archive/dlink/fpkg.py
metsw24-max: Moved the rationale into the commit message and dropped the inline comment.

On the NUL termination: the file_header has no length field for the name, unlike fpkg_header where `name` is sized by `name_len`. Each file's data is read from a fixed offset (entry start + header_len, and header_len is always 0x1C), so the name has to terminate before that point. The field is only 16 bytes, so a name that fills it leaves no room for a terminator and the format cannot represent it unambiguously, the reader just runs on into the file data. The check rejects exactly that case; the bundled samples use short names (apple.txt and so on) well within the limit.

As for what triggered it: nothing in the wild, I was going over the handler's bounds and built a crafted FPKG with a 16 byte unterminated name to confirm. The name then gets taken from the following payload and start_offset lands inside it.

qkaiser (approved):

PR #1570, opened 2026-07-01
fix(tar): accept non-standard header checksum format

elektrischermoench:
The chksum field was validated by requiring its last two bytes to be NUL+space or space+NUL, and decoded from only its first 6 bytes. Tar writers that encode the checksum as 7 octal digits followed by a single NUL (filling all 8 bytes) failed both checks: the format guard rejected the header, and even without it the 6-byte decode yielded the wrong value and mismatched the computed checksum.

Because this is applied to every header, an affected archive matches no tar chunk at all and is silently carved into raw unknown blobs instead of extracted, so nested content is never analyzed.

Parse the checksum the way tar readers do: take the octal digits up to the first NUL, ignoring surrounding space padding. Integrity is still enforced by the existing comparison against the computed checksum.

### DIFF
--- a/python/unblob/handlers/archive/tar.py
+++ b/python/unblob/handlers/archive/tar.py
@@ -1,5 +1,6 @@
 import contextlib
 import os
+import re
 import tarfile
 from pathlib import Path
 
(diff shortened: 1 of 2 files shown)


python/unblob/handlers/archive/tar.py
```diff
@@ -140,15 +140,16 @@ def calculate_chunk(self, file: File, start_offset: int) -> ValidChunk | None:
         def signed_sum(octets) -> int:
             return sum(b if b < 128 else 256 - b for b in octets)
 
-        if header.chksum[6:8] not in (b"\x00 ", b" \x00"):
+        chksum_digits = header.chksum.split(b"\x00", 1)[0].strip(b" ")
+        if not chksum_digits or any(c not in b"01234567" for c in chksum_digits):
```
qkaiser: Let's use a regex no ? Something like `r"(?P<checksum_digits>\d+)[\x00\x20]"`.

elektrischermoench: good point. Changed it! Thx.

python/unblob/handlers/archive/tar.py
```diff
@@ -140,15 +141,16 @@ def calculate_chunk(self, file: File, start_offset: int) -> ValidChunk | None:
         def signed_sum(octets) -> int:
             return sum(b if b < 128 else 256 - b for b in octets)
 
-        if header.chksum[6:8] not in (b"\x00 ", b" \x00"):
+        chksum_match = re.search(rb"(?P<digits>[0-7]+)[\x00\x20]", header.chksum)
```
qkaiser:
```suggestion
        chksum_match = re.search(rb"^(?P<digits>[0-7]+)[\x00\x20]+$", header.chksum)
```

qkaiser (approved):

PR #1579, opened 2026-07-16
fix(handlers): scale ntfs chunk end by bytes_per_sector

metsw24-max:
**NTFS chunk end sized by a fixed 512 rather than the sector size**

`calculate_chunk` ended the chunk at `start + len(header) + fsize`, but `total_sectors` excludes the trailing backup boot sector, so the volume runs one `bytes_per_sector`-wide sector past `fsize`. Hardcoding 512 for that sector overshoots on 256-byte-sector volumes (pulling the following bytes into the carved chunk, which the scan then skips) and falls short on 4096-byte ones; scaling it by `bytes_per_sector` fixes both directions and leaves the common 512-byte case unchanged, so the bundled sample still carves identically.

### DIFF
--- a/python/unblob/handlers/filesystem/ntfs.py
+++ b/python/unblob/handlers/filesystem/ntfs.py
@@ -88,7 +88,8 @@ class NTFSHandler(StructHandler):
         if fsize == 0:
             raise InvalidInputFormat("NTFS header with null disk size.")
 
-        end_offset = start_offset + len(header) + fsize
+        trailing_backup_boot_sector_size = header.bytes_per_sector
+        end_offset = start_offset + fsize + trailing_backup_boot_sector_size
         return ValidChunk(
             start_offset=start_offset,
             end_offset=end_offset,
--- a/python/unblob/handlers/filesystem/ntfs.py
+++ b/python/unblob/handlers/filesystem/ntfs.py
@@ -88,7 +88,10 @@ class NTFSHandler(StructHandler):
         if fsize == 0:
             raise InvalidInputFormat("NTFS header with null disk size.")
 
-        end_offset = start_offset + len(header) + fsize
+        # total_sectors excludes the trailing backup boot sector, so the volume
+        # spans one extra sector. That sector is bytes_per_sector wide, not a
+        # fixed 512, so scale it with the sector size rather than the header.
+        end_offset = start_offset + fsize + header.bytes_per_sector
         return ValidChunk(
             start_offset=start_offset,
             end_offset=end_offset,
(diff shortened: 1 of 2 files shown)


python/unblob/handlers/filesystem/ntfs.py
```diff
@@ -88,7 +88,10 @@ def calculate_chunk(self, file: File, start_offset: int) -> ValidChunk | None:
         if fsize == 0:
             raise InvalidInputFormat("NTFS header with null disk size.")
 
-        end_offset = start_offset + len(header) + fsize
+        # total_sectors excludes the trailing backup boot sector, so the volume
+        # spans one extra sector. That sector is bytes_per_sector wide, not a
+        # fixed 512, so scale it with the sector size rather than the header.
```
qkaiser: this belongs to the commit message

python/unblob/handlers/filesystem/ntfs.py
```diff
@@ -88,7 +88,10 @@ def calculate_chunk(self, file: File, start_offset: int) -> ValidChunk | None:
...
+        end_offset = start_offset + fsize + header.bytes_per_sector
```
qkaiser: We can make it even clearer:

```suggestion
        trailing_backup_boot_sector_size = header.bytes_per_sector
        end_offset = start_offset + fsize + trailing_backup_boot_sector_size
```

qkaiser (commented):
looks good, excellent catch ! if you have time creating NTFS integration samples for different supported sector size would be a nice addition

python/unblob/handlers/filesystem/ntfs.py
metsw24-max: Done, moved the explanation into the commit message.

python/unblob/handlers/filesystem/ntfs.py
metsw24-max: Applied, reads better.

metsw24-max:
**Integration samples for other sector sizes**

I had a go at this: built 256/1024/2048/4096 bytes-per-sector volumes with mkntfs and the chunk boundaries come out right, but 7-Zip refuses to open an NTFS volume with anything other than 512 bytes per sector (same behaviour on the macOS and Linux builds), so the extract step fails and `check_result` trips on the `ExtractCommandFailedReport`. As far as I can tell, non-512 samples can't pass CI with the current 7z extractor; the unit test covers the sector-size scaling in the meantime. Happy to add the samples if the extractor ever moves off 7z.

qkaiser (approved):

PR #1580, opened 2026-07-22
fix(lzip): detect members with odd total length

LordAizen1:
The lzip handler works out where a member ends by scanning for the `member_size` field in the trailer. The problem is it steps forward 2 bytes at a time, based on a comment that says "LZMA stream is 2 bytes aligned". That isn't true. lzip doesn't pad the LZMA stream, so a member's total length comes out odd roughly half the time. When it's odd, the `member_size` field lands on an odd offset that the 2-byte scan skips over, the loop runs off the end of the file, and the chunk gets dropped.

So right now unblob quietly misses about half of all lzip files. The smallest way to see it:

```
printf A | lzip -9
```

That gives a 37-byte member that passes `lzip -t` and decompresses fine, but unblob won't pick it up.

I tested with a batch of files made by real lzip (1.24.1) and every odd-sized one got dropped. Scanning a byte at a time instead of two picks all of them up, and nothing else changes. I added a small regression test using that 37-byte file.

There's a separate, rarer issue I left alone here: the brute-force scan can also match a stray 8 bytes inside the stream that happen to equal their own offset, which would end the chunk too early. That one is unrelated to this fix and much less common, so it felt better as its own change.

### DIFF
--- a/python/unblob/handlers/compression/lzip.py
+++ b/python/unblob/handlers/compression/lzip.py
@@ -19,8 +19,9 @@ logger = get_logger()
 
 # magic (4 bytes) + VN (1 byte) + DS (1 byte)
 HEADER_LEN = 4 + 1 + 1
-# LZMA stream is 2 bytes aligned
-LZMA_ALIGNMENT = 2
+# lzip does not pad the LZMA stream, so a member's total length can be odd;
+# scan byte-by-byte so the member_size trailer is found at any offset
+LZMA_ALIGNMENT = 1
 
 
 class LZipHandler(Handler):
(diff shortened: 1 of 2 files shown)


python/unblob/handlers/compression/lzip.py
qkaiser: this line no longer holds true, otherwise looks ready to be merged

LordAizen1:
Good catch, thanks. I updated that comment to say it reads a uint64 at every byte offset, which matches the one-byte step now. Pushed as a small follow-up commit.

qkaiser:
> Good catch, thanks. I updated that comment to say it reads a uint64 at every byte offset, which matches the one-byte step now. Pushed as a small follow-up commit.

squash them, then I'll approve and merge :)

LordAizen1:
Squashed into one commit. Thanks!

qkaiser (approved):

PR #1608, opened 2026-08-18
feat(handler): add LVM2 support

qkaiser:
Parse LVM2 physical volume metadata and expose logical volumes for recursive extraction. Cover single and multi-PV layouts with unit and integration fixtures.

---

LVM2 (Logical Volume Manager 2) is Linux's volume management layer for grouping physical volumes (PVs) into volume groups (VGs) that expose logical volumes (LVs) as resizable virtual block devices. Each PV pairs a text-format metadata area (describing the VG and its LVs) with a data area holding LV contents as fixed-size physical extents.

all the **useful information** can be found [here](https://github.com/libyal/libvslvm/blob/main/documentation/Logical%20Volume%20Manager%20(LVM)%20format.asciidoc), thanks to  Joachim Metz who wrote this wonderful documentation. :)

## Format

All binary fields are **little-endian**.

Each Physical volume (PV) self-describes through structures placed at the start of the device.

#### Label header (sector 1, 32 bytes)

- signature (8 bytes): `LABELONE`
- sector (8 bytes): sector where this label lives (usually 1)
- crc, header_size (4 + 4 bytes)
- type (8 bytes): `LVM2 001`

#### PV header (immediately after label header)

- uuid (32 bytes ASCII, dash-less; the text metadata stores it dash-separated)
- device_size (8 bytes)
- Variable-length list of data area descriptors, terminated by an all-zero descriptor
- Variable-length list of metadata area descriptors, same termination

#### Metadata area header

Found at the offset given by the first metadata area descriptor. Carries four raw location descriptors (`locns[0..3]`), points to the text metadata.

#### Text metadata

Describes the VG in ASCII text, its `physical_volumes` (each entry carries the same UUID as the PV header, dash-separated), and its `logical_volumes`. Each LV is one or more **segments**, where a segment is a mapping `{ start_extent, extent_count, type, stripe_count, stripes = [pv_name, pe_index, ...] }`.

#### Data area

Physical extents (size = `extent_size * SECTOR_SIZE`) holding LV contents. Each linear segment maps `[start_extent, extent_count)` of an LV to `[pe_index, pe_index + extent_count)` on the named PV.

## Handler ability

- **Single-PV VGs** are extracted as complete logical volumes.
- **Multi-PV VGs** produce one partial LV image per PV chunk. The data is preserved across all PV extractions, but combining the partials into a single LV image is left to the user for now.
- **Only linear segments** (striped) are supported, because they're the only shape where the LV's bytes map to a single contiguous range on a single PV.  Other shapes either: split data across multiple PVs and would need cross-chunk reassembly.

### Samples

Generated locally with `losetup` + LVM2 tools, the steps are also described in the documentation mentioned before.

## References 

- https://github.com/libyal/libvslvm/blob/main/documentation/Logical%20Volume%20Manager%20(LVM)%20format.asciidoc
- https://sourceware.org/git/?p=lvm2.git

### DIFF
--- a/python/unblob/handlers/filesystem/lvm.py
+++ b/python/unblob/handlers/filesystem/lvm.py
@@ -1,0 +1,509 @@
+import binascii
+import io
+from pathlib import Path
+from typing import BinaryIO
+
+from unblob.file_utils import (
+    Endian,
+    FileSystem,
+    InvalidInputFormat,
+    StructParser,
+    iterate_file,
+)
+from unblob.models import (
+    Extractor,
+    ExtractResult,
+    File,
+    HandlerDoc,
+    HandlerType,
+    HexString,
+    Reference,
+    StructHandler,
+    ValidChunk,
+)
+from unblob.report import ExtractionProblem
+
+C_DEFINITION = r"""
+        typedef struct lvm_label_header {
+            char     signature[8];   // "LABELONE"
+            uint64   sector;         // sector number of this label header, usually 1
+            uint32   crc;            // CRC of fields below to end of sector
+            uint32   header_size;    // size of this header; pv_header follows immediately
+            char     type[8];        // "LVM2 001"
+        } lvm_label_header_t;
+
+        typedef struct lvm_pv_header {
+            char     uuid[32];       // PV UUID, ASCII
+            uint64   device_size;    // PV size in bytes
+        } lvm_pv_header_t;
+
+        typedef struct lvm_data_area_descriptor {
+            uint64   area_offset;    // relative to start of the PV
+            uint64   area_size;      // 0 = unbounded
+        } lvm_data_area_descriptor_t;
+
+        typedef struct lvm_raw_location_descriptor {
+            uint64   data_offset;    // relative to start of metadata area
+            uint64   data_size;
+            uint32   crc;
+            uint32   flags;          // 0x1 = ignored
+        } lvm_raw_location_descriptor_t;
+
+        typedef struct lvm_metadata_area_header {
+            uint32                          crc;
+            char                            magic[16];   // "\x20LVM2\x20x[5A%r0N*>" signature
+            uint32                          version;
+            uint64                          offset;      // metadata area offset from PV start
+            uint64                          size;        // metadata area size
+            lvm_raw_location_descriptor_t   locns[4];
+            char                            padding[376];
+        } lvm_metadata_area_header_t;
+
+    """
+
+SECTOR_SIZE = 512  # LVM2 format constant
+LABEL_SCAN_SECTORS = 4
+LABEL_HEADER_SIZE = 32
+PV_HEADER_SIZE = 40
+DESCRIPTOR_SIZE = 16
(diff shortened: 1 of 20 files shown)


python/unblob/handlers/filesystem/lvm.py
```diff
@@ -0,0 +1,500 @@
...
+                None,
+            )
+            if locn is None:
+                raise InvalidInputFormat("LVM metadata area has no usable location.")
+
+            text = self._decode_metadata(file, mda_offset, mda.size, locn)
+            metadata = parse_lvm_metadata(text)
+
+            vg = self._get_vg(metadata)
+            pv_name = self._get_pv_name(vg, pv_header.uuid.decode("ascii"))
```
elektrischermoench: does it make sense to handle exception for decode?  It could throw UnicodeDecodeError.

python/unblob/handlers/filesystem/lvm.py
```diff
@@ -0,0 +1,500 @@
...
+            extent_size = vg.get("extent_size")
+            if not isinstance(extent_size, int) or extent_size <= 0:
+                raise InvalidInputFormat("LVM volume group has invalid extent size.")
+            extent_bytes = extent_size * SECTOR_SIZE
+            pe_start = data_area.area_offset
+            data_area_size = data_area.area_size or pv_header.device_size - pe_start
+
+            for lv_name, lv in vg.get("logical_volumes", {}).items():
```
elektrischermoench: vg.get("logical_volumes", {}).items() assuming dict. logical_volumes parses to int. Do we need is_instance check like for extent_size in line 199?

elektrischermoench: `vg.get("logical_volumes", {})` protects against the key being absent, but not against it being present with a non-dict value. logical_volumes = 5 is valid per parse_lvm_metadata's grammar and raises AttributeError. Should this get an` isinstance(..., dict)` guard like the `extent_size` check two lines above, or is aborting on such input acceptable here?

elektrischermoench (commented):
2 small comments/questions otherwise looks good.

python/unblob/handlers/filesystem/lvm.py
qkaiser: Fixed. Now catching the exception

python/unblob/handlers/filesystem/lvm.py
qkaiser: Fixed with an `isinstance` check for now, but I think there's a deeper issue with the lack of strong typing on what we parse out of the volume metadata. I'll check if we can somehow use pydantic in there to have strong validation + strong typing. This should make the code easier to read too.

qkaiser: We could write a `BaseModel` class following LVM standards and pass the output through its validator.

qkaiser: yeah I looked deep into it, it's not worth over-complexifying it

elektrischermoench (approved):

